Issue #524, opened 2021-02-04
Load LDAP operational attributes


hrenard:
**Is your feature request related to a problem? Please describe.**
An issue that I encountered using ldap, is that every attribute may change : uid (username), sn (last name), cn (full name), mail... In order to keep some stability, it might be useful to use the entryUUID as uniqueness field but operational attributes are not available.

**Describe the solution you'd like**
1. Load all attributes during ldap search.
```
attributes=ldap3.ALL_ATTRIBUTES
replaced by
attributes=[ALL_ATTRIBUTES, ALL_OPERATIONAL_ATTRIBUTES]
```
2. Add the option (a checkbox) to load the operational attributes.

**Describe alternatives you've considered**
Not using the entryUUID.

**Additional context**
I'm using OpenLDAP.
I can make a pull request if necessary.

BeryJu:
Hi,

I've created the PR #526 that implements this change, so you can do some testing with your OpenLDAP setup, and I want to do some tests with Active Directory too.

hrenard:
Hi, 
Thanks, that was quick ! 
I'll test it has soon as I can.

BeryJu:
Right, its merged and in master, will be released in 2021.2.1 (hopefully this weekend-ish)

PR #526, opened 2021-02-04
sources/ldap: load operational attributes

BeryJu:
#524


hrenard:
It's working for me.

BeryJu:
Tested on Active Directory and it works fine there too

PR #528, opened 2021-02-04
sources/ldap: support group to user memberships

### LINKED ISSUE #525: Support LDAP group referencing users
My current OpenLDAP setup use the groupOfNames class as a group which contains a member attributes with a list of dn referencing users. 
In that case, user don't have an attribute referencing groups.

It would be nice to have an option to switch between user reference to group reference. The membership field will reference either a user attribute or a group attribute.

Changing my LDAP setup or removing it.

**Additional context**
This probably add some complexity in the sync process.

BeryJu:
#525

Simplify group membership syncing by reading the membership from group -> users.

This also fixes the issue, where an LDAP user isn't removed from an authentik group when removed from an LDAP group.


hrenard:
I tried this with docker tag `gh-ldap-groupOfNames`, so tell me if it's a good way to test.

I'm sorry but there is still some issues :
- the distinguishedName is still null for users and groups
- groups syncs without name but I'm not sure how property mapping works for groups so it may be a miss configuration from me
- disabling group sync, I can sync users but every property is encapsulated in an array

Is there a debug mode to have more logs ?

BeryJu:
> I tried this with docker tag `gh-ldap-groupOfNames`, so tell me if it's a good way to test.

Yep that's correct, I was gonna add a notice to use this tag as its not quite obvious, but I forgot yesterday.

> I'm sorry but there is still some issues :
> 
> * the distinguishedName is still null for users and groups

I just noticed that in the LDIF you posted in the other issue, the field is `dn` not `distinguishedName`. I'll have another look, there might still be some hard-coded references to `distinguishedName`.

> * groups syncs without name but I'm not sure how property mapping works for groups so it may be a miss configuration from me

It seems for group sync I've also hardcoded the Group Name to be `name` instead of Property Mappings

> * disabling group sync, I can sync users but every property is encapsulated in an array

This seems to be different per LDAP Implementation, ldap3 seems to either return a dictionary of values or a dictionary of arrays of values.

> 
> Is there a debug mode to have more logs ?

There is, you can set `AUTHENTIK_LOG_LEVEL=debug` in your .env file.

hrenard:
Okay thanks. Let me know if I can help.

BeryJu:
> > * the distinguishedName is still null for users and groups
> 
> I just noticed that in the LDIF you posted in the other issue, the field is `dn` not `distinguishedName`. I'll have another look, there might still be some hard-coded references to `distinguishedName`.

Right, this is fixed now, it should pick either `dn` or `entryDN` from the result, tested with Active Directory and a Mock OpenLDAP Schema.

> 
> > * groups syncs without name but I'm not sure how property mapping works for groups so it may be a miss configuration from me
> 
> It seems for group sync I've also hardcoded the Group Name to be `name` instead of Property Mappings
> 

So this is still the case, I'm currently figuring out what the best way to implement this would be. What field do you use for your group name?

> > * disabling group sync, I can sync users but every property is encapsulated in an array
> 
> This seems to be different per LDAP Implementation, ldap3 seems to either return a dictionary of values or a dictionary of arrays of values.
> 

This is also fixed now, depending on whether I'm testing against AD or OpenLDAP different values are arrays, these are all flattened out now.

@hrenard

hrenard:
I use `cn` as group name.
I just tested. Everything seems to be working fine except the group name.
Great work !

BeryJu:
@hrenard With this last commit, you can control Group creation with property mappings too.

However after this is merged, I would recommend a clean re-install on the latest stable version, or a restore from backup, as I will be implementing some backwards incompatible changes on master.

PR #790, opened 2021-04-27
outpost: forwardAuth mode

BeryJu:
Send X-Auth* headers on the /akprox/auth endpoint to response instead of forwarding them to be used with traefik forwardAuth and nginx auth_request.


Akruidenberg:
Is it possible to add docker compose for traefik?

PR #960, opened 2021-06-02
Zabbix Integration Instructions

thiggin:
My attempt to document our setup with Zabbix. Please let me know if more details are needed or any recommendations, etc


BeryJu:
Hi, thanks for your contribution!

The layout and instructions look good, the only thing missing is an entry in the sidebar.js file, as seen here: https://github.com/goauthentik/authentik/blob/master/website/sidebars.js#L73

So just add like `integrations/services/zabbix/index`.

BeryJu:
Hmm, the CI Build for the site is failing for some reason, it's not finding the file for some reason. I'll look into it in a bit.

BeryJu:
Oh, its because the sidebars.js file is case sensitive, can you rename the zabbix folder to be lowercase?

PR #975, opened 2021-06-06
update docs to work with proxy

Akruidenberg:
When using Grafana with a proxy like Traefik, Authentik gaves me redirect URl errors. This is because Grafana is using a localhost adres.  Adding the env 'GF_SERVER_ROOT_URL'  to the docs will fix the issue.


BeryJu:
According to the grafana docs, the value should include protocol aswell: https://grafana.com/tutorials/run-grafana-behind-a-proxy/

```
[server]
domain = example.com
root_url = %(protocol)s://%(domain)s:%(http_port)s/grafana/
serve_from_sub_path = true
```

Akruidenberg:
Strange. Mine is working with this setup. Needs more testing maybe.

PR #1124, opened 2021-07-12
Allow for Configurable Redis Port

Starz0r:


BeryJu:
Thanks for the PR, a few minor wishes:

- Please make sure pylint passes (a quick pass with black should fix this)
- Update the port usage in the startup wait script https://github.com/goauthentik/authentik/blob/master/lifecycle/wait_for_db.py
- Wrap the `CONFIG.y` call with `int()` as when the value is set via environment variable, `CONFIG.y` will return a string.

Starz0r:
Done and done!

For the linter I did run black on my changes but it did not change anything. I assume the problems caught by pylint are the long strings, but it also didn't give any suggestions, so I'll assume that how it is currently is fine.

BeryJu:
Cheers for the fixes, if I remember correctly, the line length actually has to be updated manually as black won't do this, you can just move the `{int(CONFIG.y('redis.port'))}` to the next line before the `/`

PR #1138, opened 2021-07-13
LDAP Provider: Added auto-generated uidNumber and guidNumber generated attributes for use with SSSD and similar software.

Rynoxx:
The starting number for uid/gid can be configured iva environtment
variables and is by default 2000 which should work fine for most instances unless there are more than
999 local accounts on the server/computer.

The uidNumber is just the users Pk + the starting number.
The guidNumber is calculated by the last couple of bytes in the uuid of
the group + the starting number, this should have a low enough chance
for collisions that it's going to be fine for most use cases.

I have not added any interface stuff for configuring the environment variables as I couldn't really find my way around all the places I'd have to edit to add it and the default values should in my opinion be fine for 99% use cases.


outpost/pkg/ldap/instance_search.go
```diff
@@ -137,3 +159,32 @@ func (pi *ProviderInstance) GroupEntry(g api.Group) *ldap.Entry {
...
+			Name:   "cn",
+			Values: []string{u.Username},
+		},
+		{
+			Name:   "uid",
+			Values: []string{string(u.Uid)},
+		},
+		{
+			Name:   "objectClass",
+			Values: []string{GroupObjectClass, "goauthentik.io/ldap/group"},
```
BeryJu, line 177: Maybe we can add another objectClass to indicate that this is a fake group, like `goauthentik.io/ldap/virtual-group`?

outpost/pkg/ldap/utils.go
```diff
@@ -84,3 +89,31 @@ func (pi *ProviderInstance) GetUserDN(user string) string {
...
+
+func (pi *ProviderInstance) GetRIDForGroup(uid string) int32 {
+	data, err := hex.DecodeString(strings.Replace(uid, "-", "", -1))
+
+	if err != nil {
+		panic(err)
+	}
+
+	dataSize := len(data);
+
```
BeryJu, line 109: No quite sure how the logic works, wouldn't it be easier to convert the UUID to its integer form and use the last few digits?

outpost/cmd/ldap/server.go
```diff
@@ -19,7 +19,12 @@ const helpMessage = `authentik ldap
 Required environment variables:
 - AUTHENTIK_HOST: URL to connect to (format "http://authentik.company")
 - AUTHENTIK_TOKEN: Token to authenticate with
-- AUTHENTIK_INSECURE: Skip SSL Certificate verification`
+- AUTHENTIK_INSECURE: Skip SSL Certificate verification
+
+Additional environment variables:
```
BeryJu, line 24: This should probably be removed

BeryJu (commented):
LGTM so far, for the newly added fields you need to run `./manage.py makemigrations` to create the database migrations, and then run `make lint-fix` to ensure the formatting is right.

The other thing that could still be done is adding the new fields to the Admin UI in https://github.com/goauthentik/authentik/blob/master/web/src/pages/providers/ldap/LDAPProviderForm.ts, but I can also do that afterwards.

outpost/pkg/ldap/instance_search.go
Rynoxx: Instead of `goauthentik.io/ldap/group` or in addition to?

BeryJu: I'd say in addition.

Actually, on further inspection, wouldn't it be possible to use `pi.GroupEntry()` and pass and a custom Group object?

Rynoxx: Not a bad idea, I'll do that

outpost/pkg/ldap/utils.go
Rynoxx: Yep, it was a bit overcomplicated. I had grand plans for generating the gidNumber at first but scrapped 90% of it, this roundabout way is the scraps from that.

Doing it a simpler way now via hex to bigint conversion to just get the last 5 digits

outpost/pkg/ldap/instance_search.go
Rynoxx: I've ran makemigrations, lint-fix and added the fields to the LDAPProviderForm now

PR #1143, opened 2021-07-14
providers/ldap: fix: dn and member fields for virtual groups

Rynoxx:


outpost/pkg/ldap/utils.go
```diff
@@ -94,14 +94,12 @@ func (pi *ProviderInstance) APIGroupToLDAPGroup(g api.Group) LDAPGroup {
 }
 
 func (pi *ProviderInstance) APIUserToLDAPGroup(u api.User) LDAPGroup {
-	dn := fmt.Sprintf("cn=%s,%s", u.Username, pi.UserDN)
-
 	return LDAPGroup{
-		dn:				dn,
+		dn:				fmt.Sprintf("cn=%s,%s", u.Username, pi.GroupDN),
```
BeryJu, line 98: Any reason to not use `pi.GetUserDN` here too?

Rynoxx: Well, since it's a technically a group we shouldn't be using the users dn for the record, right?
However, I should probably rewrite the GetGroupDN function to accept the string name (like GetUserDN) and use that instead of Sprintf

BeryJu: Yeah I just noticed that I misread, my bad

Rynoxx: Refactored `pi.GetGroupDN` is now commited as well

PR #1147, opened 2021-07-15
root: optional TLS support on redis connections

Starz0r:


authentik/root/settings.py
```diff
@@ -188,11 +188,16 @@ def j_print(event: str, log_level: str = "info", **kwargs):
     "DEFAULT_SCHEMA_CLASS": "drf_spectacular.openapi.AutoSchema",
 }
 
+REDIS_PROTOCOL_PREFIX = "redis://:"
+if bool(CONFIG.y("redis.tls")):
+    REDIS_PROTOCOL_PREFIX = "rediss://:"
```
BeryJu, line 193: could be reformatted as `REDIS_PROTOCOL_PREFIX = "rediss://:" if CONFIG.y_bool("redis.tls", False) else "redis://:"` to prevent the coverage warnings

authentik/root/settings.py
```diff
@@ -188,11 +188,16 @@ def j_print(event: str, log_level: str = "info", **kwargs):
...
+
 CACHES = {
     "default": {
         "BACKEND": "django_redis.cache.RedisCache",
         "LOCATION": (
-            f"redis://:{CONFIG.y('redis.password')}@{CONFIG.y('redis.host')}:"
+            REDIS_PROTOCOL_PREFIX
+            + f"{CONFIG.y('redis.password')}@{CONFIG.y('redis.host')}:"
```
BeryJu, line 200: Please add `REDIS_PROTOCOL_PREFIX` to the f-string instead of concatenating it.

lifecycle/wait_for_db.py
```diff
@@ -40,10 +40,14 @@ def j_print(event: str, log_level: str = "info", **kwargs):
         sleep(1)
         j_print(f"PostgreSQL Connection failed, retrying... ({exc})")
 
+REDIS_PROTOCOL_PREFIX = "redis://:"
```
BeryJu, line 43: Same here as above

authentik/root/settings.py
Starz0r: I assume you mean `CONFIG.y_bool` and not merging those two lines into a single one?

authentik/root/settings.py
Starz0r: Good catch.

authentik/root/settings.py
```diff
@@ -188,11 +188,16 @@ def j_print(event: str, log_level: str = "info", **kwargs):
     "DEFAULT_SCHEMA_CLASS": "drf_spectacular.openapi.AutoSchema",
 }
 
+REDIS_PROTOCOL_PREFIX = "rediss://"
+if CONFIG.y_bool("redis.tls", False):
+    REDIS_PROTOCOL_PREFIX = "redis://"
```
BeryJu, line 193: I think the values for this are swapped around, the `y_bool` call returns a bool and the second argument is only a default, not a comparison

PR #1246, opened 2021-08-10
Update translation.md

ghost:
Instructions were unclear; took a while for me to figure it out.

# Details
Resolves #

### New Features

### Breaking Changes

## Additional

### DIFF
--- a/website/developer-docs/translation.md
+++ b/website/developer-docs/translation.md
@@ -22,6 +22,8 @@ If you want to translate the frontend to a new language, ensure the language cod
     // [...]
 ```
 
+You should have `@lingui/cli` and `@babel/core` installed in your home directory (See [LinguiJS Docs](https://lingui.js.org/tutorials/setup-react.html) for more information).
+
 Afterwards, run `npx lingui extract` to generate a base .po file.
 
 The .po files can be edited by any text editor, or using a tool such as [POEdit](https://poedit.net/).


BeryJu:
Hi, thanks for the PR! The translation setup *should* work (in theory) without having to install lingui and babel in your home directory, running `npm i` in the `/web/` folder should be enough, but I'll do some testing.

BeryJu:
@vrifox Try this:

```
In the root of the cloned repository, run `make gen-web` to generate the API Client used by the webinterface.

Afterwards, run `npm i` in the `/web` folder to install all dependencies.
```

vrifox:
If I run `make gen-web` in the root directory, it tries to run a docker container. Is that really necessary to translate the Frontend? 🤔

Thanks for the quick response @BeryJu. 🙂

BeryJu:
So it indirectly does. When you run npm i to install dependencies, it complains about the missing API client since that's a direct dependency, and of course you also need the API client to test your changes. Although it should be possible to generate a PO file without doing all of that, I agree.

BeryJu:
Alternatively you can also use transifex: https://www.transifex.com/beryjuorg/authentik/dashboard

vrifox:
Thank you for the link. I’ve tried Transifex, but it does not work for me. I simply can’t edit anything, do you have to authorize me?
And should I close this pull request or are you doing it? I’m not firm with Github. ^^

BeryJu:
Yeah so I have to authorize people to contribute on transifex but it doesn't send me an email notification when people request access so I do miss it quite a lot. I've just accepted the request to join, thanks for your help already! I'll close the PR for now since theres now instructions to do it fully locally, but yeah transifex is probably the easier way.

PR #1470, opened 2021-09-25
🐛 Fix failing traefik healthcheck

pemontto:
# Details
This fixes an issue where Traefik would not forward traffic to a proxy outpost. Traefik would perform a healthcheck on port 9000 and get a `400 Bad Request` and consider the backend/service down.

### New Features
* Updates the docker container labels for traefik to align to the healthcheck, that is `*:9300/akprox/ping`


pemontto:
Accidentally force pushed with no changes when fixing the linting issue.

PR #1490, opened 2021-09-28
Use fully qualified names to get base images.

asteven:
Signed-off-by: Steven Armstrong <steven@armstrong.cc>

# Details

The current authentik Dockerfile does not work out of the box with podman.
```
podman build --pull \
   --tag docker.io/asteven/authentik:2021.9.3 \
   --tag docker.io/asteven/authentik:latest .
[1/5] STEP 1/5: FROM python:3.9-slim-buster AS locker
[2/5] STEP 1/4: FROM node AS website-builder
Error: error creating build container: short-name "python:3.9-slim-buster" did not resolve to an alias and no unqualified-search registries are de
fined in "/etc/containers/registries.conf"
```

Additionally, and more importantly, using short names to pull container images is potentially dangerous as explained in this [article from redhat](https://www.redhat.com/en/blog/be-careful-when-pulling-images-short-name).

## Changes
Prefix all docker images with "docker.io/"

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # Stage 1: Lock python dependencies
-FROM python:3.9-slim-buster as locker
+FROM docker.io/python:3.9-slim-buster as locker
 
 COPY ./Pipfile /app/
 COPY ./Pipfile.lock /app/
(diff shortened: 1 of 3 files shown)


BeryJu:
Cheers for the PR, could you also add the prefix to the other two dockerfiles in the root of the repo?

asteven:
> Cheers for the PR, could you also add the prefix to the other two dockerfiles in the root of the repo?

Sure. Done. Just force pushed to the branch.

Issue #1476, opened 2021-09-26
[FR] SMS authentication support via SMTP or Twilio API


samip5:
**Is your feature request related to a problem? Please describe.**
It's related to an problem which happens when I want to have 2nd stage after email verification to check that user has an phone number in a specific country.

**Describe the solution you'd like**
I would like to be able to create flows where one stage is an SMS country code match and if it does match send message. 

**Describe alternatives you've considered**
KeyCloack, but I'm not a fan.

**Additional context**
Matrix server that is open to specific geographic only

PR #1595, opened 2021-10-11
New SMS provider, aka wrapper for outside API

samip5:
# Details
Extends #1476

### New Features
* Adds feature which allows Authentik to call an outside API for SMS TOTP.

## Additional
Currently not sure how I can get the static files updated that it serves.


BeryJu:
can you try setting ```yaml
web:
  load_local_files: true
``` in your local.env.yml? Think I just missed adding that in the docs cause thats a relatively new setting

samip5:
> > > > How would I show the options based on the provider? (Not used to your frontend framework.)
> > > 
> > > 
> > > you can just replace this part with an expression based on which provider is selected
> > > ```js
> > > label=${t`Twilio Account SID`}
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > you can also leave the label as is, I don't mind adding that later myself
> > 
> > 
> > Could you show an example of such expression? I want to at least change the generic ones to that.
> 
> For example, you can do
> 
> ```
> label=${this.instance?.provider === ProviderEnum.Twilio  ? t`Twilio foo` : t`Something foo`}}
> ```
> 
> which is not really that clean, I'd probably add another function somewhere to switch or w/e

It doesn't even like that syntax.

samip5:
I think it's ready for review, I know that docs haven't been updated.

BeryJu (changes requested):

PR #1754, opened 2021-11-08
docs: Add additionalHeaders to attributes list

hexeth:
Added additional headers with example of usage to docs

### DIFF
--- a/website/docs/user-group/user.md
+++ b/website/docs/user-group/user.md
@@ -22,7 +22,12 @@ See [Troubleshooting access problems](../troubleshooting/access.md), when set, t
 
 ### `additionalHeaders`:
 
-Some applications can be configured to create new users using header information forwarded from Authentik. You can forward additional header information by adding each header underneath `additionalHeaders`:
+:::info
+This field is only used by the Proxy Provider.
+:::
+
+Some applications can be configured to create new users using header information forwarded from Authentik. You can forward additional header information by adding each header 
+underneath `additionalHeaders`:
 
 #### Example:
 ```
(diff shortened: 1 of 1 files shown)


website/docs/user-group/user.md
```diff
@@ -20,6 +20,20 @@ Optional flag, when set to false, Tokens created by the user will not expire.
 
 See [Troubleshooting access problems](../troubleshooting/access.md), when set, the user gets a more detailed explanation of access decisions.
 
+### `additionalHeaders`:
+
```
BeryJu, line 24: Could you make it a bit clearer that this field is only used with the proxy provider?

For example
```
:::info
This field is only used by the Proxy Provider.
:::
```

website/docs/user-group/user.md
```diff
@@ -20,6 +20,20 @@ Optional flag, when set to false, Tokens created by the user will not expire.
...
+#### Example:
+```
+additionalHeaders:
+  REMOTE-USER: joe.smith
+  REMOTE-EMAIL: joe@jsmith.com
+  REMOTE-NAME: Joseph
+```
+
+These headers will now be passed to the application when the user logs in. Most applications will need to be configured to accept these headers. Some examples of applications that accept Auth Proxy are [Grafana](https://grafana.com/docs/grafana/latest/auth/auth-proxy/) and [Tandoor Recipes](https://docs.tandoor.dev/features/authentication/).
```
BeryJu, line 35: Very minor nitpick, but just for consistency, replace `Auth Proxy` with `Proxy Provider` or `Proxy Outpost`, as that's how its referecned in the UI and the rest of the docs.

BeryJu (changes requested):

website/docs/user-group/user.md
hexeth: done

PR #1773, opened 2021-11-11
Fixed the incorrect padding issue in apple.py

rizwan95:
Fixed the incorrect padding issue in apple.py by adding proper padding to the raw_payload.

# Details
Yes

### New Features
No

### Breaking Changes
No

## Additional
Fixed the incorrect padding issue in apple.py by adding proper padding to the raw_payload.

### DIFF
--- a/authentik/sources/oauth/types/apple.py
+++ b/authentik/sources/oauth/types/apple.py
@@ -50,7 +50,9 @@ class AppleOAuthClient(OAuth2Client):
     def get_profile_info(self, token: dict[str, str]) -> Optional[dict[str, Any]]:
         id_token = token.get("id_token")
         _, raw_payload, _ = id_token.split(".")
-        payload = loads(b64decode(raw_payload.encode().decode()))
+        padded_payload = raw_payload + "="*divmod (len (raw_payload) , 4)
+        base64decoded_payload = b64decode (padded_payload.encode().decode())
+        payload = loads(base64decoded_payload)
         return payload
 
 
--- a/authentik/sources/oauth/types/apple.py
+++ b/authentik/sources/oauth/types/apple.py
@@ -5,6 +5,7 @@ from time import time
 from typing import Any, Optional
 
 from jwt import encode
+from jwt import decode
 from structlog.stdlib import get_logger
 
 from authentik.sources.oauth.clients.oauth2 import OAuth2Client
@@ -49,10 +50,7 @@ class AppleOAuthClient(OAuth2Client):
 
     def get_profile_info(self, token: dict[str, str]) -> Optional[dict[str, Any]]:
         id_token = token.get("id_token")
-        _, raw_payload, _ = id_token.split(".")
-        padded_payload = raw_payload + "="*divmod (len (raw_payload) , 4)
-        base64decoded_payload = b64decode (padded_payload.encode().decode())
-        payload = loads(base64decoded_payload)
+        decode(id_token,options={"verify_signature": False})
         return payload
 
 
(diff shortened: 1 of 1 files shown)


BeryJu:
Is there any way to reproduce the issue this PR fixes?

Also after looking over the code again, it might be easier to just 

```python
from jwt import decode
[ .... ]

return decode(id_token)
``` instead of this

rizwan95:
> Is there any way to reproduce the issue this PR fixes?
> 
> Also after looking over the code again, it might be easier to just
> 
> ```python
> from jwt import decode
> [ .... ]
> 
> return decode(id_token)
> ```
> 
> instead of this

Yes you can just configure an Apple source provider and sign in using it. In the last step, it will fail telling that "Incorrect padding" 

As you suggested, your solution is more elegant and makes much sense.

BeryJu:
I don't remember which information is included in Apple's ID Token but I'd like to add some tests for this. Could you check with https://jwt.io/ which fields are included, and if there's anything personal besides Name?

rizwan95:
> I don't remember which information is included in Apple's ID Token but I'd like to add some tests for this. Could you check with https://jwt.io/ which fields are included, and if there's anything personal besides Name?

Yes sure!

rizwan95:
> I don't remember which information is included in Apple's ID Token but I'd like to add some tests for this. Could you check with https://jwt.io/ which fields are included, and if there's anything personal besides Name?
@BeryJu 
So here is the payload 

`{
  "iss": "https://appleid.apple.com",
  "aud": "<redacted>",
  "exp": 1636740983,
  "iat": 1636654583,
  "sub": "001786.d770f0696ea64f48a9c7b834bf536143.0359",
  "at_hash": "EKTUgiJwIUCtE65XXh0d_g",
  "email": "<redacted>",
  "email_verified": "true",
  "auth_time": 1636654580,
  "nonce_supported": true
}`

I have redacted the aud and email values

BeryJu:
Alright, nothing special in there. I'll merge it with the `jwt` `decode()` function and we'll see about tests later

rizwan95:
> Alright, nothing special in there. I'll merge it with the `jwt` `decode()` function and we'll see about tests later

Awesome!

rizwan95:
@BeryJu 
I have also fixed an another issue which was causing incorrect encoding of the client_secret. 
The "sub" must be only the client id, so I have changed self.source.consumer_key to parts[0] in oauth/types/apple.py

BeryJu:
> @BeryJu I have also fixed an another issue which was causing incorrect encoding of the client_secret. The "sub" must be only the client id, so I have changed self.source.consumer_key to parts[0] in oauth/types/apple.py

Cheers, can you add the `from jwt import decode` part? Then I'll merge the PR

rizwan95:
> > @BeryJu I have also fixed an another issue which was causing incorrect encoding of the client_secret. The "sub" must be only the client id, so I have changed self.source.consumer_key to parts[0] in oauth/types/apple.py
> 
> Cheers, can you add the `from jwt import decode` part? Then I'll merge the PR

I tried to implement the decode part (I have very little knowledge about Python)
I am getting this error ```["<class 'jwt.exceptions.InvalidAlgorithmError'>", "InvalidAlgorithmError('The specified alg value is not allowed')```

for the code 
```decode(id_token,self.source.consumer_secret,algorithms=['ES256'])```

not sure what mistake I am doing

BeryJu:
you can just do `return decode(id_token)` since we're not validating the JWT there anyways

rizwan95:
> you can just do `return decode(id_token)` since we're not validating the JWT there anyways

I have added it. I suppose that should be enough.

PR #1796, opened 2021-11-15
providers/radius: simple radius outpost

BeryJu:
A simple radius outpost that only supports username/password authentication (using the same flow executor as the ldap outpost), no support for CHAP as that would require reversible password storage, no support for TLS as there are multiple ways to implement that but none are supported by the current go library


authentik/providers/radius/models.py
```diff
@@ -0,0 +1,59 @@
...
+class RadiusProvider(OutpostModel, Provider):
+    """Allow applications to authenticate against authentik's users using Radius."""
+
+    shared_secret = models.TextField(
+        default=generate_key,
+        help_text=_("Shared secret between clients and server to hash packets."),
+    )
+
+    client_networks = models.TextField(
+        default="0.0.0.0/0",
```
samip5, line 23: Is that going to be both IPv6 and IPv4? or only IPv4?

ferferga:
@BeryJu Care to explain why this has been closed? Was eagerly waiting for this :(

BeryJu:
> @BeryJu Care to explain why this has been closed? Was eagerly waiting for this :(

I was just renaming some branches for cleanup, I forgot that closes the PR for it too.

Also this probably wont happen for a while since all radius implementations that use anything but the very basic features require access to unhashed passwords somewhere wihch I dont want to save.

MrSuicideParrot:
@BeryJu  could you share with us what is the status of the Radius outpost? What are the plans of authentik regarding it?

I've spend a week trying to configure freeradius and authentik with the ldap outpost for a VPN with DUO, and realized that working with ldap is two limiting. Even though you are only implementing clear text passwords, I think that it would be better than having to rely on the ldap outpost.

cwildfoerster:
Would love to see radius support too.

PR #1831, opened 2021-11-22
website/integrations: Add Provider/Uptime Kuma

larsl-net:
# Details
Adds page to the website with instructions for configuring Uptime Kuma

### New Features
* Instructions for setting up Uptime Kuma with the Proxy Provider

### DIFF
--- a/website/integrations/services/uptime-kuma/index.md
+++ b/website/integrations/services/uptime-kuma/index.md
@@ -52,4 +52,4 @@ To avoid that all users get admin access to Uptime Kuma create a group in authen
 
 Disble auth from Uptime Kuma, go to `Settings` > `Advanced` > `Disable Auth`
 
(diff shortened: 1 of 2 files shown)


slimshizn:
Anyway to add this while using Uptime-Kuma's integrated cloudflared setting? Thank you.

larsl-net:
@slimshizn If I understand the question correctly, you want to run Uptime Kuma which is hosted behind Cloudflare tunnels with authentik authentication.

You either have to build the authentication as in the authentik doc and cannot use the internal tunnels integration of uptime kuma, because here the traffic goes directly to Uptime Kuma and not to the authentik proxy.
Alternatively, according to Cloudflare docs, you can use your own IDP for the login provided by Cloudflare, without the authnetik proxy, in this case you probably only need to connect authentik to Cloudflare and can use the Uptime Kuma tunnels integration.

https://developers.cloudflare.com/cloudflare-one/identity/idp-integration/generic-oidc/

I don't use Cloudflare tunnels myself, so this is just how I understand it or assume it should work. Hope this helps you anyway.

PR #1972, opened 2021-12-20
Updated Gitea Integration

justsem:
Described a fix to a situation where Gitea might require an additional OIDC mapping in order to make the authentication flow function properly.

# Details
There's no active issue yet that I'm aware of.

### New Features
* N/A

### Breaking Changes
* N/A

## Additional
This change allows some users to work around an Gitea issue where Gitea expects an non-existing OIDC mapping. In this case we work around it by creating a new 'Gitea' Scope Mapping providing the data Gitea wants and configuring both Authentik and Gitea to use said scope.


BeryJu:
this shouldn't be required if you set Gitea's `OPENID_CONNECT_SCOPES` to `openid email profile`, and select all those mappings in the provider, no?

justSem:
> this shouldn't be required if you set Gitea's `OPENID_CONNECT_SCOPES` to `openid email profile`, and select all those mappings in the provider, no?

In my particular case that somehow didn't work (Gitea kept throwing the error no matter in which order/format I entered the scope list). Creating a custom scope worked in an instant.
Since I saw Gitea Issues resembling this issue with no clear solution I figured including this as optional might be useful.

BeryJu:
> > this shouldn't be required if you set Gitea's `OPENID_CONNECT_SCOPES` to `openid email profile`, and select all those mappings in the provider, no?
> 
> In my particular case that somehow didn't work (Gitea kept throwing the error no matter in which order/format I entered the scope list). Creating a custom scope worked in an instant.

Did gitea actually request the correct scopes during authorization?

> Since I saw Gitea Issues resembling this issue with no clear solution I figured including this as optional might be useful.

Yeah fair enough, I'd just rather fix it and document it properly for all setups

justSem:
> Did gitea actually request the correct scopes during authorization?

This was unclear to me - Gitea's logs said yes, Authentik's said no. - I stumbled upon the solution when I started experimenting with the values to asses what it would change in Gitea's behaviour.


> Yeah fair enough, I'd just rather fix it and document it properly for all setups

Agreed. But regardless of this error, a mapping like this should work for everyone (or I'm overlooking something) - so an user deciding to create that mapping won't break his setup with it either. (And I've mentioned it's optional of course :) )

BeryJu:
So I just did some testing with a temporary gitea install, and stock it behaves like you described (only requesting the openid scope).

However after I added 
```
[oauth2_client]
OPENID_CONNECT_SCOPES = email profile
``` to the config, it worked just as expected, with all the stock mappings.

The main reason why I am against including this workaround is it basically contains the same code as the existing property mappings, however the existing ones are managed by authenitk and will automatically be upgraded, whilst this one won't be and might cause issues in the future.

**Comment by justSem** [comment]
I've spun up a new installation and can confirm your findings - I probably messed something up elsewhere as this doesn't seem to work on my original installation.

I can update my commit later today to only alter the app.ini to request the proper scopes.

**Comment by justSem** [comment]
@BeryJu I've updated the Integrations page using your resolution, could you please check if it's fine now?

File: website/integrations/services/gitea/index.md [hunk 1]
```diff
@@ -62,6 +62,45 @@ Change the following fields
 - Icon URL: https://raw.githubusercontent.com/goauthentik/authentik/master/web/icons/icon.png
 - OpenID Connect Auto Discovery URL: https://authentik.company/application/o/gitea-slug/.well-known/openid-configuration
 
-![](./gitea1.png)
+
+(./gitea1.png)
```
BeryJu, line 66: can you undo this change?

website/integrations/services/gitea/index.md
```diff
@@ -62,6 +62,45 @@ Change the following fields
...
 
 `Add Authentication Source` and you should be done. Your Gitea login page should now have a `Sign in With` followed by the authentik logo which you can click on to sign-in to Gitea with Authentik creds.
+
+
+
+
+### Step 4 (Optional)
```
BeryJu, line 73: I wouldn't add this as 4th step but rather as part of the gitea setup under step 3. Imo it's not optional (since a stock install won't work without this), so I'd say just add the `In your Gitea instance, navigate to your app.ini and make the following changes` block.

BeryJu (changes requested):

justSem:
Done

PR #2144, opened 2022-01-24
outposts/ldap: Fix more case sensitivity issues.

### LINKED ISSUE #2120: LDAP Server (outpost): cannot browse LDAP tree from root.
When connecting to the ldap server with a client none of the children of the root CN=authentik... is accessible. Clients provide different errors:
- ldapadmin.org client "LDAP error: Operations Error!"
- Apache Directory Studio: 

>  "Error while opening connection
>  Missing schema location in RootDSE, using default schema.
> Missing schema location in RootDSE, using default schema."

The issue seems to be the RootDSE missing (i don't know what that means) and client falling back to a default DSE.

Connecting to children (CN=user, CN=groups) works but the error messages are still there.

ikogan:
# Details
Resolves #2120

## Changes
* All LDAP comparisons are now done using case insensitive methods.

### Breaking Changes
* Multiple LDAP providers with Base DNs that only differ by case will break.
   Which provider a given LDAP query will use when providers differ only in case undefined.

### DIFF
--- a/internal/outpost/ldap/bind/direct/direct.go
+++ b/internal/outpost/ldap/bind/direct/direct.go
@@ -16,6 +16,7 @@ import (
 	"goauthentik.io/internal/outpost/ldap/flags"
 	"goauthentik.io/internal/outpost/ldap/metrics"
 	"goauthentik.io/internal/outpost/ldap/server"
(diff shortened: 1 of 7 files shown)


BeryJu (approved):
LGTM

PR #2260, opened 2022-02-08
Enable 'secure' option for pwgen

lyh16:
# Details
Resolves #

### New Features
* Enables "-s"(secure) option in pwgen

### Breaking Changes
* None

## Additional
As per the [pwgen manual](https://linux.die.net/man/1/pwgen), the "-s"(secure) option instructs pwgen to generate completely random passwords, where as the default for pwgen is to generate more memorable passwords. Since the passwords generated in this part of the installation process are to be "remembered" by the dot env file, I believe that users may benefit from the additional entropy provided by the "-s" option in pwgen.

EDIT: Deleted unnecessary commas.


BeryJu:
TIL about this option, could you add it for the other pwgen examples too?

lyh16:
> TIL about this option, could you add it for the other pwgen examples too?

Done!

PR #2368, opened 2022-02-23
bad spelling

thomaslochet:
# Details
Resolves #

### New Features

### Breaking Changes

## Additional


BeryJu:
Please fix translation on transifex, as these files are read from and will be overwritten by transifex; https://www.transifex.com/beryjuorg/authentik/dashboard/

PR #2397, opened 2022-02-26
customisable user settings

### LINKED ISSUE #2121: Ability to attach flows to the "Update details" page fields.
If this is something that is already considered feel free to close this.

Currently any user has the ability to change their username, display name and email. This is fine if you trust your users,
however it can be a problem if not, because no validation (apart from checking if the new username already exists) is applied.

I am aware that the editing of those fields can be disabled. 

For fields like the username, editing can be safely disabled, since users are unlikely to want to change that anyway and changing it may even break compatibility with apps which rely on the username to authenticate the user. On the other hand, fields like the email one should be editable by the user in case they want to change their email. However, since the email field cannot be attached to a flow, no validation is performed. Thus, the user may input an invalid email (or simply misspell it). 

Similarly to the "change password" flow that can be attached to the change password button, I would like the ability to attach flows to each of the fields in the "Update details" page - username, display name and email.

Only 1 alternative I am aware of - disable editing of all 3 fields and make users ask an administrator to change it for them while exercising their own judgment to asses the validity of the new information and the consequences it might bring.

This is obviously not ideal, as it requires manual intervention, thus it does not scale well. Also it is insecure. Unless the change of details request has come from the currently registered user email or the administrator and the user know each other, it is hard to be absolute sure the request has actually come from the user.

**Additional context**
N/A

### LINKED ISSUE #2338: How to customize the user settings panel fields
**Describe your question/**
In the user settings, a user can change their username, email etc. But how can I change/add/delete fields there? They have nothing to do with flows, right?

For example, the username field said: "Required. 150 characters or fewer. Letters, digits and @/./+/-/_ only.". How can I change the verification rule? I have username policy but seems can only apply to a stage

I can set more strict username rules for enrollment flow, but after that, users can always change their username here, which shouldn't pass my rules

**Relevant infos**
No

**Screenshots**
image

**Logs**
Not relevant

 - authentik version: 2022.2.1
 - Deployment: docker-compose

**Additional context**
No

BeryJu:
Signed-off-by: Jens Langhammer <jens.langhammer@beryju.org>

closes #2338
closes #2121 

- [x] Remove update_self api endpoint
- [x] Add minimal flow executor in user interface
- [x] Add ability for prompts to use dynamic placeholders
  - [x] Add tests
  - [x] Add UI to edit expressions
  - [x] Add docs for placeholders
- [x] Ensure backwards compat by still checking user attributes (remove later)
- [x] Add custom input types for inputs which use custom UI logic
  - [x]  ak-locale input type (to replace locale selection)


sevmonster:
This should also close https://github.com/goauthentik/authentik/issues/2121

PR #2406, opened 2022-02-28
added example for custom user attributes

### LINKED ISSUE #2405: Custom User attribute in Scope Mapping
We wanted to add a custom attribute to the access token. Unfortunately we searched for a long time because we could not find an example in the documentation.

We created a pull request for the documentation.

Add  a simple example to the documentation.

**Additional context**
In the multi-tenant environment, it is desirable to include additional attributes such as the tenant ID in the access token.

CRoudsarabi:
# Details
Resolves #2405 

### New Features
* extend documentation

### DIFF
--- a/website/docs/expressions/_user.md
+++ b/website/docs/expressions/_user.md
@@ -3,8 +3,7 @@
 Example:
 
 ```python
-if not 'custom_attribute' in request.user.attributes:
-  return {'custom_attribute': 'default'}
-
-return {"custom_attribute": request.user.attributes['custom_attribute']}
+return {
+  "custom_attribute": request.user.attributes.get("custom_attribute", "default"),
+}
 ```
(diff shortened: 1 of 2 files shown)


website/docs/expressions/_user.md
```diff
@@ -0,0 +1,10 @@
+- `user`: The current user. This may be `None` if there is no contextual user. See ([User](../user-group/user.md#object-attributes))
+
+Example:
+
+```python
+if not 'custom_attribute' in request.user.attributes:
+  return {'custom_attribute': 'default'}
+
+return {"custom_attribute": request.user.attributes['custom_attribute']}
```
BeryJu, line 9: it would probably be cleaner to use something like 

```python
return {
  "custom_attribute": request.user.attributes.get("custom_attribute", "default"),
}
```

I'd also add `user.group_attributes()`, which returns a dict of all the user's group's attributes, merged

PR #2447, opened 2022-03-07
Avoiding NIL entries

a17t:
Hello,

when fetching objects via ldap, the ldap outpost ran into a NIL error in one of the used libraries, due to NIL being added to the entries.
I am not sure, whether this PR fixes the issue in a ldap standard conformant way, but the logic checking for IncludeObjectClass in the outer if and again in utils.GetContainerEntry seems redundant to me.

### DIFF
--- a/internal/outpost/ldap/search/direct/direct.go
+++ b/internal/outpost/ldap/search/direct/direct.go
@@ -252,7 +252,7 @@ func (ds *DirectSearcher) Search(req *search.Request) (ldap.ServerSearchResult, 
 	if scope >= 0 && (strings.EqualFold(req.BaseDN, ds.si.GetBaseDN()) || utils.HasSuffixNoCase(req.BaseDN, ds.si.GetBaseVirtualGroupDN())) {
 		singlevg := utils.HasSuffixNoCase(req.BaseDN, ","+ds.si.GetBaseVirtualGroupDN())
 
-		if !singlevg || utils.IncludeObjectClass(filterOC, constants.GetContainerOCs()) {
+		if !singlevg && utils.IncludeObjectClass(filterOC, constants.GetContainerOCs()) {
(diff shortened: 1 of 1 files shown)


ikogan:
I believe there's a reason for that... I'll try to have a look soon but I'm traveling for the week so I can't promise...

a17t:
> Could you let us know some more details about what data caused the nil and where?

I'll do my best, unfortunately the content of my instance changed and I cannot reproduce the exact behaviour anymore.

One of the arguments passed to  _utils.GetContainerEntry_ in line 256 of direct.go was nil. If I remember correctly it _filterOC_ .

As singlevg is false the original "or" check in line 255 does not evaluate _utils.IncludeObjectClass_ . This results in a NIL exception in _GetContainerEntry_

Currently this change anyway fixes another issue for me.

Without that change, when accessing the virtual groups like so:
_ldapsearch -H ldap://127.0.0.1:3389 -x -D "cn=***,ou=users,dc=***,dc=***" -W -b "ou=virtual-groups,dc=***,dc=***" -s one -a always -z 1000 "(objectClass=*)" "hasSubordinates" "objectClass"_

the groups are returned empty (just dn and objectclass). 

Applying the change makes the groups being returned. (Issues connecting sssd using ldap were btw. the motivation for me to look into that).

Hopes this helps

ikogan:
Hrm, I'm not sure how `filterOC` can ever become `nil`... `GetFilterObjectClass` and `parseFilterObjectClass` don't seem to have a way to generate a nil.

Anyway, looking at all of the other checks, it seems like they're `&&` so I think this PR should be good. We could probably remove the `IncludeObjectClass` check inside of `GetContainerEntry` and we wouldn't need to pass filterOC as well.

BeryJu:
Did some research too, seems like this fixes #2526 too. I'll merge this and add some tests

PR #2492, opened 2022-03-15
Update v2022.3.md

BobSilent:
minor yaml fix: update to correct version

# Details
Resolves #

### New Features

### Breaking Changes

## Additional


BeryJu:
Thanks for the PR, but I purposefully don't update the image version in the release notes (mostly because its quite annoying to automate), but also I want to switch back to having matched helm releases for authentik releases and then this section won't be needed anymore.

PR #2497, opened 2022-03-16
providers/oauth2: impl `/user/teams` endpoint for Github OAuth2

Electromaster232:
### New Features
This commit adds a functional `/user/teams` endpoint for the emulated Github OAuth2 service.
The teams a user is part of are based on the user's groups in Authentik.

Team slugs are generated based off of the team's name converted to lowercase and replacing spaces with dashes. This was done to more closely match the Github API. If you'd like to see any changes, please let me know!

### DIFF
--- a/authentik/providers/oauth2/views/github.py
+++ b/authentik/providers/oauth2/views/github.py
@@ -1,8 +1,8 @@
 """authentik pretend GitHub Views"""
-from copy import copy
 
 from django.http import HttpRequest, HttpResponse, JsonResponse
 from django.views import View
+from django.utils.text import slugify
 
 from authentik.providers.oauth2.models import RefreshToken
 
@@ -70,60 +70,58 @@ class GitHubUserTeamsView(View):
         """Emulate GitHub's /user/teams API Endpoint"""
         user = token.user
 
-        org_tpl = {
-            "id": 0,
-            "node_id": "",
-            "url": "",
-            "html_url": "",
-            "name": "",
-            "slug": "",
-            "description": "",
-            "privacy": "",
-            "permission": "",
-            "members_url": "",
-            "repositories_url": "",
-            "parent": None,
-            "members_count": 0,
-            "repos_count": 0,
-            "created_at": "",
-            "updated_at": "",
-            "organization": {
-                "login": "github",
-                "id": 1,
-                "node_id": "",
-                "url": "",
-                "repos_url": "",
-                "events_url": "",
-                "hooks_url": "",
-                "issues_url": "",
-                "members_url": "",
-                "public_members_url": "",
-                "avatar_url": "",
-                "description": "",
-                "name": "Authentik",
-                "company": "",
-                "blog": "",
-                "location": "",
-                "email": "",
-                "is_verified": True,
-                "has_organization_projects": True,
-                "has_repository_projects": True,
-                "public_repos": 0,
-                "public_gists": 0,
-                "followers": 0,
-                "following": 0,
-                "html_url": "",
-                "created_at": "",
-                "updated_at": "",
-                "type": "Organization"
-            }
-        }
-
         orgs_response = []
         for org in user.ak_groups.all():
-            _org = copy(org_tpl)
+            _org = {
+                "id": 0,
+                "node_id": "",
+                "url": "",
+                "html_url": "",
+                "name": "",
+                "slug": "",
+                "description": "",
+                "privacy": "",
+                "permission": "",
+                "members_url": "",
+                "repositories_url": "",
+                "parent": None,
+                "members_count": 0,
+                "repos_count": 0,
+                "created_at": "",
+                "updated_at": "",
+                "organization": {
+                    "login": "github",
+                    "id": 1,
+                    "node_id": "",
+                    "url": "",
+                    "repos_url": "",
+                    "events_url": "",
+                    "hooks_url": "",
+                    "issues_url": "",
+                    "members_url": "",
+                    "public_members_url": "",
+                    "avatar_url": "",
+                    "description": "",
+                    "name": "Authentik",
+                    "company": "",
+                    "blog": "",
+                    "location": "",
+                    "email": "",
+                    "is_verified": True,
+                    "has_organization_projects": True,
+                    "has_repository_projects": True,
+                    "public_repos": 0,
+                    "public_gists": 0,
+                    "followers": 0,
+                    "following": 0,
+                    "html_url": "",
+                    "created_at": "",
+                    "updated_at": "",
+                    "type": "Organization"
+                }
+            }
             _org["id"] = str(org.pk)
-            _org["slug"] = org.name.replace(" ", "-").lower()
+            _org["slug"] = slugify(org.name)
             _org["name"] = org.name
             orgs_response.append(_org)
         return JsonResponse(orgs_response, safe=False)
--- a/authentik/providers/oauth2/views/github.py
+++ b/authentik/providers/oauth2/views/github.py
@@ -1,4 +1,6 @@
 """authentik pretend GitHub Views"""
+from copy import copy
+
 from django.http import HttpRequest, HttpResponse, JsonResponse
 from django.views import View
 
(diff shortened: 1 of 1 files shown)


authentik/providers/oauth2/views/github.py
```diff
@@ -66,4 +68,63 @@ class GitHubUserTeamsView(View):
...
+                "updated_at": "",
+                "type": "Organization"
+            }
+        }
+
+        orgs_response = []
+        for org in user.ak_groups.all():
+            _org = copy(org_tpl)
+            _org["id"] = str(org.pk)
+            _org["slug"] = org.name.replace(" ", "-").lower()
```
BeryJu, line 126: use `from django.utils.text import slugify` for consistency

BeryJu (changes requested):
looks pretty good so far, small nit: I would move the dictionary in the loop to prevent having to copy it, and also so we dont have to set placeholder variables

also it would be nice to have tests for this but I dont think any of the other GitHub-emulating stuff has many tests so dw about that

authentik/providers/oauth2/views/github.py
Electromaster232: I just pushed `c00b27a` to address these. I didn't see any existing tests for the github emulation so I figured it probably wasn't super important to include some. I would be happy to write some if I get a chance at a later date, but that would probably be best for another PR (and also because I don't want this one to sit around if I don't get a chance to write them soon).

Electromaster232:
I just realized I forgot change it setting placeholder values, I'll push another commit

BeryJu (commented):
LGTM, there are some minor complaints from black and pylint

BeryJu:
so the e2e tests actually found a bug:

```
t=2022-03-17T18:53:50+0000 lvl=dbug msg="HTTP GET http://localhost:46689/user/teams?per_page=100: 200 OK [{\"id\": \"a2395088-d76d-46b6-93e4-3d2a61af3a30\", \"node_id\": \"\", \"url\": \"\", \"html_url\": \"\", \"name\": \"ayxpFneQrzFoFdRbEMTS\", \"slug\": \"ayxpfneqrzfofdrbemts\", \"description\": \"\", \"privacy\": \"\", \"permission\": \"\", \"members_url\": \"\", \"repositories_url\": \"\", \"parent\": null, \"members_count\": 0, \"repos_count\": 0, \"created_at\": \"\", \"updated_at\": \"\", \"organization\": {\"login\": \"github\", \"id\": 1, \"node_id\": \"\", \"url\": \"\", \"repos_url\": \"\", \"events_url\": \"\", \"hooks_url\": \"\", \"issues_url\": \"\", \"members_url\": \"\", \"public_members_url\": \"\", \"avatar_url\": \"\", \"description\": \"\", \"name\": \"Authentik\", \"company\": \"\", \"blog\": \"\", \"location\": \"\", \"email\": \"\", \"is_verified\": true, \"has_organization_projects\": true, \"has_repository_projects\": true, \"public_repos\": 0, \"public_gists\": 0, \"followers\": 0, \"following\": 0, \"html_url\": \"\", \"created_at\": \"\", \"updated_at\": \"\", \"type\": \"Organization\"}}]"
t=2022-03-17T18:53:50+0000 lvl=eror msg="login.OAuthLogin(get info from github)" logger=context userId=0 orgId=0 uname= error="Error getting user teams: Error getting team memberships: json: cannot unmarshal string into Go struct field GithubTeam.id of type int"
```

looks like the team ID needs to be int, you can use group.uuid.int to convert it to int automatically (allthough that gives some very big numbers)

Electromaster232:
It does generate some big ints, but I did some testing with the actual github API (their sample response only shows an id of "1" which isn't helpful in this case), and they do return big integers as well. Not as big as this will be, but also not tiny either. I don't think it will be a huge issue.

I can also confirm that with the inclusion of these changes, Authentik is compatible with [Teleport](https://github.com/gravitational/teleport), which only supports Github SSO in their open-source version. I can send in a separate PR with a setup guide for the documentation.

authentik/providers/oauth2/views/github.py
```diff
@@ -66,4 +68,57 @@ class GitHubUserTeamsView(View):
...
+                    "url": "",
+                    "repos_url": "",
+                    "events_url": "",
+                    "hooks_url": "",
+                    "issues_url": "",
+                    "members_url": "",
+                    "public_members_url": "",
+                    "avatar_url": "",
+                    "description": "",
+                    "name": "Authentik",
```
BeryJu, line 105: Sorry for the nit again; https://goauthentik.io/developer-docs/docs/writing-documentation, should be either lowercase `authentik`, or maybe even better, `request.tenant.branding_title` actually, then it's directly user configurable

Electromaster232:
Made that change. I also changed the organization "login" to be the slug of the tenant name ("login" appears to be the first part of the URL, for example `goauthentik` for this repository), instead of it being hardcoded to "github." Let me know if you want it to be something else.

authentik/providers/oauth2/views/github.py
```diff
@@ -90,7 +90,7 @@ def get(self, request: HttpRequest, token: RefreshToken) -> HttpResponse:
                 "created_at": "",
                 "updated_at": "",
                 "organization": {
-                    "login": "github",
+                    "login": slugify(request.tenant.branding_title),
```
BeryJu, line 93: not sure if we should be changing this, do you know what this field is used for?

BeryJu (approved):

BeryJu:
hmmmm 

```
t=2022-03-22T17:36:29+0000 lvl=eror msg="login.OAuthLogin(get info from github)" logger=context userId=0 orgId=0 uname= error="Error getting user teams: Error getting team memberships: json: cannot unmarshal number 111048864482795893176781688470205150957 into Go struct field GithubTeam.id of type int"
```

BeryJu:
According https://raw.githubusercontent.com/github/rest-api-description/main/descriptions/api.github.com/api.github.com.yaml it's only specified as an int so we only have until 2147483647

BeryJu:
alright, you can now use `group.num_pk` to get a short-ish integer

Electromaster232:
Done. Also, for your previous question, the "login" field matches the slugged organization name (see the following actual API response for https://github.com/orgs/electromaster-test/teams/test-2). I figured the branding title is the closest field for the org name (and also what we used for the actual org name field), as it would be more tricky to use the tenant domain as the org name (for example, how would we handle the default tenant?). The "github" value is just something I came up with when making my initial commit.
```  {
        "name": "test-2",
        "id": 5833298,
        "node_id": "T_kwDOBhdjjc4AWQJS",
        "slug": "test-2",
        "description": "test2",
        "privacy": "closed",
        "url": "https://api.github.com/organizations/102196109/team/5833298",
        "html_url": "https://github.com/orgs/electromaster-test/teams/test-2",
        "members_url": "https://api.github.com/organizations/102196109/team/5833298/members{/member}",
        "repositories_url": "https://api.github.com/organizations/102196109/team/5833298/repos",
        "permission": "pull",
        "created_at": "2022-03-23T00:45:17Z",
        "updated_at": "2022-03-23T00:45:17Z",
        "members_count": 1,
        "repos_count": 0,
        "organization": {
            "login": "electromaster-test",
            "id": 102196109,
            "node_id": "O_kgDOBhdjjQ",
            "url": "https://api.github.com/orgs/electromaster-test",
            "repos_url": "https://api.github.com/orgs/electromaster-test/repos",
            "events_url": "https://api.github.com/orgs/electromaster-test/events",
            "hooks_url": "https://api.github.com/orgs/electromaster-test/hooks",
            "issues_url": "https://api.github.com/orgs/electromaster-test/issues",
            "members_url": "https://api.github.com/orgs/electromaster-test/members{/member}",
            "public_members_url": "https://api.github.com/orgs/electromaster-test/public_members{/member}",
            "avatar_url": "https://avatars.githubusercontent.com/u/102196109?v=4",
            "description": null,
            "is_verified": false,
            "has_organization_projects": true,
            "has_repository_projects": true,
            "public_repos": 0,
            "public_gists": 0,
            "followers": 0,
            "following": 0,
            "html_url": "https://github.com/electromaster-test",
            "created_at": "2022-03-23T00:44:48Z",
            "updated_at": "2022-03-23T00:44:48Z",
            "type": "Organization"
        },
        "parent": null
    }```

DoktorAerzt:
> It does generate some big ints, but I did some testing with the actual github API (I believe their sample response only shows an id of "1" which isn't helpful in this case), and they do return big integers as well. Not as big as this will be, but also not tiny either. I don't think it will be a huge issue.
> 
> I can also confirm that with the inclusion of these changes, Authentik is compatible with [Teleport](https://github.com/gravitational/teleport), which only supports Github SSO in their open-source version. I can send in a separate PR with a setup guide for the documentation.

Could you please create the setup guide. I was trying to get Teleport to work with Authentik but it won't was working.

sover02:
> Could you please create the setup guide. I was trying to get Teleport to work with Authentik but it won't was working.

More recent versions of teleport require an enterprise license to modify the `endpoint_url:` yaml field which is meant for a Github Enterprise server.

There are probably some hacky ways to circumvent the restriction, but of the ways I can think of, none of them are something I would consider.

This PR is still a very valuable feature and appreciated - it just won't work for Teleport OSS anymore.

PR #2651, opened 2022-04-03
Update en.po

keval6b:
# Details
Just a wording change


BeryJu:
Please change translations on https://www.transifex.com/beryjuorg/authentik/

PR #2745, opened 2022-04-18
web/admin: application wizard (part 1)

BeryJu:
Provide a wizard that simplifies the creation of

- [x] application
- [ ] provider
  - [ ] oauth
    - [ ] server
    - [ ] spa/native
    - [ ] api
  - [ ] saml
    - [ ] input
    - [ ] import from file
  - [ ] proxy
    - [ ] proxy
    - [ ] forward
  - [x] ldap
    - [x] service account
  - [x] link
- [ ] outpost links
- [ ] outpost creation?
- [ ] access control (separate PR)
  - [ ] group binding


xcsdm:
For application creation, it would be nice to allow specifying/uploading an application icon file.

I'll add that 1 of 2 options would be nice.  It would be great to allow users/group binding right from the wizard.  I suspect this may be a lot of code required, so a second option might be a checkbox like "Only allow use by admin users (for testing)" or something similar.

Both of these additional options prevent users from randomly seeing an application that may not be ready for general use.

BeryJu:
this will be built and finished as multiple PRs to keep them easier to maintain and keep updated

PR #2762, opened 2022-04-24
web: add missing checkbox element in user settings

tomtom5152:
# Details
Minor: checkbox is missing in user settings.

### New Features
* Add checkbox element to user settings `PrompStage`.
* Add required CSS elements to style it correctly.

## Additional
The default `PromptStage` handles checkboxes different in `renderField`, the user settings one does not do this and cannot do it whilst maintaining horizontal styling. This change extends the existing `renderPromptInner` to add checkbox support for this edge case.

There may be a better way of doing this, this is my first contribution, all be it minor, so very much learning the frameworks and codebase.


tomtom5152:
Force pushed prettier output, forgot to run it first time before committing

PR #2786, opened 2022-04-28
Add Documentation: Weblate integration with SAML

ilijamt:


website/integrations/services/weblate/index.md
```diff
@@ -0,0 +1,90 @@
...
+
+You can of course use a custom signing certificate, and adjust durations.
+
+## Property mappings
+
+We need to create some property mappings so our application will work. After you create the property mappings, assign them to the provider.
+
+### Full name
+* Name: `Weblate - Full name`
+* SAML Attribute Name: `urn:oid:2.5.4.3`
```
BeryJu, line 37: I assume these urn attributes are hardcoded in weblate? Seeing as most of these custom mappings duplicate inbuilt mappings

BeryJu (commented):
LGTM besides the potentially duplicate mappings

website/integrations/services/weblate/index.md
ilijamt: Not in Weblate, but in social-core plugin. Without the manually created properties, the process does not work. 

*  https://github.com/python-social-auth/social-core/blob/32d7bcf41dd0f21dc71cf14cb5a7373a7e522e88/social_core/backends/saml.py#L45

These provide extra information, so we can provide more details about the user connected. Without this, the new user created has no extra information other than the email.

* https://github.com/python-social-auth/social-core/blob/32d7bcf41dd0f21dc71cf14cb5a7373a7e522e88/social_core/backends/saml.py#L17-L23
* https://github.com/python-social-auth/social-core/blob/32d7bcf41dd0f21dc71cf14cb5a7373a7e522e88/social_core/backends/saml.py#L51-L67

PR #2818, opened 2022-05-06
Addition of form_post OIDC response mode

### LINKED ISSUE #2748: FR: Support for response_mode: form_post
We have a client application that is meant to use Azure as an IDP, and I've managed to customize most of the authentik flow to be able to use Authentik as a provider. The last part of this is that the OpenID Connect/Oauth2-provider currently doesn't support response_mode: form_post, which would be needed for this compatibility.

Modify the OpenID Connect/Oauth2 provider for use with the response mode form_post

**Additional context**
I think compatibility with applications that are made for Azure authentication would enhance the usability aspect of Authentik a great deal. I'll look into getting a PR together later this week/beginning of next week when I have some spare time, just thought I'd post here as well in case anyone else is interested/working on it/has a solution already.

scheibling:
# Details
Resolves #2748 

### New Features
* Adds support for form_post response mode, required for Azure-compatibility

### Breaking Changes
None

## Additional Details
-

### DIFF
--- a/authentik/providers/oauth2/views/authorize.py
+++ b/authentik/providers/oauth2/views/authorize.py
@@ -15,11 +15,7 @@ from structlog.stdlib import get_logger
 from authentik.core.models import Application
 from authentik.events.models import Event, EventAction
 from authentik.events.utils import get_user
-from authentik.flows.challenge import (
-    Challenge, 
-    ChallengeResponse, 
-    ChallengeTypes
-)
+from authentik.flows.challenge import ChallengeTypes, HttpChallengeResponse
 from authentik.flows.models import in_memory_stage
 from authentik.flows.planner import (
     PLAN_CONTEXT_APPLICATION,
@@ -27,7 +23,7 @@ from authentik.flows.planner import (
     FlowPlan,
     FlowPlanner,
 )
-from authentik.flows.stage import StageView, ChallengeStageView
+from authentik.flows.stage import StageView
 from authentik.flows.views.executor import SESSION_KEY_PLAN
 from authentik.lib.utils.time import timedelta_from_string
 from authentik.lib.utils.urls import redirect_with_qs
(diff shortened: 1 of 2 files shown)


scheibling:
@BeryJu Right, it's working now at least! 

I added a separate class for the post_response, and I've imported the AutosubmitChallenge and AutoSubmitChallengeResponse classes from providers.saml.views.flows for now, I'm going to do some testing with this setup and see whether I can find any shortcomings.

authentik/providers/oauth2/views/authorize.py
```diff
@@ -238,6 +256,131 @@ def create_code(self, request: HttpRequest) -> AuthorizationCode:
 
         return code
 
+class OAuthPostFulfillmentStage(ChallengeStageView):
```
BeryJu, line 259: So this extra stage view shouldn't be required, imo it makes more sense to hook in here https://github.com/goauthentik/authentik/blob/master/authentik/providers/oauth2/views/authorize.py#L248, rename `redirect` to `get_response` and have it check the parameters to either return an Autosubmit Challenge or a redirect

BeryJu (changes requested):

authentik/providers/oauth2/views/authorize.py
scheibling: Yep, I agree, this was just a way of testing the concept on my part for now. I'll have a look at integrating it a bit better later today.

scheibling: Right, had some more free time this afternoon. I've removed the separate class and moved the FORM_POST handling to create_response_uri() and redirect() respectively

I'll have to spend some more time going through a couple of more scenarios where form_post might want a non-implicit response, I've only been testing with an application of ours that has a very specific capability for OIDC (Azure) so far.

https://github.com/scheibling/authentik/blob/b7fd2b38935ed63a2bd1214f6c41d6ac2e6485b9/authentik/providers/oauth2/views/authorize.py#L265-L282

scheibling: Looked at it with a little more sleep behind me, I'm going to work on the create_response_uri-part a bit more since the current code doesn't really take any of the response_type parameters into account. 

Form_post in particular can be requested with just the code and id_token parts by themselves, I'm gonna read up a bit on that part of the protocol when I get into work tomorrow. 

The original version was written as if fragment always just creates an implicit response with just an id_token and  query only provides a code, couldn't there be other combinations on those as well? 

I'll be back with an update once I'm up to speed with the latest!

scheibling: Nope, apparently the code that's already in create_implicit_response works fine for form_post as well!

PR #2828, opened 2022-05-08
stages/authenticator_validate: remember

BeryJu:
Add option only validate MFA when device has not been used in x time


Matchlighter:
@BeryJu It looks like the last 2FA verification timestamp is being stored in the DB (as the `last_t` column)? It should be stored in the/a cookie so that it is scoped to a single client/browser. Otherwise it creates a security implication that if the true user logs in and completes MFA, a malicious device has an opportunity to login as that user without being impeded by 2FA.

BeryJu:
@Matchlighter you raise a good point, it would have to be stored in a separate, signed/encrypted cookie that lasts longer than the session (maybe no expiry at all?), but yes, the way you've described it is what would currently happen. I'll add a note to the docs to make this clear to people configuring it, since the above solution is not an easy fix

BeryJu:
Actually we could set a cookie with an expiry of the configure threshold in the stage, save which device the user used to authenticate that session and then simply check for it's existence (and check the device belongs to the current user and the current stage, etc), but that should be pretty simple

BeryJu:
Yeah true the expiration would also be saved in the cookie, or alternatively the timestamp of when the verification was done, which would make changing of the threshold easier

PR #2880, opened 2022-05-16
Add Warning for Intranet Users

zaxlofful:
I was am using Authentik as a intranet only source of authentication. I went thru all the steps list here (including paying $99) and at the VERY end; the apple developer website tells you....Only public URIs are allowed and more than likely only valid TLDs, because mine is atypical.

```
There is a problem with the request entity

authentik.<domainname>.<non-standard TLD> - Domain not eligible
```

### DIFF
--- a/website/integrations/sources/apple/index.md
+++ b/website/integrations/sources/apple/index.md
@@ -10,6 +10,10 @@ Allows users to authenticate using their Apple ID.
 An Apple developer account is required for this.
 :::
 
+:::warning
+Apple only allows OAuth with public internet facing URIs, this will not work for intranet
+:::
+
 The following placeholders will be used:
 
 -   `authentik.company` is the FQDN of the authentik install.


website/integrations/sources/apple/index.md
```diff
@@ -10,6 +10,10 @@ Allows users to authenticate using their Apple ID.
 An Apple developer account is required for this.
 :::
 
+:::warning
+Apple only allows OAuth with public internet facing URIs, this will not work for intranet
```
BeryJu, line 14: `Apple mandates the use of a [registered TLD](https://en.wikipedia.org/wiki/List_of_Internet_top-level_domains), as such this source will not work with .local and other non-public TLDs.`

BeryJu (changes requested):

website/integrations/sources/apple/index.md
ZaxLofful: That works for me.

Issue #2519, opened 2022-03-17
Server Error after connecting QNAP NAS to LDAP Outpost


BobSilent:
**Short Preamble**
I am currently (hard) working to get my QNAP NAS connected to the LDAP Outpost, 
unfortunately this is not an easy task, and when I am (hopefully) finished I can provide a documentation for it (if you like)

The QNAP UI Configuration is quite limited and the documentation (QNAP + forum) as well.

**Issue**
Setting up the LDAP Outpost, created a dedicated service account to use for LDAP Connection with a token
Make sure it is working and reachable!

QNAP LDAP Connection settings in UI are quite limited and QNAP makes some not described requirements/preconditions
- LDAP Server Host
- LDAP security
- Base DN
- Root DN
- Password
- User base DN
- Group base DN

using a token (length 128 Chars) as password

QNAP continuously tries to connect to LDAP Outpost, even though it cannot, due to a bad connection message send.
it seems that password length makes the issue (I think within QNAP, or within LDAP Outpost?) 
using a smaller password, currently just 10chars is working fine (I will extend the testing here - at least for the suggested documentation)

From the authentik event log I see a long list of continuous login attempts
![image](https://user-images.githubusercontent.com/13506173/158827016-5209a878-e85a-4466-a1d7-4e9c0fa22c9e.png)

and after some time, the whole authentik application went down
![image](https://user-images.githubusercontent.com/13506173/158827139-d70a90eb-75f5-46bf-896a-f6c1d2ea33ad.png)

from the container logs I can see that Postgres is giving up after a while with 
```
FATAL:  sorry, too many clients already
```

and all containers show healthy
![image](https://user-images.githubusercontent.com/13506173/158827890-c1100a1e-add4-4b17-8859-a67cdd34eb76.png)

**Expected behavior**
I would expect on the one hand that at least in such a case the healthcheck of the server or worker would show a not healthy state, better maybe to have an postgres healthcheck?
But what i really would like that the service is not running into such a situation where a "bad configured" App using authentik would lead to this effect.

**Version and Deployment (please complete the following information):**
 - authentik version: 2022.3.2
 - Deployment: docker-compose

PR #2958, opened 2022-05-25
Add docker-compose postgresql and redis healthchecks

BobSilent:
see #2519
it will not solve the issue but it should at least check the health of the postgres and redis

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -5,6 +5,12 @@ services:
   postgresql:
     image: postgres:12-alpine
     restart: unless-stopped
+    healthcheck:
+      test: ["CMD", "pg_isready"]
+      start_period: 20s
+      interval: 30s
+      retries: 5
+      timeout: 5s
     volumes:
       - database:/var/lib/postgresql/data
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -5,6 +5,12 @@ services:
   postgresql:
     image: postgres:12-alpine
     restart: unless-stopped
+    healthcheck:
+      test: ["CMD", "pg_isready"]
+      start_period: 20s
```
BeryJu, line 10: `start_period` seems to have been removed:

```
root@beryjuorg-do:/srv/authentik# docker-compose up -d
ERROR: The Compose file './docker-compose.yml' is invalid because:
services.postgresql.healthcheck value Additional properties are not allowed ('start_period' was unexpected)
services.redis.healthcheck value Additional properties are not allowed ('start_period' was unexpected)
```

BeryJu: oh wait, it requires 3.4 of the compose file, feel free to bump it to `3.4` at the top

PR #3037, opened 2022-06-04
core: add setting to open application launch URL in a new browser tab

frprm:
# Details
* **Does this resolve an issue?** No

### New Features
* Adds a setting to open application launch URL in a new browser tab


BeryJu (commented):
Cheers for the PR! Looks pretty good, could you set `debug: true` in your local config and re-run `make gen-build`? Some of the models are only loaded when debug mode is enabled and as such, debug mode has to be enabled for generating the scheme.
Also I think there might be some small formatting things, run `make` and `make web` to autoformat them

frprm:
Thanks for taking a look at my PR! I think I've addressed the concerns you shared earlier with the commits I just pushed... Let me know if there's anything else.

PR #3079, opened 2022-06-12
website/docs: expand nginx reverse-proxy setup

sevmonster:
## Changes
Clean up the reverse-proxy documentation to provide a bit more information and rationale. The original nginx configuration was also not able to use keepalive connections to authentik since the `Connection: close` header would more than likely be sent with every request that wasn't a WebSocket handshake, since `$http_upgrade` will be empty most of the time.

## Additional
Could cut down on the wordiness but I wanted to be verbose.

### DIFF
--- a/website/docs/installation/reverse-proxy.md
+++ b/website/docs/installation/reverse-proxy.md
@@ -2,40 +2,75 @@
 title: Reverse-proxy
 ---
 
-If you want to access authentik behind a reverse-proxy, use a config like this. It is important that Websocket is enabled, so that Outposts can connect.
+:::info
+Since authentik uses WebSockets to communicate with Outposts, it does not support HTTP/1.0 reverse-proxies. The HTTP/1.0 specification does not officially support WebSockets or protocol upgrades, though some clients may allow it.
+:::
+
+If you want to access authentik behind a reverse-proxy, there are a few headers that must be passed upstream:
+- `X-Forwarded-Proto`: Tells authentik and Proxy Providers if they are being served over a HTTPS connection.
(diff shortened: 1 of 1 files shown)


website/docs/installation/reverse-proxy.md
```diff
@@ -2,40 +2,75 @@
...
+        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+        proxy_set_header Host $host;
+        # Use HTTP/1.1 keepalive by unsetting the Connection header.
+        # It is set to "close" by default.
+        # https://nginx.org/r/proxy_set_header
+        proxy_set_header Connection "";
+    }
+
+    # WebSocket
+    location /ws/client/ {
```
BeryJu, line 65: This should be `/ws/` since there's another websocket endpoint besides `/ws/client/`, namely `/ws/outpost/<uuid>/` for Outposts (also I don't get the point of having a separate location block for websockets compared to just setting the headers correctly in the main block)

sevmonster: I could only see `client` when glancing at the code/page requests, thanks.
I don't know why I didn't think of it before, but you could just rewrite the mapping to replace `close` with the empty string, and both blocks could be combined. This is just how I have been doing it in all my configs, since I never declared my own mapping and used the mapping from `nginx.conf` provided by the package maintainer. I will change it.
My previous train of thought was that on most configurations and with the current setup, the `$connection_upgrade` mapping is `close` for all other requests than ones with the `Upgrade` header set, so merging them into one would mean that those requests would not be persistent.

sevmonster:
@BeryJu how does it look now?

PR #3114, opened 2022-06-18
Fix #3076

### LINKED ISSUE #3076: [Gitea] Authentication sources with non-lowercase names do not work
When creating an authentication source from Gitea (OAuth), if one supplies a name with uppercase letters in it, the callback URL returned by Authentik will be all lowercase and thus does not match the authentication source in Gitea.

### LINKED ISSUE #3083: Jenkins login broken due to lower cased redirect URI
**Describe the bug**
Since having upgraded to, I think authentik `2022.5.2`, I can no longer login to jenkins. I made no other changes, so I suspect there is some subtle change in authentik that broke the login. Now I'm running `2022.6.1` but the problem persists.

Symptoms:
The login is processed correctly by authentik and it redirects me back to my jenkins. Jenkins however, displays a generic error message and the logs contain the following:
```
at org.jenkinsci.plugins.oic.OicSecurityRealm$3.onSuccess(OicSecurityRealm.java:394)
at com.google.api.client.auth.openidconnect.IdTokenResponse.execute(IdTokenResponse.java:120)
at com.google.api.client.auth.oauth2.TokenRequest.executeUnparsed(TokenRequest.java:287)
at com.google.api.client.auth.oauth2.TokenResponseException.from(TokenResponseException.java:105)
}
"error_description" : "Client authentication failed (e.g., unknown client, no client authentication included, or unsupported authentication method)"
"error" : "invalid_client",
{

com.google.api.client.auth.oauth2.TokenResponseException: 400 Bad Request
```

In the authentik log event stream I can see the following entry along with every login attempt to jenkins:
grafik

Note the different casing within the expected and actual redirect uri. However, I have no idea where this lower case uri expectation comes from.
When initiating the authenticat from jenkins ,it sends the following request (note the upper case chars in redirect URI):

```
GET https://<redacted>/application/o/authorize/?client_id=<redacted>&redirect_uri=https://<redacted>/securityRealm/finishLogin
```
And the provider configuration looks like this (also upper case chars):
grafik

There is no place known to me, where the lower case version of the expected redirect URI comes into play. So I suspect there is some place within authentik where the actual URI is (errroneously?) lowercased before comparing it to the expected one.

This might relate to #3076

**NOTE:**
While writing this, I tried to configure both: the correct URI and the lower case URI as "Redirect URI" and this seems to work. Putting only one OR the other results in a failed authentication attempt.

9p4:
# Details
Resolves #3076
Resolves #3083

## Additional
Ideally there would be a test to check if an uppercased redirect URI would work but that would require for the OIDC testing Docker image to be changed, since the URL is hardcoded.

### DIFF
--- a/authentik/providers/oauth2/views/token.py
+++ b/authentik/providers/oauth2/views/token.py
@@ -87,7 +87,7 @@ class TokenParams:
             provider=provider,
             client_id=client_id,
             client_secret=client_secret,
-            redirect_uri=request.POST.get("redirect_uri", "").lower(),
+            redirect_uri=request.POST.get("redirect_uri", ""),
             grant_type=request.POST.get("grant_type", ""),
             state=request.POST.get("state", ""),
             scope=request.POST.get("scope", "").split(),


9p4:
https://github.com/BeryJu/oidc-test-client/issues/20

BeryJu:
@9p4 Ty for the PR, I think this is much easier tested with a unittest than with an e2e test, but I'll add that option to the oidc-test-client regardless

PR #3299, opened 2022-07-21
Feat: use mailcow full_name as username for mailcow source

### LINKED ISSUE #3298: Get username from mailcow source
I like to get a username from mailcow. With username the enrollment for new users is more simple.

Set username to full_name provided by mailcow oauths source.

**Additional context**
For other sources the username is also set redundant to another attribute if there is no special source attribute:
azure_ad.py:
```
            "username": info.get("displayName"),
            "name": info.get("displayName"),
```

discord.py:
```
            "username": info.get("username"),
            "name": info.get("username"),
```

facebook.py:
```
            "username": info.get("name"),
            "name": info.get("name"),
```

reddit.py
```
            "username": info.get("name"),
            "name": info.get("name"),
```

l-with:
# Details
Resolves #3298

### New Features
* Adds  username for mailcow source as full_name

### DIFF
--- a/authentik/sources/oauth/types/mailcow.py
+++ b/authentik/sources/oauth/types/mailcow.py
@@ -52,6 +52,7 @@ class MailcowOAuth2Callback(OAuthCallback):
         info: dict[str, Any],
     ) -> dict[str, Any]:
         return {
+            "username": info.get("full_name"),
             "email": info.get("email"),
             "name": info.get("full_name"),
         }


BeryJu:
I would prefer to not merge this, but checking with the other source types, it seems fairly common to use the user's full name as username, even though that might not be desired

(I should probably add property mappings to this)

l-with:
> I would prefer to not merge this, but checking with the other source types, it seems fairly common to use the user's full name as username, even though that might not be desired
> 
> (I should probably add property mappings to this)

Thank you very much for merging. 

Property mappings would be fine. Furthermore it would be great to have the possibility to define a standard group for the federated users. I would like to help, but there are two constrains: 1st I have no experience with Django, 2nd I have fewer time to spend starting next week.

PR #3362, opened 2022-08-03
Make LDAP outpost ports configureable via env

### LINKED ISSUE #3360: [FR] Configureable ports for outpost via env
I'm not always able to use docker containers, but rather execute the go binary directly which doesn't work great when the default port it binds to is not allowed and/or free.

I would like a way to specify the ports to bind to via env, and if they are not set default to the current ones.

I have considered to just compile my own version of the binary, but keeping it updated is not going to be fun.

**Additional context**
Shared user shell hosting, got allocated an port and it would need to use that than the default.

samip5:
# Details
Resolves #3360

### New Features
* Added ability to configure LDAP outpost ports via env variables AUTHENTIK_LDAP_PORT and AUTHENTIK_LDAPS_PORT. Will faillback to original ones if not set.

### DIFF
--- a/internal/outpost/ldap/ldap.go
+++ b/internal/outpost/ldap/ldap.go
@@ -49,7 +49,8 @@ func (ls *LDAPServer) Type() string {
 }
 
 func (ls *LDAPServer) StartLDAPServer() error {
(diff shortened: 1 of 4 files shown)


samip5:
Superseaded by #3367.

PR #3367, opened 2022-08-03
centralise config for listeners to use same config system everywhere

BeryJu:
Signed-off-by: Jens Langhammer <jens.langhammer@beryju.org>

#3360


samip5:
Unfortunately, those env variables will not override the file that was used when building it. It would need to be able to override them via env.

PR #3372, opened 2022-08-04
website/docs: improve nginx examples

arevindh:
# Details
Fixes issues with some applications unable to identify the actual hostname and redirect to the IP after login from Authentik.

### Improves Nginx and Nginx Proxy Manager examples.
* Added commented code to pass the original hostname to app, some apps may require it to work correctly.
* Added commented code to pass original URL as X-Original-URL to host, t, some apps may require it to work correctly.


BeryJu (commented):
Setting the host header makes sense, however with `X-Original-URL` I'm not 100% sure if that should be included as `X-Original-URL` is a non-standard header that can vary application by application, so maybe only add that as a comment and mention that some applications may require it

arevindh:
@BeryJu Updated to code to comments

PR #3417, opened 2022-08-14
Patch sms api

samip5:
### Breaking Changes
* Potentially breaking change to the SMS generic payload as it was changed.

## Additional
It did not make sense to have them uppercased.


samip5:
@BeryJu If this is not going to get merged, how would you entail the templating thing to work or should I even try to make that?

BeryJu:
https://github.com/goauthentik/authentik/pull/3780

PR #3427, opened 2022-08-17
website/outposts: Kubernetes annotations for embedded outpost.

clbx:
# Details
Resolves # n/a

### New Features
* Adds a blurb in the docs to outline the importance of adding annotations to the embedded outpost. 

## Additional
I'm not sure if this is the best place or the best format for this to be in. If there is a better location or more information required, I'll add it in.


BeryJu:
Ideally all options that can be used with the embedded outpost should be marked as such, if you don't mind I'll add that

PR #3481, opened 2022-08-26
website/docs: improve Guacamole

obxjames:
### Details
**Does this resolve an issue?**
Fixes missing parts to Guac set up.

### DIFF
--- a/website/integrations/services/apache-guacamole/index.mdx
+++ b/website/integrations/services/apache-guacamole/index.mdx
@@ -28,11 +28,13 @@ Create an OAuth2/OpenID provider with the following parameters:
 Under _Advanced protocol settings_, set the following:
 
 -   Token validity: Any value to configure how long the session should last. Guacamole will not accept any tokens valid longer than 300 Minutes.
--   Signing Key: Set the key as `authentik.company`
+-   Signing Key: Set the key as `authentik Self-signed Certificate`
 
 Note the Client ID value. Create an application, using the provider you've created above.
 
 ## Guacamole
+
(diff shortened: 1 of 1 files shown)


website/integrations/services/apache-guacamole/index.mdx
```diff
@@ -28,6 +28,7 @@ Create an OAuth2/OpenID provider with the following parameters:
 Under _Advanced protocol settings_, set the following:
 
 -   Token validity: Any value to configure how long the session should last. Guacamole will not accept any tokens valid longer than 300 Minutes.
+-   Signing Key: Set the key as `authentik.company`
```
BeryJu, line 31: The default signing key is called `authentik Self-signed Certificate`

website/integrations/services/apache-guacamole/index.mdx
```diff
@@ -67,3 +68,10 @@ openid-redirect-uri=https://guacamole.company/ # This must match the redirect UR
 
   </TabItem>
 </Tabs>
+
+## Guacamole User Creation
+
+In Guacamole - if everything has configured correctly you will see in the bottom left corner `Sign in with: OpenID`
```
BeryJu, line 74: This button only shows up when multiple authentication plugins are configured, when only OIDC is configured the redirect happens directly

website/integrations/services/apache-guacamole/index.mdx
```diff
@@ -67,3 +68,10 @@ openid-redirect-uri=https://guacamole.company/ # This must match the redirect UR
...
+Before you are able to use this - using your preconfigured Admin Account - Login then create a User using your Email Address (one set in Authentik) as the username. Give the persmission as you prefer. When you then login you will now see as configured.
```
BeryJu, line 75: This should probably be an earlier step, as a note before configuring guacamole

BeryJu (changes requested):

PR #3508, opened 2022-08-31
Documentation: Improving 'Portainer' section

mtthidoteu:
Adding in a / at the end of the redirect url is crucial and failing to do so will cause a 'Redirect URL' error thrown in by authentik.
I also find it more clear to use 'portainer.company' instead of 'port.company'.

Signed-off-by: Matthieu B <66959271+mtthidoteu@users.noreply.github.com>

# Details
Resolves #

### New Features

### Breaking Changes

## Additional

### DIFF
--- a/website/integrations/services/portainer/index.md
+++ b/website/integrations/services/portainer/index.md
@@ -20,7 +20,7 @@ This is based on authentik 2021.7.3 and Portainer 2.6.x-CE. Portainer 2.6 suppor
 
 The following placeholders will be used:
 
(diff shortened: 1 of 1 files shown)


website/integrations/services/portainer/index.md
```diff
@@ -33,10 +33,11 @@ Only settings that have been modified from default have been listed.
 
 **Protocol Settings**
 
+
```
BeryJu, line 36:
```suggestion

```

PR #3557, opened 2022-09-07
Note about setup with self-signed certificates

sathrudi:
Added a note with info about using self-signed certificates with authentik.

Signed-off-by: sathrudi <85929245+sathrudi@users.noreply.github.com>

# Details
Just added a note about WikiJS configuration when using authentik with self-signed certs.

## Additional
When using self-signed certificates for authentik, WikiJS gives an error "Failed to obtain access token". This is caused by the root certificate of the authentik SSLcert not being trusted by WikiJS as detailed here: https://github.com/Requarks/wiki/discussions/3387


BeryJu (changes requested):
Please run `make website` to fix the formatting

BeryJu (approved):

PR #3577, opened 2022-09-12
web: set default value for cert select inputs

relative:
# Details
No

### New Features
* adds value/selected attribute to the fallback option part in the `until` directives used where certificate-key pairs are specified in the web interface

## Additional
Requesting a single `CertificateKeyPair` from the API can take up to 800ms (likely because it loads the private key/certificate every time the entity is loaded). This causes considerable slow-down when loading the Certificate-Key Pairs page. Additionally any forms that have select inputs that accept certificate-key pairs (e.g. SAML provider) will be blocked from submission until the request to retrieve every keypair is completed. 
If you attempt to submit a form with a cert input before the CertificateKeyPair request has completed then the associated keypair uuid will be lost. This PR solves this by setting the fallback option's value to the selected key-pair.

Perhaps a better solution would be to only request the columns that are needed for display however I don't think that is possible at the moment.

### DIFF
--- a/web/src/pages/outposts/ServiceConnectionDockerForm.ts
+++ b/web/src/pages/outposts/ServiceConnectionDockerForm.ts
@@ -98,7 +98,12 @@ export class ServiceConnectionDockerForm extends ModelForm<DockerServiceConnecti
                                     </option>`;
                                 });
                             }),
-                        html`<option>${t`Loading...`}</option>`,
+                        html`<option
+                            value="${ifDefined(this.instance?.tlsVerification)}"
+                            ?selected=${this.instance?.tlsVerification !== undefined}
+                        >
+                            ${t`Loading...`}
+                        </option>`,
                     )}
                 </select>
                 <p class="pf-c-form__helper-text">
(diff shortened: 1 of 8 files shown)


BeryJu (approved):
LGTM, feel free to create another issue about the speed of the keypair endpoint, it would definitely make sense to make the loading of each cert and private key option

BeryJu (changes requested):
My bad, the changes look good but could you run `make web` and add that to the PR

relative:
> My bad, the changes look good but could you run `make web` and add that to the PR

I ran `make web` and it didn't make any changes for me to commit. `lit-analyse` failed in the [same way as the CI did](https://github.com/goauthentik/authentik/actions/runs/3033948051/jobs/4886083307)

BeryJu:
I've disabled the `no-unknown-tag-name` rule for now as the plugin doesn't seem to interpret the tsconfig paths correctly, I think there's a new error in there that just got buried in the `no-unknown-tag-name` spam. Could you rebase on main?

BeryJu:
why did you close the PR?

relative:
I think I screwed something up when I rebased, give me a moment

relative:
Seems like my commit is now on my patch/1 branch. Not very proficient with git to be honest. 

Let me know if I should make a new PR.

BeryJu:
Looks better now, I think the initial error it was complaining about was `value="${ifDefined(this.instance?.tlsVerification)}"` and not `value=${ifDefined(this.instance?.tlsVerification)}` (note the lack of quotation marks)

relative:
Should be fixed now, lit-analyse was not showing the `no-nullable-attribute-binding` errors on my machine

BeryJu:
Yeah same here, lit-analyzer CLI isn't complaining on my machine but VS Code is, and the only way to silence those is to add `|| undefined` to all ifDefined calls for some reason?

PR #3769, opened 2022-10-12
Add setting to adjust database config for pgbouncer

### LINKED ISSUE #3358: Support HA postgresql
When using a HA installation of postgresql connection problems with the api server arise during initial load of the user dashboard.

Pgbouncer in transaction pooling mode requires custom settings in order to function correctly.
It would be nice if the user could specify that a HA installation is used and the settings are then adjusted automatically.

```
# https://docs.djangoproject.com/en/4.0/ref/databases/#transaction-pooling-server-side-cursors
DATABASES['default']['DISABLE_SERVER_SIDE_CURSORS'] = True
# https://docs.djangoproject.com/en/4.0/ref/databases/#persistent-connections
DATABASES['default']['CONN_MAX_AGE'] = None  # persistent
```

Currently the settings do not include these options:
https://github.com/goauthentik/authentik/blob/89c84f10d0e6e70e51444a9fbf18980ba25008de/authentik/root/settings.py#L279-L288

PKizzle:
# Details
Resolves #3358 

### New Features
* Add setting postgresql.use_pgbouncer to enable adjustments of the configuration for PgBouncer

### Breaking Changes
* None

## Additional
Tested on local cluster. Works as expected.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -62,7 +62,6 @@ services:
       AUTHENTIK_POSTGRESQL__USER: ${PG_USER:-authentik}
       AUTHENTIK_POSTGRESQL__NAME: ${PG_DB:-authentik}
       AUTHENTIK_POSTGRESQL__PASSWORD: ${PG_PASS}
-      AUTHENTIK_POSTGRESQL__USE_PGBOUNCER: false
       # AUTHENTIK_ERROR_REPORTING__ENABLED: "true"
     # This is optional, and can be removed. If you remove this, the following will happen
     # - The permissions for the /media folders aren't fixed, so make sure they are 1000:1000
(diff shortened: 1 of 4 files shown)


docker-compose.yml
```diff
@@ -41,6 +41,7 @@ services:
       AUTHENTIK_POSTGRESQL__USER: ${PG_USER:-authentik}
       AUTHENTIK_POSTGRESQL__NAME: ${PG_DB:-authentik}
       AUTHENTIK_POSTGRESQL__PASSWORD: ${PG_PASS}
+      AUTHENTIK_POSTGRESQL__USE_PGBOUNCER: false
```
BeryJu, line 44: As this is the same value as the default, imo it shouldn't be set here to keep the compose file small (it's already a bit much)

BeryJu (changes requested):

PR #3780, opened 2022-10-13
stages/authenticator_sms: make sms stage payload customisable

BeryJu:


samip5:
@BeryJu What about filtering the allowed phone numbers? Will this allow it too as I don't believe so?

PR #3813, opened 2022-10-18
internal: allow Sentry DSN to be configured by end user

tobz:
## Details

The existing Sentry instrumentation is fairly valuable for performance optimization purposes, but having Sentry hard-coded to report to `https://sentry.beryju.org` makes it a pain to override during local development.

### New Features

* Allow the Sentry DSN to be overridden via `AUTHENTIK_ERROR_REPORTING__SENTRY_DSN`.

## Additional

I updated the configuration docs to list the new configuration setting, and _slightly_ tweaked the wording of the surrounding documentation to explain the dichotomy between the default DSN -- useful for debugging issues in collaboration with authentik devs, anonymous performance data, etc -- and specifying a custom one.

### DIFF
--- a/authentik/lib/sentry.py
+++ b/authentik/lib/sentry.py
@@ -34,7 +34,7 @@ from authentik.lib.utils.http import authentik_user_agent
 from authentik.lib.utils.reflection import class_to_path, get_env
 
 LOGGER = get_logger()
-SENTRY_DSN = "https://a579bb09306d4f8b8d8847c052d3a1d3@sentry.beryju.org/8"
+DEFAULT_SENTRY_DSN = "https://a579bb09306d4f8b8d8847c052d3a1d3@sentry.beryju.org/8"
 
 
 class SentryWSMiddleware(BaseMiddleware):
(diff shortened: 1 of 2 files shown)


BeryJu:
I haven't made this configurable in the past, as the DSN is also in the Web Interface and the Outposts, so it would have to be exposed in the authentik config, and I'm not sure how people would feel about that.

Using it for development purposes is certainly valid, but if it's a setting it should also be set for the embedded outpost (see https://github.com/goauthentik/authentik/blob/c22dae868c0791b405fb800c57edfed0af2d9b53/cmd/server/main.go)

tobz:
I'll take a look at the link you sent and see about making it configurable there, too.

> I haven't made this configurable in the past, as the DSN is also in the Web Interface and the Outposts, so it would have to be exposed in the authentik config, and I'm not sure how people would feel about that.

It's not clear to me why an otherwise internet-accessible Sentry DSN being in the configuration would be something people don't like. There's passwords and _actual_ credentials in there. What am I missing?

PR #3899, opened 2022-10-29
Feature/simple group serializer

### LINKED ISSUE #3884: Speed-Up User Interaction by removing groups_obj from UserSerializer
We have a medium sized database of users which consists of approximately 500 users. The use groups to assign individual features to them.

When updating users or interacting with the UI, we expierence a large lag (30-60s) just retrieving the data, as the API has to resolve all groups assigned to a user and all the users within this group.

A quick search in the webfrontend yield no results where this feature would be required.

Remove this field or make the field optional from the beginning.

As a workaround i currently replaced in authentik/core/api/users.py the following line:
https://github.com/goauthentik/authentik/blob/main/authentik/core/api/users.py#L86

with a SerializerMethodField that returns an empty array

The inclusion of the drf-dynamic-fields would allow requests to include a list of necessary fields, but does not play nicely with javascript / typescript.

**Additional context**
-

d-schiffner:
# Details
This PR introduces a new serializer for retrieving groups of users. 

Background: In our instance over 500 users are assigned in a single group which results in very poor performance.
Debugging showed that this is due to the use of the user_obj, which results a query of over 500 users for a single user. Especially in the UI and the LDAP backend this results in a huge burst query on the database.

Resolves #3884

### New Features
* Adds a serializer "SimpleGroupSerializer" as part of the core/api/users routes

## Additional
I tried to test locally, however the setup (I only use docker) has proven to be very difficult.

### DIFF
--- a/authentik/core/api/users.py
+++ b/authentik/core/api/users.py
@@ -73,6 +73,27 @@ from authentik.tenants.models import Tenant
 
 LOGGER = get_logger()
 
+class SimpleGroupSerializer(ModelSerializer):
+    """Group Serializer"""
+
+    attributes = JSONField(validators=[is_dict], required=False)
+    parent_name = CharField(source="parent.name", read_only=True)
+
+    num_pk = IntegerField(read_only=True)
+
+    class Meta:
+
+        model = Group
+        fields = [
+            "pk",
+            "num_pk",
+            "name",
+            "is_superuser",
+            "parent",
+            "parent_name",
+            "attributes",
+        ]
+
 
 class UserSerializer(ModelSerializer):
     """User Serializer"""
@@ -83,7 +104,7 @@ class UserSerializer(ModelSerializer):
     groups = PrimaryKeyRelatedField(
         allow_empty=True, many=True, source="ak_groups", queryset=Group.objects.all()
     )
-    groups_obj = ListSerializer(child=GroupSerializer(), read_only=True, source="ak_groups")
+    groups_obj = ListSerializer(child=SimpleGroupSerializer(), read_only=True, source="ak_groups")
     uid = CharField(read_only=True)
     username = CharField(max_length=150)
 
--- a/authentik/core/api/users.py
+++ b/authentik/core/api/users.py
@@ -73,13 +73,11 @@ from authentik.tenants.models import Tenant
 LOGGER = get_logger()
 
 
-class SimpleGroupSerializer(ModelSerializer):
-    """Group Serializer"""
+class UserGroupSerializer(ModelSerializer):
+    """Simplified Group Serializer for user's groups"""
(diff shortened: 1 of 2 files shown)


authentik/core/api/users.py
```diff
@@ -83,7 +104,7 @@ class UserSerializer(ModelSerializer):
     groups = PrimaryKeyRelatedField(
         allow_empty=True, many=True, source="ak_groups", queryset=Group.objects.all()
     )
-    groups_obj = ListSerializer(child=GroupSerializer(), read_only=True, source="ak_groups")
+    groups_obj = ListSerializer(child=SimpleGroupSerializer(), read_only=True, source="ak_groups")
```
BeryJu, line 107: Please use `GroupMemberSerializer` in `groups.py` instead of creating a new serailizer

BeryJu (changes requested):

authentik/core/api/users.py
d-schiffner: How should the `GroupMemberSerializer` be utilized here? The `groups_obj` resolves the assigned groups (Using the `GroupSerializer`), which in turn resolve all it's users via the `GroupMemberSerializer`.

But option of replacing the `UserSerializer` is also not valid, or do I miss something here?

BeryJu: the name `GroupMemberSerializer` is a bit misleading, `GroupMemberSerializer` is a stripped down version of the GroupSerializer without a lot of the extra fields and is very close to the serializer you've added, so re-using the same serializer is a bit cleaner imo

d-schiffner: We're talking about this one, right?

https://github.com/goauthentik/authentik/blob/main/authentik/core/api/groups.py#L17-L37

I tend do disagree, as it clearly extracts user information, as written in the docs as well!
`"""Stripped down user serializer to show relevant users for groups"""`

It is a stripped down version of the `UserSerializer`, not the `GroupSerializer`!

I took inspiration when creating the Serializer from it. In other terms I fully agree not to introduce new Serializers unless necessary. I have also written options within the issue this PR is referring to. The smaller GroupSerializer was actually the one with minimal impact on the existing codebase.

The LDAP outpost uses the groups object of the users to get the actual names when creating the entries. Stripping the `groups_obj` from the `UserSerializer` would require a code change there as well including additional requests from the LDAP outpost the retrieve the group names.

BeryJu: Oh yeah my bad, I completely mis-remembered, disregard what I said above

d-schiffner:
I now managed to get the docker environment running with the tests.

In the last push i fixed the black and pylint issues

BeryJu (approved):

PR #3936, opened 2022-11-03
Added backend support for LDAP group lookups from the user

shawnweeks:
# Details
Resolves #

### New Features
* Adds feature which looks up a users group in LDAP from the user memberOf or related attribute

### Breaking Changes
* Should not be a breaking change as the default behavior has not changed.

## Additional
This is just a preliminary PR and is not complete nor ready to merge.


authentik/sources/ldap/sync/membership.py
```diff
@@ -34,7 +34,24 @@ def sync(self) -> int:
         )
         membership_count = 0
         for group in groups:
-            members = group.get("attributes", {}).get(self._source.group_membership_field, [])
+            if self._source.lookup_groups_from_user:
+                group_dn = group.get("dn", {})
```
shawnweeks, line 38: I noticed where you sometimes lookup an attribute called "distinguishedName" but you don't actually have to since the LDAP library your using always pulls the "dn" anyway. In Active Directory they always match as far as I can tell and "dn" is part of the LDAP spec.

authentik/sources/ldap/sync/membership.py
```diff
@@ -53,7 +70,7 @@ def sync(self) -> int:
                         "ak_groups__in": [ak_group],
                     }
                 )
-            )
+            ).distinct()
```
shawnweeks, line 73: I added a distinct here because I noticed it was pulling back the same user multiple times if they were in multiple groups due to the join. It seemed wrong to call group save with the same user multiple times but Django may already be handling that internally.

derlucas:
I'm very interested in this feature since i want to implement nested Groups in my AD Forest.
Is there any way i can help with the Integration of that feature?

shawnweeks:
@derlucas This will work against AD by setting the member of attribute to `memberOf:1.2.840.113556.1.4.1941:` as well and I'm testing it against both. I need to just sit down and get some test cases worked out and get some feedback. Hope to work on some test cases this weekend.

fdisamuel:
Hey! I would really love to see this feature PR merged as our company uses nested LDAP Groups heavily in their setup. Without being supported in Authentik, proper group sync from our AD server seems to be impossible...

cybertschunk:
hey,  I 'd like to continue working on the PR. Would that be possible? We really need support for nested group in our environment.

fdisamuel:
Hello @cybertschunk, it's great to see your readiness to add this feature.😊 Concerning the current PR, unfortunately, it may not be feasible to rebase due to it being 4600 commits behind the latest authentik codebase release. It might be more practical to implement the nested sync functionality in a new Pull Request using the most recent source code (I think, as a non-programmer ) However I would be happy to hear a feedback from the Authentik Dev Team regarding this topic!

PR #4036, opened 2022-11-19
Allow specifying a redis URL directly

nogweii:
# Details
I ran into this when trying to set up Authentik in my homelab. Specifically, I'm making a dedicated tiny computer run some security-sensitive software (Authentik, Vault, step-ca) and I'm hoping to minimize the number of listening TCP sockets.

### New Features
* Support overriding the generated cache location with a provided URL

## Additional
This is meant for a couple of more advanced use cases:

* Allows specifying a UNIX socket path to Redis
* Can use a specific user name (rather than the `default` user) allowing for more granular ACLs in Redis 6+

As a side effect, any future keyword arguments to `redis-py` can be passed in immediately before there are specific environment variables for each of them.

Signed-off-by: No GUI <14547+nogweii@users.noreply.github.com>

### DIFF
--- a/authentik/root/settings.py
+++ b/authentik/root/settings.py
@@ -195,7 +195,7 @@ _redis_url = (
 CACHES = {
     "default": {
         "BACKEND": "django_redis.cache.RedisCache",
-        "LOCATION": f"{_redis_url}/{CONFIG.y('redis.db')}",
+        "LOCATION": CONFIG.y('redis.url') or f"{_redis_url}/{CONFIG.y('redis.db')}",
         "TIMEOUT": int(CONFIG.y("redis.cache_timeout", 300)),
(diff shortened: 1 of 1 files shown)


nogweii:
friendly ping @BeryJu 😄

BeryJu:
This can be done via custom python settings (see https://goauthentik.io/docs/installation/configuration#custom-python-settings)

js6pak:
Using redis via an unix socket is not an advanced use case and it should be possible to do so by default

98h398hrpohpoai:
Why was this closed? js6pak is correct?
Attempting to do this through custom python settings is problematic as it seems to ignore redis changes, and the dump file doesn't process custom settings to debug.

PR #4100, opened 2022-11-28
Always add generic message to failing zxcvbn password policy

astro-stan:
Depending on the settings, sometimes a password policy that checks a password with the zxcvbn tool can fail without any message.

For example:
```
$ echo  'Awdccdw1234' | zxcvbn | jq | grep "feedback" -A 5 -B 1
Password: 
  "score": 3,
  "feedback": {
    "warning": "",
    "suggestions": []
  }
}
```

As seen above the tool does not produce any warnings or suggestions for the given password, but if the password policy is set to have a zxcvbn threshold of 3, the policy will silently fail without communicating the reason to the user. 

There are two ways to handle this:
1. Always add a generic "password is too weak" message when the policy fails.
2. Check if there are any suggestions or warnings from the zxcvbn tool and only add the generic message if not.

I personally prefer 1. This way the generic message will  be shown whenever the policy fails, and will get combined with extra "tips" whenever zxcvbn has some.

Signed-off-by: sdimovv <36302090+sdimovv@users.noreply.github.com>

# Details
N/A

### New Features
* Improve password policy messages 

### Breaking Changes
N/A

## Additional
N/A

### DIFF
--- a/authentik/policies/password/models.py
+++ b/authentik/policies/password/models.py
@@ -151,7 +151,7 @@ class PasswordPolicy(Policy):
         LOGGER.debug("password failed", check="zxcvbn", score=results["score"])
         result = PolicyResult(results["score"] > self.zxcvbn_score_threshold)
         if not result.passing:
-            result.messages += tuple((_("Password is too weak."), ))
+            result.messages += tuple((_("Password is too weak."),))
         if isinstance(results["feedback"]["warning"], list):
             result.messages += tuple(results["feedback"]["warning"])
         if isinstance(results["feedback"]["suggestions"], list):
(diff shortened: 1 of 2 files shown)


authentik/policies/password/models.py
```diff
@@ -150,6 +150,8 @@ def passes_zxcvbn(self, password: str, request: PolicyRequest) -> PolicyResult:
         results = zxcvbn(password[:100], user_inputs)
         LOGGER.debug("password failed", check="zxcvbn", score=results["score"])
         result = PolicyResult(results["score"] > self.zxcvbn_score_threshold)
+        if not result.passing:
+            result.messages += tuple(_("Password is too weak."))
```
BeryJu, line 154:
```suggestion
            result.messages += tuple((_("Password is too weak."), ))
```

BeryJu (approved):
LGTM, would appreciate a test case with the example you used

sdimovv:
Hi,

would that do?

PR #4106, opened 2022-11-28
Fixed bug causing filtering with an empty query

astro-stan:
Fixed bug allowing blueprint import to filter for existing models using an empty query.

The code only checks if the `identifiers` dict is empty, but `__query_from_identifier` skips identifier member values of type `dict` or keys == `pk`, so it is possible to produce an empty query if an `identifier` consists of just `dict` type members or "pk" key. 

Example:
```
...
- model: authentik_core.group
  identifiers:
    attributes:
      key: value
  attrs:
    name: my_group
...
```

The above is a perfectly valid blueprint entry and will pass all validation checks, but when applied will cause the `__query_from_identifier`  to produce an empty query (`Q()`), which will then cause all groups to be returned from the DB. Then the first of those groups will be replaced with the data from the blueprint instance.

Signed-off-by: sdimovv <36302090+sdimovv@users.noreply.github.com>

# Details
N/A

### New Features
N/A

### Breaking Changes
N/A

## Additional
N/A

### DIFF
--- a/authentik/blueprints/tests/test_v1.py
+++ b/authentik/blueprints/tests/test_v1.py
@@ -26,6 +26,13 @@ class TestBlueprintsV1(TransactionTestCase):
             )
         )
         self.assertFalse(importer.validate()[0])
+        importer = Importer(
+            (
+                '{"version": 1, "entries": [{"identifiers": {"attributes": {"key": "value"}}, '
+                '"attrs": {}, "model": "authentik_core.User"}]}'
+            )
+        )
+        self.assertFalse(importer.validate()[0])
 
     def test_export_validate_import(self):
         """Test export and validate it"""
--- a/authentik/blueprints/tests/test_v1.py
+++ b/authentik/blueprints/tests/test_v1.py
@@ -4,6 +4,7 @@ from django.test import TransactionTestCase
 from authentik.blueprints.tests import load_yaml_fixture
 from authentik.blueprints.v1.exporter import FlowExporter
 from authentik.blueprints.v1.importer import Importer, transaction_rollback
+from authentik.core.models import Group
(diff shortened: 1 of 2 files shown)


BeryJu (commented):
Thanks for the PR! Could you add a simple test case for this?

See https://github.com/goauthentik/authentik/blob/main/authentik/blueprints/tests/test_v1_state.py#L13 for an example

sdimovv:
Hello,

 would that do?

BeryJu:
I don't think there was a concrete reason for skipping dicts, and rather just not expecting people to use them as identifiers (and to skip implementing a function that converts all dicts from `{foo: {bar: baz}}` into `foo__bar=baz` queries

sdimovv:
I see. Well my use case was to use something irrelevant for the admin as an identifier so that they are free to change all model properties without fearing they will create a new object instead of updating the current one. 

Attributes stores arbitrary key value pairs so it made sense to use that. But thinking about it now, not all models have the attributes property this is probably not the best solution. I am interested to hear your suggestions on how this can be achieved. If you think this is worth of a discussion I can open a separate ticket so we don't overload this PR. 

In any case I will also see if I can come up with a patch for using a dict object in the query instead of skipping it.

sdimovv:
I have now added support for using dict fields as identifiers. It turned out to be surprisingly easy as PostgreSQL supports this out of the box and django exposes it via the `__contains` field suffix. I have also added a test to demonstrate the new functionality.

I  disabled the `too-many-locals` pylint warning for the `_validate_single` method as linting was failing (16/15 locals). I don't see an easy way to fix this without hurting code readability or refactoring/splitting the method. I hope that is okay.

PR #4127, opened 2022-11-30
Blueprints: Support nested custom tags in `!Find` and `!Format` tags

astro-stan:
# Details
N/A

### New Features
* Adds support for nested custom tags to the `!Find` and `!Format` tags.
This allows usages like these:
```
version: 1
context:
  flow_property: slug
  flow_property_value: my-flow
  stage_suffix: my-stage
entries:
- model: authentik_stages_password.passwordstage
  identifiers:
    name: !Format [prefixed-%s, !Context stage_suffix]
  attrs:
    configure_flow: !Find [authentik_flows.flow, [!Context flow_property, !Context flow_property_value]]
```

### Breaking Changes
N/A

## Additional
Incidentally, this also fixed a bug in the `!Format `  tag, which was previously taking the name of the context variable of a `!Context` tag, rather than resolving the tag and taking its value. The test for this functionality was also broken.

You can see the test here:
https://github.com/goauthentik/authentik/blob/d5329432fed9ed2a025ca0f7f3aa0317b44cfa85/authentik/blueprints/tests/test_v1.py#L75-L81

And the fixture here:
https://github.com/goauthentik/authentik/blob/d5329432fed9ed2a025ca0f7f3aa0317b44cfa85/authentik/blueprints/tests/fixtures/tags.yaml#L1-L10

The correct name according to the context variables should have been: `foo-bar-baz`, not `foo-foo-bar`


authentik/blueprints/v1/common.py
```diff
@@ -219,7 +226,15 @@ def __init__(self, loader: "BlueprintLoader", node: SequenceNode) -> None:
     def resolve(self, entry: BlueprintEntry, blueprint: Blueprint) -> Any:
         query = Q()
         for cond in self.conditions:
-            query &= Q(**{cond[0]: cond[1]})
+            if isinstance(cond[0], YAMLTag):
```
BeryJu, line 229: I think naming it `query_key` and `query_value` is a bit cleaner than c_0 and c_1

website/developer-docs/blueprints/v1/tags.md
```diff
@@ -10,7 +10,11 @@ If no matching entry can be found, an error is raised and the blueprint is inval
 
 #### `!Find`
 
-Example: `configure_flow: !Find [authentik_flows.flow, [slug, default-password-change]]`
+Example:
+
+`configure_flow: !Find [authentik_flows.flow, [slug, default-password-change]]`
+
+`configure_flow: !Find [authentik_flows.flow, [!Context property_name, !Context property_value]]`
```
BeryJu, line 17: I wonder if these more complicated tags can be broken up into multiple lines to make it a bit easier to read

BeryJu (approved):

authentik/blueprints/v1/common.py
sdimovv: Resolved

PR #4148, opened 2022-12-04
LDAP gracefully deleting sessions on authentik

d-schiffner:
# Details
This PR resolves an issue with the current LDAP implementation not executing a logout flow on the authentik server. This results in an ever increasing list of `authenticated_sessions` and resulting in unused cookie sessions on the server.

Furthermore this PR resolves a nasty race condition that occurs if an sessions is invalidated and in parallel a login occurs. The result is that a session exists where the `Pk=0` which in turn is not able to answer any search queries. For users this is inconvenient but this also occurs for service accounts as well. Then the ldap is not usable until the next invalidation.

### New Features
* Adds a logout execution flow (should be better parameterized)
* Adds the option to delete the flags assigned to a user
* Adds ttlcache expiry handling
* Extends non-caching login flow to immediately invalidate a session after login (cookie will be thrown away by ldap anyways)
* Adds locks for each user when using session binder

### Breaking Changes
No breaking changes are known

## Additional
* I think it would be good to make the session TTL configurable in the LDAP server. Ideally this would be pushed by the authentik server to the ldap outpost and be part of the config

Feedback is very much welcome :)

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -6,3 +6,6 @@ htmlcov
 dist/**
 build/**
 build_docs/**
+__pycache__
+.venv
+gen-images
--- a/.gitignore
+++ b/.gitignore
@@ -202,3 +202,4 @@ media/
 
 .idea/
 /gen-*/
+.vscode
--- a/cmd/ldap/server.go
+++ b/cmd/ldap/server.go
@@ -52,12 +52,6 @@ func main() {
 
 	ex := common.Init()
 	defer common.Defer()
-	go func() {
-		for {
-			<-ex
-			os.Exit(0)
-		}
-	}()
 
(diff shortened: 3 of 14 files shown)


BeryJu:
@d-schiffner I appreciate your PR, I have don't know how I completely forgot about ldap unbind when I initially created this provider. I linked a PR above that implements the logout (I wanted to implement it a bit differently than you did here). I'll need to have another look through this PR to hopefully fix the race condition you're talking about

d-schiffner:
@BeryJu I have been quite busy lately. No hard feelings here :)
I can have a look into your implementation and look for the race condition. It was quite tricky to track down, but I know what to look for (at least, I think I recall ;) )

PR #4157, opened 2022-12-05
Replace Dropdown fields in Providers/LDAP and Tokens to support filtering and more than 100 entries.

### LINKED ISSUE #3432: LDAP Provider: Search Group dropdown field does not show all entries after importing 180 groups from LDAP source
**Describe the bug**

After importing all users and 180 groups from an LDAP source, I cannot see anymore the 'authentik Admins' group in the dropdown field.  The imported groups all have uppercase first characters and the dropdown's sort is case sensitive.

**To Reproduce**
1. Import from LDAP source 180 groups all having and uppercase first character.
2. Create an LDAP Provider or edit an existing one
3. Click on the Search Group dropdown
4. Not all groups are displayed by the dropdown.

**Expected behavior**
All groups are shown/displayed for selection.

**Screenshots**
Screenshot from 2022-08-17 14-30-06

 - authentik version: 2021.8.1
 - Deployment: docker-compose

UnholyRope:
# Details
Resolves #3432 

### New Features
* Providers/LDAP: Replaces the "Search group" dropdown with the "ak-search-select" dropdown.
* Tokens: Replaces the "User" dropdown with the "ak-search-select" dropdown.

## Additional
Both forms used a dropdown menu which only displays the first 100 users/groups. Additionally those dropdowns didn't accept input, which made it impossible to select certain users/groups.
This pull request replaces those dropdowns with the  "ak-search-select" dropdown which is defined in [SearchSelect.ts](https://github.com/goauthentik/authentik/blob/main/web/src/elements/SearchSelect.ts) and already in use in [PolicyBindingForm.ts](https://github.com/goauthentik/authentik/blob/main/web/src/admin/policies/PolicyBindingForm.ts).


BeryJu:
The reason why ak-search-select isn't used everywhere is because it doesn't render correctly within a modal (the dropdown is rendered within the modal and not on top of it), which almost every form in authentik is in. There's probably some z-index or flexbox magic that needs to be done, but I haven't figured out yet what is needed

PR #4167, opened 2022-12-07
blueprints: Added conditional entry application

astro-stan:
# Details
N/A

### New Features
* Adds `!Condition` tag to the list of supported custom blueprint tags.
* Adds an optional `conditions` attribute to blueprint entries, allowing conditional application

### Breaking Changes
* N/A

## Additional
The use of `!Condition` tag with the `conditions` attribute adds flexibility.

For example, consider the following simplified blueprint:
```
version: 1
metadata:
  name: Test
context:
  enrollment_flow_slug: my-enrollment-flow
  enrollment_flow_blueprint_name: My Enrollment Flow Blueprint
entries:
- model: authentik_blueprints.metaapplyblueprint
  conditions:
  -  !Conditions [AND, !Context enrollment_flow_blueprint_name] # In this simple case just the context tag without the conditions tag is enough, but to illustrate how to build complex conditions  the `!Condition` tag is used.
  attrs:
    identifiers:
      name: !Context enrollment_flow_blueprint_name
    required: true
- model: authentik_stages_identification.identificationstage
  identifiers:
    name: test
  attrs:
    enrollment_flow: !Find [authentik_flows.flow, [slug, !Context enrollment_flow_slug]]
    ...
```

This way, the user may choose if they want an enrollment flow. If the user does not want such flow, they can set both context variables to `null`. This will skip the meta blueprint application and will set the enrollment flow to null.  However, if the user wants an enrollment flow, they can set the context variables appropriately and it will be applied as usual.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -24,7 +24,8 @@
         "!Find sequence",
         "!KeyOf scalar",
         "!Context scalar",
-        "!Format sequence"
+        "!Format sequence",
+        "!Condition sequence"
     ],
     "typescript.preferences.importModuleSpecifier": "non-relative",
     "typescript.preferences.importModuleSpecifierEnding": "index",
--- a/authentik/blueprints/tests/fixtures/conditions_fulfilled.yaml
+++ b/authentik/blueprints/tests/fixtures/conditions_fulfilled.yaml
@@ -1,0 +1,21 @@
+version: 1
+entries:
+- identifiers:
+    name: "%(id1)s"
+    slug: "%(id1)s"
+  model: authentik_flows.flow
+  conditions:
+  - true
+  attrs:
+    designation: stage_configuration
+    title: foo
+- identifiers:
+    name: "%(id2)s"
+    slug: "%(id2)s"
+  model: authentik_flows.flow
+  conditions:
+  - true
+  - true
+  attrs:
+    designation: stage_configuration
+    title: foo
--- a/authentik/blueprints/tests/fixtures/conditions_not_fulfilled.yaml
+++ b/authentik/blueprints/tests/fixtures/conditions_not_fulfilled.yaml
@@ -1,0 +1,21 @@
+version: 1
+entries:
+- identifiers:
+    name: "%(id1)s"
+    slug: "%(id1)s"
+  model: authentik_flows.flow
+  conditions:
+  - false
+  attrs:
+    designation: stage_configuration
+    title: foo
+- identifiers:
+    name: "%(id2)s"
+    slug: "%(id2)s"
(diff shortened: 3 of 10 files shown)


sdimovv:
Not sure why those 2 checks don't pass. I just added a few lines, which are all covered with tests. Not sure how it is possible to "decrease project coverage".

sdimovv:
I was thinking of adding conditionals as well, but decided to add this tag first. Then a conditional (if) tag can also be added, that will look something like this:

```
!If [ <condition>, <yaml when true>, <yaml when false>
```
The reason for this separation is so the <condition> part can accept the `!AsBool` tag, which allows for a very flexible definition of what `true` or `false` means

sdimovv:
Actually, thinking about this more, I can ditch the `!AsBool` tag entirely and implement the complex conditions logic into an `If/Condition` tag. So for example:

```
!If [123 , true, false] 
!If [!Context name , true, false] 
!If [[AND, ...], true, false] 
!If [[OR, ...], true, false] 
...
```

I can have a go at this sometime next week.

sdimovv:
Hi,

sorry for the late reply. Holidays...

I am not sure I understood what you mean. 

Do you want me to add a "conditions" property to each blueprint entry, that will act similarly to the "enabled" property of each blueprint? If so I am not sure why either an `!If` or `!Condition` tags are needed, since in this case (assuming all conditions in the conditions tag should equal `true`) we only need to evaluate if a list of expressions equals `true` or `false`. Thus the `!AsBool` tag should suffice.

sdimovv:
Ok, so this is what I came up with. 

1. Renamed the '!AsBool` tag to `!Condition`
2. Added an optional `conditions` attribute to each `BlueprintEntry`

In terms of yaml this looks something like this:

```
version: 1
entries:
- identifiers: ...
  attrs: ...
  model: ...
  conditions:
  - true
  - false
  - text
  - !Condition [AND, ...] # Same as AsBool
```

If all conditions in the `conditions` attribute match (or no conditions are provided), then the entry is processed as normal. Otherwise, the entry is skipped, similarly to the `state: created` when the entry already exists.

I called both the list of condiitons and the custom yaml tag "condition(s)" because "if" did not fit well in my opinion, but I can change if you disagree. Perhaps instead of `!Condition` it can be `!ComplexCondition` since that is what it technically is.

sdimovv:
Hi @BeryJu ,

please advise if there is something that needs reworking/adding/improving.

Apologies for being pushy. I am very keen to get this merged before 2022.12 comes out and with the holidays approaching there isn't much time left for me to work on this.

sdimovv:
Cheers!

PR #4210, opened 2022-12-14
feat(docs/nextcloud): Updated docs

samip5:
It was missing the proper syntax for urls with index.php in between.

### DIFF
--- a/website/integrations/services/nextcloud/index.md
+++ b/website/integrations/services/nextcloud/index.md
@@ -24,23 +24,21 @@ In case something goes wrong with the configuration, you can use the URL `http:/
 
 The following placeholders will be used:
 
--   `nextcloud.company` is the FQDN of the NextCloud install.
+-   `nextcloud.company` is the FQDN of the Nextcloud install.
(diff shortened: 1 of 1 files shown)


website/integrations/services/nextcloud/index.md
```diff
@@ -29,10 +29,10 @@ The following placeholders will be used:
 
 Create an application in authentik and note the slug you choose, as this will be used later. In the Admin Interface, go to Applications->Providers. Create a SAML provider with the following parameters:
 
--   ACS URL: `https://nextcloud.company/apps/user_saml/saml/acs`
+-   ACS URL: `https://nextcloud.company/index.php/apps/user_saml/saml/acs`
```
BeryJu, line 32: I think this should be a note underneath instead of changing the URLs, as this seems to depend on the nextcloud installation and if URL rewrites are enabled, something like this below the parameters:

```markdown
:::note
Depending on your nextcloud configuration, you might need to use `https://nextcloud.company/index.php/` instead of `https://nextcloud.company/`
:::
```

BeryJu (changes requested):

PR #4264, opened 2022-12-22
Blueprints: Added `!If` tag

astro-stan:
# Details
N/A

### New Features
* Adds the `!If` tag

### Breaking Changes
N/A

## Additional
As briefly discussed in #4167, I have added the ability to conditionally include YAML in blueprints.

This adds a lot of flexibility to the overall blueprint structure. 

The most basic example I can think of would be something like this:
```
entries:
...
- model: ...
  state: !If [ !Context i_want_this, present, absent]
  ...
```
Thus allowing the user to conditionally add or delete entries.


sdimovv:
I created a GPG signature, but could not find a way to sign the commits I already pushed, so I soft-reset them and commited them again with my GPG signature. Hope that is okay.

sdimovv:
All the changes I made are covered by tests. Codecov failing should be a false positive.

sdimovv:
Hi @BeryJu,

Would it be possible to merge this in before 2022.12?

PR #4266, opened 2022-12-22
blueprints: Multi documents and multi OCI layer

### LINKED ISSUE #4262: Easier fetching of blueprints in OCI registries
From the [blueprint docs for OCI storages](https://goauthentik.io/developer-docs/blueprints/#storage---oci) I gather that the user can set the path of a blueprint to an URL instead of a local path to fetch a specific blueprint from an OCI registry.

I looked at this in the UI and I don't see a way to type in the path (the path field is a dropdown list with local paths of known blueprint yaml files). Maybe I am overlooking something, but even if this is possible, it seems very tedious to do because it needs to be done for every blueprint and there may be hundreds of them in a registry.

As an alternative I want to propose the following:

Add a new environment variable(s) that can be used to specify OCI registry URLs (with/without credentials) but for the whole registry and not for an individual blueprint. Moreover, have the environment variables specify under what folder name to store the downloaded blueprints in the `/blueprints` directory. 

Then periodically delete the folder with the blueprints entirely and fetch them again from the registry. Then (re)apply them.

1. Have an separate container to download the OCI registry and place it into a folder accessible to authentik.
2. Manually download the OCI registry and place the files in an accessible to authentik directory.

**Additional context**
Alternatively the UI can have a new dialog to specify this (instead of using environment variables). But I would personally prefer the environment variable approach, as this way authentik can be bootstrapped to apply a custom set of blueprints on the first startup. Basically do something like:
1. Define registry with blueprints to download in an environment variable
2. Map them to be placed under `/blueprints/default`
3. Start authentik
4. authentik starts up fully configured, no need to delete flows/stages/etc created by the default blueprints.

BeryJu:
closes #4262

### DIFF
--- a/website/developer-docs/blueprints/index.md
+++ b/website/developer-docs/blueprints/index.md
@@ -46,10 +46,19 @@ To fetch blueprints from a private registry with authentication, credentials can
 
 Blueprints are re-fetched each execution, so when using changing tags, blueprints will automatically be updated.
 
-To push a blueprint to an OCI-compatible registry, [ORAS](https://oras.land/) can be used with this command
+To push a blueprint to an OCI-compatible registry, [ORAS](https://oras.land) can be used with this command
 
 ```
-oras push ghcr.io/<username>/blueprint/<blueprint name>:latest <yaml file>:application/vnd.goauthentik.blueprint.v1+yaml
+oras push ghcr.io/<username>/blueprint/<blueprint name>:latest \
+    <yaml file>:application/vnd.goauthentik.blueprint.v1+yaml
+```
+
+Starting with authentik 2022.12, blueprints hosted in OCI registries can contain multiple blueprints in a single package. To push an artifact with multiple files, just append more files to the command above:
+
+```
+oras push ghcr.io/<username>/blueprint/<blueprint name>:latest \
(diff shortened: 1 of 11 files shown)


sdimovv:
Bump on this so it does not close.

website/developer-docs/blueprints/index.md
```diff
@@ -46,10 +46,19 @@ To fetch blueprints from a private registry with authentication, credentials can
...
-To push a blueprint to an OCI-compatible registry, [ORAS](https://oras.land/) can be used with this command
+To push a blueprint to an OCI-compatible registry, [ORAS](https://oras.land) can be used with this command
 
 ```
-oras push ghcr.io/<username>/blueprint/<blueprint name>:latest <yaml file>:application/vnd.goauthentik.blueprint.v1+yaml
+oras push ghcr.io/<username>/blueprint/<blueprint name>:latest \
+    <yaml file>:application/vnd.goauthentik.blueprint.v1+yaml
+```
+
+Starting with authentik 2022.12, blueprints hosted in OCI registries can contain multiple blueprints in a single package. To push an artifact with multiple files, just append more files to the command above:
```
tanberry, line 56: @Jens did you mean ".. append more files to the command below"? (not above)

tanberry (changes requested):
Just the one question for the docs side.

PR #4338, opened 2023-01-02
blueprints: Add `!Enumerate`, `!Value` and `!Index` tags

astro-stan:
# Details
N/A

### New Features
* Add `!Enumerate`, `!Value` and `!Index` tags

### Breaking Changes
N/A

## Additional
These tags collectively make it possible to iterate over objects which support iteration. Any iterable Python object is supported. Such as sequences (`[]`), mappings (`{}`) and even strings.

Example:

```
configuration_stages: !Enumerate [
    !Context map_of_totp_stage_names_and_types,
    SEQ, # Output a sequence
    !Find [!Format ["authentik_stages_authenticator_%s.authenticator%sstage", !Index 0, !Index 0], [name, !Value 0]] # The value of each item in the sequence
]
```

The above example will resolve to something like this:

```
configuration_stages:
- !Find [authentik_stages_authenticator_<stage_type_1>.authenticator<stage_type_1>stage, [name, <stage_name_1>]]
- !Find [authentik_stages_authenticator_<stage_type_2>.authenticator<stage_type_2>stage, [name, <stage_name_2>]]
```
 
See doc changes for more detail and examples.

### Implementation detail
The changes add a new base class called `YAMLTagContext`, which introduces the ability for tags to have their own contexts.

In the case of this PR, the tag context is used in the `!Enumerate` tag, to hold the currently iterated item of a sequence/mapping or any other iterable.

Accessing this context from other (nested) tags however proved challenging. I managed to keep the overall `.resolve()` interface the same, but I had to introduce a new hidden attribute (`self.__tag_contexts`) and private method (`self._get_tag_context()`) to the `BlueprintEntry` class.  Those can be used to access any parent tag's context from a tag that is nested inside it. The `_get_tag_context()` method  is private because it is only useful during tag resolution. It returns an empty list when not resolving tags.

### DIFF
--- a/authentik/blueprints/tests/fixtures/tags.yaml
+++ b/authentik/blueprints/tests/fixtures/tags.yaml
@@ -3,6 +3,9 @@ context:
     foo: bar
     policy_property: name
     policy_property_value: foo-bar-baz-qux
+    sequence:
+    - foo
+    - bar
 entries:
     - model: !Format ["%s", authentik_sources_oauth.oauthsource]
       state: !Format ["%s", present]
@@ -92,6 +95,41 @@ entries:
                   ]
               if_true_simple: !If [!Context foo, true, text]
               if_false_simple: !If [null, false, 2]
+              for: !For [
+                  !Context sequence,
+                  [
+                      !For [
+                          !ForItem 1,
+                          {
+                              key1: [
+                                  !Format ["str: %s, letter: %s", !ForItem 1, !ForItem 0],
+                                  !ForItem 1,
+                                  !ForItem 0,
+                                  !For [
+                                      !ForItem 2,
+                                      {
+                                        inner_inner_for_item: !Format ["%s", !ForItem 0]
+                                      }
+                                  ]
+                              ],
+                              key2: !ForItem 0
+                          }
+                      ],
+                      !For [
+                          !ForItem 1,
+                          {
+                              key1: {
+                                  inner_key1: !Format ["letter: %s, str: %s", !ForItem 0, !ForItem 1],
+                                  inner_key2: !ForItem 0,
+                                  inner_key3: !ForItem 1
+                              },
+                              key2: !ForItem 0
+                          }
+                      ],
+                      !ForItem 0,
+                      !Format ["%s-%s", !Context foo, !ForItem 0]
+                  ]
+              ]
       identifiers:
           name: test
       conditions:
--- a/authentik/blueprints/tests/fixtures/tags.yaml
+++ b/authentik/blueprints/tests/fixtures/tags.yaml
@@ -98,53 +98,53 @@ entries:
                   ]
               if_true_simple: !If [!Context foo, true, text]
               if_false_simple: !If [null, false, 2]
-              for_mapping: !For [
+              for_mapping: !Enumerate [
                   !Context mapping,
                   MAP,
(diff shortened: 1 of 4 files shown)


sdimovv:
This should be good to go now.

@BeryJu can you spare a few minutes to take a look at this, please? Let me know if you are happy with it.

BeryJu (approved):
LGTM

(Allthough I think this is probably enough custom YAML tags for now, I don't think blueprints need to become a fully turing complete language :D)

btw @sdimovv would you happen to be interested in working on authentik full time?

sdimovv:
> I don't think blueprints need to become a fully turing complete language :D

Ah, I was afraid you might say that :D. I am working on a blueprint registry (very early stages, and no and docs yet, but the repo is public if you want to check it out). The ultimate goal is to make it easy to deploy authentik instances and customize them to your needs through blueprint context. Which is why I came up with so many new features for the blueprints.

Regarding the job opportunity, I do have a full time job unfortunately. But I might be interested in freelance work if that is on the table? I can email you or message you on LinkedIn in if interested.

BeryJu:
> > I don't think blueprints need to become a fully turing complete language :D
> 
> Ah, I was afraid you might say that :D. I am working on a blueprint registry (very early stages, and no and docs yet, but the repo is public if you want to check it out). The ultimate goal is to make it easy to deploy authentik instances and customize them to your needs through blueprint context. Which is why I came up with so many new features for the blueprints.

I've actually been planning a similar thing, the next big thing I want to add for this is either allowing people to specify a JSON schema for the context, or (the more likely option) allow people to specify prompts (in the same format as the `prompt` stage would), which would then be shown when importing a blueprint, through a wizard

> Regarding the job opportunity, I do have a full time job unfortunately. But I might be interested in freelance work if that is on the table? I can email you or message you on LinkedIn if interested.

I kinda expected that to be the response, feel free to send contact details to jens@goauthentik.io

sdimovv:
> the next big thing I want to add for this is either allowing people to specify a JSON schema for the context, or (the more likely option) allow people to specify prompts (in the same format as the prompt stage would),
HA! I have been thinking about something similar (borrowed from Truenas SCALE's "question.yaml"), but I did not want to push it hehe. 

Also, regarding the turing completeness of the blueprints, I was thinking if perhaps a version 2 using Django's template language makes sense. Only problem is that is that it will be difficult to enforce the correct indentation everywhere.

BeryJu:
> Also, regarding the turing completeness of the blueprints, I was thinking if perhaps a version 2 using Django's template language makes sense. Only problem with that (I think) is that it will be difficult to enforce the correct indentation everywhere.

I'm generally not a fan of using templating languages to modify YAML (hence I also really dislike helm charts) so I would prefer these tags to that, the other option would obviously be to allow expressions in the same way as policies do, but at some point in the future when there's more of a community around blueprints, I'd rather those not get direct code execution (the other thing I want to do for that is for the validation to return a plan, somewhat like terraform), so people can preview what will be done when applying a blueprint

sdimovv:
> not a fan of using templating languages to modify YAML

I personally don't have strong opinions on YAML templating, but I can understand why you might dislike it.  

> not get direct code execution

Oh yes, definitely a bad idea.

> the other thing I want to do for that is for the validation to return a plan, somewhat like terraform

I like this. I will definitely be interested to see it implemented.

PR #4422, opened 2023-01-12
blueprints: add mapping to sync active state from userAccountControl

liske:
# Details
*None*

### New Features
* Adds a blueprint to sync the [ADS_UF_ACCOUNT_DISABLE](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-adts/dd302fd1-0aa7-406b-ad91-2a6b35738557) bit from `userAccountControl` in Active Directory to user's active property in authentik.

### Breaking Changes
*None*

## Additional
I think it is reasonable for an application with Active Directory user sync to also support syncing the account locked/disable state. With this blueprint it should become more easy for others to get this property synced.

### DIFF
--- a/blueprints/system/sources-ldap.yaml
+++ b/blueprints/system/sources-ldap.yaml
@@ -53,6 +53,14 @@ entries:
       object_field: "attributes.sn"
       expression: |
         return list_flatten(ldap.get('sn'))
+  - identifiers:
+      managed: goauthentik.io/sources/ldap/ms-active
+    model: authentik_sources_ldap.ldappropertymapping
+    attrs:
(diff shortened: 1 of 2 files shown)


BeryJu:
This is already imported from AD into authentik, see https://github.com/goauthentik/authentik/blob/main/authentik/sources/ldap/sync/vendor/ms_ad.py#L66

liske:
> This is already imported from AD into authentik, see https://github.com/goauthentik/authentik/blob/main/authentik/sources/ldap/sync/vendor/ms_ad.py#L66

Ops, sorry - I thought that only the properties are synchronized with a configured mapping - should have tested it before building a custom mapping :facepalm:

PR #4457, opened 2023-01-17
providers/proxy: add setting to intercept authorization header

BeryJu:


samip5 (commented):
What happens if that's ticked and I want to access it w/o basic auth? I would imagine my browser would prompt for credentials?

Please keep in mind that only specific endpoints require basic auth for tvheadend + VLC setup, but the main UI still could be accessed with the normal redirect auth.

BeryJu:
> What happens if that's ticked and I want to access it w/o basic auth? I would imagine my browser would prompt for credentials?
> 
> Please keep in mind that only specific endpoints require basic auth for tvheadend + VLC setup, but the main UI still could be accessed with the normal redirect auth.

This basically has the same function as the header used to, if this is set and an Authorization header is set that is invalid, it will show a 401, if no authorization header is set a normal redirect is done

If this is not enabled, the authorization header is still checked but if it's not set or invalid to authentik a redirect is done

samip5:
That sounds confusing.

Basically if it's the first time I'm using the URL with VLC, it will try first without any auth, which that sounds like would result in a loop of redirect.

Am I understanding it correctly?

BeryJu:
It will indeed result in a redirect, as there's no way for the outpost to know whether the client can handle redirects (and is in the context of a browser) or not

If this option would make the Authorization Header required for every request (which is the only other method), then browser-based flows would always get an Authorization prompt, which kinda defats the point of SSO

PR #4482, opened 2023-01-22
Fix TOTP issuer mangling

jmptbl:
The TOTP issuer field gets populated from the tenant's branding title, but it's passed through django.utils.text.slugify, which excessively strips many useful characters from the tenant's branding title.

The issuer field only needs to be URL quoted, which is what this patch does.

Ideally the branding title should be written unquoted to OTP_TOTP_ISSUER, and django_otp will perform the URL quoting itself. I'm not sure how to set OTP_TOTP_ISSUER though.


BeryJu:
OTP_TOTP_ISSUER is set here https://github.com/goauthentik/authentik/blob/main/authentik/stages/authenticator_totp/settings.py#L6, because django_otp doesn't allow setting it dynamically when getting the URL, which also means that authentik has to make sure it's a valid URL

PR #4493, opened 2023-01-23
website/docs: improve traefik standalone docs

support-tt:
Example for Authentik Single Application Proxy with Service example because this was unclear for many users and if you dont create a middleware for every application you get the error "no app for hostname". 

Signed-off-by: support-tt <61587422+support-tt@users.noreply.github.com>

# Details
Resolves # the issue that many user asked in discord on how to create single application proxy with traefik

### New Features
* Adds documentation for single application proxy with traefik standalone.

### Breaking Changes
* none


BeryJu:
I think it would make more sense to update https://github.com/goauthentik/authentik/blob/main/website/docs/providers/proxy/_traefik_standalone.md instead of adding a new file

support-tt:
> I think it would make more sense to update https://github.com/goauthentik/authentik/blob/main/website/docs/providers/proxy/_traefik_standalone.md instead of adding a new file

Yes that also make sense. My example is a full example for "single application" with the service for authentik and the application. 
If someone wants to use traefik also for the main route to authentik then only a route/service for this is needed.

tanberry:
HI @support-tt thanks for your PR. I'd like to look at it more (and learn more), then probably ask some questions... but let's make some progress on this PR, either way.

support-tt:
I checked my commit again its for the "Single application" from the website. 
At the moment there are no examples for Single Application forward. Only for Domain level for traefik(at least for traefik). 
I think it would make sense to add examples for Single application for the Reverse proxys too. 
The user can then decide to use Domain OR Single Application level. Unfortenetly im not familiar with caddy and nginx but maybe someone can add examples for them as well.

Im sorry if im doing something wrong with the commits its my first try to commit something.

tanberry:
Thanks so much for your reply, @support-tt ! No worries, you are doing everything right! I am very new to authentik and still learning the product. ;-) I will be mostly focused on Docs, and I am very appreciative of your contribution here. Now let's figure out how best to implement it. ;-)

Since we do already have a place in the documentation for

tanberry:
Screenshot 2023-03-09 at 10 13 55 AM

support-tt:
> Screenshot 2023-03-09 at 10 13 55 AM

I would create the Nginx Traefik etc again under Single application like 
image

So the user see on the first view what is a example for Single application and what for Domain level.

tanberry:
Aha, thank you! I now see that it is perhaps due to formatting (H1 versus H2) that I wasn't understanding the current structure... the way it looks now, the sections under Domain level look like peers, not children of the Domain level entry!

So I see what you are proposing now, I think. Thanks for the screenshot of your mockup.

Single application
     NGINX
     Traefik
Domain-level
     NGINX
     Treafik
     Envoy
     Caddy     


Correct?

support-tt:
> Aha, thank you! I now see that it is perhaps due to formatting (H1 versus H2) that I wasn't understanding the current structure... the way it looks now, the sections under Domain level look like peers, not children of the Domain level entry!
> 
> So I see what you are proposing now, I think. Thanks for the screenshot of your mockup.
> 
> Single application NGINX Traefik Domain-level NGINX Treafik Envoy Caddy
> 
> Correct?

Yes correct but then we need some exaples for nginx and traefik docker-compose and ingress my example is for traefik standalone.

support-tt:
> , or does one HAVE to have the other tw

ingre

support-tt:
> Can one use only the part you are providing, the Traefik standalone, or does one HAVE to have the other two parts (traefik docker-compose and ingress)?

I misklicked .. dont wanted to close this issue -,- sorry
ingress is for Kubernetes setups
docker-compose with tags is a alternative way to standalone but both are working.

tanberry:
OK, thanks again for helping me better understand. So, I suggest that we do indeed proceed with a new section for your single-application implementation, with @BeryJu review. 

I will need to do some work to change heading levels (## versus ###) on https://github.com/goauthentik/authentik/blob/main/website/docs/providers/proxy/forward_auth.mdx. 

And either you or me will need to edit that same file (forward_auth.mdx) to add your new section (forward_auth#traefik-standalone-single). Let me know if you want to add that to your existing PR, or I can do it after we merge your PR.

Sound OK?

BeryJu:
Hi, I had a look through the config and the website, and noticed the website doesn't make it clear at all that the configs listed are for single application or for both, and since we already have a traefik config I've updated that with your PR

We'll update the structure of the website soon to make it clearer that the configs listed there aren't exclusive for domain auth

tanberry:
Thanks again for your PR, @support-tt, and also thanks @BeryJu for your solution. So with this merge, the docs now cover the "single-app" implementation only. I'll open a later PR to help make the "single app vs domain-level" distinction more clear.

PR #4648, opened 2023-02-09
crypto: Backend Support for LetsEncrypt ACME HTTP-01 Validation

### LINKED ISSUE #1835: Lets Encrypt integration
Either DNS based (probably in Go) or via HTTP Challenge (http challenge is sent to and served by outpost)

ignore-prior-instructions:
# Details
* Resolves #1835

### New Features
* Adds the capability to respond to ACME HTTP-01 challenges without a reverse proxy.

### Breaking Changes
* None

## Additional
LetsEncrypt validates domain ownership providing the user with a challenge token in the form of a file.  The file contents should be returned for HTTP GET requests located at: 
`http://<YOUR_DOMAIN>/.well-known/acme-challenge/<TOKEN>`.

- This PR creates a model to store acme http-01 challenge token, host, and response information to respond to LetsEncrypt validation challenges.  This should be set by the user.
- The `http://<YOUR_DOMAIN>/.well-known/acme-challenge/*` endpoint does not require authentication.
- Host Headers are checked to respond with the correct challenge token based on the client request's `HTTP_HOST` header value.
  - This allows for multiple domains to be validated from the same Authentik installation depending on the hostname being validated.

### TODO
- I don't have the skillset to create frontend pages for managing the new model and will need assistance to make this manageable by users via the web interface.
  - Instances can be created via Blueprint.
- REST API is not implemented in this PR (though the basic serializer is).

## Local Demonstration
```csh
curl localhost:8000/.well-known/acme-challenge/cRwdSWr
# Error 404

vim blueprints/example/letsencrypt-http-01-verification.yaml
# Edit `host` field to localhost:8000

ak apply_blueprint example/letsencrypt-http-01-verification.yaml
# Instance Created

curl localhost:8000/.well-known/acme-challenge/cRwdSWr
# Returns: AzpUGtCMuZvRpntnXBAqRdUZDoQAVyzQ
```
### References
LetsEncrypt challenge documentation: https://letsencrypt.org/docs/challenge-types/#http-01-challenge


BeryJu:
Hi, I appreciate the PR but I'm struggling to see the full usecase for this, as one would need some custom tooling to create the challenge within authentik, so there'd be some glue thats required between certbot for example and authentik.

sover02:
> Hi, I appreciate the PR but I'm struggling to see the full usecase for this, as one would need some custom tooling to create the challenge within authentik, so there'd be some glue thats required between certbot for example and authentik.

That makes sense.  I ran into this usecase when trying to use http-01 validation with certbot for a to use a letsencrypt certificate with Authentik deployed via docker-compose.  In that scenario, there was no elegant way for me to input the validation token and have it served to the validation request.

I can see a future state where Authentik is able to request the certificate through a configuration flag or the web interface and provide the validation without users needing to make DNS changes or run authentik through a reverse proxy for the validation endpoint.  In the end state, this feature would make it really easy for new users who are testing the product to opt for a "green" lock right out of the box.

In the state of this PR, you will still have to manually request and update the challenge every 90 days, but it does at least unblock the capability, and would be easy to create a script to do so for users. I can provide a script in contrib that does this if it's helpful.

I did look into the possibility reimplementing parts of the certbot workflow within the app, but the level of effort is significant. I think it's worth chasing for a green "secure-https-by-default" option and would improve the first-time user experience significantly.  This PR is a first step and serves to unblock users trying to use letsencrypt without DNS changes.

PR #4655, opened 2023-02-09
providers/saml: fix mismatched SAML SLO Urls

astro-stan:
# Details
According to the source code:

https://github.com/goauthentik/authentik/blob/b225f6f3ff6cbd5b0d69304c36ed30b81a45aa5d/authentik/providers/saml/urls.py#L25-L29

Notice no `/binding/`

### New Features
N/A

### Breaking Changes
N/A

## Additional
N/A

### DIFF
--- a/authentik/providers/saml/urls.py
+++ b/authentik/providers/saml/urls.py
@@ -23,12 +23,12 @@ urlpatterns = [
     ),
     # SLO Bindings
     path(
-        "<slug:application_slug>/slo/redirect/",
+        "<slug:application_slug>/slo/binding/redirect/",
         slo.SAMLSLOBindingRedirectView.as_view(),
         name="slo-redirect",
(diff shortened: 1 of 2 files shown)


BeryJu:
Good catch! The URL is actually supposed to have a `/binding/` part in the URL to match the SSO URLs (even though in the future I'd like to make it a single URL since these URLs are quite unwieldy but that can wait)

sdimovv:
Thanks! Yea, a unified URL in the future would be even better. 

But regarding the current state of things - are you going to merge this? If you prefer, I can (or at least try to) submit a patch here that fixes the URL instead?

BeryJu:
Yeah I'd prefer a PR to fix the URLs to be the same everywhere

sdimovv:
Should be good now

PR #4717, opened 2023-02-19
Allow tokens to be set to a specific value via the API

traverseda:
Necessary for blueprints to be able to implement features like `AK_ADMIN_PASS`, that is setting tokens via ENV variables.  This is necessary in order to do fully automated installs of outposts.

### DIFF
--- a/authentik/core/api/tokens.py
+++ b/authentik/core/api/tokens.py
@@ -53,6 +53,7 @@ class TokenSerializer(ManagedSerializer, ModelSerializer):
             "intent",
             "user",
             "user_obj",
+            "key",
             "description",
             "expires",
             "expiring",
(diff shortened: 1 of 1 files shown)


BeryJu:
Hmm, I agree with this being required for certain setups via Blueprints, but the ability to edit the key directly is currently purposefully not allowed, as tokens circumvent any kind of MFA or extra validation, so they are seen as inherently more secure and as such I'd like to keep the current behaviour of making them (by default) 60 chars.

For this usecase, it might work creating a token via blueprint without token, and then getting the token via the `!Find` tag with the token identifier, but it still wouldn't allow for all the usecases

traverseda:
>For this usecase, it might work creating a token via blueprint without token, and then getting the token via the !Find tag with the token identifier, but it still wouldn't allow for all the usecases

Definitely wouldn't work for me use case, which is using one docker-compose file to automatically deploy authentik along side some other apps that authenticate against it. 

Worth noting that this would only apply to the API, not the user interface, so it's unlikely that a user who didn't know what they were doing would set an improper token.

BeryJu:
@traverseda The PR linked above allows for setting a token's key via blueprints while disallowing it via the standard API

PR #4733, opened 2023-02-20
website/integrations: Mastodon integration

Tealk:
# Details
Extension of the instructions by the service Mastodon

### New Features
* Adds Mastodon instructions


website/integrations/services/mastodon/index.md
```diff
@@ -0,0 +1,63 @@
...
+Add the following
+
+```
+OIDC_ENABLED=true
+OIDC_DISPLAY_NAME=authentik
+OIDC_DISCOVERY=true
+OIDC_ISSUER=< OpenID Configuration Issuer>
+OIDC_AUTH_ENDPOINT=https://authentik.company/application/o/authorize/
+OIDC_SCOPE=openid,profile,email
+OIDC_UID_FIELD=preferred_username
```
BeryJu, line 51: I don't know if this field is also used as a username in mastodon, if it isn't it makes more sense to use `sub` here as the username in authentik might change

BeryJu (changes requested):

website/integrations/services/mastodon/index.md
Tealk: Can you explain in more detail what you mean by " to use `sub `here"?

BeryJu: the `sub` field in the OIDC Token is a unique identifier that is (by default) generated based on the User ID and the hashed secret key, which unlike the user's username cannot change

with the current setting, if a user changes their username, they would get a new user in mastodon

website/integrations/services/mastodon/index.md
```diff
@@ -0,0 +1,63 @@
+---
+title: Mastodon
+---
+
+<span class="badge badge--secondary">Support level: Community</span>
+
+## What is Service Name
```
BeryJu, line 7: replace placeholder

BeryJu (approved):

PR #4750, opened 2023-02-21
providers/ldap: Improve compatibility with LDAP clients

### LINKED ISSUE #4158: Is there a way to change LDAP Provider goauthentik.io attributes
Is there a way to change LDAP Provider goauthentik.io attributes? it is causing me issues with ldapsearch command, apache directory studio, and many other places.

by goauthentik.io attributes I mean the following attributes:
goauthentik.io/ldap/active
goauthentik.io/ldap/superuser
goauthentik.io/ldap/virtual-group

The slashes and the dot in those attributes are causing issues.

ldapsearch command returns:
ldap_search_ext: Bad search filter (-7)
jellyfin ldap plugin returns:
Failure: User Filter: Invalid character "/" in attribute description
Apache Directory Studio returns:
 ERR_13312_ITEM_EXPECTED An item is expected

I tried to escape with the slashes and dot but without success.

if anyone knows a workaround please share it with us.

Thanks for your time!

astro-stan:
# Details
Resolves #4158 

### New Features
* Adds 2 new LDAP fields: `goauthentik-io-ldap-active` and `goauthentik-io-ldap-superuser`.

### Breaking Changes
N/A

## Additional
The current fields (`goauthentik.io/ldap/active` and `goauthentik.io/ldap/superuser`) have special characters in them (namely `.` and `/`).  Those cannot be escaped because the fields are attribute descriptors rather than values. I am not sure if this is the actual standard or some LDAP clients are just limited. According to [this website](https://ff1959.wordpress.com/2012/03/04/characters-that-are-permitted-in-attribute-names-descriptors) however the only valid characters are alphanumeric strings and hyphens. 

#### My testing

Below are cases in which LDAP clients fail to query attribute descriptors:

1. `ldapsearch`:
    - Contain anything other than alphanumeric characters or hyphens (`[a-zA-Z\d-]`)
    - Start with a hyphen or digit (`^[\d-]`)
2.  `Jellyfin LDAP`:
    - Contain anything other than alphanumeric characters or hyphens (`[a-zA-Z\d-]`)
 
Escaping those with `\` or even their utf-8 encodings `\xXX` does not help. In the case of Jellyfin it even fails with:
```
Failure: User Filter: Escape sequence not allowed in attribute description
```

To fix this I have added 2 new fields (`goauthentik-io-ldap-active` and `goauthentik-io-ldap-superuser`) and marked the old ones as deprecated.

#### Question
There are also the `user.attributes`, which are passed directly as LDAP attributes. Some of them (for example `goauthentik.io/user/override-ips` and `goauthentik.io/user/service-account`) also contain invalid attribute descriptor characters. Should I pass each field through a sanitizing function before adding it as ldap descriptor?


sdimovv:
Not sure why the `ci-pylint` fails. On my machine it shows no errors. I also tried `make lint-fix` - no changes.

Also not sure why ldap test-e2e fails. I only added keys.  I think it is a false positive.

BeryJu:
If we're already changing the fields, they might as well have a shorter prefix, somewhat like the msDS fields in Active directory 

For the user attributes, i'm not sure tbh

sdimovv:
Thanks for taking a look at this. Just 2 questions:

- Are the old fields going to get deprecated first or straight removed? If the former, I should also remove the deprecation warning in the docs.
- Should we pass each of the `user.attributes` keys through a sanitizing function to replace/remove special characters before returning them in the ldap outpost?

BeryJu:
There's no strict deprecation policy currently, but I usually go by deprecating a feature in release x and then removing it in the next month release (if there's a big reason to remove it) or leaving it for a while

sdimovv:
Ok, do what you feel is best. Let me know if you need anything.

PR #4769, opened 2023-02-24
core: Add `resolve_dns` and `reverse_dns` functions to evaluator

### LINKED ISSUE #4615: DNS Resolution in Expressions
I would like to be able to resolve a hostname in expressions. In particular, my docker container resolves `host.docker.internal` to network address of the machine the container is running on. I would like to be able to check for requests from this hostname, but there is no way to detect this in expressions.

I would like a function to be exposed in expressions that resolves a hostname to an address or list of addresses. Such a function could be implemented in `authentik/lib/expression/evaluator.py` like this:
```py
    def expr_resolve_host(host: str) -> List[str]:
        """Resolve a hostname to a list of IP addresses."""
        
        return [
            sockaddr[0]
            for family, type, proto, canonname, sockaddr in 
            socket.getaddrinfo(host, None)
        ]
```

It would be possible for me to set up the docker network statically, and then hardcode the address, but this is fragile, and I would prefer for Docker to manage its network allocation.

**Additional context**
I currently have an expression that checks for requests for my local/trusted networks. This works well, but if I access Authentik directly from the docker host, the source address is the "host gateway" address of the docker container's network, which is not in my trusted networks. (The host address of the docker internal network.) I can imagine this functionality would also be useful for getting the addresses other containers the docker container is connected to.

At my scale and use case, it doesn't really matter, but it may be worth caching this information for larger installations. I'm not sure if `socket.getaddrinfo` does any caching of it's own, but if it decides to do real DNS resolution it could be quite slow. If an expression is getting hit tens or hundreds of times a minute this could be a serious issue. (This could alternatively be resolved with a local DNS caching server on the host.)

Any caching should be conservatively short, at most 60 seconds or so, since an change in address could also cause serious problems. A timeout may also be in order, since some use cases would prefer a fast empty response over an always correct answer. Ideally, these would be configurable with function parameters so users can determine their own caching and timeout needs.

astro-stan:
# Details
Resolves #4615 

### New Features
* Adds `resolve_dns` and `reverse_dns` functions to easy DNS and reverse DNS lookups.

### Breaking Changes
N/A

## Additional
N/A

### DIFF
--- a/authentik/lib/expression/evaluator.py
+++ b/authentik/lib/expression/evaluator.py
@@ -1,9 +1,10 @@
 """authentik expression policy evaluator"""
 import re
 import socket
+from functools import lru_cache
 from ipaddress import ip_address, ip_network
 from textwrap import indent
-from typing import Any, Iterable, List, Optional
+from typing import Any, Iterable, Optional
 
 from django.core.exceptions import FieldError
 from django_otp import devices_for_user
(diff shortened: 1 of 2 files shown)


sdimovv:
Not sure how good of an idea is to have cache based on just last calls. We should add some time cache as well. I will add it later today.

sdimovv:
I feel like 3 minutes is a more or less sensible TTL, considering you might want to query the IP of another container (which can restart and get another IP).

But change as appropriate if you disagree.

sdimovv:
Does this need anything else before getting merged?

BeryJu:
I wanted to add some tests, but they would all just end up mocking the actual socket call, so they wouldn't really be testing anything

PR #4800, opened 2023-02-27
website: add X-Frame-Options

BeryJu:


k-p-ani:
Hello @BeryJu,
    Is there away to change X-Frame-Option in Authentik, so that called from inside Iframe, browser will not block. Currently we have parent app, in that we embeded few child app using Iframe, and each of app connect to Authentik for single signon. But due to current limitation of (X-Frame-Option Deny), our apps not getting loading any we always get error message 
"To protect your security, authentik.xxxx.yyyyy.net will not allow Firefox to display the page if another site has embedded it"

Thanks in advance
Ani

PR #4822, opened 2023-03-01
stages/prompt: Add Radio Button Group, Dropdown and Text Area prompt fields

### LINKED ISSUE #2521: Additional prompt field types
Not really.

Some additional field types would make the user experience nicer and allow for new possibilities, in particular:

- List with options for checkboxes, dropdown, radio buttons, or a static list (`ul` or `ol`) that looks nice.
- Long text (textarea).

For the static list, I used a placeholder expression to generate a comma separated list.

**Additional context**
The list options are particularly useful when you really need to limit the user's choices. For example, I would like my users to be able to choose their login shell between `/bin/zsh` and `/bin/bash` but not any arbitrary path. For SSH keys, the single line text field creates some usability challenges.

### LINKED ISSUE #4352: Add radio prompt/checkbox groups
Add an ability to select only one out of multiple options. This will allow to create branches in flows. For example in a recovery flow a user can select in a prompt stage if they want to restore a forgotten password or a lost MFA method. Then later flow stages and logic can take into account the selection.

Currently not possible as checkbox controls are not able to be grouped together and no radio controls exists.

Add a new radio button prompt or add ability to group checkbox prompts together

There isn't one I think

**Additional context**
N/A

### LINKED ISSUE #4841: Add dynamic dropdowns to Promt Stage
I have written a custom Promt Stage which reminds me if i forgot to enable my time tracking on login. One feature missing is to select the Project i am working on.
It would be awesome if the promt stage would have a additional dropdown type which can be filled dynamicly (for example with the context dict).

astro-stan:
# Details
Resolves #4352
Resolves #2521
Resolves #4841 

### New Features
* Adds radio-button and dropdown and text-area prompts

### Breaking Changes
N/A

## Additional
N/A


sdimovv:
Hi @BeryJu,

I decided to implement this myself. Everything should be working as-is. There is one test that fails (`authentik/stages/prompt/tests.py::TestPromptStage::test_valid_challenge_request`) but I haven't touched anything related to that and it also seems to fail on `main`. So not sure what is up with it but it shouldn't be because of my changes.

By the way, I also have a patch for a multi-select dropdown field, but I haven't added it because I have two problems with the frontend implementation. Namely:
- Not sure how to inherit the `15em` height other selects throughout the UI seem to have
- Submitting the form seems to only get the last selected value of the multiselect dropdown instead of an array of values . Not sure how to make it get the whole array.

Would you be interested in taking a look at my multi-select dropdown implementation as well?

sevmonster:
Awesome work. This should also partially address the concerns in #2521.

authentik/stages/prompt/stage.py
```diff
@@ -54,7 +62,7 @@ def __init__(self, *args, **kwargs):
         stage: PromptStage = kwargs.pop("stage_instance", None)
         plan: FlowPlan = kwargs.pop("plan", None)
         request: HttpRequest = kwargs.pop("request", None)
-        user: User = kwargs.pop("user", None)
+        self.user: User = kwargs.pop("user", None)
```
sevmonster, line 65: What's this for?

sdimovv: Oops that is a leftover from an earlier implementation I was experimenting with. Fixed it. Thanks

sdimovv:
The last commit should fully resolve #2521.

sdimovv:
This should also resolves #4841

benedikt-bartscher:
Hey @sdimovv thanks for your work on this! Can you push your multi-select changes to a seperate branch?

sdimovv:
> Hey @sdimovv thanks for your work on this! Can you push your multi-select changes to a seperate branch?

Regarding the `multi-select`, I just realized that besides the two problems with the frontend I listed above there is also one limitation.

With the current implementation it is not possible to specify more than one selected option (the first choice) when rendering the form.  I am not sure how to fix this.

In any case, this PR started to get rather big.  I think it is best to first let @BeryJu  review what is already implemented. After that, depending on what he thinks I can push the `multi-select` patch here or in another PR for him to take a look at.

sdimovv:
Hey @BeryJu, can you please take a look at this?

authentik/stages/prompt/models.py
```diff
@@ -38,10 +39,17 @@ class FieldTypes(models.TextChoices):
...
     # Simple text field
     TEXT = "text", _("Text: Simple Text input")
+    # Long text field
+    TEXT_AREA = "text_area", _("Text area: Multiline Text Input.")
     # Simple text field
     TEXT_READ_ONLY = "text_read_only", _(
         "Text (read-only): Simple Text input, but cannot be edited."
     )
+    # Long text field
+    TEXT_AREA_READ_ONLY = "text_area_read_only", _(
```
BeryJu, line 49: Long term it might make sense to have read-only as a toggle in the model rather than as a field type, but that's not really related to this PR

authentik/stages/prompt/models.py
```diff
@@ -115,8 +136,46 @@ def serializer(self) -> Type[BaseSerializer]:
...
+            evaluator = PropertyMappingEvaluator(self, user, request, prompt_context=prompt_context)
+            try:
+                raw_choices = evaluator.evaluate(self.placeholder)
+            except Exception as exc:  # pylint:disable=broad-except
+                LOGGER.warning(
+                    "failed to evaluate prompt choices",
+                    exc=PropertyMappingExpressionException(str(exc)),
+                )
+
+        if isinstance(raw_choices, (list, tuple)):
```
BeryJu, line 160:
```suggestion
        if isinstance(raw_choices, (list, tuple, set)):
```

authentik/stages/prompt/models.py
```diff
@@ -115,8 +136,46 @@ def serializer(self) -> Type[BaseSerializer]:
...
+        return tuple(choices)
+
     def get_placeholder(self, prompt_context: dict, user: User, request: HttpRequest) -> str:
         """Get fully interpolated placeholder"""
+        if self.type in CHOICE_FIELDS:
+            # Make sure to return a valid choice as placeholder
+            choices = self.get_choices(prompt_context, user, request)
+            if choices:
+                return choices[0]
+            return ""
```
BeryJu, line 177:
```suggestion
            if not choices:
                return ""
            return choices[0]
```

imo reads a bit better like this, but fine with either one

web/src/flow/stages/prompt/PromptStage.ts
```diff
@@ -43,13 +43,31 @@ export class PromptStage extends BaseStage<PromptChallenge, PromptChallengeRespo
...
                     ?required=${prompt.required}
                     value="${placeholderAsValue ? prompt.placeholder : ""}">`;
+            case PromptTypeEnum.TextArea:
+                return `<textarea
+                    type="text"
+                    name="${prompt.fieldKey}"
+                    placeholder="${prompt.placeholder}"
+                    autocomplete="off"
+                    class="pf-c-form-control"
+                    style="min-height: 4em; max-height: 15em; resize: vertical;"
```
BeryJu, line 53: these styles should be set CSS instead of inline, also a very minor thing

web/src/flow/stages/prompt/PromptStage.ts
```diff
@@ -43,13 +43,31 @@ export class PromptStage extends BaseStage<PromptChallenge, PromptChallengeRespo
...
                     name="${prompt.fieldKey}"
                     class="pf-c-form-control"
                     readonly
                     value="${prompt.placeholder}">`;
+            case PromptTypeEnum.TextAreaReadOnly:
+                return `<textarea
+                    type="text"
+                    name="${prompt.fieldKey}"
+                    class="pf-c-form-control"
+                    style="min-height: 4em; max-height: 15em; resize: vertical;"
```
BeryJu, line 68: same as above

BeryJu (approved):
Couple of minor thoughts/suggestions but overall very good, thanks @sdimovv!

web/src/flow/stages/prompt/PromptStage.ts
sdimovv: I am not sure how to do this, frontend has never been my strong suit. Can you please show me?

web/src/flow/stages/prompt/PromptStage.ts
sdimovv: I am not sure how to do this, frontend has never been my strong suit. Can you please show me?

authentik/stages/prompt/models.py
sdimovv: Yup it seemed a bit repetitive to me, but I didn't want to introduce any unnecessary changes, especially to the UI.

web/src/flow/stages/prompt/PromptStage.ts
BeryJu: Yeah of course, see https://github.com/goauthentik/authentik/blob/63e3f6545bd0d3cf036e56d93d30add9f57ceaa0/web/src/flow/FormStatic.ts#L22

Add the import `import { css } from "lit";`

and then in the styles at the top use the
```
css``
```
template tag to add css for the `textarea {}` selector

sdimovv:
Should be okay now.

PR #4851, opened 2023-03-06
website/docs: Corrected typo and added Note about port number if using Istio/Kubern…

tanberry:
# Details
@BeryJu I was reading [this article](https://prevue.ch/news/2022-10-11-istio-authentik/) about a fellow setting up authentik, using Istio and Kubernetes. I wanted to somehow add a heads up about the port number, but I am not confident that I got it right. Is it only if there are custom decisions being made that the port number has to be for the cluster?

### DIFF
--- a/website/docs/providers/proxy/forward_auth.mdx
+++ b/website/docs/providers/proxy/forward_auth.mdx
@@ -107,7 +107,7 @@ import TraefikIngress from "./_traefik_ingress.md";
   </TabItem>
 </Tabs>
 
-## Enovy (Istio)
+## Envoy (Istio)
 
 :::info
 Requires authentik 2022.6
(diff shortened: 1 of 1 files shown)


website/docs/providers/proxy/forward_auth.mdx
```diff
@@ -117,6 +117,9 @@ Requires authentik 2022.6
 Support for this is still in preview, please report bugs on [GitHub](https://github.com/goauthentik/authentik/issues).
 :::
 
+:::info
+If you are using Istio and Kubernetes, use the port number that is exposed for your cluster.
+
```
BeryJu, line 122:
```suggestion
:::
```

The admonition in docusaurus require a starting and closing set of colons

PR #4900, opened 2023-03-10
website/docs: reformat headings in Forward auth

tanberry:
Reformat from ## to ### to show child sections under Domain level


BeryJu (approved):

tanberry:
After conversation about single app vs domain-level, I'm closing this PR and will create a new one that better reflects which is which.

PR #5000, opened 2023-03-18
website/integrations: add detail to MinIO docs

bjschafer:
# Details

## Changes

Add more information and reference links for the Scope Mapping, to clarify how it's used.

Also update to reflect that MinIO seems to require a Confidential client type now.

## Additional

The [upstream MinIO docs on OIDC](https://min.io/docs/minio/linux/reference/minio-mc-admin/mc-admin-config.html#openid-identity-management) indicate that the client secret (and thus confidential client type) are optional depending on provider. Experimentally with my single-node MinIO instance, the client secret was required and worked without further issue.

### DIFF
--- a/website/integrations/services/minio/index.md
+++ b/website/integrations/services/minio/index.md
@@ -19,6 +19,10 @@ The following placeholders will be used:
 -   `minio.company` is the FQDN of the MinIO install.
 -   `authentik.company` is the FQDN of the authentik install.
 
+### Mapping to MinIO policies
+
+The primary way to manage access in MinIO is via [policies](https://min.io/docs/minio/linux/administration/identity-access-management/policy-based-access-control.html#minio-policy). We need to configure authentik to return a list of which MinIO policies should be applied to a user.
+
 Under _Customization_ -> _Property Mappings_, create a _Scope Mapping_. Give it a name like "OIDC-Scope-minio". Set the scope name to `minio` and the expression to the following
 
 ```python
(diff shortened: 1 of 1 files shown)


website/integrations/services/minio/index.md
```diff
@@ -19,6 +19,10 @@ The following placeholders will be used:
 -   `minio.company` is the FQDN of the MinIO install.
 -   `authentik.company` is the FQDN of the authentik install.
 
+### Mapping to MinIO policies
+
+The primary way to manage access in MinIO is via [policies](https://min.io/docs/minio/linux/administration/identity-access-management/policy-based-access-control.html#minio-policy). We need to configure Authentik to return a list of which MinIO policies should be applied to a user.
```
BeryJu, line 24:
```suggestion
The primary way to manage access in MinIO is via [policies](https://min.io/docs/minio/linux/administration/identity-access-management/policy-based-access-control.html#minio-policy). We need to configure authentik to return a list of which MinIO policies should be applied to a user.
```

website/integrations/services/minio/index.md
```diff
@@ -27,9 +31,27 @@ return {
 }
 ```
 
+This mapping will result in the default MinIO `readwrite` policy being applied to all users. If you want to create a more granular mapping based on Authentik groups, use an expression like this
```
BeryJu, line 34:
```suggestion
This mapping will result in the default MinIO `readwrite` policy being applied to all users. If you want to create a more granular mapping based on authentik groups, use an expression like this
```

website/integrations/services/minio/index.md
```diff
@@ -27,9 +31,27 @@ return {
...
+      "policy": "consoleAdmin",
+}
+elif ak_is_group_member(request.user, name="Minio users"):
+  return {
+      "policy": ["readonly", "my-custom-policy"]
+}
+return None
+```
+
+Note that you can assign multiple policies to a user by returning a list, and returning `None` will map no policies to the user, resulting in no access to the MinIO instance. For more information on writing expressions, see [Expressions](/docs/property-mappings/expression) and [User](/docs/user-group/user#object-attributes) docs.
```
BeryJu, line 48: If possible use relative paths to the other docs

BeryJu (commented):
Some small changes, imo this block should be added to the doc, as they are optional and (again imo) I think they make more sense for readability at the end

bjschafer:
I'm not super familiar with Docusaurus, but I'm not sure why it doesn't like the relative links:

```
- On source page path = /integrations/services/minio/:
   -> linking to ../../../docs/property-mappings/expression.mdx (resolved as: /docs/property-mappings/expression.mdx)
   -> linking to ../../../docs/user-group/user.md#object-attributes (resolved as: /docs/user-group/user.md)
```

The relative links work fine in the GH web viewer, and they're resolving to the same absolute paths I had in my first pass at this. The [Docusaurus link docs](https://www.docusaurus.io/docs/next/markdown-features/links) also say that this relative syntax should work for file path links.

BeryJu:
Yeah docusaurus is a bit finicky when using relative paths with this setups, as /integrations and /docs are separate instances, so for relative paths the file extension has to be omitted (for some reason)

For the block I meant, I was talking about moving the `Mapping to MinIO policies` to the end, but it kinda needs to be up there as the mapping has to be created before the application to make it selectable in the provider, so nvm that

Issue #4380, opened 2023-01-07
Better management of MFA stages and flows


astro-stan:
**Is your feature request related to a problem? Please describe.**
This is probably too complicated for a single feature request, but  am not sure how to divide it into more manageable chunks as I need to explain what I am trying to achieve for this to make sense.

#### What am I trying to achieve 
I am trying to create an authentication and recovery flows that support MFA. I want them to look something like this:

1. Authentication flow:
Log in normally, then optionally present an MFA stage (or a way to setup MFA), depending on policies and/or MFA stage's not configured action.
2. Recovery flow
After email verification, allow for recovery of either a forgotten password or lost access to an MFA method (related to #4352). When recovering an MFA stage, the user will have to pass another MFA, most likely by providing a "recovery token" (static MFA token) in order to be allowed to delete or setup a new login MFA stage (depending on exact flow configuration).

#### Problems encountered during attempting to create this setup

1. MFA stages are independent and cannot be setup/chained together into a "primary"-"recovery" MFA relationship with each other.  This can be partially solved by creating an MFA setup flow that forces the user to create both MFA methods. However, this is only a partial solution (see other problems)

2. All users are free to create an unlimited number of every type of MFA from their profile pages. If there is an MFA stage with a configured setup flow, then that flow is available for use to the user without an ability to impose any restrictions. Technically, restrictions can be imposed by expression policies attached to the setup flows, but I don't think this is the best solution.

3. All users can also delete MFAs from their profile settings without any ability to impose restrictions. This can be especially damaging if a user deletes their recovery MFA method without deleting their primary one and subsequently loses access to the primary MFA (related to problem 1)

4. Regardless of the purpose of the MFA setup stage, all stages of the same type show up with the same name in the MFA enrollment dropdown menu in the user settings. For example the `static` stage shows up as `Enroll -> Static Authenticatior`. There is no way to provide a custom name such as "MFA Recovery Authenticator" for instance. 

5. Related to problem 1, when a MFA validator stage has its `non_configured_action` set to "configure", it does not invoke the configuration flow associated with the MFA setup stage, but rather just lets the user configure the MFA directly. This bypasses the configuration flow, making it only (easily) accessible from the user profile page.

**Describe the solution you'd like**
All problems resolved in such a way that:

1. An admin can chain MFA methods together in a primary-backup relationship and/or restrict access or otherwise enforce rules on when any MFA method can be created/deleted and what happens after that (does any other MFA need to be created/deleted as well?)

[...]

PR #5005, opened 2023-03-20
stages: Add ability to set user friendly names for MFA stages

astro-stan:
# Details
Resolves #

### New Features
* Add ability to set user friendly names for MFA stages

### Breaking Changes
N/A

## Additional
#4380


web/src/user/user-settings/mfa/MFADevicesPage.ts
```diff
@@ -18,17 +17,8 @@ import { until } from "lit/directives/until.js";
 import { AuthenticatorsApi, Device, UserSetting } from "@goauthentik/api";
 
 export function stageToAuthenticatorName(stage: UserSetting): string {
-    switch (stage.component) {
```
BeryJu, line 21: I don't think this should be removed, rather have the `stage.title` check at the top and if that's not set, then fallback to the current switch statement...allthough with the logic above there's no real way that stage.title could be empty I suppose, so maybe it should be set to required in the API schema

BeryJu (commented):
I agree with the goal of this, I wonder if it would make more sense to make this more specific to MFA devices (since that's the only stages a user gets listed), but this approach should give us more freedom in the future.

sdimovv:
> more specific to MFA devices (since that's the only stages a user gets listed)

I may not understand correctly, but isn't it currently this way? Only stages that have this are the TOTP, Duo, SMS, Static and Webautn. 

Password stage also uses the `ConfigurableStage`, which is why I had to create a separate abstract model for this.

web/src/user/user-settings/mfa/MFADevicesPage.ts
sdimovv: I am happy with either, but I think the setting it required in the schema makes more sense in order to avoid duplication of strings (both in frontend and backend) and also so that we can guarantee the same experience for API users.

sdimovv: Actually, the `title` attribute was already required in the schema (because the serializer has no default and is not readonly) but I added an explicit required kwarg anyway,

sdimovv:
CI main fails with:
```
ldap3.core.exceptions.LDAPSessionTerminatedByServerError: session terminated by server
```

I don't think that is related to the commits I made

PR #5010, opened 2023-03-20
website/docs: "productive" to "production" term change and version info

tanberry:
# Details
I changed the word "productive site" to "production site" and modified the section about using the .env file to override the version.

### DIFF
--- a/website/docs/installation/docker-compose.md
+++ b/website/docs/installation/docker-compose.md
@@ -71,11 +71,7 @@ docker-compose up -d
 
 The `docker-compose.yml` file statically references the latest version available at the time of downloading the compose file. Each time you upgrade to a newer version of authentik, you download a new `docker-compose.yml` file, which points to the latest available version. For more information, refer to the **Upgrading** section in the [Release Notes](../releases).
 
-```
-AUTHENTIK_IMAGE=ghcr.io/goauthentik/dev-server:${AUTHENTIK_TAG:-2023.3.1}
-```
-
-authentik is then reachable (by default) on port 9000 (HTTP) and port 9443 (HTTPS).
+By default, authentik is reachable (by default) on port 9000 (HTTP) and port 9443 (HTTPS).
 
 To start the initial setup, navigate to `https://<your server's IP or hostname>:9000/if/flow/initial-setup/`.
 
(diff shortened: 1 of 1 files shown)


website/docs/installation/docker-compose.md
```diff
@@ -69,13 +69,17 @@ docker-compose pull
 docker-compose up -d
 ```
 
-The `docker-compose.yml` file statically references the latest version available at the time of downloading the compose file, which can be overridden with the `AUTHENTIK_TAG` environment variable.
+The `docker-compose.yml` file statically references the latest version available at the time of downloading the compose file. We recommend that you edit the `.env` file and use the `AUTHENTIK_TAG` environment variable to override the default version defined in the compose file.
```
BeryJu, line 72: It's actually _not_ recommended to override the version, as the recommended upgrade process downloads the latest docker-compose file which has the correct version defined

tanberry: I think I need to better understand that exact syntax that we want people to put into the .env file, then. I suspect I am being silly, since the whole purpose of a variable is that it is, well, variable, so in the .env we wouldn't want them to specify an exact version...? Ping me when you get a moment and let's discuss, pls.

website/docs/installation/docker-compose.md
```diff
@@ -69,13 +69,17 @@ docker-compose pull
 docker-compose up -d
 ```
 
-The `docker-compose.yml` file statically references the latest version available at the time of downloading the compose file, which can be overridden with the `AUTHENTIK_TAG` environment variable.
+The `docker-compose.yml` file statically references the latest version available at the time of downloading the compose file. Each time you upgrade to a newer version of authentik, you download a new `docker-compose.yml` file, which points to the latest available version. For more information, refer to the **Upgrading** section in the [Release Notes](../releases).
+
+```
```
BeryJu, line 74: Probably should also remove this block

BeryJu: I meant to remove the code block, since it's kinda contradicting us saying that the compose file should be used for upgrades, and then we have a code block saying how to do the opposite of that

BeryJu (approved):

PR #5046, opened 2023-03-22
stages/prompt: Fix dropdown invalid choice

### LINKED ISSUE #5045: New dropdown prompts do not work with arbitrary values and cannot receive default value
**Describe the bug**
The new dropdown prompts cannot seem to handle strings as their value, and since their choices are set using the placeholder box, it's not straightforward to choose a default/placeholder value from the list.

**To Reproduce**
1. Set up some dropdown prompts with spaces in the values
2. Add them to a stage (I am using user settings) and try to save the values

**Expected behavior**
The values pass verification

**Screenshots**
image

 - authentik version: `gh-main`
 - Deployment: docker-compose

**Additional context**
While the current implementation using the placeholder field is good, it would be better to be able to provide a default/placeholder value (in my case default, for user-settings flows) in addition to the list, but this does not seem to be obviously possible when using dropdowns. I suppose you could reorder the values so the top choice is the preferred one, but that seems annoying to have to write into every dropdown placeholder expression. Since I can't get dropdowns to work I also do not know at first glance if the associated attribute value will change the selected default/placeholder implicitly, but I assume not since this does not happen with other fields.

astro-stan:
# Details
Resolves #5045 

### New Features
N/A

### Breaking Changes
N/A

## Additional
N/A

### DIFF
--- a/web/src/flow/stages/prompt/PromptStage.ts
+++ b/web/src/flow/stages/prompt/PromptStage.ts
@@ -149,7 +149,7 @@ export class PromptStage extends BaseStage<PromptChallenge, PromptChallengeRespo
                     ${prompt.choices
                         ?.map((choice) => {
                             return `<option
-                            value=${choice}
+                            value="${choice}"
                             ${prompt.placeholder === choice ? "selected" : ""}
                         >
                             ${choice}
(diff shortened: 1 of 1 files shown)


web/src/flow/stages/prompt/PromptStage.ts
```diff
@@ -149,7 +149,7 @@ export class PromptStage extends BaseStage<PromptChallenge, PromptChallengeRespo
                     ${prompt.choices
                         ?.map((choice) => {
                             return `<option
-                            value=${choice}
+                            value="${choice}"
                             ${prompt.placeholder === choice ? "selected" : ""}
```
BeryJu, line 153: I must've missed this in the previous PR

```suggestion
                            ?selected=${prompt.placeholder === choice}
```

BeryJu (approved):

PR #5095, opened 2023-03-27
stages/prompt: Add initial_data prompt field and ability to select a default choice for choice fields

astro-stan:
# Details
Resolves #5063 

### New Features
* Adds ability to specify an initial value for a prompt field (separate from a placeholder). Works on all flows not only on user-settings.
* Adds ability to select a default choice when creating a fixed choice field (`radio-button-group`, `dropdown`)

### Breaking Changes
* Strips away some of the responsibilities of the `placeholder` field. Namely - the responsibility of giving initial value to prompts in the user settings flow.

## Additional
The breaking change should not affect the default authentik config as that will be updated from the updated default blueprints automatically. However, any custom prompt fields or blueprints will need to be manually updated so that their initial value is specified in the "initial_value" field instead of in the "placeholder" field.


sdimovv:
@BeryJu, would it be possible to merge this before 2022.4 is out?

authentik/stages/prompt/tests.py
```diff
@@ -22,6 +22,7 @@
 )
 
 
+# pylint: disable=too-many-public-methods
```
BeryJu, line 25: should probably break this up into multiple files in the future, but this is fine for now

BeryJu (approved):
LGTM

looking back at this retroactively I think it would've made more sense to have an option to select some sort of "Prompt mapping" in prompts instead of specifying expressions directly, but I'm also not all too sure if that would actually be better.

Also one other thing that might be worth adding is a bit better migration, basically changing the fields in the same way as the blueprints were

sdimovv:
> For the migration I meant something like
Ah, so basically for every prompt in the DB that has "evaluate as expression" set (except fixed choice fields):
```
prompt.initial_value = prompt.placeholder
prompt.initial_value_expression = True
promp.placeholder = ""
prompt.placeholder_expression = False
```

Is that correct?

sdimovv:
I think that should do it?

sdimovv:
Hi @BeryJu ,

does this need anything else?

tanberry:
Hi @sdimovv quick question... you mention the "breaking change" of needing to manually update the any custom prompt fields or blueprints. Do you mention this in your docs in index.md? I didn't see it, but might have missed it. If it is not there, can you add a Note or some such to call it out? Thanks!

sdimovv:
> Hi @sdimovv quick question... you mention the "breaking change" of needing to manually update the any custom prompt fields or blueprints. Do you mention this in your docs in index.md? I didn't see it, but might have missed it. If it is not there, can you add a Note or some such to call it out? Thanks!

Hi @tanberry ,

Yes I don't think it is in the docs. However, since I wrote this, @BeryJu  requested to add a manual migration step. You can see this here:

https://github.com/goauthentik/authentik/blob/facdc71a6e108a694d9bfc4fc17c1d008bf54948/authentik/stages/prompt/migrations/0011_prompt_initial_value_prompt_initial_value_expression_and_more.py#L8-L22

This should automatically migrate all prompts (except the new choice prompts) that interpret the placeholder as expression to the new `initial_value` field. So the breaking change _should_ be handled automatically. I tested this to be the case on a test instance but I will be happy for you to try it out as well.

BeryJu (approved):

PR #5243, opened 2023-04-13
website/docs: flow context docs

BeryJu:
document flow context variables, when they are set and what they do

### DIFF
--- a/website/docs/flow/context/index.md
+++ b/website/docs/flow/context/index.md
@@ -15,7 +15,7 @@ Any data can be stored in the flow context, however there are some reserved keys
 
 ## Context dictionary and reserved keys
 
-This section describes the data objects (the context) that are used in authentik, and provides a list of keys, what they are used for and when they are set.
+This section describes the data (the context) that are used in authentik, and provides a list of keys, what they are used for and when they are set.
 
 :::warning
 Keys prefixed with `goauthentik.io` are used internally by authentik and are subject to change without notice, and should not be modified in policies in most cases.
--- a/website/docs/flow/context/index.md
+++ b/website/docs/flow/context/index.md
@@ -13,7 +13,7 @@ For example, in the Identification Stage (part of the default login flow), you c
 
 Any data can be stored in the flow context, however there are some reserved keys in the context dictionary that are used by authentik stages.
 
-## Context dictionary and reserved keywords
+## Context dictionary and reserved keys
 
 This section describes the data objects (the context) that are used in authentik, and provides a list of keys, what they are used for and when they are set.
 
@@ -21,7 +21,7 @@ This section describes the data objects (the context) that are used in authentik
 Keys prefixed with `goauthentik.io` are used internally by authentik and are subject to change without notice, and should not be modified in policies in most cases.
 :::
 
-### Common objects
+### Common keys
 
 #### `pending_user` ([User object](../../user-group/user.md))
 
@@ -63,7 +63,7 @@ When an unauthenticated user attempts to access a secured resource, they are red
 
 When a user authenticates/enrolls via an external source, this will be set to the source they are using.
 
-### Scenario-specific
+### Scenario-specific keys
 
 #### `is_sso` (boolean)
 
@@ -73,7 +73,7 @@ Set to `True` when the flow is executed from an "SSO" context. For example, this
 
 Set when a flow execution is continued from a token. This happens for example when an [Email stage](../stages/email/index.mdx) is used and the user clicks on the link within the email. The token object contains the key that was used to restore the flow execution.
 
-### Stage-specific
+### Stage-specific keys
 
 #### Consent stage
 
--- a/website/docs/flow/context/index.md
+++ b/website/docs/flow/context/index.md
@@ -2,9 +2,10 @@
 title: Flow Context
 ---
 
-Each flow execution has an independent *context*. This context holds all of the arbitrary data about that specific flow, data which can then be used and transformed by stages and policies.
+Each flow execution has an independent _context_. This context holds all of the arbitrary data about that specific flow, data which can then be used and transformed by stages and policies.
 
 ## Managing data in a flow context
+
 You create and manage the data for a context by configuring policies, stages, and bindings. As you plan your flow, and set up the required stages, etc. you are creating the context data for that flow.
 
 For example, in the Identification Stage (part of the default login flow), you can define whether users will be prompted to enter an email address, a username, or both. All such information about the flow's configuration makes up the context.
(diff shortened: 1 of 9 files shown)


website/docs/flow/context/index.md
```diff
@@ -0,0 +1,129 @@
...
+
+Stores the final redirect that the user's browser will be sent to after the flow is finished executing successfully. This is set when an un-authenticated user attempts to access a secured application, and when a user authenticates/enrolls with an external source.
+
+### `application` (Application object)
+
+When an unauthenticated user attempts to access a secured resource, they are redirected to an authentication flow. The application they attempted to access will be stored in this key.
+
+### `source` (Source object)
+
+When a user authenticates/enrolls via an external source, this will be set to the source they are using.
```
tanberry, line 41: Will there be a key attached to the source object? Like `source.SAML`

What do we mean by "source" here, and in general?

website/docs/flow/context/index.md
```diff
@@ -0,0 +1,129 @@
...
+
+## Scenario specific
```
tanberry, line 43:
```suggestion
## Scenario-specific objects
```

tanberry, line 43:
```suggestion
## Scenario-specific objects
```

website/docs/flow/context/index.md
```diff
@@ -0,0 +1,129 @@
...
+
+### `pending_user` (User object)
+
+`pending_user` is used by multiple stages. In the context of most flow executions, it holds the user that is executing the flow. This value is not set automatically, it is set via the [Identification stage](../stages/identification/).
+
+Stages that require a user, such as the [Password stage](../stages/password/), the [Authenticator validation stage](../stages/authenticator_validate/) and others will use this value if it is set, and fallback to the request's users when possible.
+
+### `prompt_data` (Dictionary)
+
+`prompt_data` is primarily used by the [Prompt stage](../stages/prompt/). The value of any field within a prompt stage is written to the `prompt_data` dictionary. For example, given a field with the _Field key_ `email` that was submitted with the value `foo@bar.baz` will result in the following context:
```
tanberry, line 19:
```suggestion
`prompt_data` is primarily used by the [Prompt stage](../stages/prompt/). The value of any field within a prompt stage is written to the `prompt_data` context (the dictionary of data used in that flow). For example, given a field with the _Field key_ `email` that was submitted with the value `foo@bar.baz` will result in the following context:
```

website/docs/flow/context/index.md
```diff
@@ -0,0 +1,129 @@
...
+
+An optional list of all permissions that will be given to the application by granting consent. Not supported with SAML. When used with an OAuth2 or Proxy provider, this will be set based on the configured scopes.
+
+### Autosubmit stage
+
+The autosubmit stage is an internal stage type that is not configurable via the API/Web interface. It is used in certain situations, where a POST request has to be sent from the browser, such as with SAML POST bindings. This works by using an HTML form that is submitted automatically.
+
+#### `title` (string)
+
+Optional title of the form shown to the user. Automatically set when this stage is used by the backend.
```
tanberry, line 67: Automatically set to what value? Does user ever see it?

website/docs/flow/context/index.md
```diff
@@ -0,0 +1,129 @@
...
+
+URL that the form will be submitted to.
+
+#### `attrs` (dict)
+
+Key-value pairs of the data that is included in the form and will be submitted to `url`.
+
+### User write stage
+
+#### `groups` (List of Group objects)
```
tanberry, line 79:
```suggestion
#### `groups` (list of Group objects)
```

website/docs/flow/context/index.md
```diff
@@ -0,0 +1,129 @@
+---
+title: Flow Context
+---
+
+Each flow execution has an independent context. This context can hold arbitrary data, which can be used and transformed by stages and policies.
+
+Any data can be stored in the flow context, however there are some reserved keys in the context dictionary that are used by authentik stages.
```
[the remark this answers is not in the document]
tanberry: Oh. I learned after writing this note that this page basically is the dictionary.

tanberry (commented):
Awesome new content, need to add a bit more conceptual info about what contexts are and how they work.

website/docs/flow/context/index.md
```diff
@@ -0,0 +1,129 @@
...
+
+## Common objects
+
```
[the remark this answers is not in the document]
BeryJu: Dictionary is also a data-type (https://www.programiz.com/python-programming/dictionary) 

Good point, especially the `prompt_data` dictionary has some specific keys that translate to user properties

website/docs/flow/context/index.md
```diff
@@ -0,0 +1,129 @@
...
+```json
+{
+    "prompt_data": {
+        "email": "foo@bar.baz"
+    }
+}
+```
+
+This data can be modified with policies. The data is also used by stages like [User write](../stages/user_write.md), which takes data in `prompt_data` and writes it to `pending_user`.
```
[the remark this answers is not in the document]
BeryJu: No, all of these can be modified with policies

website/docs/flow/context/index.md
```diff
@@ -0,0 +1,129 @@
...
+
+### `redirect` (string)
+
+Stores the final redirect that the user's browser will be sent to after the flow is finished executing successfully. This is set when an un-authenticated user attempts to access a secured application, and when a user authenticates/enrolls with an external source.
```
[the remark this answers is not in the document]
BeryJu: the final redirect URL that the user is redirected to after the flow is done

website/docs/flow/context/index.md
```diff
@@ -0,0 +1,129 @@
...
+
+### `application` (Application object)
+
+When an unauthenticated user attempts to access a secured resource, they are redirected to an authentication flow. The application they attempted to access will be stored in this key.
```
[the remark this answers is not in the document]
BeryJu: The key is always just `application` with the value being the application object

website/docs/flow/context/index.md
BeryJu: same as above

Sources are https://goauthentik.io/docs/terminology#source, i.e. federation sources, like log into authentik with github

website/docs/flow/context/index.md
BeryJu: Usually in a format of `Redirecting to $app...`. The user sees it very briefly as the redirect happens

This stage is a bit different as an admin can't configure it via the API/web interface, it's an "in-memory" stage that gets added by authentik itself when it's required

website/docs/flow/context/index.md
```diff
@@ -0,0 +1,172 @@
...
+
+Stores the final redirect URL that the user's browser will be sent to after the flow is finished executing successfully. This is set when an un-authenticated user attempts to access a secured application, and when a user authenticates/enrolls with an external source.
+
+#### Identifications tage
+
+##### `pending_user_identifier` (string)
+
+If _Show matched user_ is disabled, this key will hold the user identifier entered by the user in the identification stage.
+
+Stores the final redirect that the user's browser will be sent to after the flow is finished executing successfully. This is set when an un-authenticated user attempts to access a secured application, and when a user authenticates/enrolls with an external source.
```
tanberry, line 56:
```suggestion
Stores the final redirect URL that the user's browser will be sent to after the flow is finished executing successfully. This is set when an un-authenticated user attempts to access a secured application, and when a user authenticates/enrolls with an external source.
```

tanberry (approved):
LGTM!

PR #5296, opened 2023-04-18
web/admin: fix blueprint instance list without metadata or labels

cskwrd:
`Object.hasOwn()` throws an exception if the first argument is `null` or `undefined`. This change ensures that the first argument is never `null` or `undefined`.

# Details
This will let the blueprints list load when custom blueprints are added.

## Additional
The underlying issue of setting the metadata column in the blueprints table still needs to be resolved.

### DIFF
--- a/web/src/admin/blueprints/BlueprintListPage.ts
+++ b/web/src/admin/blueprints/BlueprintListPage.ts
@@ -126,7 +126,11 @@ export class BlueprintListPage extends TablePage<BlueprintInstance> {
     row(item: BlueprintInstance): TemplateResult[] {
         let description = undefined;
         const descKey = "blueprints.goauthentik.io/description";
-        if (Object.hasOwn(item.metadata?.labels ?? new Object(), descKey)) {
+        if (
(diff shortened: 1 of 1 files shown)


web/src/admin/blueprints/BlueprintListPage.ts
```diff
@@ -126,7 +126,7 @@ export class BlueprintListPage extends TablePage<BlueprintInstance> {
     row(item: BlueprintInstance): TemplateResult[] {
         let description = undefined;
         const descKey = "blueprints.goauthentik.io/description";
-        if (Object.hasOwn(item.metadata?.labels, descKey)) {
+        if (Object.hasOwn(item.metadata?.labels ?? new Object(), descKey)) {
```
BeryJu, line 129:
```suggestion
        if (item.metadata && item.metadata.labels && Object.hasOwn(item.metadata?.labels, descKey)) {
```

I think this reads slightly better

BeryJu (approved):

PR #5325, opened 2023-04-20
ci: add docker tag without timestamp

rissson:
* Add a docker tag to the main dev image with only the branch name and the sha

* Build the outposts dev images for arm64 as well.

### DIFF
--- a/.github/actions/comment-pr-instructions/action.yml
+++ b/.github/actions/comment-pr-instructions/action.yml
@@ -35,7 +35,7 @@ runs:
             ```shell
             AUTHENTIK_IMAGE=ghcr.io/goauthentik/dev-server
             AUTHENTIK_TAG=${{ inputs.tag }}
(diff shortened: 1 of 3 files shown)


.github/actions/comment-pr-instructions/action.yml
```diff
@@ -35,15 +35,15 @@ runs:
             ```shell
             AUTHENTIK_IMAGE=ghcr.io/goauthentik/dev-server
             AUTHENTIK_TAG=${{ inputs.tag }}
-            AUTHENTIK_OUTPOSTS__CONTAINER_IMAGE_BASE=ghcr.io/goauthentik/dev-%(type)s:gh-%(build_hash)s
```
BeryJu, line 38: This is on purpose as that placeholder is interpreted by authentik to use its matching build bash

BeryJu (approved):

PR #5339, opened 2023-04-21
ci: build on branch commit instead of merge commit

rissson:
An alternative would be to still build on a merge commit, but use the SHA from the original branch commit as the docker tag.

### DIFF
--- a/.github/actions/docker-push-variables/action.yml
+++ b/.github/actions/docker-push-variables/action.yml
@@ -51,12 +51,14 @@ runs:
         version_family = ".".join(version.split(".")[:-1])
         safe_branch_name = branch_name.replace("refs/heads/", "").replace("/", "-")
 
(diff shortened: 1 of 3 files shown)


.github/actions/docker-push-variables/action.yml
```diff
@@ -51,12 +51,14 @@ runs:
         version_family = ".".join(version.split(".")[:-1])
         safe_branch_name = branch_name.replace("refs/heads/", "").replace("/", "-")
 
+        sha = os.environ["GITHUB_SHA"] if not "${{ github.event.pull_request.head.sha }}" else "${{ github.event.pull_request.head.sha }}"
```
BeryJu, line 54: I suppose there isn't a way to do this purely with environment variables

rissson: unfortunately no :/

BeryJu (approved):

PR #5341, opened 2023-04-21
ci: docker build: enable GHA cache

rissson:
Enable build caching for Docker images. Might save a lot of time, at the expense of cache size, which is quite limited at the moment. A lot of the current cache seems to be `golangci-lint`, which we could probably disable as we're already skipping the pkg cache anyway.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -91,7 +91,7 @@ RUN apt-get update && \
     apt-get remove --purge -y build-essential pkg-config libxmlsec1-dev libpq-dev python3-dev && \
     apt-get autoremove --purge -y && \
     apt-get clean && \
-    rm -rf /tmp/* /var/lib/apt/lists/* /var/tmp/ && \
+    rm -rf /tmp/* /var/lib/apt/lists/* /var/tmp/ /root/.cache && \
     adduser --system --no-create-home --uid 1000 --group --home /authentik authentik && \
     mkdir -p /certs /media /blueprints && \
     mkdir -p /authentik/.ssh && \
--- a/Dockerfile
+++ b/Dockerfile
@@ -46,8 +46,8 @@ RUN --mount=type=bind,target=/go/src/goauthentik.io/go.mod,src=./go.mod \
 COPY ./cmd /go/src/goauthentik.io/cmd
 COPY ./authentik/lib /go/src/goauthentik.io/authentik/lib
 COPY ./web/static.go /go/src/goauthentik.io/web/static.go
-COPY ./web/robots.txt /go/src/goauthentik.io/web/robots.txt
-COPY ./web/security.txt /go/src/goauthentik.io/web/security.txt
+COPY --from=web-builder /work/web/robots.txt /go/src/goauthentik.io/web/robots.txt
+COPY --from=web-builder /work/web/security.txt /go/src/goauthentik.io/web/security.txt
 COPY ./internal /go/src/goauthentik.io/internal
 COPY ./go.mod /go/src/goauthentik.io/go.mod
 COPY ./go.sum /go/src/goauthentik.io/go.sum
--- a/Dockerfile
+++ b/Dockerfile
@@ -110,8 +110,6 @@ LABEL org.opencontainers.image.revision ${GIT_BUILD_HASH}
 
 WORKDIR /
 
-COPY --from=geoip /usr/share/GeoIP /geoip
-
 # We cannot cache this layer otherwise we'll end up with a bigger image
 RUN apt-get update && \
     # Required for runtime
@@ -139,6 +137,7 @@ COPY --from=python-deps /work/venv /venv
 COPY --from=web-builder /work/web/dist/ /web/dist/
 COPY --from=web-builder /work/web/authentik/ /web/authentik/
 COPY --from=website-builder /work/website/help/ /website/help/
+COPY --from=geoip /usr/share/GeoIP /geoip
 
 USER 1000
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -75,9 +75,10 @@ RUN --mount=type=secret,id=GEOIPUPDATE_ACCOUNT_ID \
 # Stage 5: Python dependencies
 FROM docker.io/python:3.11.5-bookworm AS python-deps
 
-WORKDIR /work
+WORKDIR /work/poetry
 
 ENV VENV_PATH="/work/venv" \
+    POETRY_VIRTUALENVS_CREATE=false \
     PATH="/work/venv/bin:$PATH"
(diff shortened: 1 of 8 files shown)


rissson:
With docker cache/without golangci cache: https://github.com/goauthentik/authentik/actions/runs/4765983066

rissson:
Docker build w/o buildkit cache: https://github.com/goauthentik/authentik/actions/runs/4783512742
Docker build w/ buildkit cache: https://github.com/goauthentik/authentik/actions/runs/4783620386

rissson:
TODO: add npm caching in the main image, add poetry caching in the main image (if possible)

BeryJu (commented):
TODO: some cleanup, cache pip packages too, then we can add it even if it doesn't give us a massive speed increase

Dockerfile
```diff
@@ -1,36 +1,41 @@
...
 FROM --platform=${BUILDPLATFORM} docker.io/node:20 as website-builder
 
+ENV NODE_ENV=production
+
+WORKDIR /work/website
+
+COPY ./website/package.json /work/website/package.json
+COPY ./website/package-lock.json /work/website/package-lock.json
+
+RUN --mount=type=cache,target=/root/.npm \
```
rissson, line 11: Should use a mount=bind

Dockerfile
```diff
@@ -1,36 +1,41 @@
...
 
-# Stage 3: Poetry to requirements.txt export
-FROM docker.io/python:3.11.4-slim-bookworm AS poetry-locker
+COPY ./web/package.json /work/web/package.json
+COPY ./web/package-lock.json /work/web/package-lock.json
 
-WORKDIR /work
-COPY ./pyproject.toml /work
-COPY ./poetry.lock /work
+RUN --mount=type=cache,target=/root/.npm \
```
rissson, line 30: Mount bind as well

ldap.Dockerfile
```diff
@@ -3,9 +3,17 @@ FROM docker.io/golang:1.21.0-bookworm AS builder
 
 WORKDIR /go/src/goauthentik.io
 
-COPY . .
+COPY go.mod .
+COPY go.sum .
+COPY gen-go-api .
+RUN --mount=type=cache,target=/go/pkg/mod \
```
rissson, line 9: Could use mount bind as well

proxy.Dockerfile
```diff
@@ -1,21 +1,34 @@
...
 
-COPY ./web /static/
-
 ENV NODE_ENV=production
 WORKDIR /static
-RUN npm ci --include=dev && npm run build-proxy
+
+COPY web/package.json .
+COPY web/package-lock.json .
+RUN --mount=type=cache,target=/static/.npm \
```
rissson, line 9: Mount bind, let's use /root/.npm as well as the rest

proxy.Dockerfile
```diff
@@ -1,21 +1,34 @@
...
 # Stage 2: Build
 FROM docker.io/golang:1.21.0-bookworm AS builder
 
 WORKDIR /go/src/goauthentik.io
 
-COPY . .
+COPY go.mod .
+COPY go.sum .
+COPY gen-go-api .
+RUN --mount=type=cache,target=/go/pkg/mod \
```
rissson, line 24: Mount bind

radius.Dockerfile
```diff
@@ -3,9 +3,17 @@ FROM docker.io/golang:1.21.0-bookworm AS builder
 
 WORKDIR /go/src/goauthentik.io
 
-COPY . .
+COPY go.mod .
+COPY go.sum .
+COPY gen-go-api .
+RUN --mount=type=cache,target=/go/pkg/mod \
```
rissson, line 9: Mount bind

Dockerfile
```diff
@@ -98,22 +118,25 @@ RUN apt-get update && \
...
 COPY --from=web-builder /work/web/authentik/ /web/authentik/
 COPY --from=website-builder /work/website/help/ /website/help/
 
 USER 1000
 
-ENV TMPDIR /dev/shm/
-ENV PYTHONUNBUFFERED 1
-ENV PATH "/usr/local/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/lifecycle"
+ENV TMPDIR=/dev/shm/ \
+    PYTHONDONTWRITEBYTECODE=1 \
```
rissson, line 137: I added this to be able to set the rootfs as readonly, any thoughts @BeryJu?

rissson:
New image size: 563MB, probably from getting rid of pip's cache

Dockerfile
```diff
@@ -99,22 +127,25 @@ RUN apt-get update && \
...
+COPY ./poetry.lock /
 COPY ./schemas /schemas
 COPY ./locale /locale
 COPY ./tests /tests
 COPY ./manage.py /
 COPY ./blueprints /blueprints
 COPY ./lifecycle/ /lifecycle
-COPY --from=go-builder /work/bin/authentik /bin/authentik
+COPY --from=go-builder /go/authentik /bin/authentik
+COPY --from=python-deps /work/venv /venv
```
BeryJu, line 138: I'm still not really a fan of copying an entire virtualenv into the container but if it makes the build faster and doesn't add (much) size I suppose I don't really have an argument against it

Dockerfile
BeryJu: Depends, certain paths still need to be writeable, like `~/.ssh` for the outpost docker integration, and maybe some other ones that I'm forgetting

Dockerfile
```diff
@@ -76,20 +110,14 @@ LABEL org.opencontainers.image.revision ${GIT_BUILD_HASH}
 
 WORKDIR /
 
-COPY --from=poetry-locker /work/requirements.txt /
-COPY --from=poetry-locker /work/requirements-dev.txt /
```
BeryJu, line 80: We still need an option to install the dev-dependencies in the final image as that's needed for `ak test-all` (https://github.com/goauthentik/authentik/blob/main/lifecycle/ak#L88) which is used here https://github.com/goauthentik/authentik/blob/main/.github/workflows/release-tag.yml#L24

BeryJu (changes requested):

Dockerfile
rissson: Fixed in d380b396405781f104830f8aa8ceecb6efdc794b

Dockerfile
rissson: Fair enough, let's say it's a start towards immutability. We can work on those other paths later on

BeryJu (approved):

PR #5355, opened 2023-04-23
web: Fix label not clickable for checkbox and choice field in prompts

macmoritz:
# Details
Labels next to checkboxes and choice field in prompts are not clickable.
Therefor we need to click on the checkbox or choicebox directly which is rather small.

### Fix
Wrap the checkbox/choicebox in the HTML label.

### New Features
Labels next to checkboxes and choice fields are now clickable

## Additional
Using this method has some advantages over the [HTML `for` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label?retiredLocale=de#attributes):

- No need to assign an unique id to every checkbox.
- No need to use the extra attribute in the label.
- Delivered page has same usage of storage (yes, `for` would be minor, but it is a fact)


macmoritz:
Would be great if we can fix this. 
When I have time I will take a look at it.

PR #5415, opened 2023-04-28
root: configuration reloading

rissson:
EDIT: Configuration watching is now implemented. This PR is best reviewed commit by commit, I'll update this description later.

# Details

This is the first step to handle configuration reloading. With those changes, it is already possible to do so, by sending a SIGUSR2 signal to the Go server process. The next step would be to watch for changes to configuration files and call the Restart function of the GoUnicorn instance.

SIGHUP is catched by the go server and forwarded as-is to gunicorn, which causes it to restart its workers. However, that does not trigger a reload of the Django settings, probably because they are already loaded in the master, before creating any of the worker instances.

SIGUSR2, however, can be used to spawn a new gunicorn master process, but handling it is a bit trickier. Please refer to Gunicorn's documentation[0] for details, especially the "Upgrading to a new binary on the fly" section.

As we are now effectively killing the gunicorn processed launched by the server, we need to handle some sort of check to make sure it is still running. That's done by using the already existing healthchecks, making them useful not only for the application start, but also for its lifetime. If a check is failed too many times in a given time period, the gunicorn processed is killed (if necessary) and then restarted.

[0] https://docs.gunicorn.org/en/20.1.0/signals.html

Other relevant links and documentation:

Python library handling the processing swaping upon a SIGUSR2: https://github.com/flupke/rainbow-saddle/

Golang cannot easily check if a process exists on Unix systems: https://github.com/golang/go/issues/34396

### New Features

- Support file configuration reloading by sending a SIGUSR2 signal to the `server` process.
- The golang server now automatically restarts (gracefully) the underlying gunicorn process if it fails healthchecks.

### Breaking Changes

Not that I'm aware of.

## Checklist

If applicable

### DIFF
--- a/internal/gounicorn/gounicorn.go
+++ b/internal/gounicorn/gounicorn.go
@@ -1,15 +1,21 @@
 package gounicorn
 
 import (
+	"fmt"
+	"io/ioutil"
 	"net/http"
 	"os"
 	"os/exec"
 	"runtime"
+	"strconv"
+	"strings"
 	"syscall"
 	"time"
 
 	log "github.com/sirupsen/logrus"
+
 	"goauthentik.io/internal/config"
+	"goauthentik.io/internal/utils"
 	"goauthentik.io/internal/utils/web"
 )
 
@@ -18,6 +24,7 @@ type GoUnicorn struct {
 
 	log     *log.Entry
 	p       *exec.Cmd
+	pidFile *string
 	started bool
 	killed  bool
 	alive   bool
@@ -27,6 +34,7 @@ func New() *GoUnicorn {
 	logger := log.WithField("logger", "authentik.router.unicorn")
 	g := &GoUnicorn{
 		log:             logger,
+		pidFile:         nil,
 		started:         false,
 		killed:          false,
 		alive:           false,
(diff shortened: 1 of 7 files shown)


rissson:
I've started working on automatic configuration reloading, but currently the configuration discovery doesn't match between the go and python code, so I'm updating that first.

internal/gounicorn/gounicorn.go
```diff
@@ -96,8 +106,77 @@ func (g *GoUnicorn) healthcheck() {
...
+			break
+		}
+		g.log.Debugf("waiting for new gunicorn pidfile to appear at %s", newPidFile)
+	}
+	if err != nil {
+		g.log.WithError(err).Warning("failed to find the new gunicorn process, aborting")
+		return
+	}
+
+	newPidB, err := ioutil.ReadFile(newPidFile)
```
rissson, line 145: If anyone that knows go, knows how to do this without writing 1k lines, I'd be happy to know it too!

internal/gounicorn/gounicorn.go
```diff
@@ -27,6 +34,7 @@ func New() *GoUnicorn {
 	logger := log.WithField("logger", "authentik.router.unicorn")
 	g := &GoUnicorn{
 		log:             logger,
+		pidFile:         nil,
 		started:         false,
 		killed:          false,
```
rissson, line 39: Would an enum be better here?

internal/gounicorn/gounicorn.go
rissson: `os.ReadFile`

BeryJu (commented):
- Configuration changes are good to keep all the defaults in one place
- Gunicorn health checking and reloading is great
- Configuration watching isn't really needed anymore since we can do that natively now
- In a future PR maybe generalise the process manager to be re-usable for Guacamole
  - We also don't really want to have our own "init-system"
  - We could consider using supervisord 
  - Also https://github.com/ochinchina/supervisord exists which we could maybe use as a library

PR #5449, opened 2023-05-02
core: applications backchannel provider

BeryJu:
Allow applications to have backchannel providers for SCIM and LDAP, providers that augment (currently in the sense of directory sync) normal SSO. SCIM previously didn't need an application and LDAP needed it's own application, so this makes it more logical

Access limitations on the application apply to backchannel providers in the same way, however backchannel providers may not use the application slug in URLs as an application may have multiple of the same backchannel providers


website/docs/core/applications.md
```diff
@@ -54,3 +54,13 @@ Requires authentik 2022.3
 :::
 
 To give users direct links to applications, you can now use an URL like `https://authentik.company/application/launch/<slug>/`. This will redirect the user directly if they're already logged in, and otherwise authenticate the user, and then forward them.
+
+### Backchannel providers
+
+:::info
+Requires authentik 2023.5
```
tanberry, line 61:
```suggestion
Requires authentik version 2023.5 or later.
```

website/docs/core/applications.md
```diff
@@ -54,3 +54,13 @@ Requires authentik 2022.3
...
+:::
+
+Backchannel providers can augment the functionality of applications with additional protocols. The main provider of an application provides the SSO protocol used for logging into the application. Backchannel providers can be used for protocols such as [SCIM](../providers/scim/index.md) and [LDAP](../providers/ldap/index.md) to provide directory syncing.
```
tanberry, line 64:
```suggestion
Backchannel providers can augment the functionality of applications by using additional protocols. The main provider of an application provides the SSO protocol that is used for logging into the application. Then, an additional backchannel provider can be used for protocols such as [SCIM](../providers/scim/index.md) and [LDAP](../providers/ldap/index.md) to provide directory syncing.
```

website/docs/core/applications.md
```diff
@@ -54,3 +54,13 @@ Requires authentik 2022.3
...
+
+Access restrictions configured on an application apply to all its backchannel providers.
```
tanberry, line 66:
```suggestion
Access restrictions that are configured on an application apply to all of its backchannel providers.
```

tanberry (approved):
Other than not being sure about using the term "backchannel" (and whether, if we do use it it should be two word), this looks good.

Do we need to say anything about how the second provider is invisible to end user?

website/docs/core/applications.md
```diff
@@ -3,7 +3,7 @@ title: Applications
 slug: /applications
 ---
 
-Applications in authentik are the other half of providers. They exist in a 1-to-1 relationship, each application needs a provider and every provider can be used with one application.
+Applications in authentik are the other half of providers. They exist in a 1-to-1 relationship, each application needs a provider and every provider can be used with one application. Starting with authentik 2023.5, applications can use multiple, to augment the functionality of the main provider, see [Backchannel providers](#backchannel-providers).
 
 Applications are used to configure and separate the authorization / access control and the appearance in the Library page.
```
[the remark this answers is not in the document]
BeryJu: Indeed, good point. I don't think it's called Library anywhere in the UI, we should probably change that on all docs

website/docs/core/applications.md
BeryJu: We should probably go through all version notes and bring them in the same format

website/docs/core/applications.md
BeryJu: very slight nit, an application can have multiple backend providers (even of the same kind) hence I used the plural

PR #5461, opened 2023-05-03
blueprints: fix error when trying to import blueprint that already ex…

BeryJu:
…ists


authentik/blueprints/v1/tasks.py
```diff
@@ -156,6 +156,8 @@ def check_blueprint_v1_file(blueprint: BlueprintFile):
         and blueprint.meta.labels.get(LABEL_AUTHENTIK_INSTANTIATE, "").lower() == "false"
     ):
         return
+    if BlueprintInstance.objects.filter(path=blueprint.path).exists():
+        return
     if not instance:
```
rissson, line 161: Isn't that already checked here?

PR #5480, opened 2023-05-04
ci: add workflow to redirect people to transifex when they only change translations

rissson:
Small message for contributors to redirect them to transifex when they update translations in this repository.

### DIFF
--- a/.github/workflows/translation-advice.yml
+++ b/.github/workflows/translation-advice.yml
@@ -1,0 +1,34 @@
+name: authentik-translation-advice
+
+on:
+  pull_request:
+    branches:
+      - main
+    paths:
+      - "!**"
+      - "locale/**"
+      - "web/src/locales/**"
+
+jobs:
+  post-comment:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Find Comment
+        uses: peter-evans/find-comment@v2
+        id: fc
+        with:
+          issue-number: ${{ github.event.pull_request.number }}
+          comment-author: "github-actions[bot]"
+          body-includes: authentik translations instructions
+      - name: Create or update comment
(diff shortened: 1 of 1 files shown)


.github/workflows/translation-advice.yml
```diff
@@ -0,0 +1,37 @@
...
+        with:
+          comment-id: ${{ steps.fc.outputs.comment-id }}
+          issue-number: ${{ github.event.pull_request.number }}
+          edit-mode: replace
+          body: |
+            authentik translations instructions
+
+            Thanks for your pull request!
+
+            Authentik translations are handled using
```
BeryJu, line 34: Very small nitpick; authentik should be stylised lowercase here

.github/workflows/translation-advice.yml
```diff
@@ -0,0 +1,37 @@
...
+    steps:
+      - name: Find Comment
+        uses: peter-evans/find-comment@v2
+        id: fc
+        with:
+          issue-number: ${{ github.event.pull_request.number }}
+          comment-author: "github-actions[bot]"
+          body-includes: authentik translations instructions
+      - name: Create or update comment
+        uses: peter-evans/create-or-update-comment@v2
```
BeryJu, line 24: We could use the bot accounts PAT to create the comment for consistency

.github/workflows/translation-advice.yml
rissson: Fixed

.github/workflows/translation-advice.yml
rissson: Those credentials won't be available on community PRs, will they?

BeryJu: oh true I keep forgetting that

BeryJu (approved):

PR #5498, opened 2023-05-05
web/flows: rework redirect logic

BeryJu:
always use redirect stage, remove special logic from flow executor

show better message when redirect target URL isn't http or https (show notice to close the page)

# Details

    Resolves #

### New Features

### Breaking Changes

## Checklist

If applicable

### DIFF
--- a/locale/en/LC_MESSAGES/django.po
+++ b/locale/en/LC_MESSAGES/django.po
@@ -8,7 +8,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2023-05-03 14:41+0000\n"
+"POT-Creation-Date: 2023-05-05 20:11+0000\n"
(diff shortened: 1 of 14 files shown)


BeryJu:
> ### authentik translations instructions
> Thanks for your pull request!
> 
> authentik translations are handled using [Transifex](https://explore.transifex.com/authentik/authentik/). Please edit translations over there and they'll be included automatically.

@rissson yeah I didn't think of this but the bot is gonna tell us to use transifex when we update the source of truth files, not sure if we can prevent this (maybe have a ci pipeline running that extracts i18n strings and PRs them itself?)

rissson:
Ugh, that's weird, it shouldn't have commented in this case.
https://github.com/goauthentik/authentik/blob/main/.github/workflows/translation-advice.yml#L8 this should make it skip that workflow if translation files changed _and_ source of truth files changed as well. We probably missed something during our test of this :/

rissson:
> maybe have a ci pipeline running that extracts i18n strings and PRs them itself

so run `make i18n-extract` in a CI?

PR #5556, opened 2023-05-09
providers/oauth2: add user UUID as subject option

### LINKED ISSUE #5541: Feature request: UUID or custom code for OAuth2 Subject mode
I was surprised by this, but Bitwarden self-hosted currently has a limitation for SSO externalId of 50 characters (see: https://github.com/bitwarden/server/issues/2891). The current user.uid or hash of user.uid options are both great, but cause Bitwarden errors. The other stock options (username, e-mail, and UPN) give me a bit of fear, mostly around changing username/emails.

Much like many other places in Authentik, it would be neat to be able to define the OAuth2 Subject mode with a little python snippet. However, for the purposes of getting a shorter-but-unique-id, another static option for user.uuid would, I think, work pretty great.

obrienmd:
# Details

    Resolves https://github.com/goauthentik/authentik/issues/5541

### New Features

-   Adds user.uuid as an OAuth2 subject option

### Breaking Changes

-   None

## Checklist

**Note:** Frontend compilation throw a warning on typescript types, but I believe that's because package.json is pulling @goauthentik/api with static definitions from npm, rather than the newly updated local schema... Or perhaps I'm missing a step :)

## Documentation

**Note:** I think not applicable - thought I remembered something about this in docs, but it was only the SCIM bit saying if you **do** change the subject mode, you'll need to change externalId there as well (which is fully configurable via python snippet).

### DIFF
--- a/authentik/providers/oauth2/id_token.py
+++ b/authentik/providers/oauth2/id_token.py
@@ -26,6 +26,7 @@ class SubModes(models.TextChoices):
 
     HASHED_USER_ID = "hashed_user_id", _("Based on the Hashed User ID")
     USER_ID = "user_id", _("Based on user ID")
+    USER_UUID = "user_uuid", _("Based on user UUID")
     USER_USERNAME = "user_username", _("Based on the username")
     USER_EMAIL = (
         "user_email",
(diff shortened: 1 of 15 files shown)


authentik/providers/oauth2/id_token.py
```diff
@@ -96,6 +97,8 @@ def new(
             id_token.sub = token.user.uid
         elif provider.sub_mode == SubModes.USER_ID:
             id_token.sub = str(token.user.pk)
+        elif provider.sub_mode == SubModes.USER_UUID:
+            id_token.sub = str(token.user.uuid)
```
BeryJu, line 101: Fyi, by default casting a uuid to a string includes the dashes compared to using .hex, which we might want to use as this would be used for cases where a shorter sub is required

obrienmd: Ah, makes sense - works well enough for my case either way, but .hex would probably have been better :)

PR #5583, opened 2023-05-11
stages/identification: auto-redirect to source when no user fields are selected

BeryJu:
## Details

When no user_fields are selected in an identification stage, and only one source is selected, automatically redirect to that source.

### New Features

### Breaking Changes

## Checklist

If applicable


elythh:
It is only working when the identification stage is the first in the flow right ? I can't find a way to reproduce it when it's not

PR #5602, opened 2023-05-14
sources/ldap: don't remove users from other sources

BeryJu:
## Details

Currently, the LDAP membership sets the user membership for groups based on which users should be in the group according to LDAP + which users are already in the group, filtered to only allow users that are not synced from LDAP at all.

However, if multiple LDAP sources exist, a group can have a user that is synced from another LDAP source. The current implementation removes that user.

This PR changes it to set the members of a group to the users specified by LDAP + any user currently in the group, then excluding any user that is in the group that is also not specified to be in the group by LDAP.

### New Features

### Breaking Changes

## Checklist

If applicable

### DIFF
--- a/authentik/sources/ldap/sync/membership.py
+++ b/authentik/sources/ldap/sync/membership.py
@@ -1,6 +1,7 @@
 """Sync LDAP Users and groups into authentik"""
 from typing import Any, Generator, Optional
 
+from django.db import transaction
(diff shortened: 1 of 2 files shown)


BeryJu:
Upon further testing, this actually isn't possible as we don't save which source a user was synced from, hence this will always remove users from groups that are synced from other ldap sources

PR #5634, opened 2023-05-16
lifecycle: Add depends_on for worker and server container

Bojan023:
## Details

It is possible that the server or worker container start prior to redis or postgresql, rendering their health status as unhealthy. It makes sense to define an explicit start-up sequence.

    No

## Changes

- Declaration of the stack start-up sequence:
  1.  postgresql and redis container start in parallel, after both containers run:
  2.  the server and worker start in parallel

## Checklist

If applicable


BeryJu:
I _think_ there was a reason why we don't do this, let me have a quick look in the compose file

PR #5638, opened 2023-05-16
web/flows: fix authenticator_validate device select not sent to backend

### LINKED ISSUE #5020: MFA SMS setup works but verification doesnt call SMS api
During the first setup SMS API is called with the custom payload, but after first setup next time it just says SMS is sent but does not call the API and is actually waiting for totp code, which works if totp is setup.

tograss:
## Details
Resolves #5020    
The old code version always returned without doing anything. I assume the intention was to ensure "value" is not undefined and is not false.

I tried the new version with sms and totp device. Having both active at the same time or only one. I did however not test a Yubikey /Webauthn. 

## Checklist

-   [ ] The translation files have been updated (`make i18n-extract`) <- This creates a lot of changes not caused by my fix so I did  not commit this .

### DIFF
--- a/web/src/flow/stages/authenticator_validate/AuthenticatorValidateStage.ts
+++ b/web/src/flow/stages/authenticator_validate/AuthenticatorValidateStage.ts
@@ -53,10 +53,10 @@ export class AuthenticatorValidateStage
     _selectedDeviceChallenge?: DeviceChallenge;
 
     set selectedDeviceChallenge(value: DeviceChallenge | undefined) {
-        //Do nothing if value is undefined or is casted to boolean false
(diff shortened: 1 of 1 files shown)


BeryJu:
@tograss I think the error came from the fact that `value` was assigned to `_selectedDeviceChallenge` and then we compare `value` to `_selectedDeviceChallenge` and return if they are the same, which they always will be. I fixed that comparison and with that the correct request is sent to the backend when selecting a challenge

BeryJu (approved):

tograss:
Ah now I see what is intended at this point.

PR #5677, opened 2023-05-18
root: add devcontainers

BeryJu:
## Details

Improvement developer experience by having the option to use devcontainers for development

Includes all the development dependencies, a postgres and redis database, and doesn't require installing any dependencies on the host

For the actual container there's actually some pretty big changes, as currently we run a lot of things in the root folder of the container, which makes mapping them to the host harder. This PR also moves the root of authentik to `/app-root` (maybe it should be `/app`?)

This might (will?) be a breaking change for all the external mounts, which we could at the same time also move to all be subfolders under `/data` to keep things easier

somewhat relies on https://github.com/goauthentik/authentik/pull/13172

### New Features

### Breaking Changes

## Checklist

If applicable

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,5 @@
+FROM ghcr.io/goauthentik/server:latest
+
+USER root
+
+RUN pip install --no-cache-dir -r /requirements-dev.txt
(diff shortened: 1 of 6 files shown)


tanberry:
@BeryJu for any breaking changes, what docs need to be updated? Install docs?

BeryJu:
> @BeryJu for any breaking changes, what docs need to be updated? Install docs?

Most likely yeah, this is still very early stage though (and not even fully working yet) so we can do that before we merge

BeryJu:
closed by #18813

PR #5701, opened 2023-05-20
website/integrations: Update discord integration with guild and role check

Aterfax:
## Details

Adds two sections to this document describing how the required expression policies needed to check users are a member of a certain guild or a member of a certain guild with a certain role.

I've tried to keep this closely aligned with the same syntax / style used for the equivalent Github OAuth social login page.

I've tested these and both appeared to work as intended. I'd recommend that someone else also give these a second check for functionality before merging.

Please squash merge.

I tried to name this PR according to the [styleguide](https://goauthentik.io/developer-docs/#pr-naming) but it was a bit unclear what the 'package' would be.

### New Features

-   N/A - documentation update.

### Breaking Changes

- N/A - documentation update.

If applicable

### DIFF
--- a/website/integrations/sources/discord/index.md
+++ b/website/integrations/sources/discord/index.md
@@ -56,19 +56,19 @@ For more details on how-to have the new source display on the Login Page see [he
 ### Checking for membership of a Discord Guild
 
 :::info
-Ensure that your Discord OAuth source in 'Federation & Social login' has the additional  `guilds` scope added under the 'Protocol settings'.
+Ensure that the Discord OAuth source in 'Federation & Social login' has the additional `guilds` scope added under the 'Protocol settings'.
 :::
 
 Create a new 'Expression Policy' with the content below, adjusting the variables where required:
 
 ```python
-# To get the guild ID number for the parameters, open Discord, go to Settings > Advanced and enable developer mode. 
-# Right-click on the server title and select "Copy ID" to get the guild ID.
+# To get the guild ID number for the parameters, open Discord, go to Settings > Advanced and enable developer mode.
+# Right-click on the server/guild title and select "Copy ID" to get the guild ID.
 
 ACCEPTED_GUILD_ID = "123456789123456789"
-GUILD_NAME_STRING = "Whatever you want to call your server in your error message."
+GUILD_NAME_STRING = "The desired server/guild name in the error message."
 
-# Ensure flow is only run during oauth logins via Discord
+# Ensure flow is only run during OAuth logins via Discord
 if context['source'].provider_type != "discord":
     return True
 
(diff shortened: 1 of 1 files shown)


website/integrations/sources/discord/index.md
```diff
@@ -52,3 +52,91 @@ Save, and you now have Discord as a source.
...
+GUILD_NAME_STRING = "The desired server/guild name in the error message."
+
+# Only change below here if you know what you are doing.
+
+# Ensure flow is only run during OAuth logins via Discord
+if context['source'].provider_type != "discord":
+    return True
+
+# Get the user-source connection object from the context, and get the access token
+connection = context['goauthentik.io/sources/connection']
```
BeryJu, line 78:
```suggestion
connection = context.get("goauthentik.io/sources/connection")
if not connection:
  return False
```

website/integrations/sources/discord/index.md
```diff
@@ -52,3 +52,91 @@ Save, and you now have Discord as a source.
...
+access_token = connection.access_token
+
+guilds = requests.get(
+    "https://discord.com/api/users/@me/guilds",
+    headers= {
+        "Authorization": "Bearer " + access_token,
```
BeryJu, line 84:
```suggestion
        "Authorization": f"Bearer {access_token}",
```

website/integrations/sources/discord/index.md
```diff
@@ -52,3 +52,91 @@ Save, and you now have Discord as a source.
...
+    }
+).json()
+
+user_matched = False
```
BeryJu, line 88:
```suggestion
```

website/integrations/sources/discord/index.md
```diff
@@ -52,3 +52,91 @@ Save, and you now have Discord as a source.
...
+# Right-click on the server/guild title and select server settings > roles, right click on the role and click
+# "Copy ID" to get the role ID.
+
+ACCEPTED_ROLE_ID = "123456789123456789"
+ACCEPTED_GUILD_ID = "123456789123456789"
+GUILD_NAME_STRING = "The desired server/guild name in the error message."
+ROLE_NAME_STRING = "The desired role name in the error message."
+
+# Only change below here if you know what you are doing.
+GUILD_API_URL = "https://discord.com/api/users/@me/guilds/" + ACCEPTED_GUILD_ID + "/member"
```
BeryJu, line 118:
```suggestion
GUILD_API_URL = f"https://discord.com/api/users/@me/guilds/{ACCEPTED_GUILD_ID}/member"
```

website/integrations/sources/discord/index.md
```diff
@@ -52,3 +52,91 @@ Save, and you now have Discord as a source.
...
+
+# Ensure flow is only run during OAuth logins via Discord
+if context['source'].provider_type != "discord":
+    return True
+
+# Get the user-source connection object from the context, and get the access token
+connection = context['goauthentik.io/sources/connection']
```
BeryJu, line 125:
```suggestion
connection = context.get("goauthentik.io/sources/connection")
if not connection:
  return False
```

website/integrations/sources/discord/index.md
```diff
@@ -52,3 +52,91 @@ Save, and you now have Discord as a source.
...
+access_token = connection.access_token
+
+guild_member_object = requests.get(
+    GUILD_API_URL,
+    headers= {
+        "Authorization": "Bearer " + access_token,
```
BeryJu, line 131:
```suggestion
        "Authorization": f"Bearer {access_token}",
```

website/integrations/sources/discord/index.md
```diff
@@ -52,3 +52,91 @@ Save, and you now have Discord as a source.
...
+    }
+).json()
+
+user_matched = False
```
BeryJu, line 135:
```suggestion
```

BeryJu (commented):
Couple of minor adjustments to the expression code, thanks for the PR!

BeryJu (approved):

Keyinator:
In the later code If the user is not in said guild you will get the error ``builtins.KeyError: 'roles'`` since ``guild_member_object`` does not have that key. ``guild_member_object`` will be ``mem-info:{'message': 'Unknown Guild', 'code': 10004}``

Aterfax:
> In the later code If the user is not in said guild you will get the error `builtins.KeyError: 'roles'` since `guild_member_object` does not have that key. `guild_member_object` will be `mem-info:{'message': 'Unknown Guild', 'code': 10004}`
> 
> I suggest implementing some error handling for that case.

My intended idea for the role checking expression policy would be a denied access for not having the role, so not being a member of the guild is certainly a different case of that too.

If no one objects I think there should be an earlier check that the user is a member of the guild and if not, the expression ends immediately with `return False`. That said I'd need to take a look at how the Discord API responds when forming the `guild_member_object` response when a user is not a member to figure out what logic to use or to make two requests with the earlier one checking the same logic as the other guild member checking expression.

Keyinator:
> ```python
> # If the user is not in the guild return false
> guild_matched = (guild_member_object["message"] == "Unknown Guild")
> if guild_matched:
>   return false
> ```
If I am correct this will fail if the user actually is in the guild since the dictionary will not have a key for ``message``.

Imo my code above would be a nice way of handling that specific error and from my knowledge no other exception comes to mind that could throw further exceptions so we should be in the clear.

If you really wanted to be sure you could check ``'roles' in guild_member_object`` to see if they key exists. Maybe:
```
if not 'roles' in guild_member_object
  return False
```
as a last resort.
(I'd still at least implement the above code-change since that returns a more user-friendly error)

Aterfax:
As per our chat and exploration in Discord it seems like the following might be the best solution?

```python
# The response for JSON errors is held within guild_member_object['code']
# See: https://discord.com/developers/docs/topics/opcodes-and-status-codes#json
# If the user isn't in the queried guild, it gives the somewhat misleading code = 10004.
if "code" in guild_member_object:
    if guild_member_object['code'] == 10004:
        ak_message(f"User is not a member of {GUILD_NAME_STRING}.")
    else:
        ak_message(f"Error, check https://discord.com/developers/docs/topics/opcodes-and-status-codes#json for error code {guild_member_object['code']}.")
    # Policy does not match if there is any error.
    return false
```

PR #5717, opened 2023-05-22
root: add install ID

BeryJu:
## Details

Instead of using the secret key in various places as a unique identifier for hashes, use a separate identifier. This allows us to rotate the secret key in the future (using https://docs.djangoproject.com/en/4.2/ref/settings/#std-setting-SECRET_KEY_FALLBACKS)

Also this will be used in the future for enterprise licenses

By default when upgrading an existing instance (i.e. there are entries in the django_migrations table) we use the current secret_key as install ID to keep backwards compatibility with user IDs etc, if there are no migrations then we generate a new install ID

### New Features

### Breaking Changes

## Checklist

If applicable

### DIFF
--- a/authentik/core/models.py
+++ b/authentik/core/models.py
@@ -5,7 +5,6 @@ from typing import Any, Optional
 from uuid import uuid4
 
 from deepmerge import always_merger
-from django.conf import settings
 from django.contrib.auth.hashers import check_password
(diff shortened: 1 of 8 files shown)


tanberry:
FYI for future readers: neither the end-user nor an Admin. needs to do anything differently. The system will now create a new separate install ID that will be used for license activation.

PR #5743, opened 2023-05-24
web/user: fix broken search on application library

### LINKED ISSUE #5171: UX: add an actionable button to "No Applications" page
For new installations that do not yet have any applications created, users will see a "No Applications Available" message on the Applications page of the user interface. But it is not immediately obvious how to fix the problem... can be frustrating.

Let's add a button there that takes them to the page in the UI to create an app, or to the docs on how to...

Screenshot with proposal is attached. Also, see how the Policy Binding page has a CTA button; let's do similar.

Screenshot 2023-03-09 at 8 29 59 AM

kensternberg-authentik:
This is *mortifying*.  I didn't test this well enough, and apparently broke it again once I'd tested it.  This patch restores the original behavior ("no match" means "just show everything"), and fixes a small bit of semantic lint -- the "search" feature should not be assigning meaning to what it finds; it's enough to pass back the prioritized list to whatever client wanted it, and let the client decide what to do with it.

## Details

Resolves #5171

## Changes

Fixes the search feature so that the data being passed back and forth correspond to the design.

### DIFF
--- a/web/src/user/LibraryPage/LibraryPageImpl.ts
+++ b/web/src/user/LibraryPage/LibraryPageImpl.ts
@@ -85,8 +85,13 @@ export class LibraryPage extends AKElement {
             throw new Error("ak-library-search-updated must send a custom event.");
         }
         event.stopPropagation();
-        this.selectedApp = event.detail.apps[0];
-        this.filteredApps = event.detail.apps;
+        const apps = event.detail.apps;
+        this.selectedApp = undefined;
+        this.filteredApps = this.apps.results;
+        if (apps.length > 0) {
+            this.selectedApp = apps[0];
+            this.filteredApps = event.detail.apps;
+        }
     }
 
     launchRequest(event: Event) {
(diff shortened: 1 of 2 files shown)


web/src/user/LibraryPage/LibraryPageImpl.ts
```diff
@@ -125,7 +130,7 @@ export class LibraryPage extends AKElement {
     }
 
     renderSearch() {
-        return html`<ak-library-list-search .apps="{this.apps.results}"></ak-library-list-search>`;
+        return html`<ak-library-list-search .apps=${this.apps.results}></ak-library-list-search>`;
```
kensternberg-authentik, line 133: This was the real issue; everything else is just dealing with errors to the console when the field is cleared (when there is no `app[0]`).  I got this to work, then moved it here and effed-up my argument call.  Massive apologies.

BeryJu (approved):

PR #5761, opened 2023-05-25
web: Replace lingui.js with lit-localize

kensternberg-authentik:
web: replace lingui with lit/localize

\#\# Changes

This rather massive shift replaces the lingui and `t()` syntax with lit-localize, XLIFF, and the `msg()` syntax used by lit-localize.  90% of this work was mechanized; simple perl scripts found and replaced all uses of `t()` with the appropriate corresponding syntax for `msg()` and `msg(str())`.

The XLIFF files were auto-generated from the PO files.  They have not been audited, and they should be checked over by professional translators.  The actual _strings_ have not been changed, but as this was a mechanized change there is always the possibility of mis-translation-- not by the translator, but by the script.


web/src/flow/stages/prompt/PromptStage.ts
```diff
@@ -193,28 +192,29 @@ ${prompt.initialValue}</textarea
...
-                        return true;
-                    }).map((locale) => {
-                        return html`<option
-                            value=${locale.code}
-                            ?selected=${prompt.initialValue === locale.code}
-                        >
-                            ${locale.code.toUpperCase()} - ${locale.label}
-                        </option>`;
-                    })}
+                    ${options}
```
kensternberg-authentik, line 215: This is a little more surgery than usual because it displays the locales to the customer, so I needed to replace a bit of the logic. I also separated out the business logic (I see I missed one) from the component itself, which is important to me as it leads to independent testing.

kensternberg-authentik (commented):
Added a few markers.

BeryJu (approved):

rissson:
I think [this doc page](https://goauthentik.io/developer-docs/translation) needs an update with this PR

PR #5828, opened 2023-06-01
outposts/ldap: fix race condition when refreshing the provider

rissson:
## Details

Fixes the race condition causing the crash found in #4138, which doesn't actually have anything to do with the issue itself

As far as I can work out, when the outpost refreshes its list of providers, it copies over its `boundUsers`, probably to avoid having to fetch them all again, and does so by making a shallow copy of that `map`, but not the mutex associated with it. It now has multiple references to the same map, each protected by a different mutex, which under certain conditions can cause a `concurrent map read and map write` error.

The fix replaces the duo map + mutex with a reference to a `sync.Map` which does basically the same thing, but internally. We now don't need to worry about properly copying the mutex and such, as it is handled by `sync.Map`.

### New Features

None

### Breaking Changes

None

## Checklist

-   [x] N/A The API schema has been updated (`make gen-build`)

-   [x] N/A The code has been formatted (`make web`)
-   [x] N/A The translation files have been updated (`make i18n-extract`)

If applicable

-   [x] N/A The documentation has been updated
-   [x] N/A The documentation has been formatted (`make website`)

### DIFF
--- a/.github/workflows/ci-outpost.yml
+++ b/.github/workflows/ci-outpost.yml
@@ -29,7 +29,8 @@ jobs:
       - name: golangci-lint
         uses: golangci/golangci-lint-action@v3
         with:
-          args: --timeout 5000s
+          version: v1.52.2
+          args: --timeout 5000s --verbose
           skip-pkg-cache: true
   test-unittest:
     runs-on: ubuntu-latest
--- a/internal/outpost/ldap/instance.go
+++ b/internal/outpost/ldap/instance.go
@@ -9,6 +9,7 @@ import (
 	"github.com/go-openapi/strfmt"
 	"github.com/nmcclain/ldap"
 	log "github.com/sirupsen/logrus"
+
 	"goauthentik.io/api/v3"
 	"goauthentik.io/internal/constants"
 	"goauthentik.io/internal/outpost/ldap/bind"
@@ -39,7 +40,7 @@ type ProviderInstance struct {
 	outpostName         string
 	outpostPk           int32
 	searchAllowedGroups []*strfmt.UUID
-	boundUsersMutex     sync.RWMutex
+	boundUsersMutex     *sync.RWMutex
 	boundUsers          map[string]*flags.UserFlags
 
 	uidStartNumber int32
--- a/internal/outpost/ldap/refresh.go
+++ b/internal/outpost/ldap/refresh.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/go-openapi/strfmt"
 	log "github.com/sirupsen/logrus"
+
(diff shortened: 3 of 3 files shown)


rissson:
After putting the outpost with this fix under heavy load (which triggered the issue in the first place), I am unable to reproduce it.

BeryJu (commented):
I'm not too big of a fan of sync.Map, mainly due to https://pkg.go.dev/sync#Map ```The Map type is specialized. Most code should use a plain Go map instead, with separate locking or coordination, for better type safety and to make it easier to maintain other invariants along with the map content.```

We could either copy the mutex from `existing` or do this

```diff
diff --git a/internal/outpost/ldap/refresh.go b/internal/outpost/ldap/refresh.go
index 40b757c92..ce052427c 100644
--- a/internal/outpost/ldap/refresh.go
+++ b/internal/outpost/ldap/refresh.go
@@ -59,7 +59,9 @@ func (ls *LDAPServer) Refresh() error {
                users := make(map[string]*flags.UserFlags)
                if existing != nil {
                        existing.boundUsersMutex.RLock()
-                       users = existing.boundUsers
+                       for key, value := range existing.boundUsers {
+                               users[key] = value
+                       }
                        existing.boundUsersMutex.RUnlock()
                }
```

rissson:
> We could either copy the mutex from `existing` or do this

I went with the second solution, copying the map contents, which then gives me `user flags not set after bind`, so I'll change it again for copying the mutex.

BeryJu:
Hmm, copying the mutex is also not ideal as there's probably a reason most linters warn against this..however I am curious why copying the contents of the map doesn't work correctly

rissson:
> Hmm, copying the mutex is also not ideal as there's probably a reason most linters warn against this..however I am curious why copying the contents of the map doesn't work correctly

I actually don't copy the mutex, but I made it a pointer. The idea is that throughout the life of a provider instance, even after a refresh, the `boundUsers` map stays the same, and as such the mutex should only be created once

rissson:
I'll wait for the container to build and give this a good load test again

rissson:
> I'll wait for the container to build and give this a good load test again

I cannot reproduce the issue with that fix

BeryJu (approved):

rissson:
@BeryJu should I leave the ` ci: outpost: use golangci-lint v1.52.2` in there and open a follow-up PR to revert it once the issue is resolved?

BeryJu:
> @BeryJu should I leave the ` ci: outpost: use golangci-lint v1.52.2` in there and open a follow-up PR to revert it once the issue is resolved?

I think it's fine to leave in this @rissson

PR #5850, opened 2023-06-04
sources/ldap: add support for cert based auth

### LINKED ISSUE #3399: Unable to authenticate users via Google Secure LDAP
**Describe the bug**

Once Google Secure LDAP is configured as the source and users and groups have been synced, they cannot login in. 

The problem seems to be caused by a search against Google Secure LDAP with the filter "(objectClass=*)", which is not allowed for the bound specific user who is trying to log in.

Google shows the following messages:

`Successfully bind LDAP with uid=plsan*****,ou=Users,dc=******,dc=com.`
`LDAP search with (objectClass=*) failed for the following reason: INSUFFICIENT_ACCESS_RIGHTS.`

The "INSUFFICIENT_ACCESS_RIGHTS" can be also be seen in server logs (below).

**To Reproduce**
1. Configure Google Secure LDAP as source.
2. Correctly sync users and groups
3. Attempt to login to authentik with a user from Google Secure LDAP.

**Expected behavior**

Correct login flow as the user is synced and the password is right.

**Logs**

```
{"backend": "authentik.sources.ldap.auth.LDAPBackend", "event": "Attempting authentication...", "host": "test-authentik.******.com", "level": "debug", "logger": "authentik.stages.password.stage", "pid": 20, "request_id": "4f17533c1a65480eb9b43d5e9b52c730", "timestamp": "2022-08-09T09:26:33.516377"}
{"event": "LDAP Auth attempt", "host": "test-authentik.******.com", "level": "debug", "logger": "authentik.sources.ldap.auth", "pid": 20, "request_id": "4f17533c1a65480eb9b43d5e9b52c730", "source": "<LDAPSource: Google Secure LDAP>", "timestamp": "2022-08-09T09:26:33.521620"}
{"event": "Attempting Binding as user", "host": "test-authentik.*******.com", "level": "debug", "logger": "authentik.sources.ldap.auth", "pid": 20, "request_id": "4f17533c1a65480eb9b43d5e9b52c730", "timestamp": "2022-08-09T09:26:33.525593", "user": "<User: plsan****>"}
{"event": "LDAPInsufficientAccessRightsResult - 50 - insufficientAccessRights - None - None - searchResDone - None", "host": "test-authentik.********.com", "level": "warning", "logger": "authentik.sources.ldap.auth", "pid": 20, "request_id": "4f17533c1a65480eb9b43d5e9b52c730", "timestamp": "2022-08-09T09:26:34.956817"}
{"event": "Failed to bind, password invalid", "host": "test-authentik.******.com", "level": "debug", "logger": "authentik.sources.ldap.auth", "pid": 20, "request_id": "4f17533c1a65480eb9b43d5e9b52c730", "timestamp": "2022-08-09T09:26:34.957475"}
{"backend": "authentik.sources.ldap.auth.LDAPBackend", "event": "Backend returned nothing, continuing", "host": "test-authentik.*******.com", "level": "debug", "logger": "authentik.stages.password.stage", "pid": 20, "request_id": "4f17533c1a65480eb9b43d5e9b52c730", "timestamp": "2022-08-09T09:26:34.957888"}
```

 - authentik version: 2022.7.3
 - Deployment: docker-compose

**Additional context**

Tell me if additional info, logs or whatever is needed.

ChandonPierre:
## Details

    Resolves https://github.com/goauthentik/authentik/issues/3399

### New Features

-   Adds support to authenticate against an LDAP source using certificate based authentication, like [Google Secure LDAP](https://support.google.com/a/answer/9048516?hl=en)

image

### Breaking Changes

-   No known breaking change

## Checklist

If applicable


authentik/sources/ldap/models.py
```diff
@@ -112,8 +123,20 @@ def server(self, **kwargs) -> Server:
         if self.peer_certificate:
             tls_kwargs["ca_certs_data"] = self.peer_certificate.certificate_data
             tls_kwargs["validate"] = CERT_REQUIRED
+        if self.client_certificate:
+            with NamedTemporaryFile(mode="w", delete=False) as temp_cert:
```
BeryJu, line 127: I really wish there was a better way to do this than temporary files, we do something similar here https://github.com/goauthentik/authentik/blob/main/authentik/outposts/docker_tls.py, maybe it would make more sense to save a file named after the keypair into the temp folder so that it can be re-used, but a very small nitpick

authentik/sources/ldap/models.py
```diff
@@ -112,8 +123,20 @@ def server(self, **kwargs) -> Server:
...
+            with NamedTemporaryFile(mode="w", delete=False) as temp_key:
+                temp_key.write(self.client_certificate.key_data)
+                private_key_file = temp_key.name
+            tls_kwargs["local_private_key_file"] = private_key_file
+            tls_kwargs["local_certificate_file"] = certificate_file
+            tls_kwargs["validate"] = CERT_REQUIRED
         if ciphers := CONFIG.y("ldap.tls.ciphers", None):
             tls_kwargs["ciphers"] = ciphers.strip()
+        if self.sni:
+            tls_kwargs["sni"] = self.server_uri.strip()
```
BeryJu, line 139: I honestly thought this was enabled by default, I don't think this will work in all cases as we allow multiple server URIs to be specified and separated by commas (See L146), so maybe this should instead be a text field?

BeryJu: I think that's probably fair, assuming that most LDAP setups with more than 1 server have all of the servers IPs under a single DNS entry, and that DNS entry matches the certificate's CN or SAN

authentik/sources/ldap/models.py
BeryJu: Yeah that probably also applies to the docker_tls wrapper, we could set better permission bits for the temporary file, but also from an attack vector, if someone has access to read those files they can do a lot of other things too

BeryJu (approved):

PR #5861, opened 2023-06-05
providers/ldap: add StartTLS support

BeryJu:
## Details

Finally adding StartTLS support aside from SSL support, this will allow proper certificate usage based on SNI

Needs updated docs, tests

### New Features

### Breaking Changes

## Checklist

If applicable

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,7 @@ module goauthentik.io
 go 1.20
 
 require (
+	beryju.io/ldap v0.1.0
 	github.com/Netflix/go-env v0.0.0-20210215222557-e437a7e7f9fb
 	github.com/coreos/go-oidc v2.2.1+incompatible
 	github.com/garyburd/redigo v1.6.4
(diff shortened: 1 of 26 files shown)


septatrix:
Based on the docs it is not clear to me: Can StartTLS support also be disabled such that only direct TLS is allowed?

BeryJu:
Currently there isn't a way to disable StartTLS, but usually starttls is preferable to direct SSL as direct SSL doesn't allow for selecting the certificate based on the initial request

septatrix:
Does LDAPS not support SNI? I seem to remember other implementations seem to support it. This is after all also how reverse proxies do it for HTTPS...

Also I am not sure on the way STARTTLS is implemented for LDAP though unless there are some significant differences from STARTTLS for SMTP then there exist [security disadvantages](https://serverfault.com/a/648282/543564) which even the RFC talks about. Even more so, [RFC 8314](https://datatracker.ietf.org/doc/html/rfc8314#section-3) officially obsoletes all cleartext/STARTTLS versions of mail protocols in favor of implicit TLS. In short, an attacker may disrupt the STARTTLS setup in which case many clients will fall back to unencrypted LDAP. And in general it is also not possible to force clients to use TLS whereas with LDAPS I can simply choose to not expose any unencrypted LDAP at all (which is also what we currently do).

PR #5865, opened 2023-06-05
web: Add storybook

kensternberg-authentik:
## Details

Adds Storybook's preview & testing capabilities to the application

## Changes

A bug in the way CSS is associated with `Base/AKElement` has been fixed. Storybook is not so forgiving of our typechecking shenanigans as Rollup.

### New Features

Storybook!  `npm run storybook` will start an instance of Storybook.  Right now only two dialogues have Stories; it is hoped that, going forward, bugfixes with visual consequences will be accompanied by a story to show the visual component working as expected.


web/src/elements/Base.ts
```diff
@@ -68,7 +68,8 @@ export class AKElement extends LitElement {
         if ("ShadyDOM" in window) {
             styleRoot = document;
         }
-        styleRoot.adoptedStyleSheets = [...styleRoot.adoptedStyleSheets, AKGlobal];
+        const globalStyleSheet = AKGlobal instanceof CSSResult ? AKGlobal.styleSheet : AKGlobal;
```
kensternberg-authentik, line 71: Depending on this page is loaded, this check is needed.  Storybook isn't so tolerant of poorly specified APIs as some other features might be.

web/tsconfig.json
```diff
@@ -42,6 +42,7 @@
         "strictNullChecks": true,
         "allowUnreachableCode": false,
         "allowUnusedLabels": false,
+        "useDefineForClassFields": false,
```
kensternberg-authentik, line 45: Storybook currently requires that TSC emit the "classic" class initialization syntax, rather than the more modern.  This isn't a problem at present, and it was first introduced with Typescript 3.7 and hasn't broken anything yet, so we're sticking with it.

In the event that this becomes a major problem, we should be prepared to freeze and remove Storybook, and we should _not_ "write to the test harness."  Right now, we are not; I want to ensure we don't get into the habit of doing so.

BeryJu (approved):

BeryJu:
I'll setup a separate netlify page for the storybook once this is merged so we get automated PR previews and all of that

PR #5959, opened 2023-06-14
root: partial Live-updating config

BeryJu:
## Details

An attempt to realise the same goal as #5415 but without having to restart anything (if possible)

### New Features

- Adds a wrapper database backend that uses our config loader
  (Based off of https://github.com/jciskey/django_postgres_vault but needs testing)
- Makes all email-related code get credentials directly form our config loader

### Breaking Changes

## Checklist

If applicable


BeryJu:
We've now tested these changes with https://github.com/goauthentik/customer-portal/pull/36 and https://github.com/goauthentik/customer-portal/pull/29, which both work as intended. Even though this is not the cleanest solution (requiring an explicit `reload()` call to reload the config), but still accomplishes the main goal (also cc @rissson)

BeryJu:
To make this updating work non-crudely, I had to change a lot of the ConfigLoader, namely storing an `Attr` instance instead of just the value, to keep track of the source of each configuration entry
With that we can reload certain config entries as we remember their source. This should hopefully also be useful in the future when we do file watching, to get rid of the explicit reloading (#5415)

PR #5964, opened 2023-06-14
web: Storybook css import fix

kensternberg-authentik:
This is an incredibly frustrating issue, because Storybook works in `dev` mode but not in `build` mode, and that's not at all what you'd expecte from a mature piece of software.  Lit uses the native CSS adoptedStylesheets field, which takes only a constructedStylesheet. Lit provides a way of generating those, but the imports from Patternfly (or any `.css` file) are text, and converting those to stylesheets required a bit of magic.

What this means going forward is that any Storied components will have to have their CSS wrapped in a way that ensures it is managed correctly by Lit (well, to be pedantic, by the
shadowDOM.adoptedStylesheets).  That wrapper is provided and the components that need it have been wrapped.

This problem deserves further investigation, but for the time being this actually does solve it with a minimum amount of surgical pain.

## Details

    Resolves storybook not resolving in `build` mode.

## Changes

### DIFF
--- a/web/src/common/utils.ts
+++ b/web/src/common/utils.ts
@@ -1,4 +1,5 @@
 import { SentryIgnoredError } from "@goauthentik/common/errors";
+import { CSSResult, css } from "lit";
 
 export function getCookie(name: string): string {
     let cookieValue = "";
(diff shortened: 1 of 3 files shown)


web/src/common/utils.ts
```diff
@@ -115,3 +117,28 @@ export function dateTimeLocal(date: Date): string {
...
+// using the function below.
+type AdaptableStylesheet = Readonly<string | CSSResult | CSSStyleSheet>;
+type AdaptedStylesheets = CSSStyleSheet | CSSStyleSheet[];
+
+const isCSSResult = (v: any): v is CSSResult =>
+    v instanceof CSSResult && v.styleSheet !== undefined;
+
+// prettier-ignore
+export const _adaptCSS = (sheet: AdaptableStylesheet): CSSStyleSheet =>
+    (typeof sheet === "string" ? css([sheet] as unknown as TemplateStringsArray, ...[]).styleSheet
```
BeryJu, line 133: Very small nitpick, I think this makes more sense as a full function instead of a tertiary statement

web/src/user/LibraryPage/ApplicationEmptyState.ts
```diff
@@ -19,7 +20,7 @@ import PFSpacing from "@patternfly/patternfly/utilities/Spacing/spacing.css";
  * administrator, provide a link to the "Create a new application" page.
  */
 
-const styles = [
+const styles = adaptCSS([
```
BeryJu, line 23: Is this the only place in all of the components where the wrapper needs to be added?

BeryJu (approved):

PR #6096, opened 2023-06-28
providers/ldap: fix inconsistent saving of user flags on failed cached binds

d-schiffner:
This has been a very nasty bug I was trying to fix for a long time now. By pure accident I stumbled across a simple `ldapsearch` that would corrupt the state of a user.

## Details
This happens if you use both cached bind and cached search.

In the current version, if you bind a user, the SetFlags will be set according to the BindDN. If the user later logs in again but uses different credentials, i.e. a wrong password, the SetFlags will be reset (in the direct binder), login however will fail (as expected). If the user then correctly signs in again, the session will not be refreshed (as it is still valid), and `SetFlags` will not be updated, resulting in "User with PK not in local cache" messages with Pk being `0`, as the binder did not receive a Pk from the server.

This also fixes a bug that only provides data for a "newly" logged-in user, if you ask twice.

    Yes, but I did not file one, sorry.

## Changes

I changed the logic to only call `SetFlags`, if none are available, to avoid overwriting the existing information of a user that logs in. I added the Pk of -1, because I was not sure, if 0 is a valid Pk or not.

In the search/memory, I updated the logic as well:
-  clients would report as being there, because GetFlags would return non-nil, but they haven't authenticated -> this should bail   in my opinion (line 76)
-  The if-else logic is broken. In the `true` case, the `u[0]` is never set, resulting in the second bug, as the `users` object is empty and does not contain the `userinfo` that has just been identified by the logic

### New Features

none

### Breaking Changes

none

## Checklist

### DIFF
--- a/internal/outpost/ldap/search/memory/memory.go
+++ b/internal/outpost/ldap/search/memory/memory.go
@@ -112,12 +112,13 @@ func (ms *MemorySearcher) Search(req *search.Request) (ldap.ServerSearchResult, 
 						flags.UserInfo = &ms.users[i]
 					}
 				}
-
 				if flags.UserInfo == nil {
 					req.Log().WithField("pk", flags.UserPk).Warning("User with pk is not in local cache")
 					err = fmt.Errorf("failed to get userinfo")
 				}
-			} else {
+			}
+			if flags.UserInfo != nil {
+				req.Log().WithField("baseDN", req.BaseDN).WithField("userinfo", flags.UserInfo).Debug("Assign userinfo")
 				u[0] = *flags.UserInfo
 			}
 			users = &u
--- a/internal/outpost/ldap/search/memory/memory.go
+++ b/internal/outpost/ldap/search/memory/memory.go
@@ -12,6 +12,7 @@ import (
 	log "github.com/sirupsen/logrus"
 	"goauthentik.io/api/v3"
 	"goauthentik.io/internal/outpost/ldap/constants"
+	"goauthentik.io/internal/outpost/ldap/flags"
 	"goauthentik.io/internal/outpost/ldap/group"
 	"goauthentik.io/internal/outpost/ldap/metrics"
 	"goauthentik.io/internal/outpost/ldap/search"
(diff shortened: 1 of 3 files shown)


internal/outpost/ldap/search/memory/memory.go
```diff
@@ -102,23 +103,23 @@ func (ms *MemorySearcher) Search(req *search.Request) (ldap.ServerSearchResult,
...
 					}
 				}
-
-				if flags.UserInfo == nil {
-					req.Log().WithField("pk", flags.UserPk).Warning("User with pk is not in local cache")
+				if flag.UserInfo == nil {
+					req.Log().WithField("pk", flag.UserPk).Warning("User with pk is not in local cache")
 					err = fmt.Errorf("failed to get userinfo")
 				}
-			} else {
```
BeryJu, line 120: I think this else could probably stay as we check for userinfo being nil, so if that's not the case it has to be non-nil

BeryJu (approved):

internal/outpost/ldap/search/memory/memory.go
d-schiffner: No, because if we set the userinfo in the first branch, the else-branch is not executed, leading to the described bug.

Alternative would be to set in both the `u[0]` variable

BeryJu: Good point, the more I read through this code the more I'm realising it needs a refactor

d-schiffner: I needed to read it also a few times. Maybe extract a function from it, which returns either (userinfo, error)?

BeryJu: I was thinking this, but yeah there's room for more optimisation

```go
	if needUsers {
		if flag.CanSearch {
			users = &ms.users
		} else if flag.UserInfo == nil {
			for i, u := range ms.users {
				if u.Pk == flag.UserPk {
					flag.UserInfo = &ms.users[i]
					users = &[]api.User{
						ms.users[i],
					}
				}
			}
			if flag.UserInfo == nil {
				req.Log().WithField("pk", flag.UserPk).Warning("User with pk is not in local cache")
				err = fmt.Errorf("failed to get userinfo")
			}
		} else {
			users = &[]api.User{
				*flag.UserInfo,
			}
		}
	}

```

BeryJu: Good point, allthough actually thinking about it some more, I'll revert most of the refactoring, I'm a lot more comfortable with changing that much when we have some actual tests

PR #6108, opened 2023-06-30
website/integrations: Update ArgoCD documentation

acelinkio:
## Details

Updates documentation for integrating Authentik and ArgoCD.  These are minor settings updates to make following the guide more straightforward.  

Groups was dropped from scope.  Profile contains multiple attributes including group.

### DIFF
--- a/website/integrations/services/argocd/index.md
+++ b/website/integrations/services/argocd/index.md
@@ -30,6 +30,8 @@ Only settings that have been modified from default have been listed.
 In authentik, create an _OAuth2/OpenID Provider_ (under _Applications/Providers_) with these settings:
 
 -   Name: ArgoCD
+-   Authorization flow: `default-provider-authorization-explicit-consent`
+-   Client Type: `Confidential`
 -   Signing Key: Select any available key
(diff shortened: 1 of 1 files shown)


acelinkio:
The `http://argocd.company/auth/login` path was found from ArgoCD docs.  

https://argo-cd.readthedocs.io/en/stable/operator-manual/user-management/microsoft/#configure-a-new-azure-ad-enterprise-app

website/integrations/services/argocd/index.md
```diff
@@ -30,6 +30,8 @@ Only settings that have been modified from default have been listed.
 In authentik, create an _OAuth2/OpenID Provider_ (under _Applications/Providers_) with these settings:
 
 -   Name: ArgoCD
+-   Authorization flow: `default-provider-authorization-explicit-consent`
```
BeryJu, line 33: we don't usually include the authorization flow in these docs as they aren't required to be set to a certain value for this to work

```suggestion
```

acelinkio: Thanks.  Made the change and should be good now.

PR #6121, opened 2023-07-03
web, website: compress images

rissson:
## Details

Compress images. This should be lossless. Size gain: 88M -> 16M.

Somewhat part of https://github.com/goauthentik/meta/issues/20, as I'll also be excluding the `.git` folder from the docker build. 

### New Features

### Breaking Changes

## Checklist

If applicable


rissson:
Do we want to automatically push those changes on the PR branch?

.github/workflows/compress-images.yml
```diff
@@ -0,0 +1,18 @@
+name: compress-images
+
+on:
+  pull_request:
+    branches:
+      - main
```
rissson, line 6:
```suggestion
      - main
    paths:
      - "**/*.jpg"
      - "**/*.png"
```

.github/workflows/compress-images.yml
```diff
@@ -0,0 +1,18 @@
...
+
+jobs:
+  compress:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - run: |
+          sudo apt install -y imagemagick optipng 
+          find -name '*.jpg' -exec mogrify -strip -compress JPEG {} +
```
BeryJu, line 15: we already use https://sharp.pixelplumbing.com/ for the flow background (see https://github.com/goauthentik/authentik/blob/main/web/package.json#L25) so maybe we should use that here too?

Also I dont think these compression commands are idempotent so this CI pipeline might always fail?

BeryJu (commented):
I don't think we'll get much of a size improvement from the image compression (at least from the potential pain of making a CI pipeline for that), I'm curious of how much of a difference ignoring .git makes with docker

rissson:
> I don't think we'll get much of a size improvement from the image compression

The size reduction by compressing images is 72M, which makes our docker image go from 832M to 760M, i.e. a 9% improvement.

> at least from the potential pain of making a CI pipeline for that

We can also not make a CI pipeline and just have this as a one-off.

> I'm curious of how much of a difference ignoring .git makes with docker

Nothing in terms of image size now that I looked at it more closely, as we don't do `COPY ./ ./`, it would just be an improvement of not copying the `.git` directory over to the docker daemon when building.

.github/workflows/compress-images.yml
rissson: optipng is idempotent, imagemagick (either `convert` or `mogrify`) isn't

rissson:
Implement https://github.com/marketplace/actions/image-actions

rissson:
Unfortunately, I cannot customize the commit name :/

BeryJu (approved):
LGTM, commit message doesn't really matter since we squash merge anyways

I copied all the current state icons/etc to https://github.com/goauthentik/brand-assets to keep the original copies (not that it matters too much since we have svgs but anyways)

.github/workflows/image-compress.yml
```diff
@@ -0,0 +1,60 @@
...
+        id: cpr
+        with:
+          token: ${{ steps.generate_token.outputs.token }}
+          title: "*: Auto compress images"
+          branch-suffix: timestamp
+          commit-messsage: "*: compress images"
+          body: ${{ steps.compress.outputs.markdown }}
+          delete-branch: true
+          signoff: true
+      - uses: peter-evans/enable-pull-request-automerge@v3
```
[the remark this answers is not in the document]
BeryJu: 👍 I was just about to push that exact same change too lol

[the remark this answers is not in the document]
rissson: Yes, for once I was faster :tada:

PR #6257, opened 2023-07-14
website/blog: fixed bad link

tanberry:
Fixed bad markdown that broke link at end of blog.


kensternberg-authentik (approved):
Straightforward!

PR #6303, opened 2023-07-19
website/docs: Update syntax in traefik standalone example

thmo:
## Changes

This makes syntax fixes in the traefik standalone example in the "forward authentication" section of the documentation.


BeryJu:
thanks for the PR @thmo! Could you include which command/setup you used to validate the new configuration?

PR #6319, opened 2023-07-20
outposts: support json patch for Kubernetes

### LINKED ISSUE #5908: Support customizing kubernetes objects created by managed outposts
On my cluster, I have both local, on premises nodes as well as nodes running in cloud instances. The outpost deployments generated by the kubernetes integration does not support specifying a nodeSelector or affinity options.

Node placement is more of a concern when dealing with outposts that may need to be exposed outside the cluster, such as RADIUS or LDAP outposts, and where you ideally want their externalTrafficPolicy to be Local to preserve the requesting source IPs.

It might also be a concern if you're literally using the proxy outpost to... proxy... instead of using an external reverse proxy to forward-auth.

Either allowing a nodeSelector or affinity to be specified as part of the outpost configuration.

Supporting a more generic jsonpatch/strategicMergePath approach to patching it.

**Additional context**

ChandonPierre:
## Details

    Resolves #5908 

### New Features

-   Adds support to apply an [RFC 6902](https://datatracker.ietf.org/doc/html/rfc6902) compliant JSON patch to Kubernetes outpost deployments via `kubernetes_json_patch` outpost config option. 
- Provides the ability to override the refrence deployment spec with `nodeAffinity`, `tolerations`, `resources`, etc
- Changes to the JSON patch are reconciled, since a provided patch is compared against the current V1Deployment spec

### Breaking Changes

## Checklist

If applicable

### DIFF
--- a/authentik/outposts/controllers/k8s/base.py
+++ b/authentik/outposts/controllers/k8s/base.py
@@ -1,16 +1,20 @@
 """Base Kubernetes Reconciler"""
+from json import dumps
 from typing import TYPE_CHECKING, Generic, Optional, TypeVar
 
 from django.utils.text import slugify
-from kubernetes.client import V1ObjectMeta
+from jsonpatch import JsonPatchConflict, JsonPatchException, JsonPatchTestFailed, apply_patch
+from kubernetes.client import ApiClient, V1ObjectMeta
 from kubernetes.client.exceptions import ApiException, OpenApiException
 from kubernetes.client.models.v1_deployment import V1Deployment
 from kubernetes.client.models.v1_pod import V1Pod
+from requests import Response
 from structlog.stdlib import get_logger
 from urllib3.exceptions import HTTPError
 
 from authentik import __version__
 from authentik.outposts.apps import MANAGED_OUTPOST
+from authentik.outposts.controllers.base import ControllerException
 from authentik.outposts.controllers.k8s.triggers import NeedsRecreate, NeedsUpdate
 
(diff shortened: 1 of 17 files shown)


BeryJu:
@ChandonPierre thanks for the PR! I made quite a few changes with https://github.com/goauthentik/authentik/pull/6319/commits/b0617d026c48d96ae16009158b223874632bb95f (what was supposed to be a small commit), lmk if you'd change anything about that commit

BeryJu (approved):

los93sol:
Is there an example of using this in the web ui anywhere?  I've been trying to get it to work but cannot seem to get any patches to apply

los93sol:
Thanks, that appears to work, is it possible to patch the service though?  It seems to just not create the service when I do anything to it in the patches

los93sol:
Wierd, I'm trying to use this...

kubernetes_json_patches:
  service:
    - op: add
      path: /spec
      value:
        loadBalancerIP: 192.168.80.101

and it isn't showing anything in the logs on either the worker or server, but the service never gets created when I reference it that way

los93sol:
Yes, I changed it to LoadBalancer and am attempting to configure the metallb annotation to allow sharing of the ip, but so far no luck getting the patch to apply

los93sol:
kubernetes_json_patches:  
  service:  
    - op: add  
      path: /metadata/annotations/loadBalancerIPs  
      value: 192.168.80.101

Here's another attempt that seems to fail to create it, but I still am not seeing anything in the logs

PR #6397, opened 2023-07-27
api: log errors if app URLs import fail

rissson:
## Details

If an app has an error like a typo in an import, it's catched by `except ImportError` and thus relatively impossible to debug when developing.

The way I give the exception to the logger probably isn't the way to do this though.

### New Features

### Breaking Changes

## Checklist

If applicable


BeryJu (approved):

rissson:
@BeryJu I set the level to debug in order not to flood logs for apps that don't have urls

PR #6398, opened 2023-07-27
outposts: make metrics compliant with Prometheus best-practices

xunleii:
## Details

This PR don't fix a specific issue but follows this discussion: https://github.com/goauthentik/authentik/discussions/6315.

All histograms created with `promauto` are populated with nanosecond durations, whereas the default buckets are designed for [second durations](https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#pkg-variables), making them useless (everything will be stored in the "infinite" bucket).

I've also found that some metric helpers aren't "useful"; for example `The total number of providers configured` for `authentik_outpost_ldap_requests`, which isn't the case.

So this PR attempts to fix both these issues and renames the metrics with something that follows [Prometheus best practice](https://prometheus.io/docs/practices/naming/#metric-names). I haven't removed the old metrics to avoid breaking existing alerts and dashboards.

### New Features

- Add new metrics for outposts metrics 

### Breaking Changes

Nothing

## Checklist

- [X] Local tests pass (`ak test authentik/`)
- [X] The code has been formatted (`make lint-fix`)

### DIFF
--- a/internal/outpost/flow/executor.go
+++ b/internal/outpost/flow/executor.go
@@ -8,6 +8,7 @@ import (
 	"net/http/cookiejar"
 	"net/url"
 	"strings"
+	"time"
 
 	"github.com/getsentry/sentry-go"
 	"github.com/prometheus/client_golang/prometheus"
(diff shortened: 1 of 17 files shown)


BeryJu (approved):
I'm assuming adding the _seconds suffix is both part of documentation and also best practices? Otherwise we could just change the measurement of the existing metrics without having these legacy metrics

PR #6418, opened 2023-07-28
web/admin: set required flag to false for user attributes

tanberry:
## Details
 On the Create user form, the field for Attributes is not required, so I removed the asterisk by editing `web/src/admin/users/UserForm.ts` to set required flag to `false`. 

## Changes

Attributes field on Create User no longer shown as Required (which it is not).

### New Features

n/a

### Breaking Changes

n/a

## Checklist

If applicable

### DIFF
--- a/web/src/admin/users/UserForm.ts
+++ b/web/src/admin/users/UserForm.ts
@@ -145,7 +145,7 @@ export class UserForm extends ModelForm<User, number> {
             </ak-form-element-horizontal>
             <ak-form-element-horizontal
                 label=${msg("Attributes")}
-                ?required=${true}
+                ?required=${false}
                 name="attributes"
             >
                 <ak-codemirror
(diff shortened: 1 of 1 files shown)


rissson (approved):

BeryJu (changes requested):
So this kinda works and kinda doesn't

Since we default to `{}` in L153, it'll work without changing anything. However the field is still technically required, as if that value is `null` there'll be an error

tanberry:
> So this kinda works and kinda doesn't
> 
> Since we default to `{}` in L153, it'll work without changing anything. However the field is still technically required, as if that value is `null` there'll be an error

@BeryJu when you say "if that value is null" which value do you mean? Because I have created new Users without adding any attributes, so you must not mean the value entered into the Attributes section/field... when you have time I'd love to learn more about this... and actually @baloc and I will need to edit the new Create User docs if that Attributes field is indeed required.

BeryJu (approved):
We can just intercept the potential null value and set it to the default empty user attributes

cc @kensternberg-authentik lmk if there's a cleaner way to do this

PR #6436, opened 2023-07-31
root: add get_int to config loader instead of casting to int everywhere

BeryJu:
## Details

REPLACE ME

---

## Checklist

If applicable

### DIFF
--- a/authentik/lib/config.py
+++ b/authentik/lib/config.py
@@ -213,11 +213,12 @@ class ConfigLoader:
         attr: Attr = get_path_from_dict(root, path, sep=sep, default=Attr(default))
         return attr.value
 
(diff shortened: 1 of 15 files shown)


authentik/lib/config.py
```diff
@@ -213,6 +213,13 @@ def get(self, path: str, default=None, sep=".") -> Any:
         attr: Attr = get_path_from_dict(root, path, sep=sep, default=Attr(default))
         return attr.value
 
+    def get_int(self, path: str, default = 0) -> int:
+        """Wrapper for get that converts value into int"""
+        try:
+            return int(self.get(path, default))
+        except ValueError:
```
rissson, line 220: We should probably log at least a warning here

rissson (changes requested):

PR #6478, opened 2023-08-05
api: optimise pagination in API schema

BeryJu:
## Details

Instead of having the schema of the pagination in each response component, have the pagination object once and reference it

---

## Checklist

If applicable

### DIFF
--- a/authentik/api/pagination.py
+++ b/authentik/api/pagination.py
@@ -2,6 +2,43 @@
 from rest_framework import pagination
 from rest_framework.response import Response
 
+PAGINATION_COMPONENT_NAME = "Pagination"
+PAGINATION_SCHEMA = {
+    "type": "object",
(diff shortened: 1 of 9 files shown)


authentik/api/pagination.py
```diff
@@ -35,41 +72,7 @@ def get_paginated_response_schema(self, schema):
...
-                        "next",
-                        "previous",
-                        "count",
-                        "current",
-                        "total_pages",
-                        "start_index",
-                        "end_index",
-                    ],
-                },
+                "pagination": {"$ref": f"#/components/schemas/{PAGINATION_COMPONENT_NAME}"},
```
BeryJu, line 75: Ideally this wouldn't be hardcoded and we directly get the reference from the created component, but that's a bit of a challenge with the imports

PR #6482, opened 2023-08-05
providers/proxy: set outpost session cookie to httponly and secure wh…

BeryJu:
…en possible

## Details

security improvement, possible issues when a provider has an external URL configured with HTTPS but is being accessed via HTTP, but that's not supported anyways 

---

## Checklist

If applicable


rissson (approved):

rissson:
Maybe set samesite as well?

BeryJu:
> Maybe set samesite as well?

I just talked about that on discord too, SameSite could on theory break some SSO stuff, which is why the main server also sets it to Lax (allthough I think those issues only occurred with SAML)

PR #6560, opened 2023-08-17
root: always use persistent database connections

rissson:
## Details

Ideally this would avoid re-opening a database connection for every database query, even when not using pg_bouncer.

Upside: we're avoiding creating connections for every database query.

Downside: we would always have, at most, N connections open, N being the number of workers, once every worker has been hit with a web request. Also, if something breaks, the connection is not re-created automatically, if I understand correctly.

We could also set [CONN_HEALTH_CHECKS](https://docs.djangoproject.com/en/4.1/ref/settings/#std-setting-CONN_HEALTH_CHECKS) to true to mitigate that last bit, which will check the connection health once per web request.

Finally, this may help with auto-rotating secrets, in which case the connection would stay opened and healthy from when the password has been changed on postgres' side, until it reaches authentik.

---

## Checklist

If applicable

### DIFF
--- a/authentik/root/settings.py
+++ b/authentik/root/settings.py
@@ -279,14 +279,14 @@ DATABASES = {
         "SSLROOTCERT": CONFIG.get("postgresql.sslrootcert"),
         "SSLCERT": CONFIG.get("postgresql.sslcert"),
         "SSLKEY": CONFIG.get("postgresql.sslkey"),
+        # https://docs.djangoproject.com/en/4.0/ref/databases/#persistent-connections
+        "CONN_MAX_AGE": None,
     }
 }
 
 if CONFIG.get_bool("postgresql.use_pgbouncer", False):
     # https://docs.djangoproject.com/en/4.0/ref/databases/#transaction-pooling-server-side-cursors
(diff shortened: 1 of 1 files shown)


BeryJu (commented):
Looks good, allthough I think we should also set `CONN_HEALTH_CHECKS` as you already mentioned (I'm not 100% sure about the performance impact of both of these changes, I'm estimating it'll lower the avg% for response time but could worse then p99 (especially when the password rotation does happen)

rissson:
> we should also set `CONN_HEALTH_CHECKS`

Done.

> I'm not 100% sure about the performance impact of both of these changes, I'm estimating it'll lower the avg% for response time but could worse then p99 (especially when the password rotation does happen)

I've updated the description to better differentiate web request and database queries. As the check only happens once per web request, and only if the database is queried, it shouldn't be a lot of overhead, and certainly less than re-creating a brand new connection to the database.

Also, for reference, the check made by django to see if the connection is alive is [`SELECT 1`](https://github.com/django/django/blob/d25f3892114466d689fd6936f79f3bd9a9acc30e/django/db/backends/postgresql/base.py#L386)

BeryJu (approved):

PR #6566, opened 2023-08-17
web: start experimenting with the new view transitions API

BeryJu:
## Details

Start experimenting with the [View Transitions API](https://developer.mozilla.org/en-US/docs/Web/API/View_Transitions_API) (only supported in chromium-like browsers, but there is a very easy fallback to the current way of doing things)

This PR adds the view transitions for the ak-router-outlet component which turns navigations into a nice fade

It also adds it to the flow executor with the goal of making transitions between cards look better, there's some commented out code in there with attempts to make it transition the height to give off the look of the card expanding/shrinking, but in my testing it looks a little bit jank, hence commented out

there's also quite a few errors in there as typescript doesn't have types for this yet

---

## Checklist

If applicable

### DIFF
--- a/web/package-lock.json
+++ b/web/package-lock.json
@@ -71,6 +71,7 @@
                 "@trivago/prettier-plugin-sort-imports": "^4.3.0",
                 "@types/chart.js": "^2.9.41",
                 "@types/codemirror": "5.60.15",
+                "@types/dom-view-transitions": "^1.0.4",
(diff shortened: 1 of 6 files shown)


BeryJu:
For some reason the changes on this PR cause some issues with z-indexes, namely in the admin interface where the sidebar is rendered above the modal backdrop

PR #6674, opened 2023-08-29
blueprints: fix policy exception causing password stage to be skipped after upgrade

### LINKED ISSUE #6673: docker 2023.8.0 - login without password, just valid username
i just updated my docker test install and noticed, that I can login with a valid username but without a password from any device...

akadmin is valid ..
notauser is not valid and no login..

BeryJu:
## Details

closes #6673

due to changing defaults, after upgrading from some versions, a new default policy would be run at the wrong time causing it to fail, which would cause the password stage to get skipped

this fixes both the missing default value and also makes the policy more resilient to errors

---

## Checklist

If applicable

### DIFF
--- a/blueprints/default/flow-default-authentication-flow.yaml
+++ b/blueprints/default/flow-default-authentication-flow.yaml
@@ -71,10 +71,12 @@ entries:
     name: default-authentication-flow-password-stage
   attrs:
     expression: |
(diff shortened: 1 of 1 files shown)


esackbauer:
> and also makes the policy more resilient to errors

Does that mean Authentik is still prone to timing attacks, just not so much anymore?

PR #6720, opened 2023-08-31
root: connect to backend via socket

BeryJu:
## Details

I don't know why we haven't been doing this all along tbh

Instead of having the go wrapper connect to gunicorn via localhost:8000, we can just have gunicorn listen on a socket and go forward to that

This also finally fixes the issue where gunicorn logs on startup were not json format, by taking out the logging config from django's settings and making them re-usable

---

## Checklist

If applicable


BeryJu:
After some more testing with both the embedded outpost and external outpost, everything works as it should

PR #6768, opened 2023-09-05
website/dev-docs: add raw markdown template

tanberry:
## Details

added a new raw markdown template for procedural topics, tweaked wording on a few existing topics in the templates section.

---
If applicable


website/developer-docs/docs/templates/procedural-md.md
```diff
@@ -0,0 +1,45 @@
+---
+title: "Markdown template: procedural"
+---
+
+```
```
BeryJu, line 5: I dont think escaping markdown within markdown is a good idea, rather this file should just be the template and we can link to github instead to view the raw template

tanberry: That does sound better... but how do I add a pure markdown file to GitHub without Netlify and Docusaurus converting it to html?

BeryJu: You can just add it as is and let docusaurus do its thing and just not include it in the sidebar

BeryJu (approved):

Issue #6815, opened 2023-09-08
multiple workers: repeated tasks are scheduled by each worker instead of just once


a-gerhard:
**Describe the bug**
We've got 10 workers and 1 server in our setup. ldap_sync_all is scheduled 10 times in each 2 hour window (to be more accurate, 10 times within 1 hour after each full even hour). This looks like each worker is creating its own schedule (`crontab(minute=fqdn_rand("sources_ldap_sync"), hour="*/2")` according to `sources/ldap/settings.py).

This may affect other schedules as well

**To Reproduce**

1. create a setup with multiple workers (e.g., using docker-compose replication)
2. use celery flower to observe tasks
3. wait until you see the schedules

**Expected behavior**
The task is scheduled according to its schedule, and only once per time interval. Not once per worker, but once in total.

**Screenshots**
Some tasks failed because redis ran out of memory due to the amount of data that was synchronized, which also caused some other issues.
![image](https://github.com/goauthentik/authentik/assets/986924/aa130bb2-f90a-4dc1-aa9a-249679d53a20)

**Version and Deployment (please complete the following information):**

-   authentik version: 2023.8.2
-   Deployment: docker-compose

**Additional context**
This has a severe impact on the issues I described in #6810

a-gerhard:
Or use something like celery beat.

septatrix:
> Or use something like celery beat.

Sure, but stuff like "the first worker has to be configure like x and all other like y" is more complicated and error prone -- and is currently not documented anywhere (which is the bigger issue!)

PR #6837, opened 2023-09-11
root: make Celery worker concurrency configurable

a-gerhard:
To work around #6815, but also because it'd be a good idea in general, I made it possible to configure Celery worker concurrency in a single worker.


website/docs/installation/configuration.md
```diff
@@ -332,6 +332,16 @@ Configure how many gunicorn threads a worker processes should have (see https://
 
 Defaults to 4.
 
+### `AUTHENTIK_WORKER__CONCURRENCY`
+
+:::info
+Requires authentik 2023.9.1
```
BeryJu, line 338:
```suggestion
Requires authentik 2023.9.0
```

BeryJu (approved):

PR #6852, opened 2023-09-11
web: Detangling some circular dependencies in Admin and User

kensternberg-authentik:
Admin, User, and Flow should not dependend upon each other, at least not in a circular way.  If Admin and User depend on Flow, that's fine, but Flow should not correspondingly depend upon elements of either; if they have something in common, let's put them in `@goauthentik/common` or find some other smart place to store them.

This commit refactors the intentToLabel and actionToLabel functions into `@goauthentik/common/labels` and converts them to static tables for maintenance purposes.

(I've accidentally deleted the checklist.  The code has been formatted, prechecked, manually inspected and automatically tested, to the extent that that last is possible.  No documentation or labeling changes were made.)


web/.eslintrc.precommit.json
```diff
@@ -23,6 +23,7 @@
         "quotes": ["error", "double", { "avoidEscape": true }],
         "semi": ["error", "always"],
         "@typescript-eslint/ban-ts-comment": "off",
-        "sonarjs/cognitive-complexity": ["error", 9]
+        "sonarjs/cognitive-complexity": ["error", 9],
+        "sonarjs/no-nested-template-literals": "off"
```
kensternberg-authentik, line 27: I _hated_ doing this, but six of the files I changed had nested template literals _with_ `msg()` strings embedded, which makes them more or less immobile.  Frustrating as heck.

web/src/admin/admin-overview/cards/RecentEventsCard.ts
```diff
@@ -1,7 +1,8 @@
 import "@goauthentik/admin/events/EventInfo";
-import { ActionToLabel, EventGeo } from "@goauthentik/admin/events/utils";
+import { EventGeo } from "@goauthentik/admin/events/utils";
 import { DEFAULT_CONFIG } from "@goauthentik/common/api/config";
 import { EventWithContext } from "@goauthentik/common/events";
+import { actionToLabel } from "@goauthentik/common/labels";
```
kensternberg-authentik, line 5: _All_ of the changes are of this: referring to the function in a different location, and changing the casing to indicate it's a _function_ rather than a type or class.

web/src/common/labels.ts
```diff
@@ -0,0 +1,58 @@
...
+    [EventActions.ConfigurationError, msg("Configuration error")],
+    [EventActions.ModelCreated, msg("Model created")],
+    [EventActions.ModelUpdated, msg("Model updated")],
+    [EventActions.ModelDeleted, msg("Model deleted")],
+    [EventActions.EmailSent, msg("Email sent")],
+    [EventActions.UpdateAvailable, msg("Update available")],
+]);
+
+export const actionToLabel = (action?: EventActions): string =>
+    eventActionToLabel.get(action) ?? action ?? "";
```
BeryJu, line 49: ~~we should probably have a more visible fallback like we do for the severity below~~

events can also have a custom action (which always has `custom_` prefixed), so as a fallback we should probably just return the action

BeryJu (changes requested):

web/src/common/labels.ts
kensternberg-authentik: That's what this does; only if _that_ fails does it return an empty string. It's the same logic as the original, just less, ah, "iffy."  See: [web/src/admin/events/utils.ts](https://github.com/goauthentik/authentik/pull/6852/files#diff-467f1f91e11f15e16fb5d4465813d492864eddbb23f532b84abc722ab710d8bd)

BeryJu: oh yeah I misread that, LGTM

BeryJu (approved):

web/src/common/labels.ts
kensternberg-authentik: Cool, thanks. Yeah, I just started taking the `??` operator seriously recently myself, so I'm still getting used to its semantics. Sorry for going all hypergraphia on ya, I just didn't want to be misunderstood.

PR #6861, opened 2023-09-12
root: include generated ts client in dev builds [AUTH-77]

rissson:
## Details

Include generated ts client in dev builds to allow for `deploy_me` to have an up-to-date frontend.

---

## Checklist

If applicable

### DIFF
--- a/.github/workflows/release-publish.yml
+++ b/.github/workflows/release-publish.yml
@@ -27,9 +27,12 @@ jobs:
           registry: ghcr.io
           username: ${{ github.repository_owner }}
           password: ${{ secrets.GITHUB_TOKEN }}
+      - name: make empty ts client
+        run: mkdir -p ./gen-ts-client
       - name: Build Docker Image
         uses: docker/build-push-action@v4
         with:
+          context: .
           push: ${{ github.event_name == 'release' }}
           secrets: |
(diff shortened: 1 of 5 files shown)


rissson:
This should be good. From my testing the API client still is correctly built even if the `gen-ts-api` folder is empty, but I'd still like someone to check again.

.github/workflows/release-publish.yml
```diff
@@ -27,9 +27,12 @@ jobs:
           registry: ghcr.io
           username: ${{ github.repository_owner }}
           password: ${{ secrets.GITHUB_TOKEN }}
+      - name: make empty ts client
+        run: mkdir -p ./gen-ts-client
```
BeryJu, line 31: I still wish the build was possible without this, but alas

Allthough maybe we should add a `docker` target to the makefile that people can use locally, as not doing this would lead to some potentially confusing errors

BeryJu (approved):

.github/workflows/release-publish.yml
rissson: Yes, and we could include the necessary arguments, as `docker build .` doesn't work out of the box. I usually run with `DOCKER_BUILDKIT=1 docker build . --progress plain --platform linux/amd64 -t authentik:test` on a Linux machine.

BeryJu: TIL about `--progress plain`

BeryJu (approved):

PR #6866, opened 2023-09-12
web: remove `./element`⇢`./user` references

kensternberg-authentik:
web: remove `element`⇢ `./user` references

The offender here is UserDevicesList, which despite being in `elements` is only used by the admin/user/UserViewPage. The problem is that UserDevicesList, despite being in `admin`, inherits from `user`, so moving it would have created a new admin⇢user reference, and the whole point of this exercise is to get rid of references that point “up” from the foundational pieces to the views, or that refer to components in sibling applications.

After examining UserDevicesList, I realized that *every feature* of MFADevicesList had been overridden: the rows, the columns, the toolbar, and the endpoint all had custom overrides. Nothing was left of MFADevicesList after that. Even the property that the web component used had been completely changed. The only thing they had in common was that they both inherited from `Table<Device>`.

Refactoring UserDevicesList so that it inherited directly from `Table<Device>` and then moving it into `./admin/users` was the obvious and correct step.

Both used the same label table, so that went into the `common/labels` folder.

Along the way, I cleaned up a few minor details. Just little things, like the repeated invocation of:

    new AuthenticatorsApi(DEFAULT_CONFIG).authenticatorAdminMETHODDestroy({ id: device.pk });

This is repeated five times, once for each Method. By creating these:

            const api = new AuthenticatorsApi(DEFAULT_CONFIG);
            const id = { id: device.pk };

The method invocation could be just `api.authenticatorsMETHODDestroy(id)`, which is easier on the eyes. See the MFADevicesPage for the full example.

Similarly,

    return [
       new TableColumn(msg("Name"), ""),
       new TableColumn(msg("Type"), ""), 
       new TableColumn("")
    ];

is more straightforward as:

    const headers = [msg("Name"), msg("Type"), ""];
    return headers.map((th) => new TableColumn(th, ""));

We’ve labeled what we’re working with, and web developers ought to know that `th` is the HTML code for `table header`;

## Checklist

If applicable


web/src/admin/users/UserDevicesList.ts
```diff
@@ -36,41 +36,30 @@ export class UserDeviceList extends MFADevicesPage {
             });
     }
 
+    columns(): TableColumn[] {
+        const headers = [msg("Name"), msg("Type"), msg("Confirmed")];
+        return headers.map((th) => new TableColumn(th, ""));
```
BeryJu, line 41: I think this is more readable with just returning an array of TableColumn instances imo (a very subjective nit)

web/package.json
```diff
@@ -15,7 +15,7 @@
         "build-proxy": "run-s build-locales rollup:build-proxy",
         "watch": "run-s build-locales rollup:watch",
         "lint": "eslint . --max-warnings 0 --fix",
-        "lint:precommit": "eslint --max-warnings 0 --config ./.eslintrc.precommit.json $(git status --porcelain | cut -c2- | grep '^[M?]' | cut -c7- | grep -E '\\.(ts|js|tsx|jsx)$') ",
+        "lint:precommit": "eslint --max-warnings 0 --config ./.eslintrc.precommit.json $(git status --porcelain | grep '^[M?][M?]' | cut -c8- | grep -E '\\.(ts|js|tsx|jsx)$') ",
```
BeryJu, line 18: We can use https://pre-commit.com/ instead of doing all of this grepping/cutting lol

web/src/common/labels.ts
```diff
@@ -56,3 +56,11 @@ export const severityEnumToLabel = new Map<SeverityEnum | null | undefined, stri
 
 export const severityToLabel = (severity: SeverityEnum | null | undefined) =>
     severityEnumToLabel.get(severity) ?? msg("Unknown severity");
+
```
BeryJu, line 59:
```suggestion

// TODO: Add verbose_name field to now vendored OTP devices
```

BeryJu (approved):

PR #6891, opened 2023-09-14
web: detangle components from applications

kensternberg-authentik:
web: detangle `components` from `admin`.

This was the last inappropriate reference: something from `./components` referencing something in `./admin`, in this case the `ak-event-info` component. Used by both Users and Admin, moving it into `./components` was the obvious correct step.

`ak-event-info` is a look up table relating specific events in the event log to rich, textual representations; in the special case of model changes and email info, even more rich content is available in a dl/dt format. I’ve tableized the model changes and email info renderer, and I’ve extracted every event’s textual representation into its own method, converting the `switch/case` rendering statement into a `switch/case` dispatch switch. This has the virtue of isolating each unique case and making the dispatch switch short and coherent.

The conversion was done mechanistically; I gave the refactorer (Tide, in this case) instructions to duplicate the switch block and then convert every case into a method with a name patterned on the `case`. Going back to the original switch block, it was easy to duplicate the pattern matching and convert it into a dispatch switch.

And with this, there are zero cycles in the references between the different “packageable” sections of the UI. The only thing left to do is figure out how to redistribute `./elements` and `./components` in a way that makes sense for each.

## Details

REPLACE ME

---

## Checklist

If applicable


web/src/components/ak-event-info.ts
```diff
@@ -0,0 +1,484 @@
+import { DEFAULT_CONFIG } from "@goauthentik/common/api/config";
```
kensternberg-authentik, line 1: Nah, you need to see this one. I'm very proud.  :-)

web/src/components/ak-event-info.ts
```diff
@@ -0,0 +1,484 @@
...
+import PFFlex from "@patternfly/patternfly/layouts/Flex/flex.css";
+import PFBase from "@patternfly/patternfly/patternfly-base.css";
+
+import { EventActions, FlowsApi } from "@goauthentik/api";
+
+type Pair = [string, string | number | EventContext | EventModel | string[] | TemplateResult];
+
+// https://docs.github.com/en/issues/tracking-your-work-with-issues/creating-issues/about-automation-for-issues-and-pull-requests-with-query-parameters
+
+const emailMessageBody = (context: EventContext) => `
```
BeryJu, line 28: small nit: this is not for an email, this is a template that is passed to github via querystring (which is kind of a hacky solution), but it provides a link to open a github issue that includes the full stacktrace

BeryJu (approved):

BeryJu (approved):

Issue #3715, opened 2022-10-01
expand logging to get authentik crowdsec ready


cuthulino:
**Is your feature request related to a problem? Please describe.**
I am using [CrowdSec ](https://www.crowdsec.net/) for securing my exposed services.  
Authentik as my auth service is therfore exposed aswell.  
The actual logging behavior misses some information to use it with crowdsec to parse it.  
  
At the moment an invalid login attempt look like this:  
```
{"auth_via": "unauthenticated", "event": "invalid_login", "host": "auth.domain.de", "identifier": "username", "level": "info", "logger": "authentik.flows.stage", "pid": 187641, "request_id": "58c23b3a49544bbaa20dde888011a989", "stage": "default-authentication-identification", "stage_view": "authentik.stages.identification.stage.IdentificationStageView", "timestamp": "2022-10-01T20:13:50.435689"}
```

**Describe the solution you'd like**
To use the logs with crowdsec the remote ip address of the failed login attempt is needed.  
For counting and banning.  
  
For better usabillity the logtype should be changed aswell for failed authentikations, from "info" to "warning".  
Like this info logs can be hidden and failed auth can be parsed further.

It should look like this:
```
{"auth_via": "unauthenticated", "event": "invalid_login", "host": "auth.domain.de", "identifier": "username", "level": "warn", "logger": "authentik.flows.stage", "pid": 187641, "remote": "xx.xx.xx.xx" , "request_id": "58c23b3a49544bbaa20dde888011a989", "stage": "default-authentication-identification", "stage_view": "authentik.stages.identification.stage.IdentificationStageView", "timestamp": "2022-10-01T20:13:50.435689"}
```

BeryJu:
The default authentik flows do user identification and authentication separately, and the event `invalid_login` is when an invalid identifier is used to login

The actual log message for invalid credentials is

```json
{
    "action": "login_failed",
    "auth_via": "unauthenticated",
    "client_ip": "::1",
    "context": {
        "http_request": {
            "args": {
                "query": "next=%2F"
            },
            "method": "POST",
            "path": "/api/v3/flows/executor/default-authentication-flow/"
        },
        "password": "********************",
        "stage": {
            "app": "authentik_stages_password",
            "model_name": "passwordstage",
            "name": "default-authentication-password",
            "pk": "c9d44052811748b780bb9f8abdbb6d13"
        },
        "username": "akadmin"
    },
    "event": "Created Event",
    "host": "localhost:9000",
[...]

PR #6895, opened 2023-09-14
lib: add file logging to lib logging

ALameLlama:
## Details
This add an optional file logging to the logger, this will allow us to choose where we want to log the current console stream into a file.

By default, this is set as empty string and I've added checks so if this isn't set, it won't include it into the logger

This will help with allowing us to create a parse for crowdsec so we can look for the actions  login_failed and invalid_identifier

https://github.com/goauthentik/authentik/issues/3715

---

## Checklist

If applicable

### DIFF
--- a/authentik/lib/default.yml
+++ b/authentik/lib/default.yml
@@ -43,6 +43,7 @@ debug: false
 remote_debug: false
 
 log_level: info
+log_path: ""
 
 error_reporting:
   enabled: false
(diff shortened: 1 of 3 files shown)


BeryJu:
This is not supported on purpose. The intended way to do this is to use something like https://docs.docker.com/config/containers/logging/syslog/ which will send all of the containers' logs to syslog, which in most cases is stored as a file. Logging to a file within a container is very much an antipattern, because that is the job of the orchestration platform since that will handle rotation, archiving, etc

Additionally this would only write the python logs to file and not the go logs, which would miss quite a bit of extra info

PR #6913, opened 2023-09-15
sources/ldap: add warning when a property mapping returns None or bytes

### LINKED ISSUE #6889: Do not silently ignore byte and none properties in _build_object_properties
It took me quite some time to understand why my custom ldap mapping had no effect. 

In the byte-case I think an exception should be raised, in case of none maybe a warning should be logged ? 
Maybe allow custom ldap mapping to be marked as required ?

It would also help if this behavior would be highlighted in the documentaion

**Additional context**
The source is in authentik/sources/ldap/base.py in _build_object_properties line 154 - 158

BeryJu:
## Details

closes #6889

---

## Checklist

If applicable

### DIFF
--- a/authentik/sources/ldap/sync/base.py
+++ b/authentik/sources/ldap/sync/base.py
@@ -153,8 +153,10 @@ class BaseLDAPSynchronizer:
             try:
                 value = mapping.evaluate(user=None, request=None, ldap=kwargs, dn=object_dn)
                 if value is None:
+                    self._logger.warning("property mapping returned None", mapping=mapping)
                     continue
                 if isinstance(value, (bytes)):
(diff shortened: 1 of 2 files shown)


smusali (approved):

tograss:
Hi BeryJu, I'm unsure if it is ok to just warn on the None case. If it is done like this it is impossible to have some User with a property and others not. without triggering the warning. However, this is a common case and since it worked before it should not be rolled out like this. (Or that's at least what i think :-) ) Of course the users could just ignore the warning but this is not a nice solution. Maybe add a constant we can return to signal that it is expected that this property is None ?

Thanks for improving the byte case I really appreciate how you listen to your users !

Issue #6896, opened 2023-09-14
Using PGPool


boesr:
**Describe your question/**
I am running Pgpool II before my database. It was working with version 2023.6.2, but now fails. When running with `AUTHENTIK_POSTGRESQL__USE_PGBOUNCER=true` the frontend loads all data once, but just hangs afterwards. PGpool shows an error, which indicates, that the frontend (Authentik) cannot use authentication methods offered by pgpool (see https://www.pgpool.net/pipermail/pgpool-general/2022-June/008262.html).  Without that parameter I just get CSS written on the login page and the frontend isn't usable at all. The logs show the error below.

Do I have to switch to certain authentication methods or what did change from 2023.6.2 to 2023.8.x regarding that fact?

**Relevant infos**
2023.8.3 

**Screenshots**
/

**Logs**

Error when not using PGBOUNCER parameter:

```
[...]

PR #6949, opened 2023-09-21
root: Add setting to adjust database config for pgpool

boesr:
## Details

This PR resolves the connection issues of 2023.8.3 with Pgpool. It adds a parameter `AUTHENTIK_POSTGRESQL__USE_PGPOOL` to disable server side cursors without setting `CONN_MAX_AGE` to `none` (the pgbouncer configuration).

It solves #6896  

---

## Checklist

If applicable

### DIFF
--- a/authentik/lib/default.yml
+++ b/authentik/lib/default.yml
@@ -7,6 +7,7 @@ postgresql:
   port: 5432
   password: "env://POSTGRES_PASSWORD"
   use_pgbouncer: false
+  use_pgpool: false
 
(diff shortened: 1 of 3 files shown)


BeryJu (approved):
In the future I think it would probably make more sense to directly expose the `DISABLE_SERVER_SIDE_CURSORS` and `CONN_MAX_AGE` settings instead of having them behind implementation specific toggles, but since we already have an option for pgbouncer this is fine

PR #6996, opened 2023-09-26
website/docs: restructure some parts

BeryJu:
## Details

An initial attempt to make the doc structuring a bit better
- Move blueprints and API-related things into the main docs (doesn't really make sense as developer docs)
- Move source-related docs about the protocols themselves (OAuth, LDAP and SAML source themselves) into the main docs, to match what we have for providers
- also remove developer from the navbar (for now?)
- add redirects for all the paths to make sure links still work

---

## Checklist

If applicable


website/sidebarsIntegrations.js
```diff
@@ -156,16 +153,6 @@ module.exports = {
                         "sources/freeipa/index",
                     ],
                 },
-                "sources/general",
```
rissson, line 159: Perhaps we could add a pointer to their new location in `sources/index`

BeryJu: Good point, we should probably also add that for the integrations section above

PR #7045, opened 2023-10-02
web: laying the groundwork for future expansion

kensternberg-authentik:
## Details

This commit is a hodge-podge of updates and changes to the web.  Functional changes:

- Makefile: Fixed a bug in the `help` section that prevented the WIDTH from being accurately calculated if `help` was included rather than in-lined.

- ESLint: Modified the "unused vars" rule so that variables starting with an underline are not considered by the rule.  This allows for elided variables in event handlers.  It's not a perfect solution-- a better one would be to use Typescript's function-specialization typing, but there are too many places where we elide or ignore some variables in a function's usage that switching over to specialization would be a huge lift.

- locale: It turns out, lit-locale does its own context management.  We don't need to have a context at all in this space, and that's one less listener we need to attach t othe DOM.

- ModalButton: A small thing, but using `nothing` instead of "html``" allows lit better control over rendering and reduces the number of actual renders of the page.

- FormGroup: Provided a means to modify the aria-label, rather than stick with the just the word "Details."  Specializing this field will both help users of screen readers in the future, and will allow test suites to find specific form groups now.

- RadioButton: provide a more consistent interface to the RadioButton.  First, we dispatch the events to the outside world, and we set the value locally so that the current `Form.ts` continues to behave as expected.  We also prevent the "button lost value" event from propagating; this presents a unified select-like interface to users of the RadioButtonGroup.  The current value semantics are preserved; other clients of the RadioButton do not see a change in behavior.

- EventEmitter: If the custom event detail is *not* an object, do not use the object-like semantics for forwarding it; just send it as-is.

- Comments: In the course of laying the groundwork for the application wizard, I throw a LOT of comments into the code, describing APIs, interfaces, class and function signatures, to better document the behavior inside and as signposts for future work.

## Details

REPLACE ME

---

## Checklist

If applicable


Makefile
```diff
@@ -28,10 +28,13 @@ CODESPELL_ARGS = -D - -D .github/codespell-dictionary.txt \
 
 all: lint-fix lint test gen web  ## Lint, build, and test everything
```
kensternberg-authentik, line 30: Automatically derive the width of the commands that will be rendered.  Also fixes a bug where a multi-file Makefile will break the scanner.

web/.eslintrc.json
```diff
@@ -16,11 +16,21 @@
         "sourceType": "module"
     },
     "plugins": ["@typescript-eslint", "lit", "custom-elements"],
+    "ignorePatterns": ["authentik-live-tests/**"],
     "rules": {
         "indent": "off",
         "linebreak-style": ["error", "unix"],
         "quotes": ["error", "double", { "avoidEscape": true }],
         "semi": ["error", "always"],
-        "@typescript-eslint/ban-ts-comment": "off"
```
kensternberg-authentik, line 24: Allow variables that are prefixed with an underscore to exist without being used.  Allows for temporary variables and unused arguments in source code.  It's not a perfect solution-- a better one would be to use Typescript's function-specialization typing, but there are  too many places where we elide or ignore some variables in a function's usage that switching over to specialization would be a huge lift.

web/src/elements/ak-locale-context/ak-locale-context.ts
```diff
@@ -2,13 +2,11 @@ import { EVENT_LOCALE_CHANGE } from "@goauthentik/common/constants";
 import { EVENT_LOCALE_REQUEST } from "@goauthentik/common/constants";
 import { customEvent, isCustomEvent } from "@goauthentik/elements/utils/customEvents";
```
kensternberg-authentik, line 4: It turns out, lit-locale does its own context management. We don't need to provide our own context at all in this space, and that's one less listener we need to attach to the DOM.

web/src/elements/buttons/ModalButton.ts
```diff
@@ -1,7 +1,7 @@
 import { AKElement } from "@goauthentik/elements/Base";
 import { PFSize } from "@goauthentik/elements/Spinner";
```
kensternberg-authentik, line 3: A small thing, but using `nothing` instead of "html``" allows lit better control over  rendering and reduces the number of actual renders of the page.

web/src/elements/forms/FormGroup.ts
```diff
@@ -8,11 +8,26 @@ import PFForm from "@patternfly/patternfly/components/Form/form.css";
...
+/**
+ * TODO: Listen for custom events from its children about 'invalidation' events, and
+ * trigger the `expanded` property as needed.
+ */
+
 @customElement("ak-form-group")
 export class FormGroup extends AKElement {
-    @property({ type: Boolean })
+    @property({ type: Boolean, reflect: true })
     expanded = false;
```
kensternberg-authentik, line 27: Providing a means to modify the aria-label, rather than stick with the just the word  "Details." Specializing this field will both help users of screen readers in the future, and will  allow test suites to find specific form groups now.

web/src/elements/forms/Radio.ts
```diff
@@ -15,7 +17,7 @@ export interface RadioOption<T> {
 }
 
 @customElement("ak-radio")
```
kensternberg-authentik, line 19: Provide a more consistent interface to the RadioButton. First, we dispatch the  events to the outside world, and we set the value locally so that the current `Form.ts` continues  to behave as expected. We also prevent the "button lost value" event from propagating; this  presents a unified select-like interface to users of the RadioButtonGroup. The current value  semantics are preserved; other clients of the RadioButton do not see a change in behavior.

web/src/elements/forms/Form.ts
```diff
@@ -191,6 +241,9 @@ export abstract class Form<T> extends AKElement {
         return json as unknown as T;
     }
 
+    /**
+     * As far as anyone can remember, this isn't being used.
```
BeryJu, line 245: I actually meant to fix this, this function should be used almost everywhere above in `serializeForm`, see https://github.com/goauthentik/authentik/pull/6806/commits/67bb50ac88f70a3b75845ccf036e7c76848e79d9

BeryJu: (For context, with this wrapper we can set `name` of an `ak-form-element-horizontal` to `foo.bar` and it will be translated to `data['foo']['bar']`, which we need at least for the RBAC stuff)

web/src/elements/utils/eventEmitter.ts
```diff
@@ -9,16 +9,21 @@ export const isCustomEvent = (v: any): v is CustomEvent =>
 
 export function CustomEmitterElement<T extends Constructor<LitElement>>(superclass: T) {
     return class EmmiterElementHandler extends superclass {
-        dispatchCustomEvent(eventName: string, detail = {}, options = {}) {
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
kensternberg-authentik, line 12: If the detail is a scalar or array, package it directly into the custom event, rather than including it with the target and such.  This is probably over-engineering, and I should probably just remove `target` at this point.  I think it's a YAGNI issue.

kensternberg-authentik (commented):
Aside from the signposts left here, most of my updates were comments as I was learning the system.

web/src/elements/buttons/ModalButton.ts
BeryJu: we might want to make that change everywhere later

web/src/elements/forms/Form.ts
kensternberg-authentik: Oh!  Okay, that makes sense.

BeryJu (approved):

web/src/elements/buttons/ModalButton.ts
kensternberg-authentik: I do it as I touch files.  Things that don't need to be updated or fixed ... don't need to be updated or fixed.  Yet.  :-)

PR #7053, opened 2023-10-03
web: package up horizontal elements into their own components

kensternberg-authentik:
# Details

This commit introduces a number of “components.” Jens has this idiom:

       <ak-form-element-horizontal label=${msg("Name")} name="name" ?required=${true}>
           <input
               type="text"
               value="${ifDefined(this.instance?.name)}"
               class="pf-c-form-control"
               required
           />
       </ak-form-element-horizontal>

It’s a very web-oriented idiom in that it’s built out of two building blocks, the “element-horizontal” descriptor, and the input object itself. This idiom is repeated a lot throughout the code. As an alternative, let’s wrap everything into an inheritable interface:

      <ak-text-input
          name="name"
          label=${msg("Name")}
          value="${ifDefined(this.instance?.name)}
          required
      >
      </ak-text-input>

This preserves all the information of the above, makes it much clearer what kind of interaction we’re having (sometimes the `type=` information in an input is lost or easily missed), and while it does require you know that there are provided components rather than the pair of layout-behavior as in the original it also gives the developer more precision over the look and feel of the components.

*Right now* these components are placed into the LightDOM, as they are in the existing source code, because the Form handler has a need to be able to “peer into” the “element-horizontal” component to find the values of the input objects. In a future revision I hope to place the burden of type/value processing onto the input objects themselves such that the form handler will need only look for the `.value` of the associated input control.

Other fixes:

- update the FlowSearch() such that it actually emits an input event when its value changes.
- Disable the storybook shortcuts; on Chrome, at least, they get confused with simple inputs
- Fix an issue with precommit to not scan any Python with ESLint! :-)

Storybook examples for most of the new components provided have also been provided.

I’ve also provided an E2E test for the Oauth2 New Provider sequence. I used the Oauth2 provider page as my experiment in seeing if the horizontal-element wrappers could be used instead of the raw wrappers themselves, and I wanted to make sure a test existed that asserts that filling out THAT form in the ProvidersList and ProvidersForm didn’t break anything.

I have updated the WDIO tests to do just that; the test is simple, but it does exercise the `name` field of the Provider, something not needed in the Wizard because it’s set automatically based on the Application name, and it even asserts that the new Provider exists in the list of available Providers when it’s done.

## Checklist

If applicable


web/src/components/ak-textarea-input.ts
```diff
@@ -0,0 +1,58 @@
+import { AKElement } from "@goauthentik/elements/Base";
+
+import { TemplateResult, html, nothing } from "lit";
+import { customElement, property } from "lit/decorators.js";
+
+@customElement("ak-textarea-input")
+export class AkTextareaInput extends AKElement {
```
kensternberg-authentik, line 7: Do this enough and you start to see two abstraction patterns, one around the content of the Input and one around the Layout.  I'll figure out both eventually.

BeryJu (approved):

kensternberg-authentik (commented):
Looks good.  Thanks for catching my fuckup.  :-)

PR #7074, opened 2023-10-05
root: Support PyCharm's test runner

PKizzle:
## Details

PyCharm did not correctly work with the custom test runner (`authentik/root/test_runner.py`).
I have added support for running individual tests directly from the UI without having to disable the custom test runner.
`manage.py` still functions like before and pytest is used for test execution.

### DIFF
--- a/authentik/root/test_runner.py
+++ b/authentik/root/test_runner.py
@@ -1,4 +1,5 @@
 """Integrate ./manage.py test with pytest"""
+import os
 from argparse import ArgumentParser
 from unittest import TestCase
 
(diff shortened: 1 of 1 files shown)


authentik/root/test_runner.py
```diff
@@ -47,16 +47,53 @@ def __init__(self, verbosity=1, failfast=False, keepdb=False, **kwargs):
...
 
+        for label in test_labels:
+            valid_label_found = False
+            label_as_path = os.path.abspath(label)
+            if os.path.exists(label_as_path):
+                self.args.append(label_as_path)
+                valid_label_found = True
+            else:
+                path_pieces = label.split(".")
+                for i in range(-1, -3, -1):
```
BeryJu, line 81: Should probably add a comment for why these numbers

BeryJu (commented):
Pytest currently displays tests as `file::class:method` in its output (`authentik/stages/authenticator_totp/tests.py::AuthenticatorTOTPStage::test_api_delete`), which the can be passed in as-is to the current implementation. It would be great if we could keep those two things the same, i.e. have pytest log `module.class.method`; since I'm assuming the need to use `module.class.method` as argument is required to make it compatible with PyCharm?

PKizzle:
Yes exactly. It is basically a converter from `module.class.method` to `path_to_file::class::method`.

BeryJu (approved):

PR #7092, opened 2023-10-06
root: Upgrade Python syntax for version 3.11 using pyupgrade

PKizzle:
Result of running ``pyupgrade --py311-plus `find ./authentik/ -name "*.py" -type f` `` as project uses Python 3.11

Also, consider adding the pyupgrade command to the GitHub actions to automatically upgrade all new code.
Be aware that the developer is [unwilling to allow ignoring certain source code lines](https://github.com/asottile/pyupgrade/issues/345) so it may break certain functionality that requires workarounds.
In this case several manual adjustments were necessary as there are `type` collisions and the new `Element | None` type hint instead of `Optional[Element]` does not work reliably.

More information can be found here: https://github.com/asottile/pyupgrade

### DIFF
--- a/authentik/admin/api/tasks.py
+++ b/authentik/admin/api/tasks.py
@@ -124,7 +124,7 @@ class TaskViewSet(ViewSet):
             task_func.delay(*task.task_call_args, **task.task_call_kwargs)
             messages.success(
                 self.request,
-                _("Successfully re-scheduled Task %(name)s!" % {"name": task.task_name}),
+                _("Successfully re-scheduled Task {name}!".format(name=task.task_name)),
             )
             return Response(status=204)
         except (ImportError, AttributeError):  # pragma: no cover
--- a/authentik/admin/api/tasks.py
+++ b/authentik/admin/api/tasks.py
@@ -124,7 +124,7 @@ class TaskViewSet(ViewSet):
             task_func.delay(*task.task_call_args, **task.task_call_kwargs)
             messages.success(
                 self.request,
-                _("Successfully re-scheduled Task {name}!".format(name=task.task_name)),
+                _(f"Successfully re-scheduled Task {task.task_name}!"),
             )
             return Response(status=204)
         except (ImportError, AttributeError):  # pragma: no cover
--- a/authentik/api/decorators.py
+++ b/authentik/api/decorators.py
@@ -1,6 +1,7 @@
 """API Decorators"""
 from functools import wraps
-from typing import Callable, Optional
+from typing import Optional
+from collections.abc import Callable
 
 from rest_framework.request import Request
 from rest_framework.response import Response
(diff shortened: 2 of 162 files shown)


authentik/admin/api/tasks.py
```diff
@@ -122,7 +122,7 @@ def retry(self, request: Request, pk=None) -> Response:
             task_func.delay(*task.task_call_args, **task.task_call_kwargs)
             messages.success(
                 self.request,
-                _("Successfully re-scheduled Task %(name)s!" % {"name": task.task_name}),
+                _(f"Successfully re-scheduled Task {task.task_name}!"),
```
BeryJu, line 125: TIL that gettext supports extracting with f-strings

authentik/api/decorators.py
```diff
@@ -1,6 +1,6 @@
 """API Decorators"""
+from collections.abc import Callable
```
BeryJu, line 2: TIL that typing.Callable is deprecated https://docs.python.org/3/library/typing.html#typing.Callable

authentik/sources/oauth/models.py
```diff
@@ -65,9 +67,8 @@ def type(self) -> type["SourceType"]:
     def component(self) -> str:
         return "ak-source-oauth-form"
 
-    # we're using Type[] instead of type[] here since type[] interferes with the property above
     @property
-    def serializer(self) -> Type[Serializer]:
+    def serializer(self) -> SerializerType:
```
BeryJu, line 71: for this case we should probably just rename the other type attribute on the model since iirc it's just a property, but we can do that later

BeryJu (approved):
LGTM, allthough we might want to wait for #7048 and then directly go to python 3.12

PKizzle:
Shall I open a new PR with `pyupgrade` running for Python 3.12?

authentik/sources/oauth/models.py
PKizzle: Yes, that would be great to remove this workaround 👍

PKizzle: Will create an issue to keep track of this

PKizzle:
@BeryJu I have updated the PR to incorporate the changes of https://github.com/goauthentik/authentik/issues/7096.

Seems like the upgrade to Python 3.12 might take some time. Is there any reason not to merge this PR for version 3.11?
Running pyupgrade for 3.12 only introduces one additional change compared to this PR.

What is missing in this PR is a way to automatically run pyupgrade.

PKizzle:
`authentik/stages/prompt/api.py` also uses a "type" field which requires a pyupgrade workaround. I will create another issue to fix this as well.

BeryJu:
I think most of these changes were also made when we switched to ruff (#8498) which has pretty much the same policies _and_ also allows disabling it for specific lines like the XML `Element | None` not working

PR #5395, opened 2023-04-27
root: Improve Redis configuration

PKizzle:
# Details

Resolves #5077, resolves #1285

- Redis URL parser adapted from [gitea project](https://github.com/go-gitea/gitea) for Go implementation
- Redisstore adapted from [BoxGo](https://github.com/boxgo/redisstore) for Go implementation

### New Features

- **Uses one URL for configuration of Redis (shared between Go and Python)**
- The Python URL parser has been created in a way to resemble the Go version
- By default uses Redis for broker, channel, backend (current state), but also allows for custom options
- Moves part of configuration currently under redis category into new dedicated sections -> split into https://github.com/goauthentik/authentik/pull/7097
- Adds deprecation check to config parser to allow for easy transition between old and new configuration scheme
- Old Redis configuration env vars are automatically converted to a Redis config URL if no Redis config URL has been configured otherwise they are ignored
- The new Redis URL allows environment variables to be included in bash style ${VAR}

### Breaking Changes

- ~~Several new Redis configuration env vars are no longer supported as everything is handled using the URL~~

### Additional information

- Newly supported Redis configurations are only available using the new Redis config URL environment variable
- The Redis config URL requires special characters in parameters to be URL encoded (e.g. username, password)

## Todo

- [x] Breaking change shall be removed by automatically converting current config options to URL
- [x] Tests are currently missing
- [x] Find a way to fix https://github.com/django/channels_redis/issues/345
- [x] Not tested for unix socket connection
- [x] Not working for Redis cluster -> new PR required
- [x] Add documentation on how the new Redis config URL works
- [x] Split PR into ~~Dockerfile adjustments,~~ env variable deprecation logic and Redis URL parsing -> does not make sense

PR #7097, opened 2023-10-08
root: Restructure broker / cache / channel / result configuration

PKizzle:
This is part of https://github.com/goauthentik/authentik/pull/5395.

It restructures the configuration to separate Redis configuration clearly from cache and broker settings.

Furthermore, it enables custom configuration of the broker with its respective transport options, cache, channel and result backend. With this change i.e. RabbitMQ or Amazon SQS may be used as a broker.
Broker transport options need to be Base64 encoded and are ignored if not formatted correctly.

Deprecation logic has been added that automatically converts the old configuration names into the new ones and outputs a log message to encourage changing the configuration accordingly.


PKizzle:
@BeryJu This PR is ready to be reviewed in preparation for https://github.com/goauthentik/authentik/pull/7118.

authentik/root/settings.py
```diff
@@ -193,9 +192,9 @@
 
 CACHES = {
     "default": {
-        "BACKEND": "django_redis.cache.RedisCache",
-        "LOCATION": f"{_redis_url}/{CONFIG.get('redis.db')}",
-        "TIMEOUT": CONFIG.get_int("redis.cache_timeout", 300),
+        "BACKEND": CONFIG.get("cache.backend", "django_redis.cache.RedisCache"),
```
BeryJu, line 195: I'm not sure we want to expose this as a config option as we only have the one backend?

authentik/root/settings.py
```diff
@@ -253,9 +252,9 @@
 
 CHANNEL_LAYERS = {
     "default": {
-        "BACKEND": "channels_redis.core.RedisChannelLayer",
+        "BACKEND": CONFIG.get("channel.backend", "channels_redis.core.RedisChannelLayer"),
```
BeryJu, line 255: Same as above

authentik/root/settings.py
PKizzle: This gives the user the freedom to choose and use whatever backend might come in handy.
Authentik can still only offer support for Redis. For example one might want to use a database as cache instead.

BeryJu: I'm tempted to point people to https://goauthentik.io/docs/installation/configuration#custom-python-settings for this instead of having that in the native config options (mainly going by only exposing config options where we support all possible options)

authentik/root/settings.py
```diff
@@ -330,6 +331,23 @@
...
+if CONFIG.get("broker.transport_options"):
+    try:
+        _broker_transport_options_str = base64.b64decode(
+            CONFIG.get("broker.transport_options")
+        ).decode("utf-8")
+        _broker_transport_options_str = (
+            _broker_transport_options_str.strip().lstrip("{").rstrip("}")
+        )
+        _broker_transport_options_str = "{" + _broker_transport_options_str + "}"
+        _broker_transport_options.update(ast.literal_eval(_broker_transport_options_str))
```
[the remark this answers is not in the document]
PKizzle: I was not sure whether all possible `broker_transport_options` adhere to the JSON format.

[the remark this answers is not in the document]
PKizzle: I have changed it to `json.loads`.

PKizzle:
@BeryJu Could you take a look whether there are any more changes I should make?

authentik/lib/config.py
```diff
@@ -22,6 +24,24 @@
 ENV_PREFIX = "AUTHENTIK"
 ENVIRONMENT = os.getenv(f"{ENV_PREFIX}_ENV", "local")
 
+REDIS_ENV_KEYS = [
+    f"{ENV_PREFIX}_REDIS__HOST",
+    f"{ENV_PREFIX}_REDIS__PORT" f"{ENV_PREFIX}_REDIS__DB",
```
BeryJu, line 29:
```suggestion
    f"{ENV_PREFIX}_REDIS__PORT",
    f"{ENV_PREFIX}_REDIS__DB",
```

authentik/lib/config.py
```diff
@@ -180,6 +229,21 @@ def update_from_file(self, path: Path):
                 error=str(exc),
             )
 
+    def update_from_dict(self, update: dict):
+        """Update config from dict"""
+        self.__config.update(update)
+
+    @staticmethod
+    def _set_value_for_key_path(outer, path, value, sep="."):
```
BeryJu, line 237: I think this is due to the rebase, but `set_path_in_dict` should do the same thing as this

website/docs/installation/configuration.mdx
```diff
@@ -71,16 +71,38 @@ To check if your config has been applied correctly, you can run the following co
...
+-   `AUTHENTIK_CACHE__TIMEOUT`: Timeout for cached data until it expires in seconds, defaults to 300
+-   `AUTHENTIK_CACHE__TIMEOUT_FLOWS`: Timeout for cached flow plans until they expire in seconds, defaults to 300
+-   `AUTHENTIK_CACHE__TIMEOUT_POLICIES`: Timeout for cached policies until they expire in seconds, defaults to 300
+-   `AUTHENTIK_CACHE__TIMEOUT_REPUTATION`: Timeout for cached reputation until they expire in seconds, defaults to 300
+
+    :::info
+    `AUTHENTIK_CACHE__TIMEOUT_REPUTATION` only applies to the cache expiry, see [`AUTHENTIK_REPUTATION__EXPIRY`](#authentik_reputation__expiry) to control how long reputation is persisted for.
+    :::
+
+## Channel Settings
```
BeryJu, line 98: We should probably explain for the docs what "Channel" means

```suggestion
## Websocket Settings
```

authentik/lib/config.py
```diff
@@ -188,19 +252,12 @@ def update_from_env(self):
...
-            # Recursively convert path from a.b.c into outer[a][b][c]
-            current_obj = outer
-            dot_parts = relative_key.split(".")
-            for dot_part in dot_parts[:-1]:
-                if dot_part not in current_obj:
-                    current_obj[dot_part] = {}
-                current_obj = current_obj[dot_part]
             # Check if the value is json, and try to load it
             try:
                 value = loads(value)
```
BeryJu, line 257: Could we try to parse and load b64 here instead of explicitly having to call `get_dict_from_b64_json`? Maybe not such a good idea since some values might be b64 encoded that shouldn't be decoded?

BeryJu (changes requested):
Only some small nits/questions, otherwise LGTM

authentik/lib/config.py
```diff
@@ -113,6 +137,31 @@ def __init__(self, **kwargs):
...
+            if index == len(dot_parts) - 1:
+                return current_obj.pop(dot_part)
+            value = _pop_deprecated_key(current_obj[dot_part], dot_parts, index + 1)
+            if not current_obj[dot_part]:
+                current_obj.pop(dot_part)
+            return value
+
+        for deprecation, replacement in DEPRECATIONS.items():
+            if self.get(deprecation, default=UNSET) is not UNSET:
+                self.log(
```
rissson, line 157: We may also want to send notifications to the admins here.

PKizzle: How does one send notifications to admins?

authentik/lib/config.py
PKizzle: Furthermore, b64 encoded values can then only be set in env vars and not in the yaml files.

authentik/lib/config.py
PKizzle: I have remove the duplicate.

authentik/lib/config.py
PKizzle: Using `Notifications.objects.create` ?

website/docs/installation/configuration.mdx
PKizzle: I would like to keep the reference to Django Channel Layers as it makes it much easier to lookup should any questions arise.

authentik/lib/config.py
BeryJu: it would be with `Event.new` with the action set to `configuration_error`, but the config loader purposefully doesn't require any database model to be loaded so I'd prefer to keep it that way (maybe we could have a dynamic import and handle it failing?)

authentik/lib/config.py
PKizzle: Fixed.

authentik/lib/config.py
```diff
@@ -113,6 +138,40 @@ def __init__(self, **kwargs):
...
+                    + "Please update your configuration."
+                )
+                self.log(
+                    "warning",
+                    message,
+                )
+                try:
+                    from authentik.events.models import Event, EventAction
+
+                    Event.new(EventAction.CONFIGURATION_ERROR, message=message)
```
BeryJu, line 169:
```suggestion
                    Event.new(EventAction.CONFIGURATION_ERROR, message=message).save()
```

BeryJu (approved):

authentik/lib/config.py
PKizzle: Fixed.

PKizzle:
@BeryJu Are there any other changes you would like to see?

PR #7120, opened 2023-10-09
web: patternfly hints as ak-web-component

kensternberg-authentik:
## Details

Patternfly 5's "Hints" React Component, but ported to web components. The discovery that CSS Custom Properties are still available in child components, even if they're within independent ShadowDOMs, made this fairly easy to port from Handlebars to Lit-HTML.  Moving the definitions into `:host` and the applications into the root DIV of the component made duplicating the Patternfly 5 structure straightforward.

Despite the [Patternfly Elements documentation](https://patternflyelements.org/docs/develop/create/), there's a lot to Patternfly Elements that isn't well documented, such as their slot controller, which near as I can tell just makes it easy to determine if a slot with the given name is actually being used by the client code, but it's hard to tell why, other than that it provides an easy way to determine if some CSS should be included.

This is a 1:1 port of the HBR files and SCSS to a Patternfly-Elements form.  As such, there are probably a lot of edge cases using slotted content that I haven't addressed, but for the basic use cases we have in mind, it seems to work just fine.

## Checklist

If applicable


BeryJu (commented):
Looks good, the only thing its missing is a darker background when using the dark theme, maybe something like `--ak-dark-background-lighter` (I tried to make this work with pure css myself but failed to get there)

web/src/components/ak-hint/ak-hint.ts
```diff
@@ -0,0 +1,72 @@
...
+        --ak-hint__footer--child--MarginRight: var(--pf-global--spacer--md);
+
+        /* Hint Actions */
+        --ak-hint__actions--MarginLeft: var(--pf-global--spacer--2xl);
+        --ak-hint__actions--c-dropdown--MarginTop: calc(
+            var(--pf-global--spacer--form-element) * -1
+        );
+    }
+
+    :host([theme="dark"]) {
```
kensternberg-authentik, line 35: Once all the sub-components were receiving their styling instructions from this block, adding the host-with-attribute-match made it remarkably easy to get dark mode working.

web/src/components/ak-hint/ak-hint.ts
```diff
@@ -0,0 +1,72 @@
...
+        --ak-hint--BackgroundColor: var(--ak-dark-background-darker);
+--ak-hint--BorderColor: var(--ak-dark-background-lighter);
```
BeryJu, line 37:
```suggestion
        --ak-hint--BorderColor: var(--ak-dark-background-lighter);
```

BeryJu (approved):

PR #7121, opened 2023-10-09
website/integrations: grafana: add Helm and Terraform config examples

senare:
## Details

Ill try and add some example for doing this using terraform, will that be oki ?

Also added helm section

---

## Checklist

If applicable

### DIFF
--- a/website/integrations/services/grafana/index.mdx
+++ b/website/integrations/services/grafana/index.mdx
@@ -26,7 +26,6 @@ Create an application in authentik. Create an OAuth2/OpenID provider with the fo
 
 Note the Client ID and Client Secret values. Create an application, using the provider you've created above. Note the slug of the application you've created.
 
-
 ## Terraform provider
 
 ```hcl
@@ -148,27 +147,25 @@ If you are using a Helm `values.yaml` file instead, you have to set these option
 
 ```yaml
 grafana.ini:
-  auth:
-    signout_redirect_url: "https://authentik.company/application/o/<Slug of the application from above>/end-session/"
-    oauth_auto_login: true
-  auth.generic_oauth:
-    name: authentik
-    enabled: true
-    client_id: "<Client ID from above>"
-    client_secret: "<Client Secret from above>"
-    scopes: "openid profile email name login"
-    auth_url: "https://authentik.company/application/o/authorize/"
-    token_url: "https://authentik.company/application/o/token/"
-    api_url: "https://authentik.company/application/o/userinfo/"
-    # Optionally map user groups to Grafana roles
-    role_attribute_path: contains(groups[*], 'Grafana Admins') && 'Admin' || contains(groups[*], 'Grafana Editors') && 'Editor' || 'Viewer'
-
+    auth:
+        signout_redirect_url: "https://authentik.company/application/o/<Slug of the application from above>/end-session/"
+        oauth_auto_login: true
+    auth.generic_oauth:
+        name: authentik
+        enabled: true
+        client_id: "<Client ID from above>"
+        client_secret: "<Client Secret from above>"
+        scopes: "openid profile email"
+        auth_url: "https://authentik.company/application/o/authorize/"
+        token_url: "https://authentik.company/application/o/token/"
+        api_url: "https://authentik.company/application/o/userinfo/"
+        # Optionally map user groups to Grafana roles
+        role_attribute_path: contains(groups[*], 'Grafana Admins') && 'Admin' || contains(groups[*], 'Grafana Editors') && 'Editor' || 'Viewer'
 ```
 
 :::note
 For security reasons you shouldn't inline the client_secret in the values, but use a secret instead. For more information, see https://github.com/grafana/helm-charts/blob/main/charts/grafana/README.md#how-to-securely-reference-secrets-in-grafanaini
 :::
-
 
   </TabItem>
 </Tabs>
--- a/website/integrations/services/grafana/index.mdx
+++ b/website/integrations/services/grafana/index.mdx
@@ -26,6 +26,67 @@ Create an application in authentik. Create an OAuth2/OpenID provider with the fo
 
 Note the Client ID and Client Secret values. Create an application, using the provider you've created above. Note the slug of the application you've created.
 
+
+## Terraform provider
+
+```hcl
+
+data "authentik_flow" "default-provider-authorization-implicit-consent" {
+  slug = "default-provider-authorization-implicit-consent"
+}
+
+data "authentik_scope_mapping" "scope-email" {
+  name = "authentik default OAuth Mapping: OpenID 'email'"
+}
+
+data "authentik_scope_mapping" "scope-profile" {
+  name = "authentik default OAuth Mapping: OpenID 'profile'"
+}
+
+data "authentik_scope_mapping" "scope-openid" {
+  name = "authentik default OAuth Mapping: OpenID 'openid'"
+}
+
+resource "authentik_provider_oauth2" "grafana" {
+  name          = "Grafana"
+  #  Required. You can use the output of:
+  #     $ openssl rand -hex 16
+  client_id     = "my_client_id"
+
+  # Optional: will be generated if not provided
+  # client_secret = "my_client_secret"
+
+  authorization_flow  = data.authentik_flow.default-provider-authorization-implicit-consent.id
+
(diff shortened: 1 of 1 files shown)


website/integrations/services/grafana/index.mdx
```diff
@@ -26,6 +26,72 @@ Create an application in authentik. Create an OAuth2/OpenID provider with the fo
...
+  ]
+}
+
+resource "authentik_application" "application_grafana" {
+  name              = "Grafana"
+  slug              = "grafana-slug"
+  protocol_provider = authentik_provider_oauth2.provider_grafana.id
+
+  meta_launch_url = "https://grafana.company/"
+}
```
rissson, line 91:
```suggestion
data "authentik_certificate_key_pair" "default" {
  name = "authentik Self-signed Certificate"
}

data "authentik_flow" "default-provider-authorization-implicit-consent" {
  slug = "default-provider-authorization-implicit-consent"
}

data "authentik_scope_mapping" "scope-email" {
  name = "authentik default OAuth Mapping: OpenID 'email'"
}

data "authentik_scope_mapping" "scope-profile" {
  name = "authentik default OAuth Mapping: OpenID 'profile'"
}

data "authentik_scope_mapping" "scope-openid" {
  name = "authentik default OAuth Mapping: OpenID 'openid'"
}

resource "authentik_provider_oauth2" "grafana" {
  name          = "Grafana"
  # Optional: will be generated if not provided
  # client_id     = "my_client_id"
  # client_secret = "my_client_secret"

  authorization_flow  = data.authentik_flow.default-provider-authorization-implicit-consent.id

  redirect_uris = ["https://grafana.company/login/generic_oauth"]

  property_mappings = [
    data.authentik_scope_mapping.scope-email.id,
    data.authentik_scope_mapping.scope-profile.id,
    data.authentik_scope_mapping.scope-openid.id,
  ]
  
    signing_key = data.authentik_certificate_key_pair.default.id
}

resource "authentik_application" "grafana" {
  name              = "Grafana"
  slug              = "grafana"
  protocol_provider = authentik_provider_oauth2.grafana.id
}
```

Updated the terraform to only have what's needed and to match with the other instructions.

website/integrations/services/grafana/index.mdx
```diff
@@ -78,6 +145,29 @@ token_url = https://authentik.company/application/o/token/
...
+```yaml
+grafana.ini:
+  auth:
+    signout_redirect_url: "https://authentik.company/application/o/<Slug of the application from above>/end-session/"
+    oauth_auto_login: true
+  auth.generic_oauth:
+    name: authentik
+    enabled: true
+    client_id: "<Client ID from above>"
+    client_secret: "<Client Secret from above>"
```
rissson, line 163: For Helm values, I don't like telling users to put their client_secret in the values straight away. We should at least add a comment that says that this can also be passed as an environment variable with:
```yaml
envValueFrom:
  GF_AUTH_GENERIC_OAUTH_CLIENT_ID:
    secretKeyRef:
      name: grafana-oauth-credentials
      key: client_id
  GF_AUTH_GENERIC_OAUTH_CLIENT_SECRET:
    secretKeyRef:
      name: grafana-oauth-credentials
      key: client_secret
```

rissson (changes requested):

website/integrations/services/grafana/index.mdx
senare: Yes, I agree and the reasoning is sound !

however the reason I put it in there in the first place, and hardcoded values to match (instead of generated) was to make it easy to read and follow. I however have no strong feelings about it either way.

Let me see if I can make the changes, I am afraid I am not accustomed to the 'github part' of 'git' yet

senare: This gets a bit hairy ... I think there are 3 ways of doing this ... and if we pass it as env to Grafana it might break 
i will link the issue with discussion if I can find it again !

here it is [17363](https://github.com/helm/charts/issues/17363) could be old and outdated ...

So i am guessing thats the reason for the Grafana Helm docs to spec that you should add as a file ?

[Grafana Helm docs ](https://github.com/grafana/helm-charts/blob/main/charts/grafana/README.md#how-to-securely-reference-secrets-in-grafanaini)

I will look into it and try it before we doc it ... in my case I dodge the issue using 

#   $ openssl rand -hex 32

which would mean there are spec chars to break the env import, do we have any such guarantees if go with generated values ? Would that be a better option ?

senare:
I don't think we need to spec 'authentik Self-signed Certificate' it would be default as in the other cases ...

I will update and retest my configuration and update after !

rissson:
> I don't think we need to spec 'authentik Self-signed Certificate' it would be default as in the other cases ...

You may be right. However, it might be the default on the UI but not in the terraform provider.

website/integrations/services/grafana/index.mdx
rissson: We're using that way for our grafana instance so I'm pretty sure it works. We can also not specify a specific way of doing things, but just have a comment along the lines of "The client secret should not be passed directly via Helm values, but via a secret. Check Grafana Helm chart documentation for more information."

senare:
Sorry, i don't want to leave this hanging! 

- I updated with a disclaimer would that be good enough for now ?
- Also added groups (Grafana Admins, Grafana Editors, Grafana Editors) as assumed. 
- Removed the cert seams to be fine with default (when I test anyways).
- However i added client_id back in again as terraform provider has it as required.
- Rebase & Squash

Goode enough ?

website/integrations/services/grafana/index.mdx
```diff
@@ -26,6 +26,67 @@ Create an application in authentik. Create an OAuth2/OpenID provider with the fo
...
+  name = "authentik default OAuth Mapping: OpenID 'profile'"
+}
+
+data "authentik_scope_mapping" "scope-openid" {
+  name = "authentik default OAuth Mapping: OpenID 'openid'"
+}
+
+resource "authentik_provider_oauth2" "grafana" {
+  name          = "Grafana"
+  #  Required by should be fine to use
```
rissson, line 52:
```suggestion
  #  Required. You can use the output of:
```

website/integrations/services/grafana/index.mdx
```diff
@@ -80,6 +142,33 @@ api_url = https://authentik.company/application/o/userinfo/
...
+    auth_url: "https://authentik.company/application/o/authorize/"
+    token_url: "https://authentik.company/application/o/token/"
+    api_url: "https://authentik.company/application/o/userinfo/"
+    # Optionally map user groups to Grafana roles
+    role_attribute_path: contains(groups[*], 'Grafana Admins') && 'Admin' || contains(groups[*], 'Grafana Editors') && 'Editor' || 'Viewer'
+
+```
+
+Note: for security reasons you shouldn't inline the client_secret instead you should adopt solution to your security requirements. Maybe you are using Sealed Secrets, SOPS or External Secrets ?
+for more information reference 'https://github.com/grafana/helm-charts/blob/main/charts/grafana/README.md#how-to-securely-reference-secrets-in-grafanaini'
```
rissson, line 169:
```suggestion
:::note
For security reasons you shouldn't inline the client_secret in the values, but use a secret instead. For more information, see https://github.com/grafana/helm-charts/blob/main/charts/grafana/README.md#how-to-securely-reference-secrets-in-grafanaini
:::
```

rissson (changes requested):
Some rewording but otherwise LGTM

rissson (approved):

website/integrations/services/grafana/index.mdx
```diff
@@ -80,6 +142,34 @@ api_url = https://authentik.company/application/o/userinfo/
 role_attribute_path = contains(groups[*], 'Grafana Admins') && 'Admin' || contains(groups[*], 'Grafana Editors') && 'Editor' || 'Viewer'
 ```
 
+  </TabItem>
+  <TabItem value="helm">
+If you are using helm values-file instead, you have to set these options:
```
tanberry, line 147:
```suggestion
If you are using a Helm `values.yaml` file instead, you have to set these options:
```

tanberry (approved):
Thanks for the contribution, @senare !

senare:
I did run

$ make website-lint-fix

but it wont say what's not linted nor did it correct it, any advice ?

PR #7145, opened 2023-10-11
web: change 'Attributes' to 'Custom attributes' on Invitation Field

kensternberg-authentik:
## Details

Everywhere else, the label "Attributes" refers to the dictionary field that our core models have and use as a JSON repository for key/value pairs that don't need the full model's awareness or management.  Using the same label here, where the term refers to the context data of the invitation itself, may confuse people, so we've changed the label to signal its difference.

## Checklist

If applicable


BeryJu (approved):

tanberry (approved):
Thank you!

PR #7229, opened 2023-10-19
web: isolate clipboard handling

kensternberg-authentik:
## Details

We would like to use the clipboard for more than just the token copy button.  This commit enables that by separating the "Write to Clipboard" and "Write to Notifications" routines into separate functions, putting "writeToClipboard" into the utilities collection, and clarifying what happens when a customer presses the TokenCopy button.

## Checklist

If applicable


BeryJu (approved):
Looks good, I'd love to have some tests for this to make sure it works in safari and other browsers continuously

PR #7233, opened 2023-10-19
web/admin: improve user email button labels

kensternberg-authentik:
### Label consistency:

`UserViewPage` and `UserLinkPage` have the same functionality to request to view a link with which a user may access an account recovery flow. The language and error messages were different on both of those pages. This commit harmonizes the language by making the request a standalone function. It also exploits the breakout of the “write to clipboard” commit to write the link to the clipboard, and to inform the user that the clipboard has been written to, when possible.

- “Set Password”
- “View Recovery Link”
- “Email Recovery Link”

### Functionality:

Since the `UserListPage`’s “accordion” view has an offer to “Email the recovery link” to the user, it seemed appropriate to grant the same capability to the UserListPage.

### Visual consistency:

After a bit of messing around, I have also ensured that the gap between the buttons is the same in all cases, that in the columnar display the buttons are of uniform width.

## Remove `or undefined` clauses on `this.user`, as they shouldn’t be needed

Every `...render()` method in the `UserViewPage` class has a preamble guard clause:

        if (!this.user) {
            return html``;
        }

With this clause, it should not be necessary to repeatedly check the type of `this.user` throughout the rest of the method, but the nominal type is `User?`, which means that functions called from within the method need to be protected against `undefined` failure. By creating a new variable with the type after the guard clause, we ensure the type is `User` (no question!) and can safely use it without those checks.

Along the way, I replaced the empty html with `nothing` and corrected (mostly by removing) the return types.

References:

- [Lit-HTML: Prefer `nothing` over empty html or other falsey values](https://lit.dev/docs/api/templates/#nothing)
- [TypeScript: Type annotations on return types are rarely necessary](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#:~:text=Return%20Type%20Annotations&text=Much%20like%20variable%20type%20annotations,example%20doesn't%20change%20anything.)

### Caveats

NOTE: This commit is contingent upon the PR for [isolate clipboard handling](https://github.com/goauthentik/authentik/pull/7229) to be accepted, as it relies on the clipboard handler for the “write link to clipboard” feature.

## Checklist

If applicable


web/src/admin/users/UserListPage.ts
```diff
@@ -40,6 +41,56 @@ import {
     UserPath,
 } from "@goauthentik/api";
 
+export const requestRecoveryLink = (user: User) =>
```
kensternberg-authentik, line 44: Breaking these two bits of functionality out so that they can be consistent between the two pages.

web/src/admin/users/UserViewPage.ts
```diff
@@ -287,40 +303,18 @@ export class UserViewPage extends AKElement {
...
+                                    .apiRequest=${() => requestRecoveryLink(user)}
                                 >
                                     <pf-tooltip
                                         position="top"
                                         content=${msg(
                                             "Create a link for this user to reset their password",
                                         )}
                                     >
-                                        ${msg("Reset Password")}
+                                        ${msg("View Recovery Link")}
```
BeryJu, line 314: I'd probably rename it to `Create Recovery Link`

BeryJu (approved):

PR #7302, opened 2023-10-25
web: ak-multi-select component

kensternberg-authentik:
## Details

web: provide a closed control for multi-select

This commit creates a new control, using the ak-form-element-horizontal as a *CLOSED* object, for our multi-select. This control right now is limited to what we expect to be using in the wizard, but that doesn’t mean it can’t be smarter in the future.

## Checklist

If applicable


web/src/components/ak-multi-select.ts
```diff
@@ -0,0 +1,150 @@
...
+                    bubbles: true,
+                }),
+            );
+        }
+    }
+
+    selectRef: Ref<HTMLSelectElement> = createRef();
+
+    render() {
+        return html` <div class="pf-c-form">
```
kensternberg-authentik, line 122: This `div` that brings in all of the PF-Base and PF-Grid definitions turns out to be what has been missing from the formula for building proper, closed controls using Patternfly.

BeryJu (approved):
oh I misunderstood the point of this, I thought this was the full mutli-select with search and dynamic fetch, where this is a wrapper around the current way we do select[multiple]

kensternberg-authentik:
Deleting this as it was already "rolled in" as part of the wizard.  After rebasing this branch on top of main, you get:
```
$ git diff main
$ 
```

Indicating that these fixes are already part of main, and you can see them in the `main` branch.

PR #7314, opened 2023-10-26
website/integrations: Update nextcloud Admin Group Expression

smdion:
Replace user.ak_groups.all() with user.all_groups per 2023.8 release notes in Admin Group

Update Expression in Admin group to only pass groups that start with 'NC-' to NextCloud.  Add verbiage around naming for admin group.

---

## Checklist

If applicable

### DIFF
--- a/website/integrations/services/nextcloud/index.md
+++ b/website/integrations/services/nextcloud/index.md
@@ -98,13 +98,14 @@ To give authentik users admin access to your Nextcloud instance, you need to cre
 Create a custom SAML Property Mapping:
 
 -   Set the _SAML Attribute Name_ to `http://schemas.xmlsoap.org/claims/Group`.
--   Set the _Expression_ to:
+-   Set group names that you want to passthrough to start with 'NC-'
+-   Set the admin group to 'NC-Admin'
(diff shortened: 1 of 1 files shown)


rissson (changes requested):
I agree with changing `user.ak_groups.all()` to `user.all_groups()`, but I don't like the `NC-` prefix for groups that will be synced to Nextcloud. I think we should leave that up to the admin to decide.
Thoughts @tanberry @BeryJu ?

website/integrations/services/nextcloud/index.md
```diff
@@ -98,13 +98,14 @@ To give authentik users admin access to your Nextcloud instance, you need to cre
 Create a custom SAML Property Mapping:
 
 -   Set the _SAML Attribute Name_ to `http://schemas.xmlsoap.org/claims/Group`.
--   Set the _Expression_ to:
+-   Set group names that you want to passthrough to start with 'NC-'
```
tanberry, line 101: @smdion what is the agreement on having this prefix? Sounds like @BeryJu and @rissson advise against it?

tanberry (commented):
Apologies it took me so long to look at this. @smdion shall we continue discussion about the use of a prefix, and can you update the docs/PR accordingly?

PR #7118, opened 2023-10-09
root: Improve Redis config and support

PKizzle:
# Details

Resolves #5077, resolves #1285, resolves #3979,  resolves #5531

### Go
- Redis URL parser adapted from [gitea project](https://github.com/go-gitea/gitea)
- Redisstore adapted from [BoxGo](https://github.com/boxgo/redisstore)

### Python
- Kombu cluster support adapted from https://github.com/celery/kombu/pull/1021
- Celery cluster support adapted from https://github.com/hbasria/celery-redis-cluster-backend

### New Features

- **Uses one URL for configuration of Redis (shared between Go and Python)**
- The Python URL parser has been created in a way to resemble the Go version
- By default uses Redis for broker, channel, backend (current state), but also allows for custom options
- Moves part of the configuration currently under redis category into new dedicated sections -> split into https://github.com/goauthentik/authentik/pull/7097
- Adds deprecation check to config parser to allow for easy transition between old and new configuration scheme
- Old Redis configuration env vars are automatically converted to a Redis config URL if no Redis config URL has been configured otherwise they are ignored
- The new Redis URL allows environment variables to be included in bash style ${VAR}

### Breaking Changes

- None -> Old configurations are still supported however users are advised to switch to the new URL-based configuration

### Additional information

- Newly supported Redis configurations are only available using the new Redis config URL environment variable
- The Redis config URL requires special characters in parameters to be URL encoded (e.g. username, password)
- Has been successfully tested using Redis cluster, sentinel and Unix socket connections

### Todo
- [x] ~~Move from `RedisChannelLayer` to `RedisPubSubChannelLayer`~~ -> not possible https://github.com/redis/redis-py/issues/2219
- [ ] Add documentation for Redis URL parameters
- [ ] Add unittests for Redis middleware
#### Sentinel
- [x] Handle ReadOnly error when master is demoted to slave
- [x] Fix unittests
#### Cluster
- [x] Remove all usage of `KEYS`, `MGET` and `MSET` in Django and Celery
- [x] Fix unittests
- [ ] Check all connections are correctly closed e.g. in redis_middleware_channels.py

PR #7355, opened 2023-10-27
root: Improve multi arch Docker image build speed

PKizzle:
## Details
Use only the host architecture for GeoIP database updates and for Go cross-compilation.

This significantly speeds up image creation as the GeoIP database is architecture independent and does not need to be downloaded twice. Also, cross-compilation is faster than architecture emulation.

Note: The changes in this PR are part of https://github.com/goauthentik/authentik/pull/7118.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -35,7 +35,14 @@ COPY ./gen-ts-api /work/web/node_modules/@goauthentik/api
 RUN npm run build
 
 # Stage 3: Build go proxy
-FROM docker.io/golang:1.21.3-bookworm AS go-builder
+FROM --platform=${BUILDPLATFORM} docker.io/golang:1.21.3-bookworm AS go-builder
+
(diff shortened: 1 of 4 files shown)


BeryJu (approved):
LGTM, I'll merge this and review your other open PRs next week 🎉

rissson (approved):
LGTM as well. Can we apply something similar for outposts?

PKizzle:
Sure the same changes can be applied to all images that run `go build`.

rissson:
Do you want to do it?

PKizzle:
I have updated the PR.

rissson:
I think we're missing the `ldap.Dockerfile` still

PKizzle:
I have changed that Dockerfile as well.

BeryJu (approved):

PR #7363, opened 2023-10-28
outposts: Enhance config options for k8s outposts

PKizzle:
## Details
Now by leaving `kubernetes_ingress_secret_name` empty no TLS section will be added to the ingress.
This is useful for ingress configurations where a default TLS secret is used if none is specified.

~~Additionally, `kubernetes_service_ip_family_policy` and `kubernetes_service_ip_families` add more detailed control over the created service i.e. when running Authentik in a dual-stack k8s cluster setup.~~

## Todo
- [ ] Add missing tests


BeryJu (changes requested):
The bugfixes for TLS and service type make sense, but for the other options `kubernetes_json_patches` should be used, we don't want to add specific config options for every single K8s option; see https://github.com/goauthentik/authentik/issues/5908#issuecomment-1598511954

PKizzle:
I have removed the IP family configuration options in favor of `kubernetes_json_patches`.

PKizzle:
I was not able to find examples in the documentation for the `kubernetes_json_patches` feature just a link to a RFC document. Is this something that will be added in the future? The only example is currently available in the PR that introduced the feature.

BeryJu:
> I was not able to find examples in the documentation for the `kubernetes_json_patches` feature just a link to a RFC document. Is this something that will be added in the future? The only example is currently available in the PR that introduced the feature.

@PKizzle It should probably link to something like https://github.com/kubernetes-sigs/kustomize/blob/master/examples/jsonpatch.md which better demos how the JSON patches work

BeryJu (approved):

PR #7407, opened 2023-11-01
web: refactor status label to separate component

kensternberg-authentik:
## Details

A quality of life thing: `<ak-status-label good>`

There’s an idiom throughout the UI:

``` html
<ak-label color=${item.enabled ? PFColor.Green : PFColor.Red}>
      ${item.enabled ? msg("Yes") : msg("No")}
      </ak-label>
```

There are two problems with this.

- Repeating the conditional multiple times is error-prone
- The color scheme doesn’t communicate much.

There are uses for ak-label that aren’t like this, but I’m focusing on this particular use case, which occurs about 20 times throughout the UI.

Since it’s so common, let’s isolate the most common case: `<ak-status-label good />` gives you the “good” status, and `<ak-status-label/>` gives you the “bad” status, which is the default (no arguments to the function).

There wasn’t much clarity in the system for when to use orange vs red vs grey, but looking through the use cases, it became clear that Red meant fail/inaccessible, Orange meant “Warning, but not blocking,” and Grey just means “info: this thing is off”.

So let’s define that with meaning: there are three types, error, warning, and info. Which corresponds to debugging levels, but whatever, nerds grok that stuff.

So that example at the top becomes

`<ak-status-label ?good=${item.enabled}></ak-status-label>`

… and we can now more clearly understand what that conveys.

There is some heavy tension in this case: this is an easier and quicker-to-write solution to informing the user of a binary status in an iconic way, but the developer has to remember that it exists.

Providing the other types is via a flag:

- `<ak-status-label type="info" ?good=${item.enabled></ak-status-label>` will provide the grey/info icon version.
- `<ak-status-label type="warning" ?good=${item.enabled></ak-status-label>` will provide the orange/warning icon version.

Alternative "good" and "bad" messages can be provided via an attribute:

- `<ak-status-label ?good=${item.enabled} good-label="Hurray!"></ak-status-label>`

Story provided, and changes to the existing uses of the existing idiom provided.

The story just displays the various uses, which does a pretty good job of illustrating in the source code how each alternative is accessed.

image

## Checklist

If applicable


kensternberg-authentik:
By using a custom `.prettierrc.json` file, I was able to turn all of the existing idiomatic `ak-label` blocks into one-liners.  Once that was achieved, the conversion was programmatic:

```
perl -pi.bak -e '/<ak-label.*Red.*Yes/ && s/<ak-label color=\$\{(\S+).*?<\/ak-label>/<ak-status-label ?good=\${$1}><\/ak-status-label>/' $(rg -l '<ak-label.*Red.*Yes')
```

Similar fixes were needed for the `info` and `warning` variants, and for knowing which files needed `import ... ak-label` changed to `import ... ak-status-label`.   And in at least one case (CertificateKeyPairsListPage), _both_ were needed.  (It's also the only page with a custom "good" message.) The `build` and `precommit` phases made those cases easy to identify.

BeryJu (approved):

PR #7410, opened 2023-11-02
core: add user settable token durations

### LINKED ISSUE #5881: Allow end users to create app passwords with longer (or no) expiry
So I have calibre-web authenticating to authentik via LDAP (for it's OPDS function). With MFA enabled, ideally that would use an app password. However, user created app passwords have a (fixed?) lifetime of 30 minutes which is not ideal for this use case.

An option to allow users to create longer lived app passwords, or even potentially non-expiring one.

Some way to alter the default lifetime of those app passwords, but might not be ideal I suppose.

**Additional context**

jmdilly:
## Details

This PR enables users to specify a non default lifetime for users created app application token. By the way, it enforces expires value on API side.

Closes #5881 

Current limitation : maximum duration is currently not enforce in the UI but only on API side. This is because I haven't find yet an easy way to provide that piece of information to the UI without add a new API endpoint which would require additionnal work. Nevertheless, if the chosen expiration date is invalid, an error message is still thrown up to the use explaining why it has failed.
In my opinion, this enforcement within the UI could be enforced in a future improvement.

---

## Checklist

-   [] The API schema has been updated (`make gen-build`)

If applicable


website/docs/user-group-role/user/user_ref.md
```diff
@@ -70,6 +70,14 @@ Optional flag, when set to false, Tokens created by the user will not expire.
 
 Only applies when the token creation is triggered by the user with this attribute set. Additionally, the flag does not apply to superusers.
 
+### `goauthentik.io/user/token-maximum-lifetime`:
+
+Optional flag, when set, defines the maximum lifetime of user created tokens. Defaults to 30 minutes if unset.
```
tanberry, line 75:
```suggestion
Optional flag, when set, defines the maximum lifetime of user-created tokens. Defaults to 30 minutes if not set.
```

tanberry (approved):
Tiny edits, thanks for remembering to add docs!

authentik/core/models.py
```diff
@@ -41,19 +42,27 @@
...
 USER_ATTRIBUTE_SOURCES = "goauthentik.io/user/sources"
 USER_ATTRIBUTE_TOKEN_EXPIRING = "goauthentik.io/user/token-expires"  # nosec
+USER_ATTRIBUTE_TOKEN_MAXIMUM_LIFETIME = "goauthentik.io/user/token-maximum-lifetime"  # nosec
 USER_ATTRIBUTE_CHANGE_USERNAME = "goauthentik.io/user/can-change-username"
 USER_ATTRIBUTE_CHANGE_NAME = "goauthentik.io/user/can-change-name"
 USER_ATTRIBUTE_CHANGE_EMAIL = "goauthentik.io/user/can-change-email"
 USER_PATH_SYSTEM_PREFIX = "goauthentik.io"
 USER_PATH_SERVICE_ACCOUNT = USER_PATH_SYSTEM_PREFIX + "/service-accounts"
 
+DEFAULT_TOKEN_DURATION = "minutes=30"  # nosec
```
rissson, line 52: I wonder if we should make this configurable as we do with the token length.

BeryJu: we already have `default_token_length` so we can also add `default_token_duration` as config (we'll also have to validate the duration on config load)

rissson: I added this, w/o the validation for now. How do you want to handle this?

rissson: Once this is in, I'll also open a follow-up PR to migrate those options to be per-tenant.

authentik/core/api/tokens.py
```diff
@@ -48,6 +57,30 @@ def validate(self, attrs: dict[Any, str]) -> dict[Any, str]:
...
+                    max_token_lifetime_dt = timedelta_from_string(max_token_lifetime)
+                except ValueError:
+                    max_token_lifetime_dt = default_token_duration()
+
+            if "expires" in attrs and attrs.get("expires") > token_expires_from_timedelta(
+                max_token_lifetime_dt
+            ):
+                raise ValidationError(
+                    {"expires": f"Token expires exceeds maximum lifetime ({max_token_lifetime})."}
+                )
```
rissson, line 79: Should we instead set the token expiration to something like `min(attrs.get("expires"), token_expires_from_timedelta(max_token_lifetime_dt))`?

authentik/core/models.py
rissson: Actually I added this directly in here.

authentik/core/models.py
```diff
@@ -624,12 +643,17 @@ class Meta:
...
     """Base Model which can expire, and is automatically cleaned up."""
 
-    expires = models.DateTimeField(default=default_token_duration)
+    expires = models.DateTimeField(default=None, null=True)
     expiring = models.BooleanField(default=True)
 
     class Meta:
         abstract = True
 
+    def save(self, *args, **kwargs):
```
BeryJu, line 652: should probably use `pre_save` signal instead of this

PR #7432, opened 2023-11-03
providers/oauth2: Fix URL clash

### LINKED ISSUE #7419: URL clash with certain application names and OAuth2/OpenID providers
**Describe the bug**
The discovery endpoint returns the following URLs:
```
"issuer": "https://example.com/application/o/<APP-SLUG>/",
"authorization_endpoint": "https://example.com/application/o/authorize/",
"token_endpoint": "https://example.com/application/o/token/",
"userinfo_endpoint": "https://example.com/application/o/userinfo/",
"end_session_endpoint": "https://example.com/application/o/<APP-SLUG>/end-session/",
"introspection_endpoint": "https://example.com/application/o/introspect/",
"revocation_endpoint": "https://example.com/application/o/revoke/",
"device_authorization_endpoint": "https://example.com/application/o/device/", 
"jwks_uri": "https://example.com/application/o/<APP-SLUG>/jwks/",
```

However, if an app happens to have a slug of "authorize" or "token" (etc) it will result in invalid discovery endpoints being returned. For example, if an app has a slug of "authorize", the discovery endpoint returns:

```
"issuer": "https://example.com/application/o/authorize/", # Same URL
"authorization_endpoint": "https://example.com/application/o/authorize/", # Same URL
"token_endpoint": "https://example.com/application/o/token/",
"userinfo_endpoint": "https://example.com/application/o/userinfo/",
"end_session_endpoint": "https://example.com/application/o/authorize/end-session/",
"introspection_endpoint": "https://example.com/application/o/introspect/",
"revocation_endpoint": "https://example.com/application/o/revoke/",
"device_authorization_endpoint": "https://example.com/application/o/device/",
"jwks_uri": "https://example.com/application/o/authorize/jwks/",
```

This is not warned about in the docs (to my knowledge), nor it is validated in authentik in any way

**To Reproduce**

1. Create a OAuth2/OpenID provider
2. Create an app and set a slug of "authorize", "token", etc
3. Query discovery endpoint `https://example.com/application/o/<app-slug>/.well-known/openid-configuration`
4. See URLs

**Expected behavior**
Put the endpoint "verb" as prefix and move the app slug last for all endpoints:
```
https://example.com/application/o/authorize/
https://example.com/application/o/end-session/<app-slug>/
https://example.com/application/o/issuer/<app-slug>/
https://example.com/application/o/token/
https://example.com/application/o/userinfo/
https://example.com/application/o/introspect/
https://example.com/application/o/revoke/
https://example.com/application/o/device/
https://example.com/application/o/jwks/<app-slug>/
```
Also discovery endpoint becomes:
```
https://example.com/application/o/discovery/<app-slug>/.well-known/openid-configuration
```

**Screenshots**

**Logs**

-   authentik version: all
-   Deployment: all

**Additional context**

astro-stan:
## Details

I decided to try and fix #7419.

I think that should do it.

---

## Checklist

If applicable

### DIFF
--- a/authentik/providers/oauth2/api/providers.py
+++ b/authentik/providers/oauth2/api/providers.py
@@ -95,16 +95,19 @@ class OAuth2ProviderViewSet(UsedByMixin, ModelViewSet):
             "authorize": request.build_absolute_uri(
                 reverse(
                     "authentik_providers_oauth2:authorize",
+                    kwargs={"application_slug": provider.application.slug},
                 )
             ),
             "token": request.build_absolute_uri(
                 reverse(
                     "authentik_providers_oauth2:token",
+                    kwargs={"application_slug": provider.application.slug},
                 )
             ),
             "user_info": request.build_absolute_uri(
                 reverse(
                     "authentik_providers_oauth2:userinfo",
+                    kwargs={"application_slug": provider.application.slug},
                 )
             ),
             "provider_info": None,
--- a/authentik/providers/oauth2/tests/test_api.py
+++ b/authentik/providers/oauth2/tests/test_api.py
@@ -34,7 +34,7 @@ class TestAPI(APITestCase):
         )
         self.assertEqual(response.status_code, 200)
         body = loads(response.content.decode())["preview"]
-        self.assertEqual(body["iss"], "http://testserver/application/o/test/")
+        self.assertEqual(body["iss"], "http://testserver/application/o/issuer/test/")
 
     def test_setup_urls(self):
         """Test Setup URLs API Endpoint"""
@@ -43,7 +43,7 @@ class TestAPI(APITestCase):
         )
         self.assertEqual(response.status_code, 200)
         body = loads(response.content.decode())
-        self.assertEqual(body["issuer"], "http://testserver/application/o/test/")
+        self.assertEqual(body["issuer"], "http://testserver/application/o/issuer/test/")
 
     # https://github.com/goauthentik/authentik/pull/5918
     @skipUnless(version_info >= (3, 11, 4), "This behaviour is only Python 3.11.4 and up")
--- a/authentik/providers/oauth2/tests/test_api.py
+++ b/authentik/providers/oauth2/tests/test_api.py
@@ -34,7 +34,7 @@ class TestAPI(APITestCase):
         )
         self.assertEqual(response.status_code, 200)
         body = loads(response.content.decode())["preview"]
-        self.assertEqual(body["iss"], "http://testserver/application/o/issuer/test/")
+        self.assertEqual(body["iss"], "http://testserver/application/o/test/")
 
     def test_setup_urls(self):
         """Test Setup URLs API Endpoint"""
@@ -43,7 +43,7 @@ class TestAPI(APITestCase):
         )
         self.assertEqual(response.status_code, 200)
         body = loads(response.content.decode())
-        self.assertEqual(body["issuer"], "http://testserver/application/o/issuer/test/")
+        self.assertEqual(body["issuer"], "http://testserver/application/o/test/")
 
     # https://github.com/goauthentik/authentik/pull/5918
     @skipUnless(version_info >= (3, 11, 4), "This behaviour is only Python 3.11.4 and up")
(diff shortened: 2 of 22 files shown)


rissson:
Heya! Thank you for your contribution. This is however, a massive breaking change that we probably do not want to go through with. I'd rather we block applications from being created if their slug is a reserved name (like `authorize`, `token` or `userinfo`)

sdimovv:
Limiting the app naming is not a good idea. For one, this only applies to the oauth2 provider, thus limiting the app name for all providers is not fair. Neither is limiting the app name ror ONLY the oauth2 provider. Moreover, users may already have apps called with one of these names. If we "limit" the names and make assumptions for this throughout the codebase we would need a way of forcing the users to change their existing apps with reserved names.

A breaking change is a breaking change but I wouldn't necessary say this one is more "massive" than the others. authentik has been making breaking changes every few releases.

A more sensible approach (and one that I have followed before with this project) is to keep the old URLs alongside the new ones and put a deprecation warning in the docs. Then, they the old URLs can be removed completely starting with the first 2024 release. If i recall correctly that one already has other breaking changes planned.

BeryJu:
I don't really mind changing end-session and jwks URLs, but we need to be careful with changing the Issuer and the well-known URL as some applications (notably the one I remember is [Harbor](https://goharbor.io/)) use the issuer as part of the user's unique identifier, and thus changing it would duplicate users

I think blocking the use of certain slugs is not ideal but a better solution for now, i.e. blocking some slugs + adding the new URLs and using them as default where possible (except Issuer). The other option would be to use the new URLs for newly created Oauth2 providers and leave currently providers as-is, however that might also break some documentation that explicitly states URLs and we'd also need to add a new database field to keep track of which URLs to use

sdimovv:
> use the issuer as part of the user's unique identifier
🤨 Interesting. Any idea how hard is it to migrate that, given a couple months of warning?

Otherwise, I am not a fan of either of the other options, but I feel like the new set of URLs + db field approach might be more transparent for the average user. Though I agree the docs will need some extra thought. Also if we go down that path it's not clear when/if the old URLs will go away.

In any case I feel we shouldn't limit the slugs as that just introduces special cases throughout the code, not to mention it will be annoying for the users to force them to change their app names.

tanberry:
@sdimovv et al, is there some info that we can add to the docs now (before deciding which technical implementation we do) to help prevent users being confused about invalid discovery endpoints being returned? Perhaps a note explaining the invalid endpoints, or better yet a Note telling how to avoid it?

BeryJu:
The other option is to keep the current schema and simply move the other urls, i.e. authorize to also be `/application/o/slug/authorize/`, which we don't technically need since we can lookup the app and provider via client_id (which is also the reason why some URLs have the app slug in the path and some don't). This would still be an annoying change (we could support both as well), but its still a lot less breaking

sdimovv:
Are we talking about making the URLs like so:

```
https://example.com/application/o/<APP-SLUG>/
https://example.com/application/o/<APP-SLUG>/authorize/
https://example.com/application/o/<APP-SLUG>/token/aadasd
https://example.com/application/o/<APP-SLUG>/userinfo/
https://example.com/application/o/<APP-SLUG>/end-session/
https://example.com/application/o/<APP-SLUG>/introspect/
https://example.com/application/o/<APP-SLUG>/revoke/
https://example.com/application/o/<APP-SLUG>/device/
https://example.com/application/o/<APP-SLUG>/jwks/
```

If so I suppose that would work too, with the added benefit that the issuer URL will remain unchanged. It will still be a breaking change but that is unavoidable. I propose we add the new URLs and deprecate the old ones. Then we can remove them completely starting with 2024.X.X.

sdimovv:
Sorry for the delay here. I was busy with work then the holidays came... 

I've updated the URLs in the way I wrote above. I have also updated the `reverse` calls in the tests and various places I found throughout the codebase.  

Looks like the website lint is failing but the fix for that should be coming from main. I don't want to increase the diff for issues unrelated to this PR.

tanberry (approved):
From a docs pov, looks good, and thanks!

sdimovv:
I just realised this hasn't been merged. Does it need anything else?

PR #7433, opened 2023-11-04
web: refactor the table renderer for legibility 

kensternberg-authentik:
## Details

Refactor the Table component for legiibility.

This commit does not change the functionality of the Table, nor does it require any changes to existing uses of the Table.

It will probably be easier to review this by looking at the `View Code` in the upper-right-hand corner of GitHub’s reviewer; that or side-by-side, if your monitor is wide-enough.

### Background

The existing Table component is used 49 times (at last count) in authentik, and those uses are wide-ranging and complex, but they all come down to a couple of entries:

- Displaying a row of summary information
- Permitting the display of more complex (“expanded”) information
- Displaying a collection of rows
- Displaying a collection of rows grouped by some header
- Pagination of many rows
- Permitting an action on the visible rows
- *Not* blocking events that may happen on a cell or expansion
- Providing a toolbar
- Providing a display of “selected items” when using the table as a multi-select with many pages of items (chips display)
- Providing sort functionality on columns
- Providing the ability to filter the table from the back-end

This commit changes none of that.

### Actions taken

What this commit does is re-arrange the innards of Table.ts into smaller units:

- The RowGroup’s “checkbox” and “expansion” segments are pulled out into their own functions, which makes the RowGroup’s actual functionality much easier to see and understand. The same is true of the rowGroup’s selection and expansion handlers.
- Almost all in-line decisions and event handlers have been extracted and named, to make it easier to see and understand what’s happening inside what is otherwise a jumble of HTML.
- The TablePagination code was duplicated– and one of the duplicates was wrong! So I’ve deduplicated it and fixed the bug.
- In many cases, the conditional code grew organically, resulting in some pretty hard-to-understand conditions.
  - A really good example is the `itemSelectHandler`; there are two possible events that result in a change, and the consequences of that change may be that *all* checkboxes are unchecked. In all cases where there’s an add/remove option, I’ve opted to remove the specific object always (even if it’s not present!), and then add it if it’s actually an add. Logically coherent as long as the accessors are not also mutators.

### Conflicts discovered

It was not possible to redefine the `columns()` function to take anything other than a TableColumn object; I wanted to be able to replace all of the `new TableColumn("Foo")` with just `"Foo"`, building the TableColumn dynamically at construction time. Unfortunately, some of our most complex tables dynamically re-arrange the columns (RBAC, for example, draws an empty table, fetches the content, then redraws with the columns based on what was retrieved), and detecting that change and rebuilding those columns proved more difficult than anticipated. I may contemplate an alternative column specification if I find myself building a lot of tables.

Likewise, it was not possible to replace all of our uses of the empty `html` declaration with the Lit-preferred `nothing` sigil; hard-coded `TemplateResult` entries scattered throughout the code caused massive type inconsistencies, since a type of `TemplateResult | nothing` is unique thanks to `nothing`’s underlying Symbol. It is for this issue that Typescript itself recommends you “prefer allowing Typescript infer the return type.” I may revisit this issue later.

### Misc

I’ve added a `prequick` command to `package.json`; this one runs *only* the Typescript type checker, lit-analyse, and `eslint:precommit`, the last of which lints only the files touched since the last commit. This is fast, intended to support quick checks of code quality not normally displayed in the IDE.

## Checklist

If applicable


web/src/elements/table/Table.ts
```diff
@@ -107,8 +100,7 @@ export abstract class Table<T> extends AKElement {
         return false;
     }
 
-    // eslint-disable-next-line @typescript-eslint/no-unused-vars
-    renderExpanded(item: T): TemplateResult {
+    renderExpanded(_item: T): TemplateResult {
```
BeryJu, line 103: I think I left the variable name as `item` here as that's the default when inheriting from the class

web/src/elements/table/Table.ts
```diff
@@ -249,10 +241,10 @@ export abstract class Table<T> extends AKElement {
                 <td role="cell" colspan="8">
                     <div class="pf-l-bullseye">
                         ${inner
-                            ? inner
```
BeryJu, line 252: this doesn't use `inner` at all anymore?

BeryJu (commented):
couple of questions

web/src/elements/table/Table.ts
kensternberg-authentik: Good point; I've put it back the way it was.

I was trying to use the "unused variable in abstract code" idiom here, and it shouldn't matter what the name is in the child code, only the position.  But leaving the name as-is and commenting out the check is probably better for readability in the long-run.

BeryJu (approved):

PR #7453, opened 2023-11-06
website/integrations: Add custom Group/Role mapping documentation for Grafana

gmemstr:
## Details

Adds some more documentation around assigning roles in Grafana from Authentik groups. Grafana will use the profile scope's groups entry to map to its own Groups concept (from what I can tell), so we need to nest the groups somewhere Grafana won't automatically pick up so they can be assigned.

---

## Checklist

If applicable

### DIFF
--- a/website/integrations/services/grafana/index.mdx
+++ b/website/integrations/services/grafana/index.mdx
@@ -32,7 +32,7 @@ return {
 }
 ```
 
-This ensures that groups are available under `info.groups[]`, which can be used later in [Role Mapping](#Role-Mappings).
+This ensures that groups are available under `info.groups[]`, which can be used later in [Role Mapping](#role-mappings).
 
 Note the Client ID and Client Secret values. Create an application, using the provider you've created above. Note the slug of the application you've created.
 
--- a/website/integrations/services/grafana/index.mdx
+++ b/website/integrations/services/grafana/index.mdx
@@ -24,7 +24,7 @@ Create an application in authentik. Create an OAuth2/OpenID provider with the fo
 -   Signing Key: Select any available key
 -   Redirect URIs: `https://grafana.company/login/generic_oauth`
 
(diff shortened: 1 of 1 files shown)


website/integrations/services/grafana/index.mdx
```diff
@@ -24,6 +24,16 @@ Create an application in authentik. Create an OAuth2/OpenID provider with the fo
 -   Signing Key: Select any available key
 -   Redirect URIs: `https://grafana.company/login/generic_oauth`
 
+Additionally, because Grafana has its own concept of groups, we need to create a custom Scope Mapping to ensure Grafana can read the user's groups assigned within Authentik. It should contain the following expression:
```
tanberry, line 27:
```suggestion
Additionally, because Grafana has its own concept of groups, we need to create a custom Scope Mapping to ensure Grafana can read the user's groups assigned within authentik. It should contain the following expression:
```

tanberry (changes requested):

BeryJu (commented):
Hi! The default `profile` mapping already returns the `groups` which is what the current configuration use, so this shouldn't be required. Did the default configuration not work for you?

gmemstr:
@BeryJu As noted in the PR, it does not - the `groups` entry is taken by Grafana (my assumption is that it's trying to be used for Grafana's own groups, which don't map to authentik groups) and is blank by the time it's being used for role assignment, hence the nesting it in `info`.

kissi7:
> @BeryJu As noted in the PR, it does not - the `groups` entry is taken by Grafana (my assumption is that it's trying to be used for Grafana's own groups, which don't map to authentik groups) and is blank by the time it's being used for role assignment, hence the nesting it in `info`.

Hi, I can confirm that groups are propagated to grafana, but are not used to map user to role for some reason.
With proposed PR it is working just fine. Thank you very much.

PR #7482, opened 2023-11-07
web: refactor sidebar capabilities for categorical subsections

kensternberg-authentik:
## Details

web: refactor sidebar capabilities for categorical subsections

The project “Change Admin UI lists to have sublists per type” requires some initial changes to the UI to facilitate this request. The AdminSidebar is the principle target of this project, and it is embedded in the AdminInterface. To facilitate editing the AdminSidebar as an independent entity, AdminInterface has been moved into its own folder and the AdminSidebar extracted as a standalone Web Component. This removes, oh, about half the code from AdminInterface. A little cleanup with `classMap` was also committed.

The rollup config was adjusted to find the new AdminInterface location.

The Sidebar uses the global `config: Config` object to check for Enterprise capabilities. Rather than plumb all the way down through the Interface =\> AdminInterface -\> AdminSidebar, I chose to make provide an alternative way of reaching the `config` object, as a *context*. Other configuration objects (Me, UiConfig, Tenant) interfaces will be contextualized as demand warrants.

Demand will warrant. Just not yet. <sup>1</sup>

The Sidebar has been refactored only slightly; the renderers are entirely the same as they were prior to extraction. What has been changed is the source of information: when we retrieve the current version we story *only* the information, and use type information to ensure that the version we store is the version we care about. The same is true of `impersonation`; we care only about the name of the person being impersonated being present, so we don’t store anything else.

Fetches have been moved from `firstUpdated` to the constructor. No reason to have the sidebar render twice if the network returns before the render is scheduled.

Because the path used to identify the user being impersonated has changed, the `str()` references in the XLIFF files had to be adjusted. **This change is to a variable only and does not require translation.**

---

<sup>1</sup> The code is littered with checks to `me()?`, `uiConfig?`, `config?`, etc. In the *context* of being logged in as an administrator those should never be in doubt. I intend to make our interfaces not have any doubt.

---

## Checklist

If applicable


web/rollup.config.mjs
```diff
@@ -148,7 +148,7 @@ export default [
     },
     // Admin interface
     {
-        input: "./src/admin/AdminInterface.ts",
```
kensternberg-authentik, line 151: Adjust the location for the Interface root, since we're moving it.

web/src/admin/AdminInterface/AdminInterface.ts
```diff
@@ -0,0 +1,174 @@
...
+            this.user.user.isSuperuser ||
+            // TODO: somehow add `access_admin_interface` to the API schema
+            this.user.user.systemPermissions.includes("access_admin_interface");
+        if (!canAccessAdmin && this.user.user.pk > 0) {
+            window.location.assign("/if/user/");
+        }
+    }
+
+    render(): TemplateResult {
+        const sidebarClasses = {
```
kensternberg-authentik, line 120: Aside from the Sidebar-ectomy, the only other changes are the removal of imports used _only_ by the Sidebar code, and this update to the styles to make the HTML a little cleaner.

web/src/admin/AdminInterface/AdminSidebar.ts
```diff
@@ -0,0 +1,176 @@
...
+    UiThemeEnum,
+    type UserSelf,
+    Version,
+} from "@goauthentik/api";
+
+@customElement("ak-admin-sidebar")
+export class AkAdminSidebar extends AKElement {
+    @property({ type: Boolean })
+    open = true;
+
```
kensternberg-authentik, line 30: These are the only bits we care about in either of the calls to `AdminApi()` and `me()`.

web/src/elements/sidebar/SidebarItem.ts
```diff
@@ -144,47 +144,84 @@ export class SidebarItem extends AKElement {
         return this.renderInner();
     }
 
-    renderInner(): TemplateResult {
-        if (this.childItems.length > 0) {
-            return html`<li
-                class="pf-c-nav__item ${this.expanded ? "pf-m-expandable pf-m-expanded" : ""}"
+    renderWithChildren() {
```
kensternberg-authentik, line 147: The four objects here represent the four main items drawn by the sidebar.  We don't yet have the `withPathAndChildren()` yet; that's for when we get to the subtypes.

web/src/admin/AdminInterface/AdminInterface.ts
BeryJu: Could we maybe move that logic into the actual sidebar (mainly the collapsing/expanding)

BeryJu (approved):

web/src/admin/AdminInterface/AdminInterface.ts
kensternberg-authentik: Yes!  And I've done just that.

BeryJu (approved):

Issue #7435, opened 2023-11-05
Proxy providers suddenly stop working and hit a redirect loop


brujoand:
**Describe the bug**
One or two of the applications using proxy provider will stop working. They will just go for an infinite redirect loop until I recreate them from scratch in Authentik. Sometimes they can start working again seemingly at random. Looking at the logs all I see is the redirect being logged. All these apps are registerted as app.mydomain.com without any path fiddling. Ref: #6540  

The applications are mostly in the k8s cluster, and I use local dns to reach them. For instance http://frigate.automation.svc.cluster.local:5000/, with https://frigate.mydomain.com/ as the external url. The only problem I'm seeing in the worker/server logs is "context canceled" "failed to proxy to backend". Worker and Server logs from a redirect loop is over here: https://gist.github.com/brujoand/37b372545f4a5ad86828077f1f230de3

**To Reproduce**
1. Create a proxy provider
2. Create an application using the proxy provider
3. Add the application to an outputs
4. Repeat and wait for some of the applications to go into a infinite redirect loop

**Expected behavior**
There should be a simple auth flow and the application should appear, instead of a redirect loop

**Screenshots**

**Logs**
Logs are here: https://gist.github.com/brujoand/37b372545f4a5ad86828077f1f230de3

**Version and Deployment (please complete the following information):**

-   authentik version: 2023.10.2
-   Deployment: Helm

**Additional context**
Nope

hdlineage:
Yes this started happening to with me in 2023.10.2 as well. going to downgrade for 2023.8 for now.

cfouche3005:
same behaviour with forward auth

brujoand:
Can note that this behavior is also present on 2023.10.1 and 2023.10.0. One thing that puzzles me though is that logging in with the same user on two different machines yields different results. Some sites I can login to on one machine, while on another machine that site death loops and vice versa for other sites.

brujoand:
I've also tested with the images from #7385 which was referenced in #7374 and #7342 but that didn't make much difference unfortunately.

phoen88:
Adding to this, still having issues on my install as well. Was driving me crazy as a recently joined user as well. Earlier versions work but I run into other issues, namely with permissions binding.

brujoand:
After upgrading to 2023.10.3 and deleting all browser state I’m unable to reproduce this problem. Thanks!

brujoand:
Unfortunately this problem reappeared with the same symptoms as before.

[...]

Issue #7374, opened 2023-10-29
Authentication loop in auth proxied service


gmag11:
**Describe the bug**
I have several services using Caddy as proxy.
Configuring an Auth Proxy in a fresh Authentik instance creates an authentication loop just after entering Authentik credentials. 
After several minutes I can access the service normally.

I've tested Paperless-NGX and Invidious, that previously were working.

**To Reproduce**
### Caddyfile section for Paperless-NGX
```
        @paperless host paper.xxxxxx.com
        handle @paperless {
                reverse_proxy /outpost.goauthentik.io/* http://192.168.1.120:9002
                forward_auth http://192.168.1.120:9002 {
                        uri /outpost.goauthentik.io/auth/caddy

                        # capitalization of the headers is important, otherwise they will be empty
                        copy_headers X-Authentik-Username X-Authentik-Groups X-Authentik-Email X-Authentik-Name X-Authentik-Uid X-Authentik-Jwt X-Authentik-Meta-Jwks X-Authentik-Meta-Outpost X-Authentik-Meta-Provider X-Authentik-Meta-App X-Authentik-Meta-Version

                        # optional, in this config trust all private ranges, should probably be set to the outposts IP
                        trusted_proxies private_ranges
                }
                reverse_proxy http://paperless-ngx:8000
                encode gzip
                header {
                        Strict-Transport-Security max-age=15552000
                        Referrer-Policy "strict-origin-when-cross-origin"
                        X-XSS-Protection "1; mode=block"
                        X-Content-Type-Options "nosniff"
                        X-Frame-Options "SAMEORIGIN"
                        Content-Security-Policy "frame-ancestors 'self' *.xxxxxx.com"
                        -Server
                }
        }
```
### Caddifile section for Authentik
```
        @authentik host auth2.xxxxxx.com
        handle @authentik {
                reverse_proxy http://192.168.1.120:9002
                encode gzip
                header {
                        Strict-Transport-Security max-age=15552000
                        Referrer-Policy "strict-origin-when-cross-origin"
                        X-XSS-Protection "1; mode=block"
                        X-Content-Type-Options "nosniff"
                        X-Frame-Options "SAMEORIGIN"
                        Content-Security-Policy "frame-ancestors 'self' *.xxxxxx.com"
                        -Server
                }
        }
```
### docker-compose.yaml
```
version: "3.4"

services:
  postgresql:
    image: docker.io/library/postgres:12-alpine
    restart: unless-stopped
    healthcheck:
[...]

PR #7541, opened 2023-11-13
providers/proxy: Fix duplicate cookies when using file system store.

thijsa:
## Details

#7435 and #7374 when using file system store (default when using outpost docker image).

---

## Checklist

If applicable


BeryJu (approved):
Huh I thought the redirect-loop only happens with the embedded outpost

thijsa:
@BeryJu I don't use the embedded outpost but when I setup my local env I didn't get the issue using the embedded outpost. We encountered the loop on the external outpost, it sets the cookie twice, on / and on /outpost.goauthentik.io.

PR #7563, opened 2023-11-14
ci: translation-advice: avoid commenting after make i18n-extract [AUTH-145]

rissson:
## Details

REPLACE ME

---

## Checklist

If applicable

### DIFF
--- a/.github/workflows/translation-advice.yml
+++ b/.github/workflows/translation-advice.yml
@@ -7,7 +7,8 @@ on:
     paths:
       - "!**"
       - "locale/**"
-      - "web/src/locales/**"
+      - "!locale/en/**"
+      - "web/xliff/**"
 
 jobs:
   post-comment:


.github/workflows/translation-advice.yml
```diff
@@ -7,7 +7,9 @@ on:
     paths:
       - "!**"
       - "locale/**"
+      - "!locale/en/**"
       - "web/src/locales/**"
+      - "!web/src/locales/en.ts"
```
rissson, line 12: Do we need to add pseudo-LOCALE.ts here as well?

BeryJu: All of the web/src locales are generated at build time: https://github.com/goauthentik/authentik/tree/main/web/src/locales

rissson: Ah, so we can entirely remove the web stuff in here.

For my own comprehension, does the build download them from transifex then?

rissson:
Ok this only works for backend files, as all frontend files are updated when source strings are modified

.github/workflows/translation-advice.yml
BeryJu: Nothing is downloaded, the translation source files are in GitHub and get converted to typescript during build, anytime anything changes in transifex it gets PR'd to the repo

BeryJu (approved):

PR #7581, opened 2023-11-15
stages/email: use uuid for email confirmation token instead of username

BeryJu:
## Details

This fixes an issue where a user could create a token with the username of any user and prevent that user from using this stage, the issue is non-exploitable as creating a token manually with this identifier doesn't have a flow plan and hence just leads to an error

---

## Checklist

If applicable

### DIFF
--- a/authentik/stages/email/stage.py
+++ b/authentik/stages/email/stage.py
@@ -1,5 +1,6 @@
 """authentik multi-stage authentication engine"""
 from datetime import timedelta
+from uuid import uuid4
 
 from django.contrib import messages
 from django.http import HttpRequest, HttpResponse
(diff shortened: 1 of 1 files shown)


authentik/stages/email/stage.py
```diff
@@ -71,7 +72,7 @@ def get_token(self) -> FlowToken:
         valid_delta = timedelta(
             minutes=current_stage.token_expiry + 1
         )  # + 1 because django timesince always rounds down
-        identifier = slugify(f"ak-email-stage-{current_stage.name}-{pending_user}")
+        identifier = slugify(f"ak-email-stage-{current_stage.name}-{str(uuid4())}")
```
smusali, line 75: @BeryJu what is the impact of this change on other places?

BeryJu: nothing, existing tokens will continue to work, and this only applies to newly created tokens

smusali (approved):
lgtm

PR #7590, opened 2023-11-15
root: Multi-tenancy

rissson:
## Details

Opening for visibility for now, and to have a TODO I can put down somewhere. Will edit the desc when relevant

TODO:
- [x] figure out how the embedded outpost would work -> disable it
- [x] add --tenant to relevant mgmt commands
  - it's actually --schema-name, still need to modify the thing to make `public` the default
- [x] make sure the tenant API works
- [x] update the frontend with the settings page
  - [x] the save button does send the API request which works, but the form is not reset properly (it goes back to the old values instead of keeping the new ones). Also, I'm probably not using the `ak-spinner-button` correctly
- [x] add endpoints to the tenant API for management stuff, such as getting a recovery token
- [x] find out what to do about on-disk files (like icons)
  - moved to a subfolder per-tenant, migrations are written for this already
  - [x] needs some testing
- [x] write tests
  - [x] somewhat done, missing the settings API and recovery endpoints tests
- [x] rename tenants to brands in documentation
- [x] should we allow tenant management via blueprints?
  - rough guess is no, but then how to manage the tenant settings as-code (w/o terraform)
- [x] add tenants section in docs under an "Advanced topics" dropdown
- [x] document user-facing changes, so far: 
  - tenants have been renamed to brands
  - config options that have been moved from the config file to the UI: avatars, default_user_change_name, default_user_change_email, default_user_change_username, gdpr_compliance, impersonation, footer_links, reputation_expiry
  - icons are now in a `public/` subfolder
  - the shorthand parameter for `--stage` `-s` for the `ak test_email` command has been 
changed to `-S`
  - the event retention settings has been removed from brands (previously-tenants) and is now a system setting (per-tenant)
- [x] prom metrics
- [x] expression policies
  - see #8002 

---

## Checklist

If applicable


rissson:
Ok that's a weird one. Tests pass when they're run individually, but when running all of them, they fail after some of them.
This happens because the default tenant (created by the tenants migration) gets deleted at some point, for some reason, even though we explicitly deny that tenant from being deleted.

rissson:
No idea why the other tests are failing. The unit tests work on my machine when run individually. I haven't tried the e2e tests (oauth and flow) because I don't have the setup for selenium.

rissson:
After a lot of stirring in the django test client internals, I cannot get it to work with other domains, so the full tests for the settings API are commented for now, and I'm not writing tests for tenant separation yet.

authentik/api/v3/config.py
```diff
@@ -62,15 +62,15 @@ class ConfigView(APIView):
 
     permission_classes = [AllowAny]
 
-    def get_capabilities(self) -> list[Capabilities]:
+    def get_capabilities(self, request: Request) -> list[Capabilities]:
```
BeryJu, line 65: you can use `self.request` instead of passing the request as argument, fyi

authentik/blueprints/apps.py
```diff
@@ -42,6 +44,33 @@ def reconcile(self) -> None:
...
+    def reconcile_tenant(self) -> None:
+        """reconcile ourselves for tenanted methods"""
+        from authentik.tenants.models import Tenant
+
+        try:
+            tenants = list(Tenant.objects.filter(ready=True))
+        except (DatabaseError, ProgrammingError, InternalError) as exc:
+            self._logger.debug("Failed to get tenants to run reconcile", exc=exc)
+            return
+        for tenant in tenants:
```
BeryJu, line 56: I think the loop could be in the `try` block, which would get rid of the variable and the `list` call

authentik/blueprints/apps.py
```diff
@@ -13,21 +13,23 @@ class ManagedAppConfig(AppConfig):
 
     _logger: BoundLogger
 
+    RECONCILE_PREFIX: str = "reconcile_"
```
BeryJu, line 16: Now looking at it again I think there's probably a better way to do this than using `dir()` and then `getattr()`, probably something using a decorator?

Pseudo code:

```
@self.reconcile
def do_something(self):
    pass
```

(Not really part of this PR tbh)

authentik/brands/api.py
```diff
@@ -0,0 +1,142 @@
...
+            "branding_title",
+            "branding_logo",
+            "branding_favicon",
+            "flow_authentication",
+            "flow_invalidation",
+            "flow_recovery",
+            "flow_unenrollment",
+            "flow_user_settings",
+            "flow_device_code",
+            "event_retention",
```
BeryJu, line 57: This is probably one of the configuration options we can move to the tenant instead of the brand, it was only added here to not have it be a YAML-based config option

authentik/brands/api.py
```diff
@@ -0,0 +1,142 @@
...
+    matched_domain = CharField(source="domain")
+    branding_title = CharField()
+    branding_logo = CharField()
+    branding_favicon = CharField()
+    ui_footer_links = ListField(
+        child=FooterLinkSerializer(),
+        read_only=True,
+        default=get_default_ui_footer_links,
+    )
+    ui_theme = ChoiceField(
```
BeryJu, line 88: Is this gonna throw an error if any of the attributes don't exist/aren't a dict/aren't a valid option?

authentik/brands/middleware.py
```diff
@@ -0,0 +1,29 @@
...
+    get_response: Callable[[HttpRequest], HttpResponse]
+
+    def __init__(self, get_response: Callable[[HttpRequest], HttpResponse]):
+        self.get_response = get_response
+
+    def __call__(self, request: HttpRequest) -> HttpResponse:
+        if not hasattr(request, "brand"):
+            brand = get_brand_for_request(request)
+            setattr(request, "brand", brand)
+            set_tag("authentik.brand_uuid", brand.brand_uuid.hex)
```
BeryJu, line 24: This can probably be removed, I don't actually remember why it was added, it makes sense to use `set_tag` for tenants but not for brands

authentik/brands/models.py
```diff
@@ -0,0 +1,94 @@
...
+from authentik.lib.utils.time import timedelta_string_validator
+
+LOGGER = get_logger()
+
+
+class Brand(SerializerModel):
+    """Single brand"""
+
+    brand_uuid = models.UUIDField(primary_key=True, editable=False, default=uuid4)
+    domain = models.TextField(
```
BeryJu, line 21: Do we need to do any additional checking here that this is globally unique across all tenants?

authentik/events/models.py
```diff
@@ -170,7 +170,7 @@ class Event(SerializerModel, ExpiringModel):
     context = models.JSONField(default=dict, blank=True)
     client_ip = models.GenericIPAddressField(null=True)
     created = models.DateTimeField(auto_now_add=True)
-    tenant = models.JSONField(default=default_tenant, blank=True)
+    brand = models.JSONField(default=default_brand, blank=True)
```
BeryJu, line 173: I think we probably want to keep the tenant here instead of switching to brand, maybe adding brand in addition to tenant? Although most HTTP requests already have the domain, and if not we can add it in `from_http()` under `http_request`

authentik/events/signals.py
```diff
@@ -93,5 +93,5 @@ def event_post_save_notification(sender, instance: Event, **_):
 @receiver(pre_delete, sender=User)
 def event_user_pre_delete_cleanup(sender, instance: User, **_):
     """If gdpr_compliance is enabled, remove all the user's events"""
-    if CONFIG.get_bool("gdpr_compliance", True):
+    if get_current_tenant().avatars:
```
BeryJu, line 96: I dont think this is the right replacement lol

authentik/lib/default.yml
```diff
@@ -105,19 +105,14 @@ reputation:
...
-default_user_change_username: false
-
-gdpr_compliance: true
 cert_discovery_dir: /certs
 default_token_length: 60
-impersonation: true
+
+tenants:
+  enabled: false
+  api_key: ""
```
BeryJu, line 115: I'm still trying to figure out a better way of dealing with the tenant api key instead of having it in the config

authentik/root/db/base.py
```diff
@@ -1,5 +1,5 @@
 """authentik database backend"""
-from django_prometheus.db.backends.postgresql.base import DatabaseWrapper as BaseDatabaseWrapper
+from django_tenants.postgresql_backend.base import DatabaseWrapper as BaseDatabaseWrapper
```
BeryJu, line 2: I wonder if we can combine these somehow to still get the prometheus metrics

authentik/root/settings.py
```diff
@@ -18,6 +19,8 @@
 BASE_DIR = Path(__file__).absolute().parent.parent.parent
 STATICFILES_DIRS = [BASE_DIR / Path("web")]
 MEDIA_ROOT = BASE_DIR / Path("media")
+DEFAULT_FILE_STORAGE = "django_tenants.files.storage.TenantFileSystemStorage"
```
BeryJu, line 22: I wonder if we should consider some other file storage, especially for SaaS, I'd prefer not having to have a PVC mounted on all these pods

authentik/brands/__init__.py
BeryJu, line 1: I'd probably keep the brand stuff in the tenant app and not create a specific app for it, although I guess we might have to make it a separate app since its data is per tenant instead of global

BeryJu (commented):
🎉

authentik/blueprints/apps.py
rissson: The idea of having this try/catch outside of the loop was to specifically catch errors related to the tenant query, and not to whatever _reconcile is doing.

authentik/blueprints/apps.py
rissson: Yeah, it would be a lot better than filtering on function prefixes x)

rissson: Created https://github.com/goauthentik/authentik/issues/7852 for it

authentik/brands/api.py
rissson: How would we migrate the current config to tenants? I'm thinking just drop it, and start fresh, adding it to release notes as a breaking change.

authentik/brands/api.py
rissson: `get_default_ui_footer_links` should always be json already, but not specifically a list. I'll add it to the validation of the tenant model.

authentik/events/models.py
rissson: All events are already tenant-specific, so adding it as a field would not work. If you have a tenant `t_my_tenant` and the `public` tenant (the default one), you'd have two event tables, in two different schemas: `t_my_tenant.authentik_events_event` and `public.authentik_events_event`. I decided to keep the brand in here because it's what was done before.

authentik/lib/default.yml
rissson: I think this is fine for a first iteration. In the longer run, we can have an `/if/tenants` frontend, that is only accessible by superusers in the default tenant, same for the API.

authentik/root/db/base.py
rissson: This is still done in here: https://github.com/goauthentik/authentik/blob/multi-tenant-django-tenants/authentik/root/settings.py#L285:
```
ORIGINAL_BACKEND = "django_prometheus.db.backends.postgresql"
```
django-tenants will pick-up that backend and depend from it: https://github.com/django-tenants/django-tenants/blob/master/django_tenants/postgresql_backend/base.py#L60

authentik/root/settings.py
rissson: Yes, my thought was to add optional support for S3 in another PR. I think this PR is large enough as it is. We can also decide to keep the default filesystem storage as it is currently, and force the use of S3 when using tenants.

authentik/brands/__init__.py
rissson: Yeah, as brands are a per-tenant thing, they need to be in a separate app.

authentik/root/settings.py
BeryJu: Yeah definitely makes sense to do that in another PR. As much as I don't want to support multiple ways of doing something, we'd probably keep the local file storage support around for compose installs, and maybe change the current other method (of just inputting URLs) to use S3

authentik/brands/api.py
BeryJu: Sorry I meant the `ui_theme` field

authentik/brands/api.py
BeryJu: I think so too, yeah

authentik/blueprints/apps.py
BeryJu: fair enough

authentik/events/models.py
BeryJu: oh yeah d'oh of course, and the tenant info will also be logged correctly through the logging changes below

authentik/brands/__init__.py
BeryJu: maybe `authentik.tenants.brand`? This is a very small nitpick tbf

authentik/brands/models.py
rissson: Fortunately, no! As the tenant middleware is called before the brand middleware, even if the domain matches another tenant, the user will end up in the other tenant. So this domain has a lower priority than the one set in tenants.
(They also have different semantics, as the one in tenants is a perfect match whereas the one in brands is a suffix match)

authentik/brands/api.py
rissson: Done in https://github.com/goauthentik/authentik/pull/7590/commits/66ba607317a4722b6af936fc9275738f0912357c

authentik/root/settings.py
rissson: I actually reverted the revert, because I think we still need to keep tenants files separated even if on disk.

rissson: Okay, so a couple of modifications had to be made to the filesystem backend as well as django-tenants doesn't support django 5 for this.
I merged #7861 here to see what it would look like with tenants.

rissson: Ah nvm github won't let me push anymore

rissson:
image
this is done in the UI, which doesn't have enough access to check if `outposts.disable_embedded_outpost` is true
how would I go about removing that message?

rissson:
Only things missing are:
1. fixing the web UI settings page
2. release notes
3. documentation

rissson:
Reviewing this, everything in the backend looks pretty much in place.

I think the remaining CI fails are due to the API clients not being up-to-date.

PR #7617, opened 2023-11-18
sources/ldap: clean-up certs written from db

ChandonPierre:
## Details

Cert based ldap auth introduced in #5850 does not clean-up the temporary certificate files written out from the DB - when running in Kubernetes this eventually fills up `/dev/shm` and causes the worker to crashloop. 

This PR ensures the files are cleaned up in a `finally` block.

---

## Checklist

If applicable

### DIFF
--- a/authentik/sources/ldap/models.py
+++ b/authentik/sources/ldap/models.py
@@ -1,5 +1,7 @@
 """authentik LDAP Models"""
 from os import chmod
+from os.path import dirname, exists
+from shutil import rmtree
 from ssl import CERT_REQUIRED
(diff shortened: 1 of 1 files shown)


BeryJu (approved):

BeryJu (commented):
Actually I'm not sure if this works as intended, as this would clean up the certificate files while the server connection is still open; I'm not sure if there's any logic in ldap3 that will try to re-connect with the same server where this would cause issues; ideally we'd create the files upon connect and remove them upon disconnect, although that might require a custom ldap3 Server class

PR #7662, opened 2023-11-21
events/batch: add event batching mechanism [AUTH-134]

smusali:
## Details

Add event batching mechanism

---

## Checklist

If applicable

### DIFF
--- a/authentik/events/models.py
+++ b/authentik/events/models.py
@@ -172,6 +172,63 @@ class EventManager(Manager):
         return self.get_queryset().get_events_per(time_since, extract, data_points)
 
 
+class EventBatch(ExpiringModel):
+    """Model to store information about batches of events."""
+
+    batch_id = models.UUIDField(primary_key=True, default=uuid4, editable=False)
+    event_type = models.CharField(max_length=255)
+    event_app = models.CharField(max_length=255)
+    event_user = models.CharField(max_length=255)
+    start_time = models.DateTimeField(auto_now_add=True)
+    end_time = models.DateTimeField(null=True, blank=True)
+    event_count = models.IntegerField(default=0)
+    last_updated = models.DateTimeField(auto_now=True)
+    max_batch_size = models.IntegerField(default=10)
+    batch_timeout = models.IntegerField(default=60)  # Timeout in seconds
+    sent = models.BooleanField(default=False)
+
+    def add_event_to_batch(self, event):
+        """Add an event to the batch and check if it's ready to send."""
+        self.add_event(event)
+        if self.check_batch_limits():
+            self.process_batch()
+
+    @staticmethod
+    def get_or_create_batch(action, app, user):
+        """Get or create a batch for a given action."""
+        return EventBatch.objects.filter(
+            event_type=action, event_app=app, event_user=user, end_time__isnull=True
+        ).first() or EventBatch.objects.create(event_type=action, event_app=app, event_user=user)
+
+    def check_batch_limits(self):
+        """Check if the batch has reached its size or timeout limits."""
+        time_elapsed = now() - self.start_time
+        return self.event_count >= self.max_batch_size or time_elapsed >= timedelta(
+            seconds=self.batch_timeout
+        )
+
+    def add_event(self, event):
+        """Add an event to the batch."""
+        self.event_count += 1
+        self.save()
+
+    def create_batch_summary(self):
+        """Create a summary message for the batch."""
+        return f"Batched Event Summary: {self.event_type} action \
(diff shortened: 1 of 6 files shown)


authentik/events/models.py
```diff
@@ -160,6 +160,51 @@ def get_events_per(
...
+        )
+
+    def add_event(self, event):
+        """Add an event to the batch."""
+        self.event_count += 1
+        self.save()
+
+    def create_batch_summary(self):
+        """Create a summary message for the batch."""
+        return f"Batched Event Summary: {self.event_type} occurred {self.event_count} \
```
smusali, line 197: @BeryJu how much information is better to capture in the batched event notification?

authentik/events/models.py
```diff
@@ -160,6 +160,51 @@ def get_events_per(
...
+    max_batch_size = models.IntegerField(default=10)
+    batch_timeout = models.IntegerField(default=60)  # Timeout in seconds
+
+    @staticmethod
+    def get_or_create_batch(action):
+        """Get or create a batch for a given action."""
+        existing_batch = EventBatch.objects.filter(event_type=action, end_time__isnull=True).first()
+        if existing_batch:
+            return existing_batch
+        return EventBatch.objects.create(event_type=action)
```
BeryJu, line 181: One thing I would add is the `app` field to group events by, and maybe also `user`

authentik/events/models.py
```diff
@@ -331,6 +390,25 @@ def send(self, notification: "Notification") -> list[str]:
...
+        batch = EventBatch.get_or_create_batch(event.action)
+        batch.add_event(event)
+
+        summary_message = batch.process_batch()
+        if summary_message:
+            batch.delete()
+            batch_notification = Notification(
+                severity=NotificationSeverity.NOTICE,
+                body=summary_message,
+                event=None,
```
BeryJu, line 404: For the event data we can just use the most recent event, since iirc some webhook logic for the notifications uses the `event` field

authentik/events/models.py
```diff
@@ -196,8 +243,10 @@ def new(
             current = currentframe()
             parent = current.f_back
             app = parent.f_globals["__name__"]
+
         cleaned_kwargs = cleanse_dict(sanitize_dict(kwargs))
-        event = Event(action=action, app=app, context=cleaned_kwargs)
+        batch = EventBatch.get_or_create_batch(action)
```
BeryJu, line 248: shouldn't create the batch here since the event isn't saved yet, this should be created in the .save() method

authentik/events/models.py
BeryJu: the current one is fine imo, we could also go the kubernetes route "$event_name (occurred x32 times over the last x minuntes)

authentik/events/models.py
```diff
@@ -160,6 +160,51 @@ def get_events_per(
         return self.get_queryset().get_events_per(time_since, extract, data_points)
 
 
+class EventBatch(models.Model):
```
BeryJu, line 163: should inherit from the base class `ExpiringModel`

authentik/events/models.py
```diff
@@ -160,6 +160,51 @@ def get_events_per(
...
+    """Model to store information about batches of events."""
+
+    batch_id = models.UUIDField(primary_key=True, default=uuid4, editable=False)
+    event_type = models.CharField(max_length=255)
+    start_time = models.DateTimeField(auto_now_add=True)
+    end_time = models.DateTimeField(null=True, blank=True)
+    event_count = models.IntegerField(default=0)
+    last_updated = models.DateTimeField(auto_now=True)
+    max_batch_size = models.IntegerField(default=10)
+    batch_timeout = models.IntegerField(default=60)  # Timeout in seconds
```
BeryJu, line 173: We probably need a reference to the transport too to send the notifications when the task sends it out

BeryJu, line 173: We probably also need to add a status if a batch has been processed or not

BeryJu (commented):
We probably also should have a scheduled task running that looks for any pending batches that haven't been sent and have been idle for x time, and send them

authentik/events/models.py
smusali: Done!

authentik/events/models.py
smusali: Done!

authentik/events/models.py
smusali: Done!

authentik/events/models.py
smusali: Done!

PR #7664, opened 2023-11-21
added user_agent to Failed Logins Events

roney492:
Including the User-Agent in failed login attempt logs improves security monitoring. It enables the identification of suspicious or potentially malicious login attempts originating from specific devices or browsers.


BeryJu:
Thanks for the PR! Including the user-agent makes sense, although imo it might as well be included in all events created from HTTP requests. For the Client IP, that is already included in events in a separate field (`client_ip`)

PR #7690, opened 2023-11-22
website/integrations: Add note for special case on the AWS documentation

JiveOff:
## Details

In the case of users not having two strings separated by a space as their Authentik display name, SCIM would not work because the SCIM name property object requires a familyName, and authentik gets it by partitioning the display name.

As a workaround, a space can be used in the property mapping for the familyName, and thus, the formatted value must also be adapted with an extra space.

This seems like a workaround on the edgiest case, but I easily spent an hour figuring out what was wrong, hope this little doc note helps others.

---

## Checklist

If applicable


JiveOff:
Not entirely sure by the phrasing but I hope I made it clear, got stuck with this for more than a hour.

BeryJu (commented):
I think it would be easier to just change the default property mapping to have a space as default for `familyName` in case the users name doesn't contain spaces

JiveOff:
Fair enough, do I close this PR and have another one up that references this initial PR with the change to the blueprint?

PR #7692, opened 2023-11-22
web: use OAuth

### LINKED ISSUE #6385: Deny administration paths externally? /if/admin/?
**Describe your question/**
Since authentik runs as an SSO provider for external users to authenticate, it has to be exposed to the internet. 
In my use case, only local users should have administrator access. 
External users should only be using authentik to sign into preconfigured applications/services.

In order to accomplish this, which paths should be restricted to the internal network?
Rather, which paths should be exposed to the internet, so external users can only authenticate, without being able to access any administration areas of authentik?

-   authentik version: [2023.6]

BeryJu:
## Details

Migrate web interfaces (User and Admin) to use OAuth instead of session tokens

- [ ] Update api-explorer to work with Oauth since it currently relies on session auth

closes #6385

---

## Checklist

If applicable


septatrix:
What is the motivation for this?

Does this mean the tokens will now be stored in Javascript instead of as httponly, secure cookie? To the best of my knowledge that would mean a security degradation.

BeryJu:
> What is the motivation for this?
> 
> 
> 
> Does this mean the tokens will now be stored in Javascript instead of as httponly, secure cookie? To the best of my knowledge that would mean a security degradation.

The advantage is that you'll be able to set access restrictions on different interfaces, and in the future also having different OAuth scopes that give different permissions 

Ideally we wouldn't store the token in the client at all aside from in memory and re-request one when the page is refreshed

septatrix:
> The advantage is that you'll be able to set access restrictions on different interfaces, and in the future also having different OAuth scopes that give different permissions
> 
> Ideally we wouldn't store the token in the client at all aside from in memory and re-request one when the page is refreshed

That sounds like a good idea, however, I see no reason why the OAuth token couldn't be stored in a cookie instead

BeryJu:
Yeah that's also fair, tbf I haven't done much research on this and the localstorage is what this oidc library uses by default

septatrix:
It seems like the current value of the `authentik_session` token already is a JWT. Why not simply use that instead? If you want to check which scopes a user has (e.g. to omit some API calls which would fail or to hide navigation menus) you  can add those to the return value of `/api/v3/core/users/me/`

BeryJu:
> It seems like the current value of the `authentik_session` token already is a JWT. Why not simply use that instead? If you want to check which scopes a user has (e.g. to omit some API calls which would fail or to hide navigation menus) you  can add those to the return value of `/api/v3/core/users/me/`

The session cookie is a JWT for other reasons (mainly so that the go router can know if requests are authenticated without checking with the python code)

> 
> 
> 
> > I haven't done much research on this
> 
> 
> 
> Not the answer I would have expected from someone with a company whose product is security ;P

Hence the PR is not merged yet, but also everything is a trade off and regardless of that I'd set a pretty short access token, especially for the admin interface

PR #7698, opened 2023-11-23
enterprise/search: ability to use more precise search queries

BeryJu:
## Details

image

- accepts the same fields as the previous search did
- same filtering for permissions as currently
- add buttons to certain columns to "filter objects like these" based on selected column

---

## Checklist

If applicable


kensternberg-authentik (commented):
Just spotted something... odd... in the TableSearch

web/src/elements/table/TableSearch.ts
```diff
@@ -41,23 +76,24 @@ export class TableSearch extends AKElement {
             @submit=${(e: Event) => {
                 e.preventDefault();
                 if (!this.onSearch) return;
-                const el = this.shadowRoot?.querySelector<HTMLInputElement>("input[type=search]");
+                const el = this.shadowRoot?.querySelector<HTMLInputElement>("[name=search]");
```
rissson, line 79:
```suggestion
                const el = this.shadowRoot?.querySelector<HTMLTextAreaElement>("[name=search]");
```
as per the change below, I think

web/src/elements/table/TableSearch.ts
```diff
@@ -10,12 +13,40 @@ import PFFormControl from "@patternfly/patternfly/components/FormControl/form-co
...
+            syntaxHelp: null,
+
+            // optional, enable textarea auto-resize feature. If enabled,
+            // textarea will automatically grow its height when entered text
+            // doesn't fit, and shrink back when text is removed. The purpose
+            // of this is to see full search query without scrolling, could be
+            // helpful for really long queries.
+            autoResize: true,
+        });
+    }
```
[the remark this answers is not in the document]
BeryJu: Huh, do we have any other use cases of this pattern? If we don't then I was probably misremembering with this and the lit behaviour

authentik/enterprise/search/apps.py
```diff
@@ -0,0 +1,12 @@
+"""Enterprise app config"""
+
+from authentik.enterprise.apps import EnterpriseConfig
+
+
+class AuthentikEnterpriseSearchConfig(EnterpriseConfig):
+    """Enterprise app config"""
+
+    name = "authentik.enterprise.search"
+    label = "authentik_enterprise_search"
```
rissson:
```suggestion
    label = "authentik_search"
```

BeryJu: oh yeah duh

PR #7718, opened 2023-11-27
web: add third-tier "subcategories" to the sidebar

kensternberg-authentik:
## Details

This commit implements a three-tier sidebar with subcategories.

The first thing is that we’ve refactored the Sidebar into a holistic entity; rather than be built in pieces, it’s constructed declaratively from a tree of entries, much in the same way routes are, and for much the same reason<sup>1</sup>.

The AdminSidebar element only provides the list of entries to show and some of the controls necessary to show/hide the sidebar. Because the sidebar requires a rich collection of objects retrieved from the back-end, to avoid cluttering the AdminSidebar each of those sublists of TypeCreate have been isolated into their own controllers.

The SidebarTypeController isn’t even a strongly reactive controller; all it does is fetch the TypeCreate collection and notify the client object that the fetch has completed. The client can then call the `.entries()` method on the controller to get the sub-tree of entries for the TypeCreate object.

The Sidebar has been slightly (!) refactored so that it’s emphatic about what it does: it shows the brand, nav, and user sections of the sidebar. The styling has been moved to a separate file, the better to emphasize this.

The SidebarItems file is where all the magic– and a lot of ugly– is hidden.

The main purpose of the SidebarItems is to render the tree of entries passed to it. That’s it. But it also has to be able to read the URL and highlight which entry is currently being shown by the router, and it has to be able to open up all the parent objects of the “current” link so that the user gets a clear sense of where they are navigationally.

Most messy: the `reclick()` function intercepts clicks on anchors and, using the same logic as the router, decides if the router would *not* handle the navigation event. If the router would not, it takes on the responsibility for reaching into the currently visible table, modifying the filter and triggering a new `.fetch()`.

Somewhere along the way I boyscoutted another `switch` statement or two into lookup expressions.

------------------------------------------------------------------------

<sup>1</sup>  One of the reasons for this is that the Administrator Application’s sidebar, routes, and command palette will all get their data from a single source of truth, and that single source will be independent of any of those. This is a step in that direction.

## Checklist

-   [] The API schema has been updated (`make gen-build`)

If applicable


web/src/admin/AdminInterface/AdminSidebar.ts
```diff
@@ -98,117 +138,96 @@ export class AkAdminSidebar extends AKElement {
...
-            attributes?: Record<string, any> | string[] | null, // eslint-disable-line
-            children?: SidebarEntry[],
-        ];
+    get sidebarItems(): SidebarEntry[] {
+        const reload = () =>
+            new CoreApi(DEFAULT_CONFIG).coreUsersImpersonateEndRetrieve().then(() => {
+                window.location.reload();
+            });
+
+        const newVersionMessage: LocalSidebarEntry[] =
```
kensternberg-authentik, line 147: None of the "special cases" were all that special.  They all conformed to some navigational type, so they didn't have to be marked in any way, other than being part of the tree.

web/src/admin/AdminInterface/SidebarEntries/GenericTypesController.ts
```diff
@@ -0,0 +1,55 @@
+import { ReactiveControllerHost } from "lit";
```
kensternberg-authentik, line 1: This thing is both straightforward and pretty magical.  All our fetchers are simple.

web/src/admin/flows/utils.ts
```diff
@@ -6,40 +6,33 @@ export function RenderFlowOption(flow: Flow): string {
     return `${flow.slug} (${flow.name})`;
 }
 
+type FlowDesignationPair = [FlowDesignationEnum, string];
```
kensternberg-authentik, line 9: "Always leave the campsite better than the way you found it."

web/src/common/labels.ts
```diff
@@ -43,7 +45,9 @@ export const eventActionToLabel = new Map<EventActions | undefined, string>([
     [EventActions.ModelDeleted, msg("Model deleted")],
     [EventActions.EmailSent, msg("Email sent")],
     [EventActions.UpdateAvailable, msg("Update available")],
-]);
+];
+
+export const eventActionToLabel = new Map<EventActions | undefined, string>(eventActionLabels);
```
kensternberg-authentik, line 50: Oh, and this is _why_ I ended up boyscoutting; I needed the Event labels for the Events part of the Sidebar.

web/src/elements/sidebar/SidebarItems.css.ts
```diff
@@ -0,0 +1,86 @@
...
+
+import PFNav from "@patternfly/patternfly/components/Nav/nav.css";
+import PFPage from "@patternfly/patternfly/components/Page/page.css";
+import PFBase from "@patternfly/patternfly/patternfly-base.css";
+
+export const sidebarItemStyles = [
+    PFBase,
+    PFPage,
+    PFNav,
+    css`
```
kensternberg-authentik, line 11: There's a lot of custom CSS here, but none of it is particularly locked-in.  All of my customized additions use the `--pf-global` namespace for their values, in the hopes that we can keep using those without fear.

web/src/elements/sidebar/SidebarItems.ts
```diff
@@ -0,0 +1,247 @@
...
+    toggleExpand(entry: SidebarEntry) {
+        const key = entryKey(entry);
+        if (this.expanded.has(key)) {
+            this.expanded.delete(key);
+        } else {
+            this.expanded.add(key);
+        }
+        this.requestUpdate();
+    }
+
```
kensternberg-authentik, line 87: Just reiterating: this is gross.  :-)

PR #7739, opened 2023-11-30
website/docs: add expression example for geoip

Rukk385:
Added example for GeoIP
`return context["geoip"].country.iso_code == "US"`

## Checklist

If applicable

### DIFF
--- a/website/docs/policies/expression.mdx
+++ b/website/docs/policies/expression.mdx
@@ -58,6 +58,7 @@ import Objects from "../expressions/_objects.md";
     ```python
     return context["geoip"].country.iso_code == "US"
     ```
+
 -   `ak_is_sso_flow`: Boolean which is true if request was initiated by authenticating through an external provider.
 -   `ak_client_ip`: Client's IP Address or 255.255.255.255 if no IP Address could be extracted. Can be [compared](#comparing-ip-addresses), for example
 
--- a/website/docs/policies/expression.mdx
+++ b/website/docs/policies/expression.mdx
@@ -54,6 +54,10 @@ import Objects from "../expressions/_objects.md";
     -   `request.context`: A dictionary with dynamic data. This depends on the origin of the execution.
 
 -   `geoip`: GeoIP object, see [GeoIP](https://geoip2.readthedocs.io/en/latest/#geoip2.models.City)
+
+    ```python
+    return context["geoip"].country.iso_code == "US"
+    ```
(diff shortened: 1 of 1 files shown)


website/docs/policies/expression.mdx
```diff
@@ -54,6 +54,10 @@ import Objects from "../expressions/_objects.md";
     -   `request.context`: A dictionary with dynamic data. This depends on the origin of the execution.
 
 -   `geoip`: GeoIP object, see [GeoIP](https://geoip2.readthedocs.io/en/latest/#geoip2.models.City)
+
+    ```python
+    return context["geoip"].country.iso_code == "US"
+    ```
```
rissson, line 60:
```suggestion
    ```

```

rissson (approved):
Just a small nitpick, otherwise LGTM

website/docs/policies/expression.mdx
```diff
@@ -54,6 +54,11 @@ import Objects from "../expressions/_objects.md";
...
+
 -   `ak_is_sso_flow`: Boolean which is true if request was initiated by authenticating through an external provider.
```
tanberry, line 62: @Rukk385 and @rissson, does this rewrite keep the technical accuracy? "Boolean. By default, this flag is set to True if request was initiated by authenticating through an external provider."

We need to separate the fact that it is boolean from rest of info. And is it a flag, or a value? Finally, does the system set this value (by default), or the user?

website/docs/policies/expression.mdx
```diff
@@ -54,6 +54,11 @@ import Objects from "../expressions/_objects.md";
...
 -   `ak_client_ip`: Client's IP Address or 255.255.255.255 if no IP Address could be extracted. Can be [compared](#comparing-ip-addresses), for example
```
tanberry, line 63: Is there supposed to be more at the end of this line? Or do we just need a period there at end?

And can we rewrite to "If no IP Address could be extracted, this value defaults to either the client's IP Address or 255.255.255.255. You can use the [compare functions](#comparing-ip-addresses) to learn more about the IP address."

tanberry (commented):
Thanks so much @Rukk385 for your additions to the docs! Can you take a look at my suggested rewrites, and make sure they are still technically accurate? Tweak as needed. :-)

website/docs/policies/expression.mdx
rissson: This is out of scope for this PR I think.

Issue #7025, opened 2023-10-02
Authentik [worker] CPU 100% and redis.


crisukbot:
**Describe the bug**
Authentik Worker clogs the processor to 100% and eventually shuts down the entire system.
After deleting the redis folder, everything worked fine. (Maybe there's a problem with how Authentik works with Redis?)

**To Reproduce**
It's hard to explain, I started authentik and after three or four or five hours the server shut down.

**Expected behavior**
Authentik does not shut down my server

**Screenshots**
![image](https://github.com/goauthentik/authentik/assets/105280814/d26b1c69-8e40-41e2-80b7-a62ee434f475)

**Logs**
Not anymore.

**Version and Deployment (please complete the following information):**

-   authentik version: 2023.8.3
-   Deployment: Docker-compose

Zegorax:
I have the same problem. Worker takes 800% of my CPUs and crashes everything.

sammcj:
Same problem here on 2023.6.1.

- docker / docker-compose.
- minimal config (one domain forward auth).
- no errors or warnings in the logs.
- pegs 1 Ryzen 7600 CPU core at 75% constantly.

sammcj:
Running a strace on the process it looks like it's constantly trying to access a file that doesn't exist:

```
strace: Process 1838696 attached
09:20:30.185132 close(954267412)        = -1 EBADF (Bad file descriptor) <0.000004>
09:20:30.185146 close(954267411)        = -1 EBADF (Bad file descriptor) <0.000004>
09:20:30.185161 close(954267410)        = -1 EBADF (Bad file descriptor) <0.000004>
09:20:30.185175 close(954267409)        = -1 EBADF (Bad file descriptor) <0.000004>
09:20:30.185189 close(954267408)        = -1 EBADF (Bad file descriptor) <0.000004>
09:20:30.185203 close(954267407)        = -1 EBADF (Bad file descriptor) <0.000004>
09:20:30.185217 close(954267406)        = -1 EBADF (Bad file descriptor) <0.000004>
09:20:30.185233 close(954267405)        = -1 EBADF (Bad file descriptor) <0.000004>
```

sammcj:
Perhaps interesting - if I reduce the worker cpu to 0.1 I notice the following errors:

```
│   {"event": "Timed out waiting for UP message from <ForkProcess(ForkPoolWorker-50, started daemon)>", "level": "error", "logger": "celery.concu   │
[...]

PR #7762, opened 2023-12-03
root: fix 100% CPU for worker container (#7025)

DKingAlpha:
Some linux users (arch linux, for example) are running docker with default service file that set NOFILES to infiite, which will cause celery to hang for hours to days taking 100% CPU to close all fds by enumerating from NOFILES to 3.

This commit override ulimit for container without touching user docker service configuration.

For details see #7025

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -56,6 +56,10 @@ services:
     image: ${AUTHENTIK_IMAGE:-ghcr.io/goauthentik/server}:${AUTHENTIK_TAG:-2023.10.4}
     restart: unless-stopped
     command: worker
+    ulimits:
+      nofile:
+        soft: 10240
+        hard: 10240
     environment:
       AUTHENTIK_REDIS__HOST: redis
       AUTHENTIK_POSTGRESQL__HOST: postgresql


BeryJu (approved):
Interesting, thanks for digging into this! I'd like to get confirmation from some more people in #7025 if it fixes their issue before merging this. Also we should probably add a similar thing for the helm chart (https://github.com/goauthentik/helm) if possible

DKingAlpha:
Tried to ping someone, waiting for feedback.

Meanwhile you can easily reproduce by changing the ulimit to 0x3ffffff8 in decimal. That should kinda prove it.

Leptopoda (approved):
Fixes the issue for me.
```
OS: clear-linux
# ulimit -Sn
1024
# ulimit -Hn
524288
```

mobiledude:
> Fixes the issue for me.
> 
> ```
> OS: clear-linux
> # ulimit -Sn
> 1024
> # ulimit -Hn
> 524288
> ```

Any idea someone hoe I can change this for my Authentik worker on unraid? It’s been running high cpu usage for days now? Help is appreciated.

cenkalti:
Do you know why setting ulimit to a larger number fixes the issue? Is it an issue in Celery or Authentik?

BeryJu:
With #7810, #8440 and #7813 this shouldn't be an issue anymore, could you check this again with 2024.2.2 @cenkalti @mobiledude @Leptopoda @DKingAlpha

Leptopoda:
Thanks for coming back.
I can confirm that I no longer need the workaround on 2024.02.2

DKingAlpha:
I still got high cpu usage with latest 2024.2.2. Adding ulimit back to compose.yml fixed the issue for me.

live py profiler py-spy is incompatible with recent py3.12, I will find another way to identify the issue when I have time.

cenkalti:
I still experience high CPU usage with the latest 2024.2.2 version. However, I was able to resolve the issue by adding `ulimit` back to the 'compose.yml' file.

MyIgel:
I can confirm that it fixed my setup too so its imho worth merging :tada:

SpiderD555:
New authentik user here. Tried re-setting Redis, tried setting ulimits in docker-compose, unfortunately CPU still spikes at 100%, and I haven't been even able to access the main page yet :(

Janhouse:
Same here with `2024.4.2`, ulimits in the compose fixed the issue.

arthurlockman:
Adding ulimits back to compose fixed my issue on `2024.4.2`.

BeryJu:
For context, the reason why we haven't merged this PR:

- Configuring ulimit values for containers is possible with compose, but not possible with kubernetes, so it would only solve half the problem
- It's also more of a bandaid than a full solution, this ulimit adjustment shouldn't be required and changing the values seems to just prevent a bug in either our code/our usage of celery, or celery itself from happening, instead of fixing the root cause itself.

ForsakenRei:
I can confirm this fixed my issue for docker running on Oracle Linux 9.4

cubic3d:
Just deployed a fresh compose install on current Arch linux using `2024.6.3`. Celery was taking one core to 100%. Setting ulimits for nofile resolved the issue.

I'm not experiencing this in k8s running on Talos.

justin8:
Thanks! This fixed the issue for me as well.

maxnoe:
Fix worked for me as well

BeryJu:
with 2025.8 we no longer use celery so this shouldn't be required anymore

PR #7792, opened 2023-12-05
providers/scim: change familyName default

JiveOff:
## Details

In some SCIM implementations, the familyName cannot be empty (such as the AWS implementation). This quick fix adds a space for the familyName in the default SCIM property mapping and concats the givenName & familyName for the formatted value.

From the initial PR #7690 

---

## Checklist

If applicable

### DIFF
--- a/blueprints/system/providers-scim.yaml
+++ b/blueprints/system/providers-scim.yaml
@@ -11,7 +11,7 @@ entries:
       name: "authentik default SCIM Mapping: User"
       expression: |
         # Some implementations require givenName and familyName to be set
-        givenName, familyName = request.user.name, ""
+        givenName, familyName = request.user.name, " "
         # This default sets givenName to the name before the first space
         # and the remainder as family name
         # if the user's name has no space the givenName is the entire name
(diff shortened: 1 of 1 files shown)


blueprints/system/providers-scim.yaml
```diff
@@ -11,14 +11,16 @@ entries:
...
-        givenName, familyName = request.user.name, ""
+        givenName, familyName = request.user.name, " "
         # This default sets givenName to the name before the first space
         # and the remainder as family name
         # if the user's name has no space the givenName is the entire name
         # (this might cause issues with some SCIM implementations)
         if " " in request.user.name:
             givenName, _, familyName = request.user.name.partition(" ")
 
+        formatted = givenName + familyName
```
BeryJu, line 22: I don't think this is needed tbh, it'll just append a space to the name if no family name is given or will concat them both without space if one is given

JiveOff: From my testing, it seems necessary as the familyName prop must match the formatting

JiveOff: Sorry, didn't quite catch that. Isn't it useless to move it up since we already have the givenName & familyName values that are going to be used in the SCIM object ?

BeryJu: right, but in the case that there is a space in the users' name, we don't need to construct `formatted` manually as it'll be identical to what `user.name` is, we only need to default `formatted` to `user.name` + ` ` in the case there's no space in the users' name

PR #7793, opened 2023-12-05
events: add ASN Database reader

BeryJu:
## Details

REPLACE ME

---

## Checklist

If applicable


authentik/lib/default.yml
```diff
@@ -106,7 +106,11 @@ cookie_domain: null
 disable_update_check: false
 disable_startup_analytics: false
 avatars: env://AUTHENTIK_AUTHENTIK__AVATARS?gravatar,initials
-geoip: "/geoip/GeoLite2-City.mmdb"
+events:
+  processors:
+    geoip: "env://AUTHENTIK_GEOIP?/geoip/GeoLite2-City.mmdb"
```
rissson, line 111: do we want to deprecate AUTHENTIK_GEOIP, and add AUTHENTIK_GEOIP_CITY and AUTHENTIK_GEOIP_ASN instead?

authentik/policies/types.py
```diff
@@ -39,12 +40,13 @@ def __init__(self, user: User):
...
-        if not GEOIP_READER.enabled:
-            return
         client_ip = get_client_ip(request)
         if not client_ip:
             return
-        self.context["geoip"] = GEOIP_READER.city(client_ip)
+        if GEOIP_ENRICHER.enabled:
+            self.context["geoip"] = GEOIP_ENRICHER.city(client_ip)
+        if ASN_ENRICHER.enabled:
+            self.context["asn"] = ASN_ENRICHER.asn(client_ip)
```
rissson, line 49: Can enrichers have a `get_context` method instead? This pattern is used here and in the enricher itself for `enrich_event` as well

authentik/events/enrich/geoip.py
```diff
@@ -23,7 +26,7 @@ class GeoIPDict(TypedDict):
     city: str
 
 
-class GeoIPReader:
+class GeoIPEnricher(EventEnricher):
```
rissson, line 29: It seems there's a lot of duplicated stuff between the GeoIPEnricher and ASNEnricher. I get it doesn't make sense to but those in the baseenricher, but does it make sense to make a MMDBEnricher?

PR #7810, opened 2023-12-07
blueprints: only watch for fs events we're interested in

rissson:
## Details

I looked into this after #7025. Turns out we were not using the built-in features for filtering filesystem events. With this we would be.

Not tried out yet.

---

## Checklist

If applicable

### DIFF
--- a/authentik/blueprints/v1/tasks.py
+++ b/authentik/blueprints/v1/tasks.py
@@ -60,7 +60,12 @@ def start_blueprint_watcher():
     if _file_watcher_started:
         return
     observer = Observer()
-    observer.schedule(BlueprintEventHandler(), CONFIG.get("blueprints_dir"), recursive=True)
+    observer.schedule(
+        BlueprintEventHandler(),
+        CONFIG.get("blueprints_dir"),
+        recursive=True,
+        event_filter=(FileCreatedEvent, FileModifiedEvent),
+    )
     observer.start()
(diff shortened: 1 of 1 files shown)


authentik/blueprints/v1/tasks.py
```diff
@@ -62,29 +62,42 @@ def start_blueprint_watcher():
     if _file_watcher_started:
         return
     observer = Observer()
-    observer.schedule(BlueprintEventHandler(), CONFIG.get("blueprints_dir"), recursive=True)
+    observer.schedule(
+        BlueprintEventHandler(),
+        CONFIG.get("blueprints_dir"),
+        recursive=True,
+        event_filter=(FileCreatedEvent, FileModifiedEvent),
```
BeryJu, line 69: I wasn't able to find any reference to `event_filter` in the watchdog docs

rissson: Oh yeah, there isn't anything in the reference docs. I looked at the code directly:

https://github.com/gorakhargosh/watchdog/blob/master/src/watchdog/observers/api.py#L306
https://github.com/gorakhargosh/watchdog/blob/master/src/watchdog/observers/inotify.py#L218

Apparently it hasn't been released yet, that's maybe why it isn't in the published docs.

BeryJu (approved):

PR #7831, opened 2023-12-09
root: Fix cache related image build issues

PKizzle:
## Details

This PR fixes issues with the caching of dependencies when building multi arch images.

In particular the locking of apt caused issues when the same cache was reused for multiple target architectures.
Adding a custom id for each architecture fixes this as a separate cache will be created for each of them.

This is enabled by [a recent change in buildkit](https://github.com/moby/buildkit/issues/815) that added the ability to use variables in the mount options.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,3 +1,5 @@
+# syntax=docker/dockerfile:1
+
 # Stage 1: Build website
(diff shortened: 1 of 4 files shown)


rissson (approved):

rissson:
@BeryJu we'll need to do the same for the RAC outpost

PR #7854, opened 2023-12-11
web: fix turnstile types after update

kensternberg-authentik:
## Details

After running 'npm update' on the dev tree, the build started to fail with these options and types no longer being set correctly in the source tree.

I have explicitly included the Turnstile object as a sub-component of Window, and modified the CaptchaStage to understand the TurnstileObject and TurnstileOptions, and the build now completes.  As these are purely type-level changes, the functionality of the system is unchanged, and testing (by hand, _sigh_) indicates that the visible behavior is consistent with previous iterations.

## Checklist

If applicable


web/src/global.d.ts
```diff
@@ -1,4 +1,5 @@
 declare module "*.css";
+
```
kensternberg-authentik, line 2: Prettier _should have_ had an opinion here!  :-)

web/src/flow/stages/captcha/CaptchaStage.ts
```diff
@@ -110,9 +115,9 @@ export class CaptchaStage extends BaseStage<CaptchaChallenge, CaptchaChallengeRe
         if (!Object.hasOwn(window, "turnstile")) {
             return false;
         }
-        window.turnstile.render(container, {
+        (window as unknown as TurnstileWindow).turnstile.render(container, {
             sitekey: this.challenge.siteKey,
-            size: "invisible",
+            size: "invisible" as TurnstileOptions["size"],
```
kensternberg-authentik, line 120: @BeryJu Could you please affirm this?  The `as TurnstilesOptions["size"]` clause was necessary here because "invisible" is not a valid setting according to the options in the provided type file.  If `invisible` is valid because you've made it so, let me know and I'll a comment to that effect.

BeryJu: I don't think thats valid (at least not anymore), so we can probably remove it

BeryJu (commented):
As https://github.com/goauthentik/authentik/pull/7797 is still open, could you also lock pyright to the previous version as it's currently breaking CI

BeryJu (approved):

PR #7902, opened 2023-12-18
website/docs: add operations category

rissson:
## Details

As discussed. Cc. @tanberry 

---

If applicable


rissson:
I'm still not happy with the `running authentik` nomenclature. I still think that `Operating` is a better term.

rissson:
Changed "running authentik" to "operations" per our last weekly discussion

PR #7923, opened 2023-12-19
website/integrations: multiple integration edits

ZuluWhiskey:
## Details

- This PR does not close an issues
- After recently contributing the Immich documentation, I took a look at the other integrations within the Miscellaneous section of the website. I noticed that there were differences in documentation across the different services and thought it may be a good idea to try and update and standardise things a little. In this PR I've changed:
  - Home Assistant
    - I've tried to update formatting and also updated the guidance as I felt it was unclear
    - It was listed as `Home-Assistant` rather than `Home Assistant`, I've not seen it referenced with a hyphen elsewhere
  - FreshRSS
    - I've updated the authentik section only
  - Gravitee
    - I've updated the authentik and service sections
  - Immich (I spotted aspects in other services which I feel would improve this)
    - Added some notes
I felt it best to not change too much in one PR before I get guidance from the project on whether this is wanted and/or useful

---

## Checklist

If applicable

### DIFF
--- a/website/integrations/services/home-assistant/index.md
+++ b/website/integrations/services/home-assistant/index.md
@@ -24,6 +24,24 @@ The following placeholders will be used:
 -   `hass.company` is the FQDN of the Home Assistant install.
 -   `authentik.company` is the FQDN of the authentik install.
 
+## authentik configuration
+
+1. Create a **Proxy Provider** under **Applications** > **Providers** using the following settings:
+
+    - **Name**: Home Assistant
+    - **Authentication flow**: default-authentication-flow
+    - **Authorization flow**: default-provider-authorization-explicit-consent
+    - **External Host**: Set this to the external URL you will be accessing Home Assistant from
+    - **Internal Host**: `http://hass.company:8123`
+
+2. Create an **Application** under **Applications** > **Applications** using the following settings:
+
+    - **Name**: Home Assistant
+    - **Slug**: homeassistant
+    - **Provider**: Home Assistant (the provider you created in step 1)
+
+3. Create an outpost deployment for the provider you've created above, as described [here](../../../docs/outposts/). Deploy this Outpost either on the same host or a different host that can access Home Assistant. The outpost will connect to authentik and configure itself.
+
 ## Home Assistant configuration
 
 1. Configure [trusted_proxies](https://www.home-assistant.io/integrations/http/#trusted_proxies) for the HTTP integration with the IP(s) of the Host(s) authentik is running on.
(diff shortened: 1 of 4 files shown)


ZuluWhiskey:
I meant to say, if you'd prefer I chunk these PRs up per integration then let me know and I'll ensure to do so in future as I plan on making my way through the other integrations (as long as you're happy with what I'm doing of course!)

rissson:
> I meant to say, if you'd prefer I chunk these PRs up per integration then let me know and I'll ensure to do so in future as I plan on making my way through the other integrations (as long as you're happy with what I'm doing of course!)

I hope Jens agrees with me, but of course it's fine. It's absolutely great that you're taking the time to standardize our docs and we very much appreciate it!

rissson (approved):
I'm gonna go ahead and approve this, but I'd still like @tanberry to take a last look to make sure this is indeed the direction we want to go in

ZuluWhiskey:
Just to say, I'll try to keep working on these but will wait until there's confirmation on the approach I'm taking etc. 👍

website/integrations/services/home-assistant/index.md
```diff
@@ -13,54 +13,57 @@ title: Home-Assistant
...
+                X-ak-hass-user: hassusername
+            ```
+        3. Then configure the Home Assistant custom component to use this header:
+            ```yaml
+            auth_header:
+                username_header: X-ak-hass-user
+            ```
+
+## authentik configuration
+
```
tanberry, line 54: Let's move the authentik config section to be first, so that it is consistent with [the majority] of the other integrations. Basically, let's keep this order:
What is XYZ
Preparation
authentik configuration
XYZ configuration

website/integrations/services/immich/index.md
```diff
@@ -19,24 +19,26 @@ The following placeholders will be used:
...
+      :::note
+      Please note that the following URIs are just examples. Be sure to include all of the domains / URLs that you will use to access Immich.
+      ::: - app.immich:/ - https://immich.company/auth/login - https://immich.company/user-settings
     - **Signing Key**: authentik Self-signed Certificate
     - Leave everything else as default
 2. Open the new provider you've just created.
 3. Make a note of the **OpenID Configuration Issuer**.
 
-## Immich Configuration
+## Immich configuration
```
tanberry, line 41: Nice, thank you for this level of attention to detail. :-)

tanberry:
Hi @ZuluWhiskey huge apologies for my taking so long to review! My bad. This is really a terrific PR, much appreciated indeed. I have noticed a lot of inconsistencies between the diff Integrations docs; we should probably create a template for new docs.

I made only one comment in your PR that needs your action... I think that on the Home Assistant one, we want to switch the order of sections so that authentik comes first.

Also, what are your thoughts about those sections in general? Does one always have to do configuration on the authentik side?

Again, thanks so much, this is 100% in the direction that we want to move with the Integrations docs: consistent ordering of the sections, headers using sentence-case capitalization, etc.

tanberry (changes requested):
Thanks again! Just one requested change, on the Home Assistant one, to reorder the sections so that authentik is first. Everything else looks great!

website/integrations/services/home-assistant/index.md
ZuluWhiskey: Agreed, good spot! Sorted now 😄

ZuluWhiskey:
> Hi @ZuluWhiskey huge apologies for my taking so long to review! My bad. This is really a terrific PR, much appreciated indeed. I have noticed a lot of inconsistencies between the diff Integrations docs; we should probably create a template for new docs.

Agreed with regards to the need for a template. Obviously one already exists but I think if we want consistency then ideally either we add more to the template or improve the guidance for writing documentation on the website?

> Also, what are your thoughts about those sections in general? Does one always have to do configuration on the authentik side?

I would assume that authentik would always need to be configured in one way or another in order to work with a new integration. And I guess we have to consider that for some users, this may be their first time setting up anything within authentik so they may not have any pre-existing configuration.
In this vein however, I did notice when making these edits that there were consistent steps with only minor differences - For example the creation of an **OAuth2/OpenID Provider**. But I'm unsure that we could template this and apply it across the board 🤔 

> Again, thanks so much, this is 100% in the direction that we want to move with the Integrations docs: consistent ordering of the sections, headers using sentence-case capitalization, etc.

No worries at all, glad I can give something back to the project!

tanberry (approved):

tanberry:
Thank you again @ZuluWhiskey sorry it took me so long!

Issue #5851, opened 2023-06-04
Issue with changing passwords with eDirectory


bdallen:
**Describe the bug**
Unable to change passwords using eDirectory for LDAP backend

**To Reproduce**

1. Login as user
2. Click Change Password
3. Enter both passwords
4. Click OK to change password
5. Get Request Denied Message

**Expected behavior**
Expected to allow password change and continue

**Supporting Error Mesage**
Event info

Action
    Configuration error
App
    authentik.sources.ldap.signals
User
    [xxxxxxxxxx](https://xxxxxx.levelup.solutions/if/admin/#/identity/users/4)
Created
    05/06/2023, 7:49:26 am
Client IP
    -
Tenant
    Tenant fallback

Context
{
    "source": {
        "pk": "dc9fc888725147b9aa85b318e7d47236",
        "app": "authentik_sources_ldap",
        "name": "xxxxxxxxx",
        "model_name": "ldapsource"
    },
    "message": "Failed to change password in LDAP source due to remote error: 16, NDS error: no such value (-602), noSuchAttribute"
}
User
{
    "pk": 4,
    "email": "xxxx@xxxx",
    "username": "xxxxxxxx"
}

**Logs**
root@BNE-PROD-IDP001:~/authentik# cat logs.txt | grep NDS
authentik-server-1  | {"auth_via": "session", "event": "failed to set LDAP password", "exc": "LDAPNoSuchAttributeResult - 16 - noSuchAttribute - None - NDS error: no such value (-602) - modifyResponse - None", "host": "xxxxxx.levelup.solutions", "level": "warning", "logger": "authentik.sources.ldap.signals", "pid": 22, "request_id": "42b4a761fa044d96ad820ec17179dcac", "timestamp": "2023-06-04T22:06:43.481174"}
[...]

PR #7997, opened 2023-12-26
sources/ldap: fix Issue with changing passwords with eDirectory

bdallen:
## Details

Bugfix for Issue #5851

Had issue with an unknown attribute when Authentik by default used Microsoft AD for password writeback, which threw an uncaught exception. 

To fix this one, I just added LDAPNoSuchAttributeResult to the exception catch statement to process standard LDAP change password extensions which worked with Microfocus eDirectory perfectly.

### DIFF
--- a/authentik/sources/ldap/password.py
+++ b/authentik/sources/ldap/password.py
@@ -4,7 +4,7 @@ from re import split
 from typing import Optional
 
 from ldap3 import BASE
-from ldap3.core.exceptions import LDAPAttributeError, LDAPUnwillingToPerformResult
+from ldap3.core.exceptions import LDAPAttributeError, LDAPUnwillingToPerformResult, LDAPNoSuchAttributeResult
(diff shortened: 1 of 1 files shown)


authentik/sources/ldap/password.py
```diff
@@ -4,7 +4,7 @@
 from typing import Optional
 
 from ldap3 import BASE
-from ldap3.core.exceptions import LDAPAttributeError, LDAPUnwillingToPerformResult
+from ldap3.core.exceptions import LDAPAttributeError, LDAPUnwillingToPerformResult, LDAPNoSuchAttributeResult
```
BeryJu, line 7:
```suggestion
from ldap3.core.exceptions import (
    LDAPAttributeError,
    LDAPNoSuchAttributeResult,
    LDAPUnwillingToPerformResult,
)
```

BeryJu (approved):

PR #8002, opened 2023-12-27
expressions: add restricted evaluator

rissson:
## Details

Very much a WIP, I've only tested it a bit, and it seems to work with our existing blueprints. It also throws errors when trying to import something. Haven't tested stuff like `open`.

We also probably need to come up with a list of stuff to import "by default". For now I have the following, feel free to edit this description to add more:
- `json`
- `yaml`
- `re`

---

## Checklist

If applicable


authentik/lib/expression/evaluator.py
```diff
@@ -221,13 +251,8 @@ def handle_error(self, exc: Exception, expression_source: str):  # pragma: no co
...
     def validate(self, expression: str) -> bool:
         """Validate expression's syntax, raise ValidationError if Syntax is invalid"""
-        param_keys = self._context.keys()
         try:
-            compile(
-                self.wrap_expression(expression, param_keys),
-                self._filename,
-                "exec",
-            )
+            self.compile(expression)
```
rissson, line 255: It's a bit of a shame that this won't validate whether the expression will run. For instance, it won't check that `import module` will work, because that relies on the builtin `__import__` being there, and we purposefully remove it when we call `exec`.

authentik/lib/expression/evaluator.py
```diff
@@ -180,6 +197,18 @@ def wrap_expression(self, expression: str, params: Iterable[str]) -> str:
         full_expression += f"\nresult = handler({handler_signature})"
         return full_expression
 
+    def compile(self, expression: str) -> Any:
+        """Parse expression. Raises SyntaxError or ValueError if the syntax is incorrect."""
+        param_keys = self._context.keys()
+        compiler = (
+            compile_restricted if CONFIG.get_bool("epxressions.restricted", False) else compile
```
rissson, line 204:
```suggestion
            compile_restricted if CONFIG.get_bool("expressions.restricted", False) else compile
```

rissson: Maybe that's why it was working :upside_down_face:

rissson:
TODO: look into making a `Runtime` abstraction
TODO: use python client instead of importing models

rissson:
Note to self: pre-importing django models won't prevent access to other tenants' data, as you can do Model.objects.raw() and type whatever sql you'd like.

PR #8004, opened 2023-12-28
web: provide dual-list multiselect with pagination

### LINKED ISSUE #4880: Outpost is not able to select more than 100 Applications
**Describe the bug**
The embedded outpost application select box is only able to list 100 items.

**To Reproduce**
once you edit the outpost the applications selector is lock to 100 ordered alphabetically.

293CC595-88CE-4194-A76B-B3AFFC39630D

 - authentik version:  2023.2.2
 - Deployment: docker-compose

kensternberg-authentik:
web: provide a “select / select all” tool for the dual list multiselect

**This commit**

This commit provides the following new features for dual list multiselect:

- The “available” pane, which has all of the entries that are available to be selected. Items that are already selected will remain, but they’re marked with a checkmark and can neither be selected or moved.
- The “selected” pane, which has *all* of the entries that have been selected.
- The Pagination control, which in this case only sends an event upstream.

**Plan**:

The plan is to have a master control that marries the available-pane, selected-pane, select-controls, and pagination-controls into a single component that receives the list of “currently visible” available entries and keeps the list of “currently selected” entries, as well as a pass-through for the pagination value that allows it to hide the pagination control if there is only one page.

A master component *above that* will provide the list of currently visible entries and, at need, read the value of the master control object for the “selected” list. That component will mostly be data-only; it’s render will probably just be `<slot></slot>`; its duty will be only to map entries to string keys Lit can use, and to provide the lists we want to provide and the pagination ranges we want to show.

Some judicious use of grid will allow me size the controls properly with/without the pagination control.

Status and Title are going to be in the master control.

There is already a planned fallback control; the multi-select experience on mobile is actually excellent, and we should exploit that appropriately.

closes #4880

## Checklist

If applicable


BeryJu (commented):
there's a couple bugs left with the dark mode that we should fix before merging

web/src/admin/outposts/OutpostForm.ts
```diff
@@ -20,14 +22,76 @@ import {
...
 } from "@goauthentik/api";
 
+interface ProviderBase {
+    pk: number;
+    name: string;
+    assignedBackchannelApplicationName?: string;
+    assignedApplicationName?: string;
+}
+
+interface ProviderData {
```
BeryJu, line 37: Could probably replace this with `import { PaginatedResponse } from "@goauthentik/app/elements/table/Table";` which is a generic type

BeryJu (approved):
LGTM, especially with the dark theme changes

We'll have to do a separate PR to replace all the other multiselect occurrences

PR #8013, opened 2023-12-29
outpost: improved set secret answers for flow execution

sirux88:
## Details

Set the flow answer for StageAuthenticatorValidate equal to StagePassword if code-based MFA is not enabled.
This change is valid for the radius outpost handle function.

Reason for this PR is:
I need to validate a user's MFA (in my case a TOTP token) against authentik which is not possible in its current state
(Password validation is done by an other system)

A possible but  very dirty workaround would be to prepend a semicolon to the MFA token. But my system doesn't allow this.

As a sidemark:
If necessary this could be implemented for ldap outpost aswell.

---

## Checklist

If applicable

### DIFF
--- a/internal/outpost/radius/handle_access_request.go
+++ b/internal/outpost/radius/handle_access_request.go
@@ -21,14 +21,7 @@ func (rs *RadiusServer) Handle_AccessRequest(w radius.ResponseWriter, r *RadiusR
 	fe.Params.Add("goauthentik.io/outpost/radius", "true")
 
 	fe.Answers[flow.StageIdentification] = username
-	fe.Answers[flow.StagePassword] = rfc2865.UserPassword_GetString(r.Packet)
-	if r.pi.MFASupport {
-		fe.CheckPasswordInlineMFA()
-	} else {
-		// If code-based MFA is disabled StageAuthenticatorValidate answer is set to StagePassword answer.
-		// This allows flows with only a mfa stage
-		fe.Answers[flow.StageAuthenticatorValidate] = fe.Answers[flow.StagePassword]
-	}
+	fe.SetSecrets(rfc2865.UserPassword_GetString(r.Packet), r.pi.MFASupport)
 
 	passed, err := fe.Execute()
 	if err != nil {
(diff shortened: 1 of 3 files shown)


internal/outpost/radius/handle_access_request.go
```diff
@@ -24,6 +24,10 @@ func (rs *RadiusServer) Handle_AccessRequest(w radius.ResponseWriter, r *RadiusR
 	fe.Answers[flow.StagePassword] = rfc2865.UserPassword_GetString(r.Packet)
 	if r.pi.MFASupport {
 		fe.CheckPasswordInlineMFA()
+	} else {
+		// If code-based MFA is disabled StageAuthenticatorValidate answer is set to StagePassword answer.
+		// This allows flows with only a mfa stage
+		fe.Answers[flow.StageAuthenticatorValidate] = fe.Answers[flow.StagePassword]
```
BeryJu, line 30: I think this should either be changed for LDAP and Radius, and/or maybe it should be an option to configure...although I don't think this would be a breaking change to the current behaviour

sirux88: No this isn't a breaking change at all but it improves usability of the radius outpost.

I see three options to get this done better:
1) Implement this as an option (as you suggested) for both `LDAP` and `Radius` providers. 
But I think this would confuse users because of the already implemented code-based MFA option
2) Analyze the connected flow and fill answers accordingly. I'm not quite sure if this can be done at all and if it would cover all possible flows
3) Refactor [CheckPasswordInlineMFA](https://github.com/goauthentik/authentik/blob/0e1646ca1b651b7b0eeca02a62afa7620d9d690f/internal/outpost/flow/solvers_mfa.go#L15C33-L15C33) to something like `SetupFlowAnswers (string password, bool mfa)` and use it for both outposts. 

In my eyes we should go for option 3) but I leave the descision up to you @BeryJu

sirux88:
I implemented `SetSecrets` function for radius and ldap outposts @BeryJu

sirux88:
Is there I can help you with or provide to get this merged @tanberry ?

tanberry:
Hi @sirux88, in reviewing the build fails I found a tiny typo (you can just approve my suggestion and it will be added to your PR), and then @BeryJu is going to give it another technical review. Thanks, we'll get it merged soon!

sirux88:
> Hi @sirux88, in reviewing the build fails I found a tiny typo (you can just approve my suggestion and it will be added to your PR), and then @BeryJu is going to give it another technical review. Thanks, we'll get it merged soon!

I can't find your suggestion. Did I miss something?
The only typo I could find is "acccording" with an "c" to much. 
Is this what you are talking about @tanberry ? If so I'll fix it.

BeryJu (approved):

PR #8047, opened 2024-01-02
website/integrations: add openproject doc

benszau:
Added integration documentation for [openproject](https://www.openproject.org/).
---

## Checklist

If applicable

### DIFF
--- a/website/integrations/services/openproject/index.md
+++ b/website/integrations/services/openproject/index.md
@@ -1,0 +1,91 @@
+---
+title: OpenProject
+---
+
+<span class="badge badge--secondary">Support level: Community</span>
+
+## What is OpenProject
+
+> OpenProject is a open source project management software.
+>
+> -- https://www.openproject.org/
+
+## Preparation
+
+The following placeholders will be used:
+
+-   `openproject.company` is the FQDN of the Service install. (Remove this for SaaS)
+-   `authentik.company` is the FQDN of the authentik install.
+
+## authentik Configuration
+
+Create a [OAuth2/OpenID provider](https://goauthentik.io/docs/providers/oauth2) with the following parameters:
+
+-   Client Type: `Confidential`
+-   Scopes: `openid`, `email` and `profile`
+-   Signing Key: Select any available key
+-   Redirect URIs: `https://openproject.company/auth/authentik/callback`
+
+Note the `Client ID` and `Client Secret` values.
+Create an [application](https://goauthentik.io/docs/applications), using the provider you've created above and set a slug.
+In this example and the following configuration files, the slug `openproject` is used.
+
+### Add family name
+OpenProject uses `First name` and `Last name` but Authentik does only provide a name by default (e.g. name="foo bar" instead first_name="foo", last_name="bar"), you can modify the `authentik default OAuth Mapping: OpenID 'profile'` to provide first and last name.
+To do that, you need to:
+-   log in as `admin`
+-   open the `admin interface`
+-   navigate to `Customisation` -> `Property Mappings`
+-   uncheck `Hide managed mappings`
+-   edit the `authentik default OAuth Mapping: OpenID 'profile'` mapping
+- add the following lines:
+   ```
+    "family_name": request.user.name.rsplit(" ", 1)[-1],
+    "given_name": request.user.name.rsplit(" ", 1)[0],
+   ```
+
+Now, the fields for first and last name will get properly set in OpenProject.
+
+
+## OpenProject Configuration
+
+OpenProject can be installed in different ways (see the [documentation](https://www.openproject.org/docs/installation-and-operations/installation/)).
+For this configuration, the [docker-based installation using docker-compose](https://www.openproject.org/docs/installation-and-operations/configuration/#docker) is used.
+
+
+As described in the [installation guide](https://www.openproject.org/docs/installation-and-operations/installation/docker/#quick-start), the first step is to clone the [openproject-deploy repository](https://github.com/opf/openproject-deploy/tree/stable/13/compose).
+Following the [instructions of the openproject-deploy repository](https://github.com/opf/openproject-deploy/tree/stable/13/compose#openproject-installation-with-docker-compose), create a copy of the provided `.env.example` and adjust its content:
+-   Set `OPENPROJECT_HOST__NAME` to `openproject.company`
+
+The next step is to add some more configuration lines to the `.env` file:
+
+```
+# sso auth
+OPENPROJECT_OMNIAUTH__DIRECT__LOGIN__PROVIDER="Authentik"
+
+# The name of the login button in OpenProject, you can freely set this to anything you like
+OPENPROJECT_OPENID__CONNECT_AUTHENTIK_DISPLAY__NAME="Authentik"
+OPENPROJECT_OPENID__CONNECT_AUTHENTIK_HOST="authentik.company"
+OPENPROJECT_OPENID__CONNECT_AUTHENTIK_IDENTIFIER="< insert the `Client ID` you have copied in the authentik configuration step >"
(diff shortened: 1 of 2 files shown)


website/integrations/services/openproject/index.md
```diff
@@ -0,0 +1,91 @@
...
+## Preparation
+
+The following placeholders will be used:
+
+-   `openproject.company` is the FQDN of the Service install. (Remove this for SaaS)
+-   `authentik.company` is the FQDN of the authentik install.
+
+## authentik Configuration
+
+Create a [OAuth2/OpenID provider](https://goauthentik.io/docs/providers/oauth2) with the following parameters:
```
rissson, line 22:
```suggestion
Create a OAuth2/OpenID provider with the following parameters:
```
We usually don't like like this, but I'm sure @tanberry will be able to provide some more useful feedback

website/integrations/services/openproject/index.md
```diff
@@ -0,0 +1,91 @@
...
+
+-   Client Type: `Confidential`
+-   Scopes: `openid`, `email` and `profile`
+-   Signing Key: Select any available key
+-   Redirect URIs: `https://openproject.company/auth/authentik/callback`
+
+Note the `Client ID` and `Client Secret` values.
+Create an [application](https://goauthentik.io/docs/applications), using the provider you've created above and set a slug.
```
rissson, line 30:
```suggestion
Create an application, using the provider you've created above and set a slug.
```

website/integrations/services/openproject/index.md
```diff
@@ -0,0 +1,91 @@
...
+-   navigate to `Customisation` -> `Property Mappings`
+-   uncheck `Hide managed mappings`
+-   edit the `authentik default OAuth Mapping: OpenID 'profile'` mapping
+- add the following lines:
+   ```
+    "family_name": request.user.name.rsplit(" ", 1)[-1],
+    "given_name": request.user.name.rsplit(" ", 1)[0],
+   ```
+
+Now, the fields for first and last name will get properly set in OpenProject.
```
rissson, line 47: In another life, I setup an OpenProject instance, and we would just shove the whoole name in first name, and leave last name empty. It did work a treat, but we can leave the example like this.

rissson (changes requested):
Heya, thanks for your contribution! A few nitpicks, but all in all looks very good.

IIRC, we also setup an LDAP outpost with our OP instance, so that users would be provisioned straight away, and didn't need to have logged in prior to them being assigned to tasks. I don't remember whether that requires a paying version though. If you end up setting this up, I'd love for it to be documented as well!

website/integrations/services/openproject/index.md
```diff
@@ -0,0 +1,91 @@
+---
+title: OpenProject
+---
+
+<span class="badge badge--secondary">Support level: Community</span>
+
+## What is OpenProject
+
+> OpenProject is a open source project management software.
```
tanberry, line 9: Ideally, there would be a few more sentences here, perhaps a use case, how people typically use this software.

website/integrations/services/openproject/index.md
tanberry: True, we don't usually link to the docs like this from an integration-specific doc, but in this case I think it is fine. The next one, linking to our Applications docs, does not work as well...
So let's leave this link in, but next one, please.

website/integrations/services/openproject/index.md
tanberry: The reason we don't want to link to the Applications docs (right now) is because the docs do not yet have steps for how to *create* an Application, and that is what this line 30 is about. We are about to update our Applications docs to include procedurals for using the new Wizard to create apps/providers. 

So let's remove this link, please.

website/integrations/services/openproject/index.md
```diff
@@ -0,0 +1,91 @@
...
+-   Scopes: `openid`, `email` and `profile`
+-   Signing Key: Select any available key
+-   Redirect URIs: `https://openproject.company/auth/authentik/callback`
+
+Note the `Client ID` and `Client Secret` values.
+Create an [application](https://goauthentik.io/docs/applications), using the provider you've created above and set a slug.
+In this example and the following configuration files, the slug `openproject` is used.
+
+### Add family name
+OpenProject uses `First name` and `Last name` but Authentik does only provide a name by default (e.g. name="foo bar" instead first_name="foo", last_name="bar"), you can modify the `authentik default OAuth Mapping: OpenID 'profile'` to provide first and last name.
```
tanberry, line 34:
```suggestion
OpenProject uses `First name` and `Last name` but authentik only provides a name by default (e.g. name="foo bar" instead first_name="foo", last_name="bar"), you can modify the `authentik default OAuth Mapping: OpenID 'profile'` to provide first and last name.
```

website/integrations/services/openproject/index.md
```diff
@@ -0,0 +1,91 @@
...
+To do that, you need to:
+-   log in as `admin`
+-   open the `admin interface`
```
tanberry, line 37:
```suggestion
-   open the Admin interface
```

website/integrations/services/openproject/index.md
```diff
@@ -0,0 +1,91 @@
...
+-   navigate to `Customisation` -> `Property Mappings`
```
tanberry, line 38:
```suggestion
-   navigate to **Customization -> Property Mappings**
```

tanberry: For UI elements, we use bold text. Code format is used for commands, file names, and other system-related nouns.

website/integrations/services/openproject/index.md
```diff
@@ -0,0 +1,91 @@
...
+-   uncheck `Hide managed mappings`
```
tanberry, line 39:
```suggestion
-   uncheck **Hide managed mappings**
```

website/integrations/services/openproject/index.md
```diff
@@ -0,0 +1,91 @@
...
+    "given_name": request.user.name.rsplit(" ", 1)[0],
+   ```
+
+Now, the fields for first and last name will get properly set in OpenProject.
+
+
+## OpenProject Configuration
+
+OpenProject can be installed in different ways (see the [documentation](https://www.openproject.org/docs/installation-and-operations/installation/)).
+For this configuration, the [docker-based installation using docker-compose](https://www.openproject.org/docs/installation-and-operations/configuration/#docker) is used.
```
tanberry, line 53:
```suggestion
For this configuration, the [docker-based installation using Docker Compose](https://www.openproject.org/docs/installation-and-operations/configuration/#docker) is used.
```

website/integrations/services/openproject/index.md
```diff
@@ -0,0 +1,91 @@
...
+
+
+As described in the [installation guide](https://www.openproject.org/docs/installation-and-operations/installation/docker/#quick-start), the first step is to clone the [openproject-deploy repository](https://github.com/opf/openproject-deploy/tree/stable/13/compose).
+Following the [instructions of the openproject-deploy repository](https://github.com/opf/openproject-deploy/tree/stable/13/compose#openproject-installation-with-docker-compose), create a copy of the provided `.env.example` and adjust its content:
+-   Set `OPENPROJECT_HOST__NAME` to `openproject.company`
+
```
tanberry, line 59: @benszau Do you think it would be good to add a note that they should rename the sample file `.env.example` to be just `.env" or will users know that?

tanberry (changes requested):
@benszau huge apologies for slow review on my part... Thank you so much for this contribution! I made a couple of edits, and one question. We look forward to merging this as soon as you address those few points, and thanks again!

website/integrations/services/openproject/index.md
tanberry: Ooops, and now we do have instructions! ;-) So let's leave the link, but edit it to point to https://docs.goauthentik.io/docs/applications/manage_apps.

@benszau I hope it is ok with you if we make the changes and go ahead and merge. Or let us know if you want to do so, or add more content, etc.

PR #8110, opened 2024-01-09
website/integrations: fix Grafana role mapping docs

cornfeedhobo:
## Details

This commit removes the redundant property mapping and updates the role_attribute_path examples accordingly.

### DIFF
--- a/website/integrations/services/grafana/index.mdx
+++ b/website/integrations/services/grafana/index.mdx
@@ -17,29 +17,20 @@ The following placeholders will be used:
 -   `grafana.company` is the FQDN of the Grafana install.
 -   `authentik.company` is the FQDN of the authentik install.
 
-Create an application in authentik. Create an OAuth2/OpenID provider with the following parameters:
+Create an OAuth2/OpenID provider with the following parameters:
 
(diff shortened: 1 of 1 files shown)


website/integrations/services/grafana/index.mdx
```diff
@@ -56,16 +47,6 @@ data "authentik_scope_mapping" "scope-openid" {
   name = "authentik default OAuth Mapping: OpenID 'openid'"
 }
 
-resource "authentik_scope_mapping" "scope-grafana-roles" {
```
rissson, line 59: This should also be removed below in the `property_mappings` list.

rissson (approved):
Looks pretty good overall, a very-much needed cleanup indeed.

rissson (changes requested):
Whoops, wrong button

website/integrations/services/grafana/index.mdx
cornfeedhobo: I think you clicked the wrong line, but I think I understood what you meant, line 86.

rissson (approved):

rissson:
Thank you for your contribution!

PR #8126, opened 2024-01-10
root: add name to docker-compose

raiford:
## Details

This gives the docker compose file a service name.  Some docker compose translation tools require this such as `compose2nix`.

---

## Checklist

If applicable

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,4 +1,5 @@
 ---
+name: authentik
 version: "3.4"
 
 services:


rissson:
Heya! Thanks for your contribution, I can't find anything in the compose spec about the `name` top-level key. Do you know if it has any implication on how docker-compose behaves?

BeryJu:
I was looking into the same thing and it seems the spec allows so called extensions even on a top level of the file, however they are recommended to have an x- prefix if I understand correctly

raiford:
Oh interesting.  I think I have seen this on a couple of other recent docker-compose files that I imported and I just assumed it was more common.  I did find this doc: https://docs.docker.com/compose/compose-file/04-version-and-name/

rissson:
After testing manually, this does indeed change the docker-compose behavior. It has the same effect as passing the `-p` flag to docker-compose commands. Given that that would be a breaking change, I'm not particularly fond of making this change. You can however give the `-project` option to compose2nix to achieve the same behavior.

raiford:
Yep, I found the `-project` flag and started using it but thought I'd send this PR in case y'all wanted this.

PR #8138, opened 2024-01-12
website/docs: remove references to legacy Docker Compose v1

Daniel15:
## Details

Docker Compose v2 was released in 2020. v1 was deprecated in 2021 and no longer receives any updates as of July 2023. v2 uses `docker compose` rather than `docker-compose`.

I updated all references *except* for those in older releases.

---

## Checklist

If applicable

### DIFF
--- a/website/developer-docs/setup/frontend-dev-environment.md
+++ b/website/developer-docs/setup/frontend-dev-environment.md
@@ -45,7 +45,7 @@ Depending on platform, some native dependencies might be required. On macOS, run
 
     This makes the local web files and the config file available to the authentik server.
 
-5. Run `docker-compose up -d` to apply those changes to your containers.
+5. Run `docker compose up -d` to apply those changes to your containers.
 6. `cd web`
 7. Run `npm i` and then `npm run watch` to start the build process.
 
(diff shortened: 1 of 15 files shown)


BeryJu (commented):
Thanks! Does changing the commands to compose v2 do the correct name migration (i.e. authentik_server_1 to authentik-server-1)?

tanberry:
As @BeryJu said above, Thanks @Daniel15 ! This contribution is very appreciated. We'll wait to hear back on @BeryJu 's question before merging, but looks good and again thank you.

Daniel15:
> Does changing the commands to compose v2 do the correct name migration (i.e. authentik_server_1 to authentik-server-1)?

That's my understanding. I updated a Mailcow server from docker-compose v1 to v2 around a year ago and didn't have any issues around naming. docker-compose relies on a label (`com.docker.compose.project`) rather than container naming to determine the containers associated with a docker-compose project, so `docker-compose up -d` using v1 then `docker compose down` using v2 should work fine too.

tanberry (approved):
Thanks so much @Daniel15 for digging through all of the places in the docs to update this command! Based on your reply to @BeryJu I am going to go ahead and merge this; sorry for the delay. We appreciate your help.

PR #8149, opened 2024-01-12
web: add RAC Provider to the list of providers understood by the wizard

kensternberg-authentik:
## Details

This commit adds the RAC Provider to the list of providers understood by the Application Wizard.

It also creates a new, simple alert that knows how to look up the enterprise requirements and chooses to fill itself in with a notice saying "A license is required for this provider," or nothing.  That harmonizes the display across both wizards, and reduces the demands on the wizards themselves to "know" about enterprise features.

## Checklist

    -   [X] No new text was added

If applicable


web/src/admin/applications/wizard/methods/rac/ak-application-wizard-authentication-for-rac.ts
```diff
@@ -0,0 +1,119 @@
...
+                                        >
+                                            ${mapping.name}
+                                        </option>`,
+                                )}
+                            </select>
+                            <p class="pf-c-form__helper-text">
+                                ${msg("Hold control/command to select multiple items.")}
+                            </p>
+                        </ak-form-element-horizontal>
+                        <ak-form-element-horizontal label=${msg("Settings")} name="settings">
```
BeryJu, line 105: We could probably not show this here? since ux wise it's much easier to set these settings via a mapping

BeryJu (approved):

web/src/admin/applications/wizard/methods/rac/ak-application-wizard-authentication-for-rac.ts
BeryJu: Only here is fine

PR #8150, opened 2024-01-12
website/docs: Blog for the new year

tanberry:
Happy new year blog


kensternberg-authentik (approved):
I don't actually see anything in here that stands out at me as needing revision.  It's straightforward and clean, and it communicates well.  I know it's more vibes and less technical, but that's fine; I usually get itchy when someone is trying to communicate with me ("Communication is when what you read or hear changes what you will do"), and this doesn't.  I mean, it's obviously necessary, but other than "We had a good year," it doesn't move me to take any action, so I don't see many problems with it in this form.

PR #8153, opened 2024-01-12
web: provide a context for checking the status of the enterprise license

kensternberg-authentik:
## Details

There are a few places (currently 5) in our code where we have checks for the current enterprise licensing status of our product. While not particularly heavy or onerous, there's no reason to repeat those same lines, and since our UI is always running in the context of authentik, may as well make that status a client-side context in its own right. The status will update with an EVENT_REFRESH request.

A context-aware custom alert has also been provided; it draws itself (or `nothing`) depending on the state of the license, and the default message, "This feature requires an enterprise license," can be overriden with the `notice` property.

These two changes reduce the amount of code needed to manage our license alerting from 67 to 38 lines code, and while removing 29 lines from a product with 54,145 lines of code (a savings of 0.05%, oh boy!) isn't a miracle, it does mean there's a single source of truth for "Is this instance enterprise-licensed?" that's easy to access and use.

The internationalization files (including pseudolocale!) have been updated.

The text, "Provider require enterprise," has been changed to the default message.  It was ungrammatical and, when used on the PropertyMappings page, didn't make sense in context.

---

## Checklist

If applicable


web/src/elements/AuthentikContexts.ts
```diff
@@ -1,9 +1,13 @@
 import { createContext } from "@lit-labs/context";
```
kensternberg-authentik, line 1: **1**: Add the context name.

web/src/elements/Interface/Interface.ts
```diff
@@ -1,7 +1,9 @@
+import { DEFAULT_CONFIG } from "@goauthentik/common/api/config";
```
kensternberg-authentik, line 1: **2**: Retrieve and provide the context, `_licenseSummary`.

web/src/elements/Interface/licenseSummaryProvider.ts
```diff
@@ -0,0 +1,25 @@
+import { authentikEnterpriseContext } from "@goauthentik/elements/AuthentikContexts";
```
kensternberg-authentik, line 1: **3**: Make the context easily accessible to others, including a specialized accessor for the most commonly accessed field.

web/src/admin/common/ak-license-notice.ts
```diff
@@ -0,0 +1,24 @@
+import "@goauthentik/elements/Alert";
```
kensternberg-authentik, line 1: **4**: Provide a notification mechanism for this particular setting.

After this, all the other changes are swapping out the ad-hoc request mechanisms to use the context instead.

kensternberg-authentik (commented):
A guided tour has been provided!

BeryJu (approved):

PR #8177, opened 2024-01-15
enterprise: add full audit log [AUTH-458]

BeryJu:
## Details

Audit logging that shows which fields were changed

needs:

- [x] Testing
- [ ] more sanitization
- [x] potentially replacing deepdiff with a simpler diff
- [x] Some way to ignore certain model fields being changed (`last_login` and `password` on user for example since they have dedicated events (`login` and `password_set`))

---

## Checklist

If applicable


authentik/enterprise/apps.py
```diff
@@ -32,3 +32,9 @@ def check_enabled(self):
         from authentik.enterprise.models import LicenseKey
 
         return LicenseKey.get_total().is_valid()
+
+    def reconcile_install_middleware(self):
```
rissson, line 36:
```suggestion
    def reconcile_global_install_middleware(self):
```

authentik/enterprise/audit/apps.py
```diff
@@ -0,0 +1,19 @@
...
+
+class AuthentikEnterpriseAuditConfig(EnterpriseConfig):
+    """Enterprise app config"""
+
+    name = "authentik.enterprise.audit"
+    label = "authentik_enterprise_audit"
+    verbose_name = "authentik Enterprise.Audit"
+    default = True
+
+    def reconcile_install_middleware(self):
```
rissson, line 15:
```suggestion
    def reconcile_global_install_middleware(self):
```

authentik/enterprise/apps.py
rissson: Also, I don't think this should be here, it only needs to be in the enterprise_audit app

authentik/enterprise/audit/apps.py
```diff
@@ -0,0 +1,19 @@
...
+from django.conf import settings
+
+from authentik.enterprise.apps import EnterpriseConfig
+
+
+class AuthentikEnterpriseAuditConfig(EnterpriseConfig):
+    """Enterprise app config"""
+
+    name = "authentik.enterprise.audit"
+    label = "authentik_enterprise_audit"
```
rissson, line 11: Did we want to stick with a label w/o `enterprise` in it in the event we move this to the OSS version?

authentik/enterprise/apps.py
```diff
@@ -32,3 +32,9 @@ def check_enabled(self):
         from authentik.enterprise.models import LicenseKey
```
rissson, line 32: So above this I copied the `lru_cache` in the tenancy PR, but we should remove it

rissson (changes requested):

authentik/enterprise/audit/apps.py
rissson: By always install the middleware, I meant adding it to settings.py (or apps.py which should overload it with the logic in settings.py)

BeryJu: adding it to middleware in settings doesn't overwridte the main middleware list but rather just appends, so we have to do this here

rissson: Ah yes, indeed, I missed that

rissson (approved):

PR #8244, opened 2024-01-20
bug-fix for azure ad

roney492:


BeryJu:
This is already fixed in the version-2023.10 branch (check out the `ghcr.io/goauthentik/dev-server:gh-version-2023.10` image), and these changes will add a backwards incompatible change

roney492:
@BeryJu , May i know which branch is used for docker images?

BeryJu:
the branch version-2023.10, but this will also be in a bug fix release next week

PR #8278, opened 2024-01-23
web: clean up UserInterface in prep for OAuth and Silo Projects

kensternberg-authentik:
While looking through @BeryJu's Oauth-for-authentik stuff, changes to `UserInterface` triggered the "harder" eslint pass, which said "UserInterface exceeds permitted complexity (9)." I couldn't disagree; it had lots of conditionals.

This commit:

- Changes no functionality; it's just cleanup.
- Breaks UserInterface into business and presentation layers
- The presentation layer:
  - Further breaks the presentation into a frame and conditional components. Each condition is now proceeded by a simple guard clause.
  - Taps into the event listener set-up for toggles, eliminating their local scope/window duplication
  - Extracts in-line complex expressions into isolated and scoped functions to name them and make them easier to find and read.
  - Extracts the custom CSS into its own named variable, again, making it easier to find and read.
- The business layer:
  - Builds the window-level event listeners at connection, and disconnects them correctly, allowing this whole interface to be used in a SPA. (Not _as_ a SPA, but _in_ a SPA!)
  - Asserts a reliable contract at the presentation layer; there should be no question "Session" and "UIConfig" are available before rendering.
  - Renames `firstUpdated` to `fetchConfigurationDetails`, and calls it in the constructor. There ought to be no circumstances where this object is constructed outside a working environment; no sense in waiting until you've done a `render() { nothing }` pass to fetch details.

Oddities: There are a pair of `<!-- -->` HTML comments in the framing `render()`; those are there just to stop prettier from slamming a string of conditional renders all into one line, making them harder to read.

---

## Checklist

-   [N] Local tests pass (`ak test authentik/`)

-   [N] The API schema has been updated (`make gen-build`)

-   [N] The translation files have been updated (`make i18n-extract`)

If applicable

-   [N] The documentation has been updated
-   [N] The documentation has been formatted (`make website`)


web/src/user/UserInterface.ts
```diff
@@ -67,118 +124,62 @@ export class UserInterface extends Interface {
...
+    @property({ type: Boolean, reflect: true })
+    apiDrawerOpen = false;
+
+    @property({ type: Number })
+    notificationsCount = 0;
+
+    @property({ type: Object })
+    brand!: CurrentBrand;
+
+    get userDisplayName() {
```
kensternberg-authentik, line 149: Turns this into a switch _expression_, so Typescript can type it and assert conditional exhaustion correctly!

web/src/user/UserInterface.ts
```diff
@@ -192,85 +193,18 @@ export class UserInterface extends Interface {
...
-                                          href="#/settings"
-                                      >
-                                          <pf-tooltip position="top" content=${msg("Settings")}>
-                                              <i class="fas fa-cog" aria-hidden="true"></i>
-                                          </pf-tooltip>
-                                      </a>
-                                  </div>`
-                                : html``}
+                            ${this.renderApiDrawer()}
+                            <!-- -->
```
kensternberg-authentik, line 204: These are here just to space things out; otherwise, `prettier` wants to string all these together into a single line, which was harder to read.

web/src/user/UserInterface.ts
```diff
@@ -364,4 +273,202 @@ export class UserInterface extends Interface {
             </div>
         </ak-locale-context>`;
     }
+
+    renderApiDrawer() {
+        if (!this.uiConfig.enabledFeatures.apiDrawer) {
+            return nothing;
+        }
+
+        const onClick = (ev: Event) => {
```
kensternberg-authentik, line 282: The current version has _two_ handlers for EVENT_API_DRAWER_TOGGLE, one attached to `UserInterface`, another attached to `window`.  No sense in keeping both; this eliminates one of them.

web/src/user/UserInterface.ts
```diff
@@ -364,4 +273,202 @@ export class UserInterface extends Interface {
...
+
+    @state()
+    me?: SessionUser;
+
+    constructor() {
+        super();
+        this.ws = new WebsocketClient();
+        this.fetchConfigurationDetails();
+        configureSentry(true);
+        this.toggleNotificationDrawer = this.toggleNotificationDrawer.bind(this);
```
kensternberg-authentik, line 407: All this binding is annoying, but necessary; it allows us to connect and disconnect these handlers from `window` at need.  Which would also allow us, if we wanted, to switch between "User" and "Admin" apps without actually leaving the page or reloading the environment, since we're now cleaning up after ourselves.

web/src/user/UserInterface.ts
```diff
@@ -364,4 +273,202 @@ export class UserInterface extends Interface {
...
+
+    get isFullyConfigured() {
+        return !!(this.uiConfig && this.me);
+    }
+
+    render() {
+        if (!this.isFullyConfigured) {
+            return nothing;
+        }
+
```
kensternberg-authentik, line 464: I like this because it makes it very clear what the presentation responds to: here are the state variables that will trigger a re-render.

web/src/user/UserInterface.ts
```diff
@@ -192,85 +193,18 @@ export class UserInterface extends Interface {
...
                     <div class="pf-c-page__header-tools">
                         <div class="pf-c-page__header-tools-group">
-                            ${this.uiConfig.enabledFeatures.apiDrawer
-                                ? html`<div
-                                      class="pf-c-page__header-tools-item pf-m-hidden pf-m-visible-on-lg"
-                                  >
-                                      <button
-                                          class="pf-c-button pf-m-plain"
-                                          type="button"
-                                          @click=${() => {
```
kensternberg-authentik, line 209: This duplicates the functionality of the `window.EVENT_API_DRAWER_TOGGLE` functionality above, so replacing this with "send the `EVENTS_API_DRAWER_TOGGLE` event to `window`" eliminates that duplication.

kensternberg-authentik (commented):
Added the guided tour.

web/src/user/UserInterface.ts
BeryJu: would probably rename this to `renderApiDrawerTrigger` since it doesn't render the actual drawer

web/src/user/UserInterface.ts
```diff
@@ -364,4 +269,202 @@ export class UserInterface extends Interface {
...
+        return html`<div class="pf-c-page__header-tools-item pf-m-hidden pf-m-visible-on-lg">
+            <button class="pf-c-button pf-m-plain" type="button" @click=${onClick}>
+                <pf-tooltip position="top" content=${msg("Open API drawer")}>
+                    <i class="fas fa-code" aria-hidden="true"></i>
+                </pf-tooltip>
+            </button>
+        </div>`;
+    }
+
+    renderNotificationDrawer() {
```
BeryJu, line 294: Same as above

web/src/user/UserInterface.ts
```diff
@@ -364,4 +269,202 @@ export class UserInterface extends Interface {
...
+        return html` <div class="pf-c-page__header-tools-item">
+            <a class="pf-c-button pf-m-plain" type="button" href="#/settings">
+                <pf-tooltip position="top" content=${msg("Settings")}>
+                    <i class="fas fa-cog" aria-hidden="true"></i>
+                </pf-tooltip>
+            </a>
+        </div>`;
+    }
+
+    renderAccessAdmin() {
```
BeryJu, line 341:
```suggestion
    renderAdminInterfaceLink() {
```

web/src/user/UserInterface.ts
BeryJu: Switching between all these interfaces without a full reload would definitely be very nice to have

BeryJu (approved):
LGTM

PR #8299, opened 2024-01-25
web/admin: revamped rbac and user settings tabs

### LINKED ISSUE #6760: UI Elements Overlap (Directory > Users)
**Describe the bug**
UI Elements Overlap Under the Directory > Users Menu when you interact with the "Hide deactivated user toggle switch"

**To Reproduce**

1. Log into authentik
2. Click on 'Admin interface'
3. Click on 'Directory' to expand
4. Click on 'Users'
5. Click on 'Hide deactivated user' (in either windowed browser or full screen)
6. See blue ring around 'Hide deactivated user' toggle switch overlap with other UI elements

**Expected behavior**
no UI overlap

**Screenshots**
When browser (Chrome) is fullscreen:
image

When browser (Chrome) is windowed:
image

**Logs**
N/A

-   authentik version: 2023.8.2

**Additional context**
N/A

### LINKED ISSUE #7675: Button to clear all notifications at the bottom of application list
**Describe the bug**
When the list of applications needs to be scrolled to the bottom, the button to clear all notifications appears at the bottom of that list, instead of at the bottom of the notification container

**To Reproduce**

1. Have enough application groups to require the user to scroll down to see them all
2. Click on the notification icon
3. Scroll down to the bottom of the page
4. See that the clear all button is at the bottom.

**Expected behavior**
I'd expect to see the clear all button to be at the bottom of the notification container, and not the bottom of the page

**Screenshots**
What I see:
image
image

What I'd expect to see:
image

**Logs**
N/A

-   authentik version: 2023.10.4

**Additional context**

BeryJu:
## Details

A whole load of minor-ish fixes

- rename role permission tables to match user tables
- use separate tabs
- rework role page to match user page
- show all of users permission tabs on one main tab
- move stuff into vertical subtab
- fix some leftover html?
- switch non-embedded permission page to use vertical tabs
- web/admin: fix duplicate RBAC preview banner on permission modal

closes #7675
closes #6760

---

## Checklist

If applicable


web/src/admin/events/EventListPage.ts
```diff
@@ -82,9 +83,9 @@ export class EventListPage extends TablePage<Event> {
             html`<div>${actionToLabel(item.action)}</div>
                 <small>${item.app}</small>`,
             EventUser(item),
-            html`<span>${item.created?.toLocaleString()}</span>`,
+            html`<div>${getRelativeTime(item.created)}</div>
```
kensternberg-authentik, line 86: Nifty.  I tend to reverse these: "Dec 10, 2023 (23 days ago)", but I think the human brain appreciates having a "distance from here / distance from now" along with a location or date.

web/src/admin/roles/RoleViewPage.ts
```diff
@@ -83,73 +81,61 @@ export class RoleViewPage extends AKElement {
...
+                <div class="pf-l-grid pf-m-gutter">
+                    <div
+                        class="pf-c-card pf-l-grid__item pf-m-12-col pf-m-3-col-on-xl pf-m-3-col-on-2xl"
+                    >
+                        <div class="pf-c-card__title">${msg("Role Info")}</div>
+                        <div class="pf-c-card__body">
+                            ${renderDescriptionList([
+                                [msg("Name"), this._role.name],
+                                [msg("Edit"), this.renderUpdateControl(this._role)],
+                            ])}
```
kensternberg-authentik, line 115: Nice, small description list renderer.  :-)

web/src/elements/rbac/ObjectPermissionsPage.ts
```diff
@@ -76,4 +86,78 @@ export class ObjectPermissionPage extends AKElement {
                 </section>
             </ak-tabs>`;
     }
+
+    renderCoreUser() {
```
kensternberg-authentik, line 90: @BeryJu broke out `renderControlButtons` above in the UserView page, and it's a good practice; break out big pieces of functionality, especially when they're rendered conditionally, and name them.

kensternberg-authentik (approved):
I approve your changes, but I made two changes myself, so I'll need your assent.  :-)

web/src/admin/roles/RoleViewPage.ts
BeryJu: can we turn this into a component? Currently when using this we'll also have to manually import the descriptor CSS

web/src/admin/roles/RoleViewPage.ts
```diff
@@ -71,7 +72,7 @@ export class RoleViewPage extends AKElement {
         });
     }
 
-    render(): TemplateResult {
+    render() {
```
BeryJu, line 75: any reason for removing the type annotations?

web/src/admin/events/EventListPage.ts
tanberry: My brain is quite happy to get that relative info.

web/src/admin/roles/RoleViewPage.ts
kensternberg-authentik: We can, but you already have in existing cases, so I went with a lazier solution.  I do intend to create a `ak-description-list` component someday.  This is a compromise; I think it's a good one that doesn't introduce any new semantics or CSS shadow boundary issues, and it puts markers in the ground for where to target our description lists in the future.

web/src/elements/rbac/ObjectPermissionsPage.ts
kensternberg-authentik: Two reasons, actually: the first is that [Typescript generally recommends against it](https://www.typescriptlang.org/docs/handbook/2/functions.html), as inference of return type is very good and the only reason to do it is if you don't trust the function body itself to generate the correct return type, and the other is that Lit's re-render algorithm is smarter with `nothing` rather than `html\`\`` , but doesn't conform to TemplateResult without a lot of fussiness.

web/src/admin/roles/RoleViewPage.ts
kensternberg-authentik: Two reasons, actually: the first is that [Typescript generally recommends against it](https://www.typescriptlang.org/docs/handbook/2/functions.html), as inference of return type is very good and the only reason to do it is if you don't trust the function body itself to generate the correct return type, and the other is that Lit's re-render algorithm is smarter with `nothing` rather than `html\`\`` , but doesn't conform to TemplateResult without a lot of fussiness.

PR #8377, opened 2024-01-31
sources/ldap: add ability to disable password write on login

### LINKED ISSUE #6122: Suppress authentik password cache for LDAP sources
Some users have strict requirements where and how password are stored.
For LDAP sources (not sure about other external sources) Authentik however writes the passwords in its own (hashed) format back to internal tables. This potentially reduced password security and in the case that always binding to LDAP is enabled also does not offer any performance benefits regardless.

An option to suppress writing the password to authentik tables for external sources (LDAP).

Because we liked authentik so far during an internal testing phase I currently consider adding a Postgres trigger to prevent any writebacks to the password column of the user table (except for the akadmin user). This is obviously fragile, hacky, and is not possible for setups with both local and external users.

**Additional context**
[Discord discussion](https://discord.com/channels/809154715984199690/1062045018367467531)

rissson:
## Details

Closes #6122

---

## Checklist

If applicable


authentik/sources/ldap/models.py
```diff
@@ -99,6 +99,11 @@ class LDAPSource(Source):
         help_text=_("Property mappings used for group creation/updating."),
     )
 
+    password_login_update_internal_password = models.BooleanField(
```
rissson, line 102: Maybe we need a better name for this

septatrix: I called this `update_local_user_password_after_bind` in my local branch^^

rissson: Yeah I'm trying to use a name that will work for both LDAP and kerberos

authentik/sources/ldap/models.py
```diff
@@ -99,6 +99,11 @@ class LDAPSource(Source):
...
+        default=True,
```
septatrix, line 103: Personally I would advocate for having this default to `False` as having this active implies some non-obvious side effects and pitfalls. This should be a conscious opt-in with careful considerations and only once people experience performance bottlenecks without it.

Implications of having this enabled:
- Updating the LDAP password does not invalidate the local password, this means that old password are still valid indefinitely
- Logging in via LDAP credentials overwrites the local password if users have different pwds for both
- The security measures made to secure the passwords (salting, peppering, hash/PBKDF function) in Authentik might be different from the LDAP backend. Users might not be aware of this...

rissson: I understand your point, but on the other hand that would mean breaking a behavior that existed for 4 years

septatrix: Then the migration should be configured such that existing sources keep the existing behavior but new ones will have this disabled

rissson: There we go.

septatrix (approved):
Perfect. I have no strong opinion about the variable name. They are not exposed to the user so a more technical name using the phrasing of the source (i.e. bind for ldap) would also be fine for me but keeping it consistent across sources which also will use this could also be appealing.

web/src/admin/sources/ldap/LDAPSourceForm.ts
```diff
@@ -86,6 +86,28 @@ export class LDAPSourceForm extends BaseSourceForm<LDAPSource> {
                     <span class="pf-c-switch__label">${msg("Enabled")}</span>
                 </label>
             </ak-form-element-horizontal>
+            <ak-form-element-horizontal name="passwordLoginUpdateInternalPassword">
+                <label class="pf-c-switch">
+                    <input
+                        class="pf-c-switch__input"
+                        type="checkbox"
+                        ?checked=${first(this.instance?.passwordLoginUpdateInternalPassword, true)}
```
septatrix, line 94: I am not sure what `first` does but I guess the last parameter should then also be `false` to default to this being unset for new sources?

authentik/sources/ldap/models.py
```diff
@@ -99,6 +99,11 @@ class LDAPSource(Source):
         help_text=_("Property mappings used for group creation/updating."),
     )
 
+    password_login_update_internal_password = models.BooleanField(
+        default=False,
```
BeryJu, line 103: Should probably be set to `True` by default to retain the same behaviour as the current version

rissson: See the other discussion about this. The migrations set it to true, and then back to false so that existing behavior is preserved and new sources will get false

BeryJu: I think we should still keep it to true for the new sources.

rissson: Changed back to have the default to true in the model, the migrations and the frontend.

septatrix: Could you please explain your reasoning for wanting this to be true? LDAP binds are generally fast enough do not outweigh the several pitfalls and negative effects on security. You can see my arguments above

authentik/sources/ldap/models.py
septatrix: Sync seems to be the wrong word here as this is not done regularly but only on successful binds. If one is happy to use different terms for this mechanism in LDAP and kerberos I think that something like `update_{local,internal}_{user,password}_after_{bind,login}` is probably most clear as to what it does.

septatrix (changes requested):
I know I am in no position to make request but I would *really* like to have my concerns of having this enabled by default regarding usability, pitfalls and security addressed, please...

BeryJu:
Yeah I suppose after going through this again we can probably set it to False for new sources as there isn't really a strong case for leaving it enabled by default (the only thing being a future built in scheduled HIBP check, which could just check for this

We should probably also add all of the points from https://github.com/goauthentik/authentik/pull/8377#discussion_r1473313128 to the documentation, so we can update them as they are resolved (especially the first one)

rissson:
> Yeah I suppose after going through this again we can probably set it to False for new sources as there isn't really a strong case for leaving it enabled by default 

Changed it back again.

septatrix:
> Yeah I suppose after going through this again we can probably set it to False for new sources as there isn't really a strong case for leaving it enabled by default (the only thing being a future built in scheduled HIBP check, which could just check for this
> 
> We should probably also add all of the points from [#8377 (comment)](https://github.com/goauthentik/authentik/pull/8377#discussion_r1473313128) to the documentation, so we can update them as they are resolved (especially the first one)

I agree in both aspects. For now this should default to false for newly created sources and the drawbacks of enabling should be documented such that one can weight the ups and downs.

Regarding the HIBP check: How would that even work with stored password (hashed and salted)? This seems like it would only work upon logging in (when we have access to the raw password) or if the leaked database used exactly the same hash function (including number of rounds etc) and salt?

Regarding my first point (password update) this should already work for sources which have a `pwdLastSet` property. Though sadly this is not really documented anywhere. Also this would again conflict if users might have a different password set in Authentik themselves.

At some point I was thinking if it would make more sense to drop this from the providers and instead make this into a flow step. I.e. upon successful login one would issue a user update and set the received password. This would likely make the updating more clear, reduce code duplication and offer more customizability (e.g. skip the update of admins where the redundant password store and delay until a password gets invalidated might be considered a security risk).

authentik/sources/ldap/models.py
rissson: we could go with `update_internal_password_after_login` indeed

rissson:
Documentation updated to document this feature and associated security risks.

septatrix (approved):

septatrix:
Any outlook on getting this merged soon?

BeryJu (approved):

PR #8435, opened 2024-02-06
events: fix SystemTask timestamps and scheduling

BeryJu:
## Details

Turns out `default_timer` does indeed not return a timestamp but a more relative time, so we use a real timestamp for start/end but still track the duration with `default_timer`

This also fixes a couple issues with prefilling, and also system tasks on schedules sometimes not running depending on when instances of authentik start/stop

---

## Checklist

If applicable


smusali (approved):

authentik/events/models.py
```diff
@@ -620,8 +620,9 @@ class SystemTask(SerializerModel, ExpiringModel):
     name = models.TextField()
     uid = models.TextField(null=True)
 
-    start_timestamp = models.FloatField()
-    finish_timestamp = models.FloatField()
+    start_timestamp = models.DateTimeField()
+    finish_timestamp = models.DateTimeField()
+    duration = models.FloatField()
```
rissson, line 625: I really want to play with https://docs.djangoproject.com/en/5.0/ref/models/fields/#django.db.models.GeneratedField for this, but we'd loose precision :/

rissson (approved):

authentik/events/models.py
BeryJu: TIL about that field, I wonder if we could use that for `User.uid`

PR #8483, opened 2024-02-12
blueprints: use reconcile decorator instead of relying on function name prefix

### LINKED ISSUE #7852: blueprints: find another way to run reconcile than with dir()
Now looking at it again I think there's probably a better way to do this than using `dir()` and then `getattr()`, probably something using a decorator?

Pseudo code:

```
@self.reconcile
def do_something(self):
    pass
```

(Not really part of this PR tbh)

_Originally posted by @BeryJu in https://github.com/goauthentik/authentik/pull/7590#discussion_r1422312556_

BeryJu:
## Details

closes #7852
---

## Checklist

If applicable

### DIFF
--- a/authentik/blueprints/apps.py
+++ b/authentik/blueprints/apps.py
@@ -2,6 +2,7 @@
 
 from importlib import import_module
 from inspect import ismethod
+from typing import Callable
 
 from django.apps import AppConfig
 from django.db import DatabaseError, InternalError, ProgrammingError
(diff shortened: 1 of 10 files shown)


authentik/blueprints/apps.py
```diff
@@ -61,7 +63,19 @@ def _reconcile(self, prefix: str) -> None:
             except (DatabaseError, ProgrammingError, InternalError) as exc:
                 self.logger.warning("Failed to run reconcile", name=name, exc=exc)
 
-    def reconcile_tenant(self) -> None:
+    @staticmethod
+    def reconcile_tenant(func: Callable):
+        """Mark a function to be called on startup (for each tenant)"""
+        setattr(func, "_authentik_managed_reconcile", ManagedAppConfig.RECONCILE_TENANT_CATEGORY)
```
rissson, line 69: If there are many reconcile functions for this category, would one get overridden?

BeryJu: Yeah you can only have one category per reconcile function but that also matches the current behaviour

rissson: Fair enough, I didn't look into the code too much

rissson: And if needed we can always turn it into a list in the future

rissson (approved):
I'll approve, but let's merge after 2024.2

PR #8498, opened 2024-02-13
root: early spring clean for linting

BeryJu:
## Details

- Remove pylint since it keeps throwing false positives at random in CI pipelines requiring yet more `pylint: disable=` comments on everything
- Remove pyright since we had to disable quite a few checks and lock it to an older version to make it not complain about django and celery
- Instead use ruff for everything (keeping black and isort for now...?)

---

## Checklist

If applicable


rissson:
Also, I removed isort and replaced it with ruff as well.

PR #8579, opened 2024-02-19
web: move authentik to a sub-repository

kensternberg-authentik:
This commit moves the build of the web into a sub-repository and refactors the build instructions to understand that move.  The simple result at this time will be just a move and a re-arrangement of the build paths so that the final product is still deposited in the `./dist` folder.

Only a few "functional" changes have been made:

- The core package has been renamed '@goauthentik/authentik' to distinguish itself from the root package name '@goauthentik/web'.
- The codespell paths have been updated
- The path to the imported markdown documentation has been updated.
- TSConfig.json has been split into a "base" and "build" pair. This is necessary because our future builds will use a slightly different configuration from our linters, as the former do not need the typing information (their job is to strip it away, leaving us with pure JavaScript), and the latter very definitely need it to make our IDE's work correctly.
- Rollup.config.js has also been split into a "base" and "build" pair. The base has been slightly refactored, with heavily concretized paths for the project ROOT and project DIST folders. This informs the moved authentik project where to put the built release.

The NPM -> Lage relationship was done mechanically: I copied our targets from the existing package.json into lage, then used some regex magic to tell lage "call all these targets in our packages," of which we have exactly ONE right now.

I also had to fix yet another xliff consistency error.  *Sigh*.

## Checklist

If applicable


kensternberg-authentik:
This was part of the monorepo experiment, and is no longer needed.

PR #8699, opened 2024-02-26
web: replace rollup with esbuild

kensternberg-authentik:
## Details

This commit replaces rollup with esbuild.

The biggest fix was to alter the way CSS is imported into our system; esbuild delivers it to the browser as text, rather than as a bundle with metadata that, frankly, we never use. ESBuild will bundle the CSS for us just fine, and interpreting those strings *as* CSS turned out to be a small hurdle. Code has been added to AKElement and Interface to ensure that all CSS referenced by an element has been converted to a Browser CSSStyleSheet before being presented to the browser.

A similar fix has been provided for the markdown imports. The biggest headache there was that the re-arrangement of our documentation broke Jen’s existing parser for fixing relative links. I’ve provided a corresponding hack that provides the necessary detail, but since the Markdown is being presented to the browser as text, we have to provide a hint in the markdown component for where any relative links should go, and we’re importing and processing the markdown at runtime. This doesn’t seem to be a big performance hit.

The entire build process is driven by the new build script, `build.mjs`, which starts the esbuild process as a service connected to the build script and then runs the commands sent to it as fast as possible. The biggest “hack” in it is actually the replacement for rollup’s `rollup-copy-plugin`, which is clever enough I’m surprised it doesn’t exist as a standalone file-copy package in its own right.

I’ve also used a filesystem watch library to encode a “watcher” mechanism into the build script. `node build.mjs --watch` will work on MacOS; I haven’t tested it elsewhere, at least not yet.

`node build.mjs --proxy` does what the old rollup.proxy.js script did.

The savings are substantial. It takes less than two seconds to build the whole UI, a huge savings off the older ~45-50 seconds I routinely saw on my old Mac. It’s also about 9% smaller.

The trade-offs appear to be small: processing the CSS as StyleSheets, and the Markdown as HTML, at run-time is a small performance hit, but I didn’t notice it in amongst everything else the UI does as it starts up.

Manual chunking is gone; esbuild’s support for that is quite difficult to get right compared to Rollup’s, although there’s been a bit of yelling at ESbuild over it. Codemirror is built into its own chunk; it’s just not *named* distinctly anymore.

The one thing I haven’t been able to test yet is whether or not the polyfills and runtim shims work as expected on older browsers.

**Additional changes**

`build-locales`: To avoid rebuilding the `.ts` files from the locales collection on every pass, the build-locales script has been wrapped in a script that checks if the code is up-to-date already.  If it is, `build-locales` will not be re-run.  This script will also fail (in a build-crashing way) if an XLiff file required by the `lit-localize.config` file is missing.

`check-spelling`: The command line for this was incredibly long and visually cluttering in package.json, so I've wrapped it in a convenient script, one that can automatically find the dictionaries, using the `git rev-parse` trick.

`precommit`: Again, a very long command line, and a complex ad-hoc Unix script; I've wrapped this one in a script that has better documentation for the class of files we'll be checking with SonarJS along with the usual ESlint rules.

`pseudolocalize` and `import-maps`: These were in TypeScript, but for our purposes that's one extra bit of code that we didn't need to execute at build time.  Helper scripts are going to be in modular JavaScript (`.mjs`) for the foreseeable future, as they don't have or provide dependencies on our actual product.

## Checklist

If applicable


BeryJu (approved):
🎉 Finally a quick build

@kensternberg-authentik I had to re-add some rollup dependencies to make storyboard work, not sure if we can (or even want to) switch that to esbuild

PR #8727, opened 2024-02-28
root: fix container build

BeryJu:
## Details

Due to some change in poetry (I assume https://github.com/python-poetry/poetry/releases/tag/1.8.0?) it was no longer enough to call poetry from a python interpreter from a virtualenv, but requires activating the virtualenv

---

## Checklist

If applicable

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -104,9 +104,10 @@ RUN --mount=type=bind,target=./pyproject.toml,src=./pyproject.toml \
     --mount=type=cache,target=/root/.cache/pip \
     --mount=type=cache,target=/root/.cache/pypoetry \
     python -m venv /ak-root/venv/ && \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -106,7 +106,7 @@ RUN --mount=type=bind,target=./pyproject.toml,src=./pyproject.toml \
     python -m venv /ak-root/venv/ && \
     pip3 install --upgrade pip && \
     pip3 install poetry && \
-    poetry install --only=main --no-ansi --no-interaction
+    bash -c "source ${VENV_PATH}/bin/activate && poetry install --only=main --no-ansi --no-interaction --no-root"
```
rissson, line 109: The source should probably be done before pip3 invocations as well?

PR #8775, opened 2024-03-01
root: ensure consistent install_id

BeryJu:
## Details

Even though we use an advisory lock to prevent parallel migrations, it can happen that multiple install_id entries are created. This itself doesn't cause any issues, however this PR ensures that always the same install_id value will be returned

---

## Checklist

If applicable

### DIFF
--- a/authentik/root/install_id.py
+++ b/authentik/root/install_id.py
@@ -7,6 +7,8 @@ from psycopg import connect
 
 from authentik.lib.config import CONFIG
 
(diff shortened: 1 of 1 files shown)


rissson:
Will this cause issues for existing deployments if somehow two install_ids exist?

rissson (approved):

BeryJu:
> Will this cause issues for existing deployments if somehow two install_ids exist?

It might, but it'll only bring up issues that would've come up later anyways (since their install ID might change later in a non-determinate way)

PR #8799, opened 2024-03-04
Source flow manager property mappings UI revisions

kensternberg-authentik:
web: streamline new property mappings UI features

There are four major changes here between the source-flow-manager-property-mappings branch @rissson put up and this one:

- A BasePropertyMappingForm already exists; the form containing the CodeMirror stuff was duplicated across five source types, so I’ve put it into the parent class instead. This eliminated all of the duplicate imports as well.
- The complexity of the `OAuthSourceForm.send()` method exceeded the maximum specified in eslint. I’ve refactored it and isolated elements that weren’t needed into named and enclosed functions to drop it below that maximum.
- It’s no longer necessary to import `config` and fetch it; your capabilities are known when you log in and are kept in the global context. If the parent class is marked `WithCapabilitiesConfig` (or if you add that yourself), then the class has a method `can(capability: CapabilitiesEnum): boolean`.
- The idiom Jens uses to trigger “When this property is set, fetch this data and set it in a state of its own to trigger a new refresh,” is not recommended. The section [Creating custom property accessors](https://lit.dev/docs/v2/components/properties/#accessors-custom) says so explicitly, recommending instead that you monitor when a property is set using the [`willUpdate()`](https://lit.dev/docs/v2/components/lifecycle/#willupdate) method, and if the property you’re monitoring changes trigger a new fetch. Let Lit do its thing about refreshing on a state change. I’ve made that change, with named fetchers, here.

(I also updated it to use the latest `main`, so there are some other changes.)

## Checklist

If applicable


tanberry (approved):
From docs perspective looks good.

kensternberg-authentik:
It looks like most of the recommendations have been taken to heart anyway, so I'm pulling this.

PR #8812, opened 2024-03-05
stages/email: Disable autoescape for text templates

Chasethechicken:
Some Flows send links via e-mail. \
Those links have to be properly escaped to be embedded in the HTML-e-mail, e.g. like this: \
`https://sso.fachschaften.org/if/flow/auth/?next=%2F&amp;flow_token=<token>` \
But Authentik also sends the content in plaintext. In that case the link should not be html-escaped: \
`https://sso.fachschaften.org/if/flow/auth/?next=%2F&flow_token=<token>`

This currently this does not happen, so when users copy the link from the plaintext-mail, it does not work, because it contains the html-escaped link.

We solved this by adding Django's `{% autoescape off %}` tag to all `.txt`-templates.

An alternative could be to disable autoescape in code, that way custom text templates would automatically no longer be escaped.

---

## Checklist

If applicable

### DIFF
--- a/authentik/stages/email/templates/email/account_confirmation.txt
+++ b/authentik/stages/email/templates/email/account_confirmation.txt
@@ -4,6 +4,6 @@
 
 {{ url }}
 
---
+-- 
 Powered by goauthentik.io.
 {% endautoescape %}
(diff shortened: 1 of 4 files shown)


BeryJu (commented):
I'm kind of torn between setting autoescape to off and using `{{ url|safe }}`, but I suppose both of them work fine

Also could you re-add the trailing space for the `--`, this was added to conform to https://datatracker.ietf.org/doc/html/rfc3676#section-4.3

Chasethechicken:
Thank you for your feedback. I put the trailing whitespace my editor removed back.

I think turning off autoescape is more intuitive, because we wouldn't want to escape anything else in plaintext e-mails, correct?

PR #8856, opened 2024-03-08
web: fix esbuild issue with style sheets

kensternberg-authentik:
Getting ESBuild, Lit, and Storybook to all agree on how to read and parse stylesheets is a serious pain. This fix better identifies the value types (instances) being passed from various sources in the repo to the three *different* kinds of style processors we're using (the native one, the polyfill one, and whatever the heck Storybook does internally).

Falling back to using older CSS instantiating techniques one era at a time seems to do the trick. It's ugly, but in the face of the aggressive styling we use to avoid Flashes of Unstyled Content (FLoUC), it's the logic with which we're left.

In standard mode, the following warning appears on the console when running a Flow:

```
Autofocus processing was blocked because a document already has a focused element.
```

In compatibility mode, the following **error** appears on the console when running a Flow:

```
crawler-inject.js:1106 Uncaught TypeError: Failed to execute 'observe' on 'MutationObserver': parameter 1 is not of type 'Node'.
    at initDomMutationObservers (crawler-inject.js:1106:18)
    at crawler-inject.js:1114:24
    at Array.forEach (<anonymous>)
    at initDomMutationObservers (crawler-inject.js:1114:10)
    at crawler-inject.js:1549:1
initDomMutationObservers @ crawler-inject.js:1106
(anonymous) @ crawler-inject.js:1114
initDomMutationObservers @ crawler-inject.js:1114
(anonymous) @ crawler-inject.js:1549
```

Despite this error, nothing seems to be broken and flows work as anticipated.

## Details

REPLACE ME

---

## Checklist

If applicable

### DIFF
--- a/web/package-lock.json
+++ b/web/package-lock.json
@@ -17059,6 +17059,7 @@
         },
         "node_modules/tree-sitter-json": {
             "version": "0.20.2",
+            "hasInstallScript": true,
             "license": "MIT",
             "optional": true,
             "dependencies": {
@@ -17067,6 +17068,7 @@
         },
         "node_modules/tree-sitter-yaml": {
             "version": "0.5.0",
+            "hasInstallScript": true,
             "license": "MIT",
             "optional": true,
             "dependencies": {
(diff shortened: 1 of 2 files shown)


BeryJu (approved):

BeryJu (commented):
I have noticed one issue, when debug is enabled and the sentry spotlight thingy is enabled, you get this 
image

PR #8858, opened 2024-03-09
admin: Handle latest  version unknown in admin dashboard

### LINKED ISSUE #8857: False version status in admin dashboard for AirGapped environments
**Describe the bug**
In an AirGapped environment with `AUTHENTIK_DISABLE_UPDATE_CHECK=true`, or when the version check has not yet been performed, the version tile on the admin dashboard will always state its `Up-to-date!` which may not actually be the case.

**To Reproduce**

1. Deploy a fresh Authentik installation on an older version (e.g. `2023.10.7`) and ensure  `AUTHENTIK_DISABLE_UPDATE_CHECK=true` is set.
2. Go to the admin dashboard
3. Observe the version tile stating proudly that its up-to-date.

**Expected behavior**
Not incorrectly stating its up-to-date as that would give a false sense of security.

**Screenshots**
image

**Logs**
N/a

-   authentik version: 2024.2.0

**Additional context**
N/a

RoboMagus:
## Details

This PR adds an additional check to the latest version to see if it is a valid version and not a dummy default value.
This is used to remove the false statement that Authentik is 'Up-to-date!' in environments where the update check is disabled.

Closes #8857

---

## Checklist

If applicable

### DIFF
--- a/authentik/admin/api/version.py
+++ b/authentik/admin/api/version.py
@@ -10,7 +10,7 @@ from rest_framework.response import Response
 from rest_framework.views import APIView
 
 from authentik import __version__, get_build_hash
-from authentik.admin.tasks import VERSION_CACHE_KEY, update_latest_version
+from authentik.admin.tasks import VERSION_CACHE_KEY, VERSION_NULL, update_latest_version
 from authentik.core.api.utils import PassiveSerializer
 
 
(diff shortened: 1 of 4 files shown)


rissson:
Could we not just check for `v0.0.0` in the web UI instead of adding a new field to the API?

RoboMagus:
> Could we not just check for v0.0.0 in the web UI instead of adding a new field to the API?

That's also an option IMO, but I wouldn't like to add any magic values that are used in more than one place.

BeryJu (approved):

PR #8888, opened 2024-03-12
Update _envoy_istio.md

valkenburg-prevue-ch:
## Details
Added a comment about allowing the http authorization headers to upstream, necessary in an istio meshConfig if there are proxy providers which inject http basic auth headers.

---

## Checklist

If applicable

### DIFF
--- a/website/docs/providers/proxy/_envoy_istio.md
+++ b/website/docs/providers/proxy/_envoy_istio.md
@@ -20,6 +20,9 @@ spec:
                   headersToUpstreamOnAllow:
                       - set-cookie
                       - x-authentik-*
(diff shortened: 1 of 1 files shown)


tanberry (commented):
Thank you @valkenburg-prevue-ch , we appreciate this contribution. 

Question: if users need to add the auth header, they will need to uncomment line 25, correct?

valkenburg-prevue-ch:
> Question: if users need to add the auth header, they will need to uncomment line 25, correct?

That is correct. I wasn't sure how to best present that. Please let me know if there is a better way to communicate that.

tanberry (approved):
I think the way you have it is good, I just wanted to double-check. Thanks again for adding this important detail!

PR #8906, opened 2024-03-14
website/docs: update example ldapsearch commands

Trident101:
## Details

Updated the ldapsearch command to no longer use the deprecated -h -p options.

---

## Checklist

If applicable

### DIFF
--- a/website/docs/providers/ldap/generic_setup.md
+++ b/website/docs/providers/ldap/generic_setup.md
@@ -84,8 +84,7 @@ sudo yum install openldap-clients -y # CentOS-based systems
 ```
 ldapsearch \
   -x \
-  -h <LDAP Outpost IP address> \
-  -p 389 \ # Production should use SSL 636
+  -H ldap://<LDAP Outpost IP address>:<Portnumber 389/636, production should use SSL 636> \
   -D 'cn=ldapservice,ou=users,DC=ldap,DC=goauthentik,DC=io' \
   -w '<ldapuserpassword>' \
   -b 'DC=ldap,DC=goauthentik,DC=io' \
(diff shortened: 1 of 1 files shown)


website/docs/providers/ldap/generic_setup.md
```diff
@@ -84,8 +84,7 @@ sudo yum install openldap-clients -y # CentOS-based systems
 ```
 ldapsearch \
   -x \
-  -h <LDAP Outpost IP address> \
-  -p 389 \ # Production should use SSL 636
+  -H ldap://<LDAP Outpost IP address>:<Portnumber 389/636, production should use SSL 636> \
```
BeryJu, line 87: (I don't remember how this is treated with the old parameters)

When using SSL the URI schema needs to be `ldaps://`

BeryJu (changes requested):

website/docs/providers/ldap/generic_setup.md
Trident101: I've tried using the commands with the -h -p parameters, but it simply failed due to not knowing those parameters (anymore) as they're deprecated. How and if SSL is handled with the old parameters, i don't know. I left out the ldaps;// part as (in my eyes) this is just for a quick test. I can add it if it's preferred to be documented.

BeryJu:
```suggestion
  -H ldap://<LDAP Outpost IP address>:<Port number 389> \ # In production it is recommended to use SSL, which also requires `ldaps://` as the protocol and the SSL port 
```

BeryJu (approved):

PR #8993, opened 2024-03-21
website/docs: improve flow inspector docs

BeryJu:
## Details

REPLACE ME

---

## Checklist

If applicable

### DIFF
--- a/website/docs/flow/inspector.md
+++ b/website/docs/flow/inspector.md
@@ -5,8 +5,13 @@ title: Inspector
 The flow inspector, introduced in 2021.10, allows administrators to easily figure out how custom flows work, inspect the current context and debug issues.
 
 :::info
-When running a flow with the inspector enabled, the flow is still executed normally. This means that for example, a `user_write` stage will _actually_ write user data.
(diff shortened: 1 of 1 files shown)


website/docs/flow/inspector.md
```diff
@@ -5,9 +5,14 @@ title: Inspector
 The flow inspector, introduced in 2021.10, allows administrators to easily figure out how custom flows work, inspect the current context and debug issues.
 
 :::info
-When running a flow with the inspector enabled, the flow is still executed normally. This means that for example, a `user_write` stage will _actually_ write user data.
+When running a flow with the inspector enabled, the flow is still executed normally. This means that for example, a [User write](../flow/stages/user_write.md) stage will _actually_ write user data.
 :::
 
+### Accessing the inspector
```
rissson, line 11: So should we allow users with the debug attribute as well?

BeryJu: it should probably be an RBAC permission instead of just flat-out superuser, but the reason why this works the way it does is because you need to be logged in a super user to then start the flow with inspector as otherwise the user wouldn't be authenticated

tanberry (approved):
Thank you, much better!

PR #9044, opened 2024-03-27
web: a few minor bugfixes and lintfixes

kensternberg-authentik:
## Details

web: just a few minor bugfixes and lintfixes

While investigating the viability of using ESLint 9, I found a few bugs.

The one major bug was found in the error handling code, where a comparison was automatically invalid and would never realize “true.”

A sequence used in our Storybook support code to generate unique IDs for applications and providers had an annoying ambiguity:

    new Array(length).fill(" ")

Lint states (and I agree):

> It’s not clear whether the argument is meant to be the length of the array or the only element. If the argument is the array’s length, consider using `Array.from({ length: n })`. If the argument is the only element, use `[element]`.”

It’s the former, and I intended as much.

Aside from those, a few over-wrought uses of the spread operator were removed.

---

## Checklist


web/src/common/errors.ts
```diff
@@ -16,7 +16,7 @@ export async function parseAPIError(error: Error): Promise<APIErrorTypes> {
     if (!(error instanceof ResponseError)) {
         return error;
     }
-    if (error.response.status < 400 && error.response.status > 499) {
+    if (error.response.status < 400 || error.response.status > 499) {
```
kensternberg-authentik, line 19: "Less than 400 AND greater than 499" will never resolve.  Oops!

web/src/elements/buttons/TokenCopyButton/ak-token-copy-button.stories.ts
```diff
@@ -12,7 +12,7 @@ import AKTokenCopyButton from "./ak-token-copy-button";
 
 function makeid(length: number) {
     const sample = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789";
-    return new Array(length)
+    return Array.from({ length })
```
kensternberg-authentik, line 15: Lint:

> It's not clear whether the argument is meant to be the length of the array or the only element. If the argument is the array's length, consider using `Array.from({ length: n })`. If the argument is the only element, use `[element]`."

BeryJu (approved):

PR #9048, opened 2024-03-27
website/docs: email stage: fix example translation error

schnadoslin:
fixed configuration_error 
django.template.exceptions.TemplateSyntaxError: 'blocktrans' doesn't allow other block tags (seen 'blocktrans') inside it

👋 Hi there! 

## Details
website/docs : flow/stages/email)

Given template on website was faulty.


BeryJu (commented):
I don't see why this change is needed as within the `blocktrans` block there is no other block?

tanberry:
True, @BeryJu it's not really needed as the block is now, but allows for future changes. Let's merge it. Thanks @schnadoslin for this; at very least I learned a bit of Django. ;-)

tanberry (approved):

PR #9119, opened 2024-04-03
providers/proxy: avoid erroring on logout with session_id is None

rissson:
## Details

https://id.goauthentik.io/if/admin/#/events/log/b8be48b8-ea3d-46f0-9a10-45f0d460e972

---

## Checklist

If applicable

### DIFF
--- a/authentik/providers/proxy/tasks.py
+++ b/authentik/providers/proxy/tasks.py
@@ -23,8 +23,10 @@ def proxy_set_defaults():
 
 
 @CELERY_APP.task()
-def proxy_on_logout(session_id: str):
+def proxy_on_logout(session_id: str | None):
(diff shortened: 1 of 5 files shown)


authentik/providers/proxy/tasks.py
```diff
@@ -23,8 +23,10 @@ def proxy_set_defaults():
 
 
 @CELERY_APP.task()
-def proxy_on_logout(session_id: str):
+def proxy_on_logout(session_id: str | None):
```
BeryJu, line 26: The reason why this wasn't handled is because it should never happen, I'll need to do some digging to see the reason this happens

BeryJu (approved):

PR #9138, opened 2024-04-04
web: ak-checkbox-group for short, static, multi-select events

kensternberg-authentik:
web: ak-checkbox-group for short, static, multi-select events

## Details

Implements a checkbox groups web component, wholly independent of the API (although it does implement the ‘data-ak-control’ protocol, including the `json()` method that makes it easier to send the data to the Form handler). The controller works much like multi-select: `value` returns an array of strings, the `name` attribute associated with whatever it is you’re asking about.

The `required` property only works if you give the whole item a name, as if it were an input. Otherwise, it does nothing.

Giving it a `name` also activates the browser standard `formAssociated` protocol; it works just fine for ordinary HTML forms, and presents to that protocol the `FormValue` type, so any form using it will automagically convert it into the CGI (Common Gateway Interface) format of, to use the example from Storybook:

    ak-test-checkgroup-input=funky&ak-test-checkgroup-input=invalid

Note that the classic CGI format is not automatically key/value; keys can appear multiple times, and indicate that the value is an array of strings. Most modern appservers understand this format. Some do not.

There’s a full and complete JSDOC-like comment documenting the component. I have even provided CSSPart sections for everything: the wrapper, each line, the input and its associated label. The brave or foolhardy can mangle the CSS to their hearts’ content without having to know a thing about Patternfly.

image

## Checklist


web/package.json
```diff
@@ -14,8 +14,8 @@
         "build": "run-s build-locales esbuild:build",
         "build-proxy": "run-s build-locales esbuild:build-proxy",
         "watch": "run-s build-locales esbuild:watch",
-        "lint": "cross-env NODE_OPTIONS='--max_old_space_size=8192' eslint . --max-warnings 0 --fix",
-        "lint:precommit": "cross-env NODE_OPTIONS='--max_old_space_size=8192' node scripts/eslint-precommit.mjs",
+        "lint": "cross-env NODE_OPTIONS='--max_old_space_size=16384' eslint . --max-warnings 0 --fix",
```
BeryJu, line 17: maybe we should just migrate to https://github.com/biomejs/biome/ at some point

kensternberg-authentik: I would agitate for Oxlint.  Or maybe _both_; Oxlint makes an excellent (and very fast) pre-test.

web/src/elements/ak-checkbox-group/ak-checkbox-group.ts
```diff
@@ -0,0 +1,212 @@
...
+
+@customElement("ak-checkbox-group")
+export class CheckboxGroup extends AkElementWithCustomEvents {
+    static get styles() {
+        return [
+            PFBase,
+            PFForm,
+            PFCheck,
+            css`
+                .pf-c-form__group-control {
```
BeryJu, line 87: Taken from https://github.com/goauthentik/authentik/blob/main/web/src/elements/forms/Radio.ts#L40, idk if we want to dedupe that @kensternberg-authentik

BeryJu (approved):

web/src/elements/ak-checkbox-group/ak-checkbox-group.ts
kensternberg-authentik: Not yet, I don't think.  Good catch.  Thanks so much!

PR #9154, opened 2024-04-05
website/docs: add more info and links about enforciing unique email addresses

tanberry:
This is a good first step in making it easier for readers to find info about unique email addresses, I think.

I ended up not moving the entire topic, but instead adding a link to it on the "Manage users" page. Since the topic is really about a policy (that happens to be about user emails), I think it is best to keep it under policies.

I don't want to move it until we look at all the options, and maybe use an <import> function to have it in several places. Because this [Examples page](https://docs.goauthentik.io/docs/flow/examples/snippets) also is a fine place to mention it.

### DIFF
--- a/website/docs/user-group-role/user/user_basic_operations.md
+++ b/website/docs/user-group-role/user/user_basic_operations.md
@@ -4,6 +4,8 @@ title: Manage users
 
 The following topics are for the basic management of users: how to create, modify, delete or deactivate users, and using a recovery email.
 
+[Policies](../../policies/index.md) can be used to further manage how users are authenticated. For example, in authentik email addresses are (by default) not required to be unique, but you can use a policy to [enforce unique email addresses](../../policies/working_with_policies/unique_email.md).
+
 ### Create a user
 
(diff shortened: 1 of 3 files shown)


website/docs/user-group-role/user/user_basic_operations.md
```diff
@@ -4,6 +4,8 @@ title: Manage users
 
 The following topics are for the basic management of users: how to create, modify, delete or deactivate users, and using a recovery email.
 
+[Policies](../../policies/index.md) can be used to further manage how users are authenticated. For example, in authentik email addresses are (by default) not required to be unique, but you can use a policy to [enforce unique email addresses](../../policies/working_with_policies/unique_email.md).
```
kensternberg-authentik, line 7: "in authentik email addresses" looked like one whole phrase about "authentik email addresses."  Suggestion: "For example, by default authentik does not require email addresses be unique, but you can use a policy to ..."  

Our lack of capitalization of our name is a partial weakness, as it doesn't set us off as a proper noun the way we're trained to read. It sometimes takes a second read to tease apart the implied meaning.

kensternberg-authentik (approved):
One suggestion. Otherwise, it looks good.

PR #9200, opened 2024-04-09
web: preserve selected list when provider updates

kensternberg-authentik:
web: preserve selected list when provider updates

## Details

The impulse to preserve the functionality of the system given a change of provider was… admirable, but unnecessary in this case. A premature optimization that doesn’t make a difference. Observations:

1.  change from the client will bring a new `selected`. But changes from the outside shouldn’t happen once the interactive experience is “settled.”
2.  the client is perfectly capable of listening to the `change` event and reading the content of the value list for selecteds. If the client is going to change the provider, it should provide the most up-to-date copy of selecteds as well.
3.  We set the selecteds from two locations: from the client on start-up, and from the “selected” pane during user interaction. Anything more is risk. I shouldn’t have taken that risk.

## Checklist


web/src/elements/ak-dual-select/ak-dual-select-provider.ts
```diff
@@ -72,16 +70,14 @@ export class AkDualSelectProvider extends CustomListenerElement(AKElement) {
 
     willUpdate(changedProperties: PropertyValues<this>) {
         if (changedProperties.has("searchDelay")) {
-            this.doSearch = debounce(this.doSearch.bind(this), this.searchDelay);
+            this.doSearch = debounce(
```
kensternberg-authentik, line 73: Although it's unlikely the searchDelay will change multiple times during the course of *one* instance of this component's lifecycle, it's still bad practice to create the possibility of multiple wrappers.  Always start from the prototype in these cases!

web/src/elements/ak-dual-select/ak-dual-select-provider.ts
```diff
@@ -72,16 +70,14 @@ export class AkDualSelectProvider extends CustomListenerElement(AKElement) {
...
+                AkDualSelectProvider.prototype.doSearch.bind(this),
+                this.searchDelay,
+            );
         }
 
         if (changedProperties.has("provider")) {
             this.pagination = undefined;
-            const previousProvider = changedProperties.get("provider");
```
kensternberg-authentik, line 80: _Sigh_.  "Premature optimization is the root of all evil."

BeryJu (approved):

PR #9272, opened 2024-04-15
events: add indexes

BeryJu:
## Details

REPLACE ME

---

## Checklist

If applicable


authentik/events/models.py
```diff
@@ -305,6 +305,16 @@ def __str__(self) -> str:
     class Meta:
         verbose_name = _("Event")
         verbose_name_plural = _("Events")
+        indexes = [
+            models.Index(fields=["action"]),
+            models.Index(fields=["user"]),
```
rissson, line 310: Would user__username be enough?

rissson (approved):

authentik/events/models.py
BeryJu: I thought about that too, but we might want to filter on user.pk too so might as well index the entire field

rissson: Oh yeah I didn't notice that it was only pk, username and email in there.

Issue #4191, opened 2022-12-09
Token endpoint returning wrong value for expires_in (from refresh token instead of access token)


pawel-pacewicz-altium:
**Bug description**
Authentik endpoint /application/o/token/ seems to be returning the value of a refresh token "expires_in" instead of the access token's.
The application using authentik's SSO capabilities is relying on "expires_at" to refresh the access token proactively. Because information received is incorrect, it won't try to refresh until the token is expired and fail on a request when the access token did expire.

**steps to reproduce:**
0. (Set-up) Set Access token validity to 6m, refresh token to 14d.
1. Request new token by querying the URL (can be done from the browser): `https://authentik/if/flow/default-provider-authorization-explicit-consent/?response_type=code&client_id=5e6f645d1836e5f696b31fc136a5dde1f734f73f&redirect_uri=_REDIRECT_URI_&scope=SCOPE`
2. Copy the code
3. Request token through Postman using the code above
4. Result: `"expires_in": 1209600`  (e.g. 14d).
![image](https://user-images.githubusercontent.com/108667766/203544197-4f107fc9-9329-4a32-a803-8d321d8f4751.png)

Confirmed that the token itself also contains wrong information:
![image](https://user-images.githubusercontent.com/108667766/203545124-586e4656-57b8-4d37-ab07-053e14ac2979.png)

**Expected behavior**
The Access token should return a validity of 360s for validity set to 6m.

**Version and Deployment (please complete the following information):**
 - authentik version 2022.11.3
 - Deployment: docker

**Additional context**
This was claimed to be resolved under commit #3306003, but the issue still persists with the latest update of Authentik - confirmed as of Dec 9th, 2022.

pawel-pacewicz-altium:
Hi,
Any chance for an update as to a potential resolution plan?
Thanks in advance! :)

BeryJu:
So there's some confusion coming from the fact the UI setting is labeled `Access token validity` when it controls the value of `accessCodeValidity`, and the value of that is accordingly used for the OAuth Authorization code instead of access/refresh token.

Now this does highlight the bigger issue in this case, being that both access and refresh tokens share the same database model and currently share the same expiry, which of course defeats the point of a refresh token. To fix this, a larger refactor of the OAuth provider is required, but that is in progress.

lsjostro:
[...]

PR #9275, opened 2024-04-15
providers/oauth2: fix refresh_token grant returning incorrect id_token

BeryJu:
## Details

#4191

---

## Checklist

If applicable

### DIFF
--- a/authentik/providers/oauth2/tests/test_token.py
+++ b/authentik/providers/oauth2/tests/test_token.py
@@ -208,7 +208,7 @@ class TestToken(OAuthTestCase):
                 "token_type": TOKEN_TYPE,
                 "expires_in": 3600,
                 "id_token": provider.encode(
-                    refresh.id_token.to_dict(),
+                    access.id_token.to_dict(),
                 ),
             },
         )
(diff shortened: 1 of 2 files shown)


lsjostro:
Can we add proper integration tests for this?

BeryJu:
@lsjostro do you have any suggestion on a client application that could be used for this? The OpenID compliance test didn't complain about this

I'll still merge this as-is for the time being so it can be included in 2024.4

lsjostro:
I'm more searching for why this hasn't been noticed since the last fix 2 years ago? How can we prevent it will not happen in the future? issuing tokens valid for 30 days without letting users know is a big regression if you ask me. Writing e2e/integration tests and validating configuration vs token expire times etc would help? I dont think the compliance test would have caught this right? But yeah merge it and get it released asap is of course most important right now 👍🏻 and thanks for the quick fix!

BeryJu:
@lsjostro I'm assuming that most applications don't use the `id_token`  part of a refresh_token grant as even the OIDC spec says this is optional, and hence it hasn't been noticed

lsjostro:
that's not true, first app that pops up is the k8s api server?

BeryJu:
sorry I meant from the apps I've used with the OIDC provider, I've not set up my K8s clusters to use the authentik OIDC provider yet

PR #9301, opened 2024-04-16
web: fix locale loading being skipped

Gunsmithy:
## Details

Hey there,

I noticed on authentik that no matter what browser locale I had set or what locale I passed to the query string, the locale would always be loaded as 'en'.

After looking at the code, I attributed a potential fix to the one line I reference in this PR. I did a Chrome local override and this fixed the issue with the query string being ignored, but it still doesn't seem to consider the browser headers.

The lack of consideration for browser headers may be due to an issue with `autoDetectLanguage` where it never defaults `requestedCode` to `TOMBSTONE` because `requestedCode` is always set to `en` from `DEFAULT_LOCALE` so it's not really an optional.

I'm at work so I didn't have sufficient time to determine if this is an acceptable fix for at least the query string issue, nor run any local tests, so this PR is more meant to serve as an Issue for review/discussion.

I'll see if I can explore it more in the future if I can find some time to set up a proper local dev environment.

Let me know if I can provide any more additional info.

Thanks,

Dylan

---

## Checklist

If applicable

### DIFF
--- a/web/src/elements/ak-locale-context/ak-locale-context.ts
+++ b/web/src/elements/ak-locale-context/ak-locale-context.ts
@@ -88,7 +88,7 @@ export class LocaleContext extends LitElement {
         }
         locale.locale().then(() => {
             console.debug(`authentik/locale: Loaded locale '${code}'`);
-            if (this.getLocale() === code) {
+            if (this.getLocale() === requestedLocale) {
                 return;
             }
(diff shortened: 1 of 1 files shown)


kensternberg-authentik (approved):
Good catch, and thanks, @Gunsmithy.

PR #9302, opened 2024-04-16
events: fix incorrect user logged when using API token authentication

BeryJu:
## Details

This was introduced in 2024.2 by incorrectly caching the user from before the request instead of taking it from the request when an action was actually done. This works fine with the session auth as the session middleware happens before the event middleware, however since the token authentication modifies the user as part of the view and not the middleware, the user was updated too late.

This also adds a test that makes sure this doesn't happen again.

---

## Checklist

If applicable


kensternberg-authentik (approved):
At a syntactical level, this doesn't look like brain science.

PR #9336, opened 2024-04-18
web/admin: fix error in admin interface due to un-hydrated context

BeryJu:
## Details

REPLACE ME

---

## Checklist

If applicable


kensternberg-authentik (approved):
This is the sort of thing I was hoping to avoid; I really don't want our UI to be dependent upon values that aren't derived from API, so that it can be deployed and used standalone.  For the time being, though, the environment contains these values, so this is acceptable.

PR #9341, opened 2024-04-18
web: fix locale prioritization scheme

### LINKED ISSUE #4562: Locale Auto-detect doesn't work
**Describe the bug**
Using a browser configured in French, but Authentik is in English when locale is "Auto-detect".

**To Reproduce**
1. Got to Authentik with a French browser
2. View Authentik UI in English

**Expected behavior**
Authentik UI in French.

**Logs**
Logs of Firefox and Chrome debug console
`authentik(early): version 2023.1.2, apiBase /api/v3 [config.ts:105:8]
authentik/locale: No matching locale for en-us [locale.ts:147:20]
authentik/locale: Activating detected locale 'en' [locale.ts:143:20]`

 - 2023.1.2
 - docker compose

**Additional context**
Tested with multiple browser (Chrome Desktop 109.0.5414.120, Chrome Android 109.0.5414.117, Firefox 109.0)

kensternberg-authentik:
web: fix locale prioritization scheme

## Details

The locale priority algorithm had two problems: first, the order was incorrect, allowing the global default from globalAK() to override a lot of more precise settings; second, the algorithm would take outside locale overrides from the event handler, which was not necessary.

This commit revises the locale prioritization scheme. It continues to watch for “change of locale” events from all sources (URL, browser, and user/brand/site internal settings), but if the event carries a suggested locale, that suggestion is ignored. Instead, when a change of locale event occurs, it re-runs the algorithm in priority order.

That order is:

- The URL query parameter `locale=`
- The User’s stated preference in `CurrentUser.attributes`
- The Browser’s stated locale
- The Brand’s stated preference in `CurrentBrand.attributes`
- The authentik instance’s setting `from window.globalAK()`
- The default locale complied into the UI at build time.

Note to @tanberry: We should note this order somewhere in the documentation, so that users are not “surprised” that their user preference (set in User Interface ⇒ Settings ⇒ User Details ⇒ Locale) is not overriden by the browser’s preference. (The setting they need is “Based on your browser” to make browser locale detection work.)

closes #4562

## Checklist


web/src/elements/ak-locale-context/ak-locale-context.ts
```diff
@@ -41,6 +42,9 @@ export class LocaleContext extends LitElement {
 
     setLocale: LocaleSetter;
 
+    @state()
+    userLocale = "";
```
kensternberg-authentik, line 46: We don't have a good "subscribe to user changes," so I'm having to keep this one local.

web/src/elements/ak-locale-context/ak-locale-context.ts
```diff
@@ -59,8 +63,11 @@ export class LocaleContext extends LitElement {
 
     connectedCallback() {
         super.connectedCallback();
-        const localeRequest = autoDetectLanguage(this.locale);
-        this.updateLocale(localeRequest);
+        new CoreApi(DEFAULT_CONFIG)
+            .coreUsersMeRetrieve()
+            .then((user) => (this.userLocale = user?.user?.settings?.locale ?? ""))
+            .catch(() => {});
```
kensternberg-authentik, line 69: Can't use `me()` here, it'll cause an infinite loop since it'll 403 when not logged in and *force* a redirect to the login page.  Which will then loop.  Doing it by hand allows us to avoid that.  We get the `EVENT_REFRESH` event anyway when we transition from the Flow to the Admin or User interfaces.

web/src/elements/ak-locale-context/helpers.ts
```diff
@@ -45,12 +45,13 @@ export function localeCodeFromUrl(param = "locale") {
 const isLocaleCandidate = (v: unknown): v is string =>
     typeof v === "string" && v !== "" && v !== TOMBSTONE;
 
-export function autoDetectLanguage(requestedCode?: string): string {
+export function autoDetectLanguage(userReq = TOMBSTONE, brandReq = TOMBSTONE): string {
```
kensternberg-authentik, line 48: This much neediness is a code smell; this should be moved into the element itself.  But not now.

web/src/elements/ak-locale-context/ak-locale-context.ts
```diff
@@ -1,18 +1,19 @@
...
 import type { LocaleGetter, LocaleSetter } from "./configureLocale";
-import {
-    DEFAULT_LOCALE,
-    autoDetectLanguage,
-    getBestMatchLocale,
-    localeCodeFromUrl,
-} from "./helpers";
+import { DEFAULT_LOCALE, autoDetectLanguage, getBestMatchLocale } from "./helpers";
+
+const LocaleContextBase = WithBrandConfig(LitElement);
```
kensternberg-authentik, line 16: Subscribe to the brand changes here.

web/src/elements/ak-locale-context/ak-locale-context.ts
BeryJu: Overall I think it would probably be a better pattern to not have `me()` do a redirect and just add a mixin that redirects to login if not authenticated, but that can be a separate PR

BeryJu (approved):

PR #9386, opened 2024-04-23
web: provide better feedback on Application Library page about search results

kensternberg-authentik:
## Details

web: improve state management of Fuze application search

This commit rewrites a bit (just a bit, really!) of the relationship between `ak-library-application-impl` and `ak-library-application-search`.

The “show only apps with launch URLs filter” has been moved up to the retrieval layer; there was no reason for the renderer to repeatedly call a *required* filter; just call it on the list of applications once and be done.

The search component exchanges the two-state guesswork and custom events for a concrete three-state solution and *private* events. The search handler now sends the events “reset,” “updated,” and the new “updated and empty,” which we could not previously track.

By limiting the Impl layer to only those apps with launchUrls, we can now distinguish between “all apps,” and “filtered apps,” and understand that when “all apps” is empty we have no apps, and when “filtered apps” is empty the search has returned nothing.

I also tried to add a lot more comments.

In keeping with ES2020, I’ve put `.js` extensions on all the local imports.

In keeping with a variety of [best practice recommendations](https://webcomponents.today/best-practices/), I’ve renamed web component files to match the custom element they deploy:

    ak-library-application-search-empty.ts
    19:@customElement("ak-library-application-search-empty")

    ak-library-impl.ts
    44:@customElement("ak-library-impl")

    ak-library.ts
    30:@customElement("ak-library")

    ak-library-application-list.ts
    34:@customElement("ak-library-application-list")

    ak-library-application-empty-list.ts
    22:@customElement("ak-library-application-empty-list")

    ak-library-application-search.ts
    46:@customElement("ak-library-application-search")

The only effect(s) external to the changes in this vertical is that the Route() had to be updated, and I have done that.


web/src/user/LibraryPage/ak-library-application-list.ts
```diff
@@ -12,7 +12,7 @@ import PFBase from "@patternfly/patternfly/patternfly-base.css";
 
 import type { Application } from "@goauthentik/api";
 
-import type { AppGroupEntry, AppGroupList } from "./types";
+import type { AppGroupEntry, AppGroupList } from "./types.js";
```
BeryJu, line 15: should be we importing these files are `.js`? If anything we probably shouldn't for consistency

kensternberg-authentik: The standard is quickly becoming "If it's going to be a JS file, [always use a .js extension](https://www.typescriptlang.org/docs/handbook/modules/theory.html#module-format-detection)."  I'd like to keep doing that from now on.  Typescript has extensive documentation on it.  Browser stuff should be ES2020 and beyond.

BeryJu (approved):

PR #9479, opened 2024-04-26
root: add primary-replica db router

BeryJu:
## Details

REPLACE ME

---

## Checklist

If applicable


authentik/root/settings.py
```diff
@@ -315,7 +315,23 @@
     # https://docs.djangoproject.com/en/4.0/ref/databases/#persistent-connections
     DATABASES["default"]["CONN_MAX_AGE"] = None  # persistent
 
-DATABASE_ROUTERS = ("django_tenants.routers.TenantSyncRouter",)
+for replica in CONFIG.get_keys("postgresql.read_replicas"):
```
rissson, line 318: Let's also add support for use_pgpool and use_pgbouncer

rissson (changes requested):

authentik/root/settings.py
rissson: actually let's not

rissson (approved):

website/docs/installation/configuration.mdx
```diff
@@ -77,6 +77,20 @@ To check if your config has been applied correctly, you can run the following co
...
+-   `AUTHENTIK_POSTGRESQL__READ_REPLICAS__0__NAME`: same as above
+-   `AUTHENTIK_POSTGRESQL__READ_REPLICAS__0__USER`: same as above
+-   `AUTHENTIK_POSTGRESQL__READ_REPLICAS__0__PORT`: same as above
+-   `AUTHENTIK_POSTGRESQL__READ_REPLICAS__0__PASSWORD`: same as above
+-   `AUTHENTIK_POSTGRESQL__READ_REPLICAS__0__SSLMODE`: same as above
+-   `AUTHENTIK_POSTGRESQL__READ_REPLICAS__0__SSLROOTCERT`: same as above
+-   `AUTHENTIK_POSTGRESQL__READ_REPLICAS__0__SSLCERT`: same as above
+-   `AUTHENTIK_POSTGRESQL__READ_REPLICAS__0__SSLKEY`: same as above
+
+Note that `USE_PGBOUNCER` and `USE_PGPOOL` are inherited from the main database configuration and are not overridable per read replica. By default, if read replicas are configured, the main database is not used for reads. If you'd like it to be included for reads, add it as a read replica.
```
BeryJu, line 92: I'm not sure on this; I think it both makes sense to use the primary for reading but I also think it makes sense to exclude it

Aside from that, we should also explicitly state that you can rotate the postgres credentials and authentik will pick it up, however we should also note that you cant add/remove replicas without a restart

rissson: > I'm not sure on this; I think it both makes sense to use the primary for reading but I also think it makes sense to exclude it

Yeah, I defaulted to not include it because that's how it was written initially. I think it's fine as-is, and we can let the user make the call of whether they want to include it or not.

> Aside from that, we should also explicitly state that you can rotate the postgres credentials and authentik will pick it up, however we should also note that you cant add/remove replicas without a restart

Will do.

PR #9489, opened 2024-04-28
Caddy V2 Basic Auth missing header

alhazmy13:
## Details

This PR is to fix the issue with the missing authorization header for the basic authentication

---

## Checklist

If applicable

### DIFF
--- a/website/docs/providers/proxy/_caddy_standalone.md
+++ b/website/docs/providers/proxy/_caddy_standalone.md
@@ -10,7 +10,7 @@ app.company {
         uri /outpost.goauthentik.io/auth/caddy
 
         # capitalization of the headers is important, otherwise they will be empty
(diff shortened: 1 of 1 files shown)


rissson (changes requested):
This not only applies to Caddy and the `Authorization` header, but to all reverse proxies and [custom headers](https://docs.goauthentik.io/docs/providers/proxy/custom_headers)
I'd rather we add a note in that page that for extra headers and sending basic auth, the admin needs to add those to their reverse proxy configuration, based on what we already show in there.

alhazmy13:
Dear @rissson  Thank you for your feedback! I've added a section to the documentation covering the forwarding of the basic auth, based on your suggestion. Please review the updated content.

rissson:
I would have added that text to `website/docs/providers/proxy/custom_headers.md`. @tanberry thoughts?

PR #9491, opened 2024-04-28
website/integrations: netbox: add missing scope configuration

Eising:
While trying the proposed documentation, I found that the group mapping did not work.

After troubleshooting for a while, it seems that Netbox does not seem to request the `roles` scope at all, so the Scope mapping is actually never used (or at least that is how I understand the problem).

Changing the scope of the example provided to `profile` instead makes it work perfectly.

This PR contains just this change in the documentation.


rissson (changes requested):
You need to set `SOCIAL_AUTH_OIDC_SCOPE = ["openid", "profile", "email", "roles"]` in the Netbox configuration for the `roles` scope to be requested. That configuration is indeed missing from the documentation. I'd rather we add that than use a well-known scope that doesn't fit the use case.

tanberry:
Hi @Eising did you want to make the change that @rissson suggested? That would be great if so, otherwise we can do, just let us know. Thanks for pointing out this discrepency in the docs!

Eising:
> Hi @Eising did you want to make the change that @rissson suggested? That would be great if so, otherwise we can do, just let us know. Thanks for pointing out this discrepency in the docs!

I will do so momentarily. Thanks for reminding me!

Eising:
Updated with the suggested fix.

rissson (approved):

PR #9557, opened 2024-05-03
providers/scim: fix SCIM ID incorrectly used as primary key

BeryJu:
## Details

Previously we incorrectly used the SCIM ID given by the external system as a primary key on the SCIMUser and SCIMGroup objects, causing them to be globally unique. This PR fixes that by adding the `scim_id` column to hold this value, and set the primary key to a UUID

---

## Checklist

If applicable


kensternberg-authentik (approved):
Looks standard. I'm glad to see the `unique_together` in the metadata, but should SCIM_ID also be "not empty" or adhere to a pattern for validation?

PR #9558, opened 2024-05-03
core: include version in built JS files

BeryJu:
## Details

To prevent caching issues, give the JS files a unique name

Note: this current implementation does not take the build hash into consideration as for development we set that to `dev` (when debug is `true`), and to detect that in JS we'd have to implement a config parser in that too

---

## Checklist

If applicable


kensternberg-authentik (approved):
What a pain in the... neck.

BeryJu:
Of course this breaks e2e tests since the package json hasnt changed and as such it's using the cached frontend build...sigh

authentik/core/templatetags/authentik_core.py
```diff
@@ -0,0 +1,24 @@
...
+
+@register.simple_tag()
+def versioned_script(path: str) -> str:
+    """Wrapper around {% static %} tag that supports setting the version"""
+    returned_lines = [
+        f'<script src="{static_loader(path.replace("%v", __version__))}" type="module"></script>',
+        # Legacy method of loading scripts used as a fallback, without the version in the filename
+        # TODO: Remove after 2024.6 or later
+        (
+            f'<script src="{static_loader(path.replace("-%v", ""))}?'
```
BeryJu, line 20: This is both a fix/workaround for e2e tests but also for rolling deployments as especially during upgrades

kensternberg-authentik (approved):
Still looks good to me.  :-)

web/build.mjs
```diff
@@ -9,12 +10,21 @@ import { fileURLToPath } from "url";
 
 const __dirname = fileURLToPath(new URL(".", import.meta.url));
 
+// Use the package.json file in the root folder, as it has the current version information.
+const authentikProjectRoot = execFileSync("git", ["rev-parse", "--show-toplevel"], { encoding: "utf8" }).replace(
```
BeryJu, line 14: I'm not sure if that'll work in the container build, since we don't have .git available there currently? (Although we might want to change that since I've also been meaning to try out https://github.com/codecov/codecov-javascript-bundler-plugins)

rissson: That, plus we don't have git binaries either

web/build.mjs
```diff
@@ -10,11 +10,15 @@ import { fileURLToPath } from "url";
...
-);
+let authentikProjectRoot = __dirname + "../";
+try {
+    // Use the package.json file in the root folder, as it has the current version information.
+    authentikProjectRoot = execFileSync("git", ["rev-parse", "--show-toplevel"], {
+        encoding: "utf8",
+    }).replace("\n", "");
+} catch (exc) {
+    // We probably don't have a .git folder, which could happen in container builds
+}
```
kensternberg-authentik, line 21: Oh! I wasn't aware there wasn't a clearly defined project root in container builds. In that case you're right and we shouldn't even bother with using the `git rev-parse` trick to find it; I would just remove the whole thing and live with what's on line 13.

kensternberg-authentik (approved):

PR #9598, opened 2024-05-06
core: fix task clean_expiring_models removing valid sessions with using database storage

rissson:
## Details

REPLACE ME

---

## Checklist

If applicable

### DIFF
--- a/authentik/core/tasks.py
+++ b/authentik/core/tasks.py
@@ -2,7 +2,9 @@
 
 from datetime import datetime, timedelta
 
+from django.conf import ImproperlyConfigured
 from django.contrib.sessions.backends.cache import KEY_PREFIX
+from django.contrib.sessions.backends.db import SessionStore as DBSessionStore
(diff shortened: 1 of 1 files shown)


authentik/core/tasks.py
```diff
@@ -39,16 +42,32 @@ def clean_expired_models(self: SystemTask):
...
+                except Exception as exc:
+                    LOGGER.debug("Failed to get session from cache", exc=exc)
+                if not value:
+                    session.delete()
+                    amount += 1
+            case "db":
+                if (
+                    DBSessionStore.get_model_class()
+                    .objects.filter(session_key=session.session_key, expire_date__gt=now())
+                    .first()
```
BeryJu, line 61: instead of this we can probably do `.exists()` since we don't actually need to fetch any of the data

rissson: that makes perfect sense

BeryJu (approved):

PR #9648, opened 2024-05-08
web: fix value handling inside controlled components

kensternberg-authentik:
# web: fix value handling inside controlled components

## Details

This is one of those stupid bugs that drive web developers crazy. The basics are straightforward: when you cause a higher-level component to have a “big enough re-render,” for some unknown definition of “big enough,” it will re-render the sub-components. In traditional web interaction, those components should never be re-rendered while the user is interacting with the form, but in frameworks where there’s dynamic re-arrangement, part or all of the form could get re-rendered at any mmoment. Since neither the form nor any of its intermediaries is tracking the values as they’re changed, it’s up to the components themselves to keep the user’s input– and to be hardened against property changes coming from the outside world.

So static memoization of the initial value passed in, and aggressively walling off the values the customer generates from that field, are needed to protect the user’s work from any framework’s dynamic DOM management. I remember struggling with this in React; I had hoped Lit was better, but in this case, not better enough.

The protocol for “is it an ak-data-control” is “it has a `json()` method that returns the data ready to be sent to the authentik server.” I missed that in one place, so that’s on me.


web/src/elements/ak-checkbox-group/ak-checkbox-group.ts
```diff
@@ -112,10 +113,14 @@ export class CheckboxGroup extends AkElementWithCustomEvents {
     @queryAll('input[type="checkbox"]')
     checkboxes!: NodeListOf<HTMLInputElement>;
 
+    @state()
```
kensternberg-authentik, line 116: Storing the internal "values" independently of the "value" field, which is really just `defaultValue` for our purposes.

web/src/elements/forms/Form.ts
```diff
@@ -80,7 +80,7 @@ export function serializeForm<T extends KeyUnknown>(
         }
 
         if ("akControl" in inputElement.dataset) {
-            assignValue(element, inputElement.value, json);
+            assignValue(element, (inputElement as unknown as AkControlElement).json(), json);
```
kensternberg-authentik, line 83: And with the taking control of `value`, this is how we ensure the values passed to authentik are correct.

BeryJu (approved):

PR #9651, opened 2024-05-09
web: break application view into constituent parts

kensternberg-authentik:
## Details

As part of the project to make the verticals more controllable and responsive, this commit breaks the ApplicationView into two different parts: The API layer and the rendering layer.

The rendering layer is officially dumb beyond words; it knows nothing at all about Applications, RBAC, or Outposts; it just draws what it’s told to draw. It has parts inside that have their own reactivity, but that reactivity means nothing to the renderer.

The Renderer itself is broken into two: The LoadingRenderer works when there is no application, and the regular Renderer is build when there is. Typescript’s check makes it impossible to attempt to use the standard renderer when there is no application, so all of the `this.application?` checks just… go away.

A *huge* section of the View is the control card, which offers the user the power to visit the provider, provide access to the backchannel providers, edit the application, run an access check against a given user, and launch the application. All of these features were heavily obscured by a blizzard of dg/dl/dt/dd html objects that made it hard to see what was in there. Each “description” pair has been broken out into a tuple of Term and Description, with filters to remove the ones that aren’t applicable whenever an application doesn’t have, for example, a launch url, or backchannel providers, and a utility function I wrote *ages* ago renders the description list syntax for me without my having to do it all by hand.

The nice thing about this work is that it now allows me to *see* where in the ApplicationView code to focus my efforts on providing activation hooks for the “create a new policy,” “assign a new permission to a user,” or “edit an application” commands that should be accessible by the palette, or even from the sidebar.

The other nice thing is that it reveals just *where* in our code to focus our efforts on revamping our styling, and making it better for ourselves and our users.

There’s a reason I call this my *legibility project*.

It didn’t even take that long… about 2½ hours, and I’m only going to get faster at it as the needs of the different components become clear.


kensternberg-authentik:
Probably going to take a different tack in the future.

PR #9668, opened 2024-05-09
web/admin: rework initial wizard pages and add grid layout

BeryJu:
Signed-off-by: Jens Langhammer <jens@goauthentik.io>

## Details

REPLACE ME

---

## Checklist

If applicable


web/src/admin/sources/SourceWizard.ts
```diff
@@ -7,73 +7,35 @@ import { DEFAULT_CONFIG } from "@goauthentik/common/api/config";
...
 
 import PFButton from "@patternfly/patternfly/components/Button/button.css";
-import PFForm from "@patternfly/patternfly/components/Form/form.css";
-import PFRadio from "@patternfly/patternfly/components/Radio/radio.css";
 import PFBase from "@patternfly/patternfly/patternfly-base.css";
 
 import { SourcesApi, TypeCreate } from "@goauthentik/api";
 
 @customElement("ak-source-wizard-initial")
-export class InitialSourceWizardPage extends WizardPage {
```
kensternberg-authentik, line 26: It's always nice to see a change that just makes a whole bunch of code disappear.  :-)

web/src/elements/wizard/TypeCreateWizardPage.ts
```diff
@@ -0,0 +1,136 @@
...
+import PFBase from "@patternfly/patternfly/patternfly-base.css";
+
+import { TypeCreate } from "@goauthentik/api";
+
+export enum TypeCreateWizardPageLayouts {
+    list,
+    grid,
+}
+
+export abstract class TypeCreateWizardPage extends WithLicenseSummary(WizardPage) {
```
kensternberg-authentik, line 22: The only thing I could wish is that... this only shows a collection of radios.  I could wish is was relatively standalone and *evented* rather than relying on a pair of callbacks for its behavior.  Then it wouldn't need all that inheritance; you could just `<ak-type-create-wizard-page .types=${this.providerTypes} ?hidden={currentPage == 0} @select=${ ... set host type }></ak-type-create-wizard-page>`.

web/src/elements/wizard/TypeCreateWizardPage.ts
```diff
@@ -0,0 +1,136 @@
...
+                        ${requiresEnterprise
+                            ? html`<ak-license-notice></ak-license-notice>`
+                            : nothing}
+                    </span>
+                </div>`;
+            })}
+        </form>`;
+    }
+
+    render(): TemplateResult {
```
kensternberg-authentik, line 128: `return (this.layout === TypeCreateWizardPageLayouts.grid) ? this.renderGrid() : this.renderList()` ? :-)

web/tsconfig.json
```diff
@@ -2,7 +2,6 @@
     "extends": "./tsconfig.base.json",
     "compilerOptions": {
         "paths": {
-            "@goauthentik/authentik/*": ["src/*"],
```
kensternberg-authentik, line 5: 🎉

kensternberg-authentik (approved):
I approve of these changes, especially the abstraction of the "choose your type", although you're still thinking in an OO way that "routes around" some of the DOM-ness (and at other times is mind-bogglingly into wild things like filling all the slots and then providing only one of them for viewing at any time, which confuses the heck out of the debugger, which I think is nifty).

Anyway, it's nice to see a whole frackload of copy/paste disappear into a simple component.

web/src/elements/wizard/TypeCreateWizardPage.ts
BeryJu: The initial implementation worked like that, however it was quite a bit more annoying to use as with how the page is used there isn't a direct reference to the wizard host which would be required to advance pages...although I suppose that could be fixed with an `@query` attribute

web/src/elements/wizard/TypeCreateWizardPage.ts
BeryJu: Thought about that but I think this reads nicer

web/src/elements/wizard/TypeCreateWizardPage.ts
kensternberg-authentik: Yeah, the current architecture is much more OO oriented, less like traditional HTML.  It's fine, it works, and I have long learned to say "That's not how I would do it, but it gets the job done and it's not wrong."

web/src/admin/applications/wizard/auth-method-choice/ak-application-wizard-authentication-method-choice.choices.ts
```diff
@@ -25,167 +25,152 @@ type ModelConverter = (provider: OneOfProvider) => ModelRequest;
...
-}
-
-export const providerModelsList = _providerModelsTable.map(mapProviders);
+        iconUrl: "/static/authentik/sources/scim.png",
+    },
+];
 
 export const providerRendererList = new Map<string, ProviderRenderer>(
-    _providerModelsTable.map(([modelName, _0, _1, renderer]) => [modelName, renderer]),
+    providerModelsList.map((tc) => [tc.formName, tc.renderer]),
```
kensternberg-authentik, line 173: I'm glad this wasn't too confusing.  If data is uniform, I like tuples, and I like annotating them.

kensternberg-authentik (approved):
I like a lot of the changes in the UI!  Thanks!

PR #9681, opened 2024-05-10
web: provide a test framework

kensternberg-authentik:
## Details

As is typical of a system where a new build engine is involved, this thing is sadly fragile. Use the wrong import style in wdio.conf.js and it breaks; there are several notes in tsconfig.test.conf and wdio.conf.ts to tell eslint or tsc not to complain, it’s just a different build with different criteria, the native criteria don’t apply.

On the other hand, writing tests is easy and predictable. We can test behaviors at the unit and component scale in a straightforward manner, and validate our expectations that things work the way we believe they should.


BeryJu (approved):

rissson (approved):
Approved on the CI side

kensternberg-authentik:
> Approved on the CI side

@rissson I'm really anxious about this one because of [the spotlight/storybook issue](https://github.com/getsentry/spotlight/issues/419), which leads to having nonsense like this addition to the PR I did yesterday to make it pass CI, a [live-patch to compiled code](https://github.com/goauthentik/authentik/pull/9681/files#diff-e275d0bc023b6efbd839e26ab72671add8dd3752e602829c1b66deae97afdf28).

rissson:
> > Approved on the CI side
> 
> @rissson I'm really anxious about this one because of [the spotlight/storybook issue](https://github.com/getsentry/spotlight/issues/419), which leads to having nonsense like this addition to the PR I did yesterday to make it pass CI, a [live-patch to compiled code](https://github.com/goauthentik/authentik/pull/9681/files#diff-e275d0bc023b6efbd839e26ab72671add8dd3752e602829c1b66deae97afdf28).

The things is we only require `ci-web-mark` to pass for PRs and such to be able to proceed. So even if the `test` job fails, we won't be blocked on other things.

PR #9726, opened 2024-05-14
web: replace handmade list in Admin Overview with generator, storybook generator, fix storybook, fix bug in list's parent component

kensternberg-authentik:
web: update storybook, storybook a few things, fix a few things

## Details

After examining how people like Adobe and Salesforce do things, I have updated the storybook configuration to provide run-time configuration of light/dark mode (although right now nothing happens), inject the correct styling into the page, and update the preview handling so that we can see the components better. We’ll see how this pans out.

I have provided stories for the AggregateCard, AggregatePromiseCard, and a new QuickActionsCard. I also fixed a bug in AggregatePromiseCard where it would fail to report a fetch error. It will only report that “the operation falied,” but it will give the full error into the console.

**As an experiment**, I have changed the interpreter for `lint:precommit` and `build:watch` to use [Bun](https://bun.sh/) instead of NodeJS. We have observed significant speed-ups and much better memory management with Bun for these two operations. Those are both developer-facing operations, the behavior of the system undur current CI/CD should not change.

And finally, I’ve switched the QuickActionsCard view in Admin-Overview to use the new component. Looks the same. Reads *way* easier. :-)

If applicable


web/scripts/eslint.mjs
```diff
@@ -0,0 +1,63 @@
+#!/usr/bin/env node --max_old_space_size=65536
```
kensternberg-authentik, line 1: Scripted so we can run it against Bun, if we like.

web/src/admin/admin-overview/AdminOverviewPage.ts
```diff
@@ -64,6 +69,14 @@ export class AdminOverviewPage extends AKElement {
         ];
     }
 
+    quickActions: QuickAction[] = [
```
kensternberg-authentik, line 72: This and the next (red) block are what I started out to do!

web/src/elements/cards/AggregateCard.ts
```diff
@@ -8,8 +8,16 @@ import PFCard from "@patternfly/patternfly/components/Card/card.css";
 import PFFlex from "@patternfly/patternfly/layouts/Flex/flex.css";
 import PFBase from "@patternfly/patternfly/patternfly-base.css";
 
+export interface IAggregateCard {
+    icon?: string;
+    header?: string;
+    headerLink?: string;
+    subtext?: string;
+    leftJustified?: boolean;
```
kensternberg-authentik, line 16: The Lit and Open-WC "best practices" guidelines state that boolean values should _never_ be true by default; since "isCenter" was the default behavior (see lines 26-34 immediately below) and was true by default, I've re-arranged the logic so that it's now, well, the default behavior, and "left-justified" in the component's attributes will disable that default behavior. 

Only one aggregate card ever uses the left-justified form anyway, and it's, yes, QuickActionsCard.

web/src/elements/cards/AggregatePromiseCard.ts
```diff
@@ -1,22 +1,34 @@
...
     @property({ attribute: false })
     promise?: Promise<Record<string, unknown>>;
 
-    async promiseProxy(): Promise<TemplateResult> {
+    async promiseProxy(): Promise<TemplateResult | typeof nothing> {
         if (!this.promise) {
-            return html``;
+            return nothing;
+        }
+        try {
```
kensternberg-authentik, line 23: We weren't checking for if the promise ever failed.  Now we are.

web/tsconfig.json
```diff
@@ -2,17 +2,17 @@
     "extends": "./tsconfig.base.json",
     "compilerOptions": {
         "paths": {
-            "@goauthentik/authentik/*": ["src/*"],
```
kensternberg-authentik, line 5: ESlint had never been run against this file. It complained that the "relative locations" had a poor syntax, so, now they don't.

web/package.json
```diff
@@ -10,12 +10,12 @@
         "build-locales:repair": "prettier --write ./src/locale-codes.ts",
         "esbuild:build": "node build.mjs",
         "esbuild:build-proxy": "node build.mjs --proxy",
-        "esbuild:watch": "node build.mjs --watch",
```
BeryJu, line 13: we should probably either keep the node commands and add equivalents with bun and update the docs to say that bun is required for frontend (in that case we should probably switch everything to bun though)

BeryJu (approved):

PR #9727, opened 2024-05-14
website/docs: next release notes: add upgrade instructions for postgres to 16

### LINKED ISSUE #9628: Q: Using Postgres V16 in docker-compose instead of V12
I noticed in the docker-compose.yml file that the Postgres image is set to version 12. Is there any specific reason why a newer version isn't recommended in the Docker Compose file? Considering that [version 12 will reach its end of life in six months](https://endoflife.date/postgresql), would it be wise to create the database on version 16 or attempt to upgrade the current one? Also, how does this decision impact security?

dominic-r:
## Details

Provides documentation to upgrade authentik-postgresql to PG v16

closes #9628 

---

## Checklist

If applicable


jehy:
Great PR! Just several hours ago I independently upgraded to postgres 16, there were no problems and it seems like authentik is working much faster.

rissson:
We'll probably include this in the release notes, so `website/docs/releases/2024/next.md`, which you can create from the [template](https://github.com/goauthentik/authentik/blob/main/website/docs/releases/_template.md).

rissson:
> Great PR! Just several hours ago I independently upgraded to postgres 16, there were no problems and it seems like authentik is working much faster.

Yeah we've been testing 16 in CI pretty much since in came out, no issue. Interesting that it seems to be faster though!

rissson:
> > Great PR! Just several hours ago I independently upgraded to postgres 16, there were no problems and it seems like authentik is working much faster.
> 
> Thanks! Did you update with an existing database or did you recreate it? I'm currently facing the error `psycopg.errors.DuplicateTable: relation "authentik_events_event" already exists` and i'm looking for ways to make the upgrade easier. From wha I saw, it's impossible (unless i only saw the errors and not the success) to restore database from upgrade_backup_12.sql without purging it first due to already existing table errors.

You shouldn't need to re-create a database, restoring your dump file should be enough. `pg_dump` has arguments to delete objects if they already exist (`-c`, `-C`). The tables probably got re-created because authentik was running and thus executed its migrations as if it was a new install. As such, it should be shut down during the whole operation.

rissson:
> Regarding K8s, someone would have to write the documentation as I don't really have any experience with it

We distribute postgres 15 with the helm chart, so we won't be upgrading that one soon.

website/docs/releases/2024/v2024.5.md
```diff
@@ -0,0 +1,71 @@
+---
+title: Release 2024.5
```
rissson, line 2: Nice try, but we'll probably not release in May. We usually use `2024.next` until we know the release month

jehy:
@4d62ext 

> Thanks! Out of curiosity, did you update with an existing database or did you recreate it?

I simply exported my database to sql dump and then imported it to a new postgres instance, all went smoothly. May be you were running authentik with new database and it already created new database object before you imported dump?

jehy:
I don't think that cutting connection to database is really simpler.

Step by step, my approach was:

1. Add new postgres16 instance docker compose. It is not connected to authentik, so it stays empty.
2. Dump old database using docker and pg CLI
3. Stop old database to make sure that I won't accidentally break it, just to make sure I can roll back to it.
4. Import dump to new postgres instance
5. Change authentik database host to this new instance

rissson:
> I don't think that cutting connection to database is really simpler.

It is what is usually recommended for postgres upgrades though. A nice addition with doing this is that you ensure that no new data is written by authentik to the database between the dump, and the database switch from the old to the new one.

rissson (approved):

website/docs/releases/2024/v2024.next.md
```diff
@@ -0,0 +1,72 @@
...
+
+### PostgreSQL minimum supported version upgrade
+
+authentik now requires PostgreSQL version 14 or later. We recommend upgrading to the latest version if you are running an older version.
+
+The provided Helm chart defaults to PostgreSQL 15. If you are using the Helm chart with the default values, no action is required.
+
+The provided docker-compose file was updated with PostgreSQL 16. You can follow the procedure below to upgrade.
+
+#### docker-compose PostgreSQL upgrade instructions
```
BeryJu, line 22: I think like we did with the postgres upgrade in K8s, these steps should be on a separate file that can be linked to, like https://github.com/goauthentik/authentik/blob/main/website/docs/troubleshooting/postgres/upgrade_kubernetes.md

website/docs/releases/2024/v2024.next.md
```diff
@@ -0,0 +1,54 @@
...
+
+## Breaking changes
+
+### PostgreSQL minimum supported version upgrade
+
+authentik now requires PostgreSQL version 14 or later. We recommend upgrading to the latest version if you are running an older version.
+
+The provided Helm chart defaults to PostgreSQL 15. If you are using the Helm chart with the default values, no action is required.
+
+The provided docker-compose file was updated with PostgreSQL 16. You can follow the procedure [here](../../troubleshooting/postgres/upgrade_docker.md) to upgrade.
```
tanberry, line 20: We should consider saying "Compose file" because Docker has deprecated `docker-compose`. 

In general, when referring to the product called Docker Compose, we capitalize and do not use a hyphen. When referring to a command, using courier font, then use lower-case and no hyphen `docker compose`.  For the file, when you are just referring to it without naming it, let's go with "Compose file".  If naming it, then use code format and use the exact name of file (`docker-compose.yml`).

tanberry, line 20:
```suggestion
The provided Compose file was updated with PostgreSQL 16. You can follow the procedure [here](../../troubleshooting/postgres/upgrade_docker.md) to upgrade.
```

website/docs/releases/2024/v2024.next.md
```diff
@@ -0,0 +1,54 @@
...
+
+## New features
+
+## Upgrading
+
+authentik now requires PostgreSQL version 14 or later. We recommend upgrading to the latest version if needed. Follow the instructions [here](../../troubleshooting/postgres/upgrade_docker.md) if you need to upgrade PostgreSQL with docker-compose.
+
+### docker-compose
+
+To upgrade, download the new docker-compose file and update the Docker stack with the new version, using these commands:
```
tanberry, line 30:
```suggestion
To upgrade, download the new Compose file and update the Docker stack with the new version, using these commands:
```

website/docs/troubleshooting/postgres/upgrade_docker.md
```diff
@@ -0,0 +1,45 @@
+---
+title: Upgrade PostgreSQL on Docker compose
```
tanberry, line 2:
```suggestion
title: Upgrade PostgreSQL on Docker Compose
```

website/docs/troubleshooting/postgres/upgrade_docker.md
```diff
@@ -0,0 +1,45 @@
...
+### Backup your existing database
+
+If you use Docker volumes you can run the following command: `docker volume create authentik_database_backup && docker run --rm -v authentik_database:/from -v authentik_database_backup:/to alpine sh -c 'cd /from && cp -a . /to'`. You can find the name of the `authentik_database` volume with `docker volume ls`.
+
+If your data is a file path: `cp -a /path/to/v12-data /path/to/v12-backup`
+
+### Delete your old database
+
+:::::danger
+Do not execute this step without checking that step 3 completed successfully
```
tanberry, line 22:
```suggestion
Do not execute this step without checking that the backup (previous step) completed successfully.
```

website/docs/troubleshooting/postgres/upgrade_docker.md
```diff
@@ -0,0 +1,45 @@
...
+
+### Modify your docker-compose.yml file
+
+Update the PostgreSQL service image from `docker.io/library/postgres:12-alpine` to `docker.io/library/postgres:16-alpine`.
+
+Add `network_mode: none` to prevent connections being established to the database during the upgrade.
+
+### Recreate the database container
+
+Pull new images and re-create the postgresql container: `docker compose pull && docker compose up --force-recreate -d postgresql`
```
tanberry, line 37:
```suggestion
Pull new images and re-create the PostgreSQL container: `docker compose pull && docker compose up --force-recreate -d postgresql`
```

website/docs/troubleshooting/postgres/upgrade_docker.md
```diff
@@ -0,0 +1,45 @@
...
+
+Apply your backup to the new database: `cat upgrade_backup_12.sql | docker compose exec postgresql psql -U authentik`
+
+Remove the network configuration you made in step 5.
```
tanberry, line 41:
```suggestion
Remove the network configuration setting `network_mode: none` that you added to the Compose file in the previous step.
```

tanberry (changes requested):
Great info and steps to have, thanks @4d62ext !

website/docs/releases/2024/v2024.next.md
tanberry: Oh, haha, I just saw this @BeryJu . Here's my documentation opinion:

Content that is transitory typically belongs either in Rel Notes or a KB... not in regular docs (because it will soon be outdated). My vote is that if this is a breaking change for some users, and it won't be needed by the majority of users past the release of 2024.6, then let's leave it in Rel Notes.

However, if there will be users discovering for many more months to come that they need to upgrade to either 15 or 16, let's move it to Troubleshooting as Jens suggests.

It just could cause our Troubleshooting section to become quite crowded if we always add Upgrade info there. Ah, I think I will [eventually] finally create a new doc just about Upgrading, and that will solve the problem. ;-)

website/docs/troubleshooting/postgres/upgrade_docker.md
```diff
@@ -0,0 +1,45 @@
+---
+title: Upgrade PostgreSQL on Docker Compose
+---
+
+### Dump your database
+
+Dump your existing database with `docker compose exec postgresql pg_dump -U authentik -d authentik -C > upgrade_backup_12.sql`.
```
rissson, line 7:
```suggestion
Dump your existing database with `docker compose exec postgresql pg_dump -U authentik -d authentik -cC > upgrade_backup_12.sql`.
```

website/docs/releases/2024/v2024.next.md
rissson: > Content that is transitory typically belongs either in Rel Notes or a KB... not in regular docs (because it will soon be outdated). My vote is that if this is a breaking change for some users, and it won't be needed by the majority of users past the release of 2024.6, then let's leave it in Rel Notes.

Agreed, but those upgrade instructions are very unlikely to change for the next time users need to upgrade postgresql, so it makes sense to have them on hand somewhere to link to in the future

rissson:
> @BeryJu something like this? tried adding headers like the example you provided

rissson:
I'm so sorry, I misscliked.

> something like this? tried adding headers like the example you provided

Yes!

rissson (approved):

website/docs/troubleshooting/postgres/upgrade_docker.md
```diff
@@ -0,0 +1,45 @@
+---
+title: Upgrade PostgreSQL on Docker Compose
```
rissson, line 2: You also need to add this to the `website/sidebars.js` file, along with the other postgres upgrade item.

rissson (changes requested):

rissson (approved):

tanberry (approved):
Thanks again @4d62ext !

PR #9735, opened 2024-05-15
core/oauth2: don't set state in responses if not supplied

alxndrsn:
I think this fixes a bug in OAuth2 / OpenID Connect (OIDC) implementation, as reported at https://github.com/getodk/central-backend/pull/1135#issue-2274114687

I guess this PR would ideally include tests which would fail without the code changes.


BeryJu:
Thanks for the PR! I'll have to do some reading but iirc as part of OIDC the state parameter is required? In that case we'd have to enforce it being set when the openid scope is set

alxndrsn:
> iirc as part of OIDC the state parameter is required?

I may be reading the wrong place, but https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest reads:

> state
>    RECOMMENDED. Opaque value used to maintain state between the request and the callback. Typically, Cross-Site Request Forgery (CSRF, XSRF) mitigation is done by cryptographically binding the value of this parameter with a browser cookie.

lognaturel:
Hi! Any chance this would be a straightforward review for you @BeryJu or someone else given the spec clarification above? Thanks!

authentik/providers/oauth2/views/authorize.py
```diff
@@ -558,7 +557,8 @@ def create_response_uri(self) -> str:
             if self.params.response_mode == ResponseMode.QUERY:
                 query_params = parse_qs(uri.query)
                 query_params["code"] = code.code
-                query_params["state"] = [str(self.params.state) if self.params.state else ""]
+                if self.params.state != None:
```
chocolateimage, line 705: Would use `is not None` instead.

PR #9931, opened 2024-05-31
core: groups: optimize recursive children query

rissson:
## Details

Went down from 3 queries to 1.

---

## Checklist


authentik/core/models.py
```diff
@@ -135,38 +160,6 @@ def is_member(self, user: "User") -> bool:
         """Recursively check if `user` is member of us, or any parent."""
         return user.all_groups().filter(group_uuid=self.group_uuid).exists()
 
-    def children_recursive(self: Self | QuerySet["Group"]) -> QuerySet["Group"]:
```
BeryJu, line 138: I think for backwards compat we might want to keep this function around as a wrapper, if we can (I think there's a non-0% change it's used in some policies)

authentik/core/models.py
```diff
@@ -96,6 +97,28 @@ class UserTypes(models.TextChoices):
     INTERNAL_SERVICE_ACCOUNT = "internal_service_account"
 
 
+class GroupQuerySet(CTEQuerySet):
+    def with_children_recursive(self):
```
BeryJu, line 101: _I_ know what this does, but maybe add a couple more comments

authentik/core/models.py
```diff
@@ -96,6 +97,28 @@ class UserTypes(models.TextChoices):
...
+            ).union(
+                cte.join(Group, group_uuid=cte.col.parent_id)
+                .annotate(
+                    relative_depth=models.ExpressionWrapper(
+                        cte.col.relative_depth
+                        + models.Value(1, output_field=models.IntegerField()),
+                        output_field=models.IntegerField(),
+                    )
+                )
+                .filter(relative_depth__lt=20),
```
BeryJu, line 114: Should probably move this limit of 20 to a constant somewhere (this also applies to the old code tbh)

BeryJu (approved):

PR #9932, opened 2024-05-31
stages/authenticator_webauthn: optimize device types creation

rissson:
## Details

Small improvement to avoid long-running transactions.

---

## Checklist

If applicable


BeryJu (approved):
That's a good way to fix it, I would've probably added a lock to it instead

I think we might still want to combine the tasks as a system task and set it on a schedule so we don't have to dispatch it on startup since even with the optimisation that's not great

Issue #9701, opened 2024-05-13
upgrade form 2023.10.4 to 2024.4. Containers server and worker restart


Julien-Quidam:
**Describe the bug**
I upgrade images the server and the worker 
ghcr.io/goauthentik/server:2023.10.4 to ghcr.io/goauthentik/server:latest

Containers server and worker restart again and again

**To Reproduce**
1 - stop all containers
authentik-worker-1
authentik-server-1
authentik-redis-1
authentik-postgresql-1

2 - change image on authentik-worker-1 and authentik-server-1

3 - restart all containers

Ok : 
authentik-redis-1
authentik-postgresql-1

Nok:
authentik-worker-1
authentik-server-1

Containers server and the worker restart again and again. No error in log.

**Expected behavior**
Containers server and worker start

**Logs**
```console
{"event": "Loaded config", "level": "debug", "logger": "authentik.lib.config", "timestamp": 1715607634.2827733, "file": "/authentik/lib/default.yml"}
{"event": "Loaded environment variables", "level": "debug", "logger": "authentik.lib.config", "timestamp": 1715607634.283426, "count": 9}
{"event": "Starting authentik bootstrap", "level": "info", "logger": "authentik.lib.config", "timestamp": 1715607634.7895787}
{"event": "PostgreSQL connection successful", "level": "info", "logger": "authentik.lib.config", "timestamp": 1715607634.813482}
{"event": "Redis Connection successful", "level": "info", "logger": "authentik.lib.config", "timestamp": 1715607634.8301268}
{"event": "Finished authentik bootstrap", "level": "info", "logger": "authentik.lib.config", "timestamp": 1715607634.8303654}
{"event": "Booting authentik", "level": "info", "logger": "authentik.lib.config", "timestamp": 1715607638.2763298, "version": "2024.4.1"}
{"event": "Enabled authentik enterprise", "level": "info", "logger": "authentik.lib.config", "timestamp": 1715607638.3674917}
{"event": "Loaded app settings", "level": "debug", "logger": "authentik.lib.config", "timestamp": 1715607638.3692343, "path": "authentik.enterprise.settings"}
{"event": "Loaded app settings", "level": "debug", "logger": "authentik.lib.config", "timestamp": 1715607638.3716676, "path": "authentik.admin.settings"}
{"event": "Loaded app settings", "level": "debug", "logger": "authentik.lib.config", "timestamp": 1715607638.3725333, "path": "authentik.crypto.settings"}
[...]

PR #9937, opened 2024-06-01
Fix Bug #9701

bnounours:
Hello,
This PR fix a minimum version for lxml to 5.2.2. Indeed the version 5.2.1 raise a core dump on the server and worker image of Authentik as described in #9701

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -118,7 +118,7 @@ gunicorn = "*"
 jsonpatch = "*"
 kubernetes = "*"
 ldap3 = "*"
-lxml = "*"
+lxml = "^5.2.2"
 msgraph-sdk = "*"
 opencontainers = { extras = ["reggie"], version = "*" }
 packaging = "*"


BeryJu:
We don't have a fixed lock in the pyproject file, however the versions themselves are locked in the poetry.lock file: https://github.com/goauthentik/authentik/blob/main/poetry.lock#L2214-L2215

I'm curious as to how something like that could've slipped into a release since:
- as lxml is imported so early, it would affect everyone
- we run the full test suite in the final container images and nothing was broken there

bnounours:
Hi @BeryJu,
Sorry I did not see the poetry.lock :)
I agree it is strange in fact we see not all people are affected, few people reacted to this version.

Maybe it is related to my hardware that make lxml crash...
Anyway If you want me to perform some tests don't hesitate if it can help.

How do you want to proceed ? Do I change the merge to change the poetry.lock or I close it as you already did the change ?

bnounours:
I tried to import lxml on my server which is on AMD directly on the host and the behavior is the same => Crash. On my personal computer with intel it works ...

BeryJu:
so since 2024.6 changing this is not as easy anymore, since we use a custom base image for FIPS compliance, hence changing the lxml version here won't make a difference (see `pip install --force-reinstall /wheels/*` in the dockerfile)

We'll need to add some intel/amd tests to our base image in addition to the current x86/arm64 tests

PR #9958, opened 2024-06-03
stages/authenticator_webauthn: migrate device type import to systemtask and schedule

rissson:
## Details

Follow-up for https://github.com/goauthentik/authentik/pull/9932#pullrequestreview-2092051031

---

## Checklist

If applicable

### DIFF
--- a/authentik/stages/authenticator_webauthn/settings.py
+++ b/authentik/stages/authenticator_webauthn/settings.py
@@ -1,0 +1,17 @@
+"""Stage authenticator webauthn Settings"""
+
+from celery.schedules import crontab
+
+from authentik.lib.utils.time import fqdn_rand
+
+CELERY_BEAT_SCHEDULE = {
+    "stages_authenticator_webauthn_import_mds": {
+        "task": "authentik.stages.authenticator_webauthn.tasks.webauthn_mds_import",
+        "schedule": crontab(
+            minute=fqdn_rand("webauthn_mds_import"),
+            hour=fqdn_rand("webauthn_mds_import", 24),
+            day_of_week=fqdn_rand("webauthn_mds_import", 7),
+        ),
+        "options": {"queue": "authentik_scheduled"},
+    },
+}
--- a/authentik/stages/authenticator_webauthn/tasks.py
+++ b/authentik/stages/authenticator_webauthn/tasks.py
@@ -8,6 +8,8 @@ from django.core.cache import cache
 from django.db.transaction import atomic
 from fido2.mds3 import filter_revoked, parse_blob
 
+from authentik.events.models import TaskStatus
+from authentik.events.system_tasks import SystemTask, prefill_task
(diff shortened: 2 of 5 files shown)


authentik/stages/authenticator_webauthn/tasks.py
```diff
@@ -27,9 +29,13 @@ def mds_ca() -> bytes:
...
 
 
-@CELERY_APP.task()
-def webauthn_mds_import(force=False):
-    """Background task to import FIDO Alliance MDS blob into database"""
+@CELERY_APP.task(
+    bind=True,
+    base=SystemTask,
+)
+@prefill_task
```
BeryJu, line 36: I don't remember if this needed to be in a specific order related to the celery decorator or not, but if this works then it's probably the correct order

authentik/stages/authenticator_webauthn/settings.py
```diff
@@ -0,0 +1,13 @@
+"""Stage authenticator webauthn Settings"""
+
+from celery.schedules import crontab
+
+from authentik.lib.utils.time import fqdn_rand
+
+CELERY_BEAT_SCHEDULE = {
+    "stages_authenticator_webauthn_import_mds": {
+        "task": "authentik.stages.authenticator_webauth.tasks.webauthn_mds_import",
+        "schedule": crontab(minute=fqdn_rand("webauthn_mds_import"), hour="*/4"),
```
BeryJu, line 10: I think we could probably do even less frequently as the files only change with a new authentik build

BeryJu (approved):

authentik/stages/authenticator_webauthn/settings.py
rissson: Yeah I just picked another settings.py to copy from. What would you suggest?

authentik/stages/authenticator_webauthn/tasks.py
rissson: I picked it from somewhere else, so it should work

authentik/stages/authenticator_webauthn/settings.py
BeryJu: Every 24 hours? Could probably even do every week since even upon upgrade there should rarely be important things in here 


Actually that reminds me, we do the check for overdue system tasks on startup, would that also trigger this task on a first startup? If not we should probably change the date set in the prefill decorator to also run those tasks on first startup

rissson: > would that also trigger this task on a first startup

From my understanding of the code it should. I'll try it out nonetheless.

rissson: > > would that also trigger this task on a first startup
> 
> I'll try it out nonetheless.

Yeah it does, so we just need to increase the time to like 1 week or something.

BeryJu (approved):

PR #9959, opened 2024-06-03
stages/captcha: rework

### LINKED ISSUE #5971: How to use reCaptcha v3 with authentik
**Describe your question/**
How do I protect user login with a captcha ?
**Relevant infos**
I tried to protect user logins with a captcha. Therefore I created a new captcha stage and added the stage to the default-login-flow. However I was only able to create a reCaptcha Enterpreise api key (v3). I entered the credentials and it seams working, the reCaptcha logo appears on the login site. However, accesses is never blocked and I'm not forced to solve a captcha, even when creating a test key which always returns 0 (the highest risk) 
Further I struggle to see how authentiks reputation policy comes into play with reCaptcha v3 
Any tipp how to get a captcha in front of the login is greatly appreciated

BeryJu:
## Details

Adds a bunch of very dearly needed functionality to the captcha stage
- Score Min/Max threshold (by default set to match recaptcha recommendation)
- Continue/stop on error (when continuing, you can make custom decisions based on score, for example enable an MFA validation stage)
- Better testing

closes #5971

---

## Checklist

If applicable


web/src/admin/stages/captcha/CaptchaStageForm.ts
```diff
@@ -78,6 +78,40 @@ export class CaptchaStageForm extends BaseStageForm<CaptchaStage> {
                             )}
                         </p>
                     </ak-form-element-horizontal>
+                    <ak-form-element-horizontal
+                        label=${msg("Score minimum threshold")}
+                        ?required=${true}
+                        name="scoreMinThreshold"
+                    >
+                        <input
```
rissson, line 86: Maybe use `ak-number-input`?

rissson (approved):

PR #9977, opened 2024-06-04
api: use custom json renderer for speed

rissson:
## Details

---

## Checklist

If applicable


authentik/api/renderers.py
```diff
@@ -0,0 +1,64 @@
...
+from django.utils.functional import Promise
+from rest_framework.compat import coreapi
+from rest_framework.renderers import JSONRenderer as BaseJSONRenderer
+
+
+def default(obj: Any) -> Any:
+    """
+    Render object to JSON.
+
+    Adaptation of https://github.com/encode/django-rest-framework/blob/master/rest_framework/utils/encoders.py
```
BeryJu, line 18: Instead of having all the cases in here, couldn't we just call the upstream function except for types that it isn't needed for?

rissson: We could, the main issue is that the upstream one returns `super().default(obj)` which we may not want.

rissson: No, actually we could

rissson: Yeah, I tried, I can't get it to work with generators :/

BeryJu: maybe we can have a look at (or even just use) https://github.com/brianjbuck/drf_orjson_renderer, since they don't seem to need this (also that repo switches the request parser to orjson too which we might as well do too)

BeryJu (approved):

PR #10086, opened 2024-06-12
Fix of bad translation

vromero:
Quick fix for a bad translation


rissson:
Thanks for your pull request!

authentik translations are handled using [Transifex](https://explore.transifex.com/authentik/authentik/). Please edit translations over there and they'll be included automatically.

PR #10088, opened 2024-06-12
root: add configuration option to enable fips

rissson:
## Details

REPLACE ME

---

## Checklist

If applicable

### DIFF
--- a/lifecycle/gunicorn.conf.py
+++ b/lifecycle/gunicorn.conf.py
@@ -7,7 +7,6 @@ from pathlib import Path
 from tempfile import gettempdir
 from typing import TYPE_CHECKING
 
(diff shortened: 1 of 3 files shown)


lifecycle/gunicorn.conf.py
```diff
@@ -30,10 +30,11 @@
 
 defuse_stdlib()
 
-try:
-    backend._enable_fips()
-except InternalError:
-    pass
+if CONFIG.get_bool("compliance.fips.enabled", False):
+    try:
```
BeryJu, line 34: I think with it being an explicit option to enable, when the option is enabled and we can't active it, we should throw that error instead of just continuing

BeryJu (approved):

PR #10090, opened 2024-06-12
web: Add enterprise / FIPS notification to the AdminOverviewPage

kensternberg-authentik:
## Details

This commit creates a new mini-card widget that shows FIPS compliance, and re-aligns the display on the Overview page to show that widget when the customer has an Enterprise license.  The following four images show the widget active (a valid Enterprise license was detected) when FIPS is active, when it's not, in both wide and narrow views.

image
image
image
image


web/src/admin/admin-overview/AdminOverviewPage.ts
```diff
@@ -201,4 +156,70 @@ export class AdminOverviewPage extends AKElement {
...
+        return html`<div class=${classMap(classes)}>
+                <ak-admin-status-system> </ak-admin-status-system>
+            </div>
+            <div class=${classMap(classes)}>
+                <ak-admin-status-version> </ak-admin-status-version>
+            </div>
+            <div class=${classMap(classes)}>
+                <ak-admin-status-card-workers> </ak-admin-status-card-workers>
+            </div>
+            ${isEnterprise
```
rissson, line 181: Following #10091, this check will also need to happen for outposts in `web/src/admin/outposts/OutpostHealth.ts`

kensternberg-authentik: It appears to already be there?
```
        if (this.outpostHealth.fipsEnabled) {
            versionString = msg(str`${versionString} (FIPS)`);
        }
```

rissson: I meant checking for enterprise as well. But given that that field will be null, it should be fine

rissson (approved):

PR #10138, opened 2024-06-17
outposts: make refresh interval configurable

rissson:
## Details

REPLACE ME

---

## Checklist


authentik/outposts/api/outposts.py
```diff
@@ -89,7 +89,8 @@ def validate_providers(self, providers: list[Provider]) -> list[Provider]:
     def validate_config(self, config) -> dict:
         """Check that the config has all required fields"""
         try:
-            from_dict(OutpostConfig, config)
+            parsed = from_dict(OutpostConfig, config)
+            timedelta_string_validator(parsed.refresh_interval)
```
BeryJu, line 93: I wonder if we can set this up as a validator that's called within from_dict

BeryJu: yeah ok lets not add another dependency just for that

Issue #9866, opened 2024-05-26
ak server stuck in migration Applying authentik_core.0029_provider_backchannel_applications_and_more...


EdungDivinefavour:
**Describe the bug**
I followed all the steps in the Authentik full development setup [here](https://docs.goauthentik.io/developer-docs/setup/full-dev-environment), and when I run ak serve, it runs forever and seems to get stuck In the step authentik_core.0029_provider_backchannel_applications_and_more...

**To Reproduce**
1. Follow the steps in the documentation, ensure to use the docker-compose.yml from the scripts folder as mentioned in the documentation

2. Run `ak server`

**Expected behavior**
I expect the server to start running without any issues

**Logs**
```
(authentik-py3.12) edungdivinefavour@Edungs-MacBook-Pro AuthentikClone % ak server
{"event":"Loaded config","level":"debug","path":"inbuilt-default","timestamp":"2024-05-25T21:27:40-03:00"}
{"event":"Loaded config","level":"debug","path":"/Users/edungdivinefavour/Documents/AuthentikClone/authentik/lib/default.yml","timestamp":"2024-05-25T21:27:40-03:00"}
{"event":"Loaded config","level":"debug","path":"/Users/edungdivinefavour/Documents/AuthentikClone/local.env.yml","timestamp":"2024-05-25T21:27:40-03:00"}
{"event":"Loaded config from environment","level":"debug","timestamp":"2024-05-25T21:27:40-03:00"}
INFO[0000] Starting Debug server                         listen="0.0.0.0:9900" logger=authentik.go_debugger
DEBU[0000] Starting gunicorn                             args="[dev_server]" cmd=./manage.py logger=authentik.router.unicorn
INFO[0000] Starting Metrics server                       listen="0.0.0.0:9300" logger=authentik.router.metrics
INFO[0000] Starting HTTP server                          listen="0.0.0.0:9000" logger=authentik.router
DEBU[0000] starting healthcheck                          logger=authentik.router.unicorn
INFO[0000] Starting HTTPS server                         listen="0.0.0.0:9443" logger=authentik.router
{"event": "Loaded config", "level": "debug", "logger": "authentik.lib.config", "timestamp": 1716683260.37202, "file": "/Users/edungdivinefavour/Documents/AuthentikClone/authentik/lib/default.yml"}
[...]

PR #10172, opened 2024-06-19
events: provide the deleted model state in the event context #10169

local-it-dev:
## Details
This PR closes #10169
It implements the sanitized model state as dict inside the `model_deleted` event.
Another option would be to implement it as `diff` like in the `model_created` event: https://github.com/goauthentik/authentik/blob/a448107feb5261cf133830148a938ea2368d3c9b/authentik/enterprise/audit/middleware.py#L109
As the `diff` field is implemented in `enterprise/audit/middleware.py` I've done the same with the `model_state`. 
I'm not sure if `events/middleware.py` would be the better place for it, but `enterprise/audit/middleware.py` already provides the `serialize_simple()` function.

---

## Checklist

  - It passes the test if I apply the changes to the last release 2024.4.2 https://github.com/goauthentik/authentik/commit/6802614fbf878cec1d872bd01da922aa95b136fd
  - I can't get the current main running because of https://github.com/goauthentik/authentik/issues/9866 and the tests also won't pass for the current main (even without my changes)

If applicable

### DIFF
--- a/authentik/enterprise/audit/middleware.py
+++ b/authentik/enterprise/audit/middleware.py
@@ -110,6 +110,17 @@ class EnterpriseAuditMiddleware(AuditMiddleware):
             thread_kwargs["diff"] = diff
         return super().post_save_handler(request, sender, instance, created, thread_kwargs, **_)
 
+
+    def pre_delete_handler(self, request: HttpRequest, sender, instance: Model, **_):
+        """Signal handler for all object's pre_delete"""
+        if not should_log_model(instance):  # pragma: no cover
+            return None
+        # Get current state
+        model_state = self.serialize_simple(instance)
+        sanitized_model_state = sanitize_item(model_state)
+        thread_kwargs = {"model_state": sanitized_model_state}
+        return super().pre_delete_handler(request, sender, instance, thread_kwargs, **_)
+
     def m2m_changed_handler(  # noqa: PLR0913
         self,
         request: HttpRequest,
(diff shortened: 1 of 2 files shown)


local-it-dev:
Can someone please give some feedback? Is there any way to get this feature upstream? Something I should change or add?

fheisler:
@local-it-dev thank you for this straightforward and sensible PR! Our only hold-up in merging this is that it touches `/enterprise` code, so in terms of licensing, we will likely have to get a [CLA](https://en.wikipedia.org/wiki/Contributor_License_Agreement) in place; we're looking into the easiest way to do so.

local-it-dev:
> @local-it-dev thank you for this straightforward and sensible PR! Our only hold-up in merging this is that it touches `/enterprise` code, so in terms of licensing, we will likely have to get a CLA in place; we're looking into the easiest way to do so.

Thank you for your response. If it's easier in terms of licensing I could also place the code in `events/middleware.py` and may reimplement `serialize_simple()` there or import it from  `enterprise/audit/middleware.py` (if this is not a licensing issue). 
Will the enterprise code like  `enterprise/audit/middleware.py`, that is already used in the community version, always be part of the community version or is does it may happen at some point that it can only be accessed with an enterprise licence?

local-it-dev:
> While it is always included (since we only create one package that has everything included), when no enterprise license (the license in authentik not the code license) is activated, that middleware is not used

But at the moment this middleware is also used without a license. So as I understand in future it could only be used with a license. Than I would place the code in `events/middleware.py`

PR #10177, opened 2024-06-19
web: enable custom-element-manifest and DOM/JS integration checking.

kensternberg-authentik:
## Details

web: re-enable custom-element-manifest and enable component checking in Typescript

This commit includes a monkeypatch to allow custom-element-manifest (CEM) to work correctly again despite our rich collection of mixins, reactive controllers, symbol-oriented event handlers, and the like. With that monkeypatch in place, we can now create the CEM manifest file and then exploit it so that IDEs and the Typescript compilation pass can tell when a component is being used incorrectly; when the wrong types are being passed to it, or when a required attribute is not initialized.


web/tsconfig.base.json
```diff
@@ -48,6 +48,13 @@
                     "no-unknown-property": "off",
                     "no-unknown-attribute": "off"
                 }
+            },
+            {
+                "name": "@genesiscommunitysuccess/custom-elements-lsp",
+                "designSystemPrefix": "ak-",
```
kensternberg-authentik, line 54: This enables TSC and TSServer to identify when web components are not being invoked correctly.  I've limited the CEM check to only our own components.

web/custom-elements.json
BeryJu, line 1: do we need to include this file in git? I assume the file would change everytime a component is added/removed or an attribute is added/removed/etc

BeryJu (approved):

PR #10195, opened 2024-06-21
root: add setting to configure primary django hasher

boesr:
## Details

This pull request makes the primary django hasher configurable via the .env file. I would like to switch the internal hash algorithm to Argon2, due to its compatibility with OpenLDAP.  There is no breaking change and the parameter is optional. If not set, the standard django hasher hierarchy is used. I couldn't test it with the current main since the `ak server` commands gets stuck during migration processes (see https://github.com/goauthentik/authentik/issues/9866). It is tested with 2024.04.

---

## Checklist

-   [ ] Local tests pass (`ak test authentik/`) -> no because main is currently broken, tests are failing before changing anything, at least for me

-   [ ] ~The API schema has been updated (`make gen-build`)~ -> no API changes

-   [ ] ~The code has been formatted (`make web`)~ -> no frontend changes

If applicable

### DIFF
--- a/authentik/lib/default.yml
+++ b/authentik/lib/default.yml
@@ -121,6 +121,8 @@ compliance:
 
 cert_discovery_dir: /certs
 
+django_primary_hasher: null
+
 tenants:
   enabled: false
   api_key: ""
--- a/authentik/root/settings.py
+++ b/authentik/root/settings.py
@@ -211,6 +211,16 @@ CACHES = {
         "REVERSE_KEY_FUNCTION": "django_tenants.cache.reverse_key",
     }
 }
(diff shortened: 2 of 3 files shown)


rissson:
I'm curious as to the use case of this. Do you have custom scripts or things that sync that password back to LDAP?

boesr:
> I'm curious as to the use case of this. Do you have custom scripts or things that sync that password back to LDAP?

That will be the fact since the performance of integrated ldap provider isn't fast / stable enough for our use case (user login to central server systems via sssd). Besides we only need to provide some users via LDAP. I thought about adding a filter to the integrated LDAP provider, but I think OpenLDAP is more stable / faster.

boesr:
@rissson I there a chance of getting this merged?

PR #10208, opened 2024-06-24
website/integrations: add offline_access scope to DokuWiki

pkuehne:
Since 2024.2, authentik requires the offline_access scope to be enabled and requested in order for DokuWiki to request a refresh token.

## Details

Documentation Update Only

With the current guidance on the Integrations page for DokuWiki, authentik won't issue refresh tokens to the application. This results in a) DokuWiki warning about this after every login b) a need to re-authenticate as soon as the access token expires

By adding the _offline access_ scope to the authentik and DokuWiki configurations, DokuWiki can request tokens in the background, allowing you to remain logged in. I would imagine this is what most users would want.

---

## Checklist

If applicable


rissson:
I'm pretty sure dokuwiki isn't the only one that requires this. From my discussions with other users, Grafana and ArgoCD (and probably others) are also candidates for this change. @BeryJu @tanberry thoughts?

tanberry:
Thanks @pkuehne for this contribution, and saving users' annoyance. ;-) 

@marc, yes agree, if this can be applied to other integrations that would be good.

website/integrations/services/dokuwiki/index.md
```diff
@@ -60,6 +60,9 @@ In authentik, under _Providers_, create an _OAuth2/OpenID Provider_ with these s
 
 Note the _client ID_ and _client secret_, then save the provider. If you need to retrieve these values, you can do so by editing the provider.
 
+You will need to include the _offline access_ scope both in authentik and DokuWiki to allow Dokuwiki to use refersh tokens. Otherwise you will need to
```
tanberry, line 63: suggested rewrite: "To prevent users from needing to log in again as soon as the access token expires, include the _offline access_ scope in both authentik and DokuWiki. This scope allows Dokuwiki to use refresh tokens."

Also I think there is a line break between lines 63 and 64...?

tanberry (approved):
Suggested rewrite,  then yeah let's get this merged!

website/integrations/services/dokuwiki/index.md
pkuehne: I've rewritten it as suggested and remove the line break. 👍

tanberry: Hi @pkuehne can you run `make website` and then push again, so that prettier gets run? The build is failing on that check. Or let us know if you have and it is still failing... thanks!

pkuehne: Hi @tanberry, I ran it again on my branch, but it succeeded and there is nothing to push.

image

PR #10212, opened 2024-06-24
web: lintpicking

kensternberg-authentik:
## Details

While working on other projects, a few small lint issues came up, and their resolutions:

- Duplicate imports of a web component
  - Duplicates eliminated.
- A switch statement with a single condition
  - Replaced with a single `if` expression. If we need further differentiation in the future, we can go back to a switch statement or use a pattern matcher.
- Empty returns
  - Removed the empty returns, and changed the return type to an inferred `Promise<void>`, since that is the explict return type implied by a naked `return;` in an async function.

All of these made eslint complain, so I have edited the code to meet our standards. They are all quite small.


tanberry (approved):
makes sense as I understand the description. and doesn't impact docs, so I approve. ;-)

BeryJu (approved):

PR #10217, opened 2024-06-25
web: add  HTMLTagNameElementMaps to everything to activate lit analyzer

kensternberg-authentik:
## Details

web: the most boring PR in the universe: Add HTMLTagNameElementMap to everything.

The [instructions for lit-analyzer](https://www.npmjs.com/package/lit-analyzer) read:

> Below you will see an example of what to add to your library typescript definition files if you want type checking support for a given html tag name.

```
declare global {
  interface HTMLElementTagNameMap {
    "my-element": MyElement;
  }
}
```

This commit does exactly that: it adds HTMLTagNameElementMap entries to every web component in the front end. Activating and associating the HTMLTagNamElementMap with its class has enabled lit-analyzer to reveal a *lot* of basic problems within the UI, the most popular of which is “missing import.” We usually get away with it because the object being imported was already registered with the browser elsewhere, but it still surprises me that we haven’t gotten any complaints over things like:

    ./src/flow/stages/base.ts
    Missing import for <ak-form-static>
    96:  <ak-form-static
    no-missing-import

Given how early and fundamental that seems to be in our code, I’d have expected to hear *something* about it.

I have not enabled most of the possible checks because, well, there are just a ton of warnings when I do. I’d like to get in and fix those.

Aside from this, I have also *removed* `customElement` declarations from anything declared as an `abstract class`. It makes no sense to try and instantiate something that cannot, by definition, be instantiated. If the class is capable of running on its own, it’s not abstract, it just needs to be overridden in child classes. Before removing the declaration I did check to make sure no other piece of code was even *trying* to instantiate it, and so far I have detected no failures. Those elements were:

- elements/forms/Form.ts
- elements/wizard/WizardFormPage.ts

The one that blows my mind, though, is this:

    src/elements/forms/ProxyForm.ts
    6-@customElement("ak-proxy-form")
    7:export abstract class ProxyForm extends Form<unknown> {

Which, despite being `abstract`, is somehow instantiable?

    src/admin/outposts/ServiceConnectionListPage.ts:    <ak-proxy-form
    src/admin/providers/ProviderListPage.ts:    <ak-proxy-form
    src/admin/sources/SourceWizard.ts:    <ak-proxy-form
    src/admin/sources/SourceListPage.ts:    <ak-proxy-form
    src/admin/providers/ProviderWizard.ts:    <ak-proxy-form type=${type.component}></ak-proxy-form>
    src/admin/stages/StageListPage.ts:    <ak-proxy-form

I’ve made a note to investigate.

### Tooling

I’ve started a new folder where all of my one-off tools for *how* a certain PR was run. It has a README describing what it’s for, and the first tool, `add-htmlelementtagnamemaps-to-everything`, is its first entry. That tool is also documented internally.

### Coda

    I've got a little list,
    I've got a little list,
    Of all the code that would never be missed,
    The duplicate code of cute-and-paste,
    The weak abstractions that lead to waste,
    The embedded templates-- you get the gist,
    There ain't none of 'em that will ever be missed,
    And that's why I've got them on my list!


web/src/admin/AdminInterface/AdminInterface.ts
```diff
@@ -158,3 +158,9 @@ export class AdminInterface extends EnterpriseAwareInterface {
         ></ak-locale-context>`;
     }
 }
+
```
kensternberg-authentik, line 161: Warning: Aside from the _last two_ files in this PR, this one and the next 359 after it are all exactly this.  If you want to see _how_ this was done, scroll down and read those last two.

BeryJu (approved):

PR #10291, opened 2024-06-28
web: restore hasLaunchUrl to client-side criteria for filtering apps

kensternberg-authentik:
## Details

I misunderstood where this information was coming from.  I'm sorta surprised and annoyed at myself for pasting the algorithm with changes, and not having the curiosity to wonder why they were different.  As it is, I'm surprised that "hasLaunchUrl" is filtered on the client side, where users (at least, savvy users) could still see the "hidden" apps in the browser's memory.


web/src/user/LibraryPage/ak-library-impl.ts
```diff
@@ -136,7 +137,10 @@ export class LibraryPage extends AKElement {
         const selected = this.selectedApp?.slug;
         const layout = this.uiConfig.layout as string;
         const background = this.uiConfig.background;
-        const groupedApps = groupBy(this.filteredApps, (app) => app.group || "");
+        const groupedApps = groupBy(
```
kensternberg-authentik, line 140: Added the filter here.

web/src/user/LibraryPage/ak-library-impl.ts
```diff
@@ -152,6 +156,10 @@ export class LibraryPage extends AKElement {
         ></ak-library-application-search>`;
     }
 
+    renderNoAppsFound() {
```
kensternberg-authentik, line 159: Just giving this a clear name.

web/src/user/LibraryPage/ak-library-impl.ts
```diff
@@ -162,10 +170,9 @@ export class LibraryPage extends AKElement {
                 ?isadmin=${this.isAdmin}
             ></ak-library-application-empty-list>`;
         }
-        if (this.filteredApps.length === 0) {
-            return html`<ak-library-application-search-empty></ak-library-application-search-empty>`;
-        }
-        return this.renderApps();
+        return this.filteredApps.some(appHasLaunchUrl) // prettier-ignore
```
kensternberg-authentik, line 173: This is what was missing from the commit.  Dammit.  Also, changed from using `find` in the pre-commit version to `some` here. `Some` is more efficient (doesn't allocate a reference), is type-correct (I want a true/false value, not `app`/undefined) and microseconds faster.  :-)

rissson (approved):

PR #10319, opened 2024-07-01
web: provide default endpoint api configuration

kensternberg-authentik:
## Details

Finds 19 places where the exact same configuration is used to describe a table’s API endpoint, and replaces that configuration with a provided default from a parent class.

While examining the logs for our build, I noted that this particular sequence is duplicated multiple times throughout our code base, accounting for a bloat of 169 lines or so of the estimated 5552 lines of bloat. By providing a default endpoint configuration and substituting it (mechanically) wherever the default is required, we reduce our code duplication issue from 9.26% of the codesabe to 8.99%.

… which is a start.


web/src/admin/blueprints/BlueprintListPage.ts
```diff
@@ -1,7 +1,6 @@
 import "@goauthentik/admin/blueprints/BlueprintForm";
```
kensternberg-authentik, line 1: With the exception of the _addition_ of the `defaultEndpointConfig` definition [in Table.ts]( https://github.com/goauthentik/authentik/pull/10319/files#diff-7ae4e7a01898c3c9928af2e9c458a00be883a13867432c0b135c6ef770ac0168), everything is just deleting duplication; I don't have a copy of the script for this, sorry, I did it in elisp.

BeryJu (approved):

PR #10336, opened 2024-07-02
core: applications api: add option to only list apps with launch url

rissson:
## Details

REPLACE ME

---

## Checklist

If applicable

### DIFF
--- a/authentik/core/api/applications.py
+++ b/authentik/core/api/applications.py
@@ -147,6 +147,13 @@ class ApplicationViewSet(UsedByMixin, ModelViewSet):
                 applications.append(application)
         return applications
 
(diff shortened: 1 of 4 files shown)


authentik/core/api/applications.py
```diff
@@ -216,6 +228,8 @@ def list(self, request: Request) -> Response:
         if superuser_full_list and request.user.is_superuser:
             return super().list(request)
 
+        only_with_launch_url = str(request.query_params.get("only_with_launch_url", "false")).lower()
```
BeryJu, line 231: I still there has to be a better way to parse this value than doing manual string parsing...but I suppose not part of this PR..I know we can create serializers and pass `request.query_params` into that and use that but that's a whole lot more code

BeryJu (approved):

PR #10337, opened 2024-07-02
web: provide 'show password' button

kensternberg-authentik:
## Details

web: provide `show password` on login page

### What

Provide a `show password` icon, text, and button for the password field both in the IdentificationStage and the PasswordStage. Essentially the same code for both, although the id of the password field is unique to each. 

This code is derived from the Patternfly 4 example. The HTML example is broken, but the React one was 
straightforward to read and port.

At Jens's suggestion, I have added a field to the PasswordStage model, `allow_show_password`, and threaded
its availability through both the challenge and admin configuration details.  It can now be on a 
per-password-stage basis.  Currently `false` by default.  

### Why

Requested by users. Seems to be a common thing anyway.


BeryJu (approved):
🎉 I wonder if we can dedupe more code between password stage and identification stage with password enabled

authentik/stages/identification/stage.py
```diff
@@ -199,6 +200,8 @@ def get_challenge(self) -> Challenge:
                 "primary_action": self.get_primary_action(),
                 "user_fields": current_stage.user_fields,
                 "password_fields": bool(current_stage.password_stage),
+                "allow_show_password": bool(current_stage.password_stage)
+                and current_stage.password_stage.allow_show_password,
```
kensternberg-authentik, line 204: I wouldn't normally argue with `black` but, man, that is ugly.

web/src/flow/stages/identification/IdentificationStage.ts
```diff
@@ -44,23 +45,39 @@ export class IdentificationStage extends BaseStage<
...
+        return [
+            PFBase,
+            PFAlert,
+            PFInputGroup,
+            PFLogin,
+            PFForm,
+            PFFormControl,
+            PFTitle,
+            PFButton,
             /* login page's icons */
```
kensternberg-authentik, line 64: Although this is in response to my adding PFInputGroup, I moved this comment _out_ of the CSS; there's no reason to ship it with what the user sees.

authentik/stages/identification/models.py
```diff
@@ -38,10 +38,11 @@ class IdentificationStage(Stage):
         help_text=_(
             (
                 "When set, shows a password field, instead of showing the "
-                "password field as seaprate step."
+                "password field as separate step."
```
kensternberg-authentik, line 41: Genuinely surprised codespell didn't throw a fit here.

PR #10411, opened 2024-07-08
providers/proxy: fix Traefik label generation for v3

### LINKED ISSUE #9786: Automatic Outpost Deployment via Docker Integration prodouces invalid Traefik Labels
**Describe the bug**
When deploying a proxy outpost labels are added automatically which add the outpost to Traefik. For each `single-forward` provider a `Host(<host>)` label has to be used. Authentik generates these as `Host(<host>, <host>, ...)` which is **not possible** anymore since Traefik v3.0 (See [here](https://doc.traefik.io/traefik/v3.0/migration/v2-to-v3/?ref=containous.ghost.io#new-v3-syntax-notable-changes) and [here](https://community.traefik.io/t/host-rule-does-not-support-multiple-hostnames-anymore/21518)). This leads to an invalid deployment of the outpost which makes it defacto unusable for users with Traefik >v3.0 since the label can not be disabled as far as I can tell.

**To Reproduce**
Deploy a proxy outpost with multilple providers on a system with Traefik > v3.0 which uses the docker label integration

**Expected behavior**
Authentik should generate a valid label.

**Logs**
```"error while adding rule Host(`<redacted>`,`<redacted>`,`<redacted>`) && PathPrefix(`/outpost.goauthentik.io`): error while adding rule and: error while adding rule Host: unexpected number of parameters; got 3, expected one of [1]"```

-   authentik version: 2024.4.2

DriftDevil:
Fixes #9786 

## Details

When deploying a proxy outpost, labels are added automatically which add the outpost to Traefik. For each single-forward provider a Host(<host>) label has to be used. Authentik generates these as Host(<host>, <host>, ...) which is not possible anymore since Traefik v3.0. This leads to an invalid deployment of the outpost.
This PR remedies that and created labels that separate the Host. i.e. (Host(<host>) || Host(<host>))

### DIFF
--- a/authentik/providers/proxy/controllers/docker.py
+++ b/authentik/providers/proxy/controllers/docker.py
@@ -28,7 +28,7 @@ class ProxyDockerController(DockerController):
         labels = super()._get_labels()
         labels["traefik.enable"] = "true"
         labels[f"traefik.http.routers.{traefik_name}-router.rule"] = (
-            f"Host({','.join(hosts)}) && PathPrefix(`/outpost.goauthentik.io`)"
+            f"({' || '.join([f'Host(`{host}`)' for host in hosts])}) && PathPrefix(`/outpost.goauthentik.io`)"
         )
         labels[f"traefik.http.routers.{traefik_name}-router.tls"] = "true"
         labels[f"traefik.http.routers.{traefik_name}-router.service"] = f"{traefik_name}-service"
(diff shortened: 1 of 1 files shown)


BeryJu (approved):
I'm assuming this fixed syntax also works with traefik 2.x?

rissson:
> I'm assuming this fixed syntax also works with traefik 2.x?

It does yes.

PR #10181, opened 2024-06-20
website/docs: Fix nginx proxy_pass directive documentation

fotinakis:
## Details

As per the [nginx docs](https://docs.nginx.com/nginx/admin-guide/web-server/reverse-proxy/), `proxy_pass` sets the baseline for the reverse proxied URLs and will prepend the `proxy_pass` URL.

In the nginx docs in authentik, it's currently misdocumented and causes a duplicate `/outpost.goauthentik.io` path to be added during the rewrite, causing 404s with the forward auth (single domain) proxy setup and likely others?

I believe this might be the cause of a few github issues noticing 404s on the outpost, namely things like https://github.com/goauthentik/authentik/issues/9122 and maybe others.

Previous behavior (incorrect):

1. nginx proxy receives request for `https://some-proxy.company/outpost.goauthentik.io/auth/nginx`
2. nginx reverse proxies this to `http://outpost.company:9000/outpost.goauthentik.io/outpost.goauthentik.io/auth/nginx` (incorrect)

New behavior:

1. nginx proxy receives request for `https://some-proxy.company/outpost.goauthentik.io/auth/nginx`
2. nginx reverse proxies this to `http://outpost.company:9000/outpost.goauthentik.io/auth/nginx`

PR #10422, opened 2024-07-09
website/docs: update nginx docs for embedded outposts

fotinakis:
A followup to https://github.com/goauthentik/authentik/pull/10181 — that docs PR was partially correct, but does not work for the embedded outpost.

The way it works is:
* Manual outpost: when traffic flows to the proxy `/outpost.goauthentik.io/<PATH>` path, it should be stripped off and then sent to `outpost:company:9000/<PATH>`  (this is what the docs show now)
* Embedded outpost: when traffic flows to the proxy `/outpost.goauthentik.io/<PATH>` nginx proxy must prepend this path back again when sending it to the embedded outpost at `authentik.company:9000/outpost.goauthentik.io/<PATH>`

This PR clarifies how to make it work for both cases.

### DIFF
--- a/website/docs/providers/proxy/_nginx_proxy_manager.md
+++ b/website/docs/providers/proxy/_nginx_proxy_manager.md
@@ -48,9 +48,16 @@ location / {
 
 # all requests to /outpost.goauthentik.io must be accessible without authentication
 location /outpost.goauthentik.io {
-    proxy_pass              http://outpost.company:9000;
-    # ensure the host of this vserver matches your external URL you've configured
-    # in authentik
+    # When using the embedded outpost, use:
+    proxy_pass              http://authentik.company:9000/outpost.goauthentik.io;
+    # For manual outpost deployments:
+    # proxy_pass              http://outpost.company:9000;
+
+    # Note: ensure the Host header matches your external authentik URL:
+    proxy_set_header        Host $host;
+
+    # Ensure the host of this vserver matches your external URL you've configured
+    # in authentik.
     proxy_set_header        Host $host;
     proxy_set_header        X-Original-URL $scheme://$http_host$request_uri;
     add_header              Set-Cookie $auth_cookie;
--- a/website/docs/providers/proxy/_nginx_proxy_manager.md
+++ b/website/docs/providers/proxy/_nginx_proxy_manager.md
@@ -56,9 +56,6 @@ location /outpost.goauthentik.io {
     # Note: ensure the Host header matches your external authentik URL:
     proxy_set_header        Host $host;
 
-    # Ensure the host of this vserver matches your external URL you've configured
-    # in authentik.
-    proxy_set_header        Host $host;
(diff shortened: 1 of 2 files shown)


tanberry:
Thanks so much for this contribution, @fotinakis ! We will get a tech review on it and get it merged.

website/docs/providers/proxy/_nginx_proxy_manager.md
```diff
@@ -48,9 +48,16 @@ location / {
...
-    proxy_pass              http://outpost.company:9000;
-    # ensure the host of this vserver matches your external URL you've configured
-    # in authentik
+    # When using the embedded outpost, use:
+    proxy_pass              http://authentik.company:9000/outpost.goauthentik.io;
+    # For manual outpost deployments:
+    # proxy_pass              http://outpost.company:9000;
+
+    # Note: ensure the Host header matches your external authentik URL:
+    proxy_set_header        Host $host;
```
rissson, line 57: This line is already present just below

rissson (changes requested):

website/docs/providers/proxy/_nginx_proxy_manager.md
fotinakis: Fixed

rissson (approved):

BeryJu (commented):
The PR looks good, but I'm a bit confused as needing two different configuration settings shouldn't be required, since the code for the routing is identical between embedded and standalone outpost

rissson:
> The PR looks good, but I'm a bit confused as needing two different configuration settings shouldn't be required, since the code for the routing is identical between embedded and standalone outpost

I think the idea is just to point out that specifically for the embedded outpost, you need to point the reverse proxy to the authentik server

BeryJu:
> > The PR looks good, but I'm a bit confused as needing two different configuration settings shouldn't be required, since the code for the routing is identical between embedded and standalone outpost
> 
> I think the idea is just to point out that specifically for the embedded outpost, you need to point the reverse proxy to the authentik server

That part makes sense, but either both of them should have `/outpost.goauthentik.io` or `/` as path, there shouldn't be a difference as the URL handling is the same in both outpost modes. I haven't tested this myself cc @fotinakis

fotinakis:
@BeryJu (sorry, missed this PR notification) — I actually did test this and found experimentally it is correct. It makes sense from this perspective: a standalone outpost accepts outpost traffic at `/*`, but the authentik server accepts outpost traffic at `/outpost.goauthentik.io/*` — so if you stick an nginx proxy in front of your app, you must differentiate where it is sending outpost traffic depending on if you're using standalone or embedded outposts.

You could potentially hide this behavior by making standalone outposts also accept traffic with the `/outpost.goauthentik.io/*` prefix — then you could condense these configs down to one because no path transformation would be necessary.

fotinakis:
@BeryJu this was approved by @rissson  — can you guys help get this merged before it gets stale? Thank you!

PR #10440, opened 2024-07-10
web: move to wireit as the build runner language

kensternberg-authentik:
## Details

web: make `sfe` a legitimite subpackage and use `wireit` to control the build

### What

- Move sfe to a `packages` subfolder: this is a more standard format for subpackages
- `Move sfe/index.ts` to `sfe/src/index.ts`: this is a more standard layout for a package
- Adjusted paths is `package.json` and `sfe/rollup.config.js` accordingly.
- Add prettier and safety linting to `sfe`.
- fix a naming issues in `build-locales`, highlighted by eslint
- fix some minor linting issues is `build-locales`
- add comments to `build-locales`, to make it clear what it does
- updated the README and LICENSE files
- start using `wireit` heavily as the task-runner definition language

### Why

Primarily, to look professional and pave the way for future enhancements.

Aside from the standardization and so forth, the primary goal here is to move our task runner to wireit. Wireit offers a number of intriguing abilities with respect to caching, building, and testing, such as an ability to `watch` our folders and files and automatically re-run the build when the relevant code changes, without having to rebuild the copied content or sub-packages such as `sfe`.

The ability to pass in environment variables without needed `cross-env` makes code that required it much easier to read.

Commands that take a long time can be prefixed with the environment variable `${NODE_RUNNER}`, which then would allow you to default to using `node`, but by setting `NODE_RUNNER` in your shell you could specify `bun` (or `deno`, maybe, but I haven’t tested it with `deno`). `bun` runs the `eslint` pass in about three-quarters the time `node` takes.

This commit exists primarily to ensure that the build runs as expected under CI, and the result is as expected under CI.

Wireit was produced by Google and is used by Adobe Spectrum Components, Patternfly Components, Material Web, Red Hat Design, ING Bank’s Lion Components Library, and the Lit-Element teams, so I’m confident that it’s robust and reliable as a build runner.

## Checklist


Dockerfile
```diff
@@ -30,22 +30,17 @@ WORKDIR /work/web
 
 RUN --mount=type=bind,target=/work/web/package.json,src=./web/package.json \
     --mount=type=bind,target=/work/web/package-lock.json,src=./web/package-lock.json \
-    --mount=type=bind,target=/work/web/sfe/package.json,src=./web/sfe/package.json \
-    --mount=type=bind,target=/work/web/sfe/package-lock.json,src=./web/sfe/package-lock.json \
+    --mount=type=bind,target=/work/web/packages/sfe/package.json,src=./web/packages/sfe/package.json \
```
kensternberg-authentik, line 33: In a well-managed monorepo, especially using a version harmonizer like `syncpack`, most sub-packages won't have `package-lock.json` files.

web/scripts/build-locales.mjs
```diff
@@ -3,9 +3,16 @@ import fs from "fs";
...
+ * Determines if all the Xliff translation source files are present and if the Typescript source
+ * files generated from those sources are up-to-date. If they are not, it runs the locale building
+ * script, intercepting the long spew of "this string is not translated" and replacing it with a
+ * summary of how many strings are missing with respect to the source locale.
+ */
+
 const localizeRules = JSON.parse(fs.readFileSync("./lit-localize.json", "utf-8"));
 
-function compareXlfAndSrc(loc) {
+function translationFileIsUpToDateWithXlf(loc) {
```
kensternberg-authentik, line 15: I didn't intend to update this, but while working on this, one `wireit` pass had a mis-defined lint pass that made this stand out as bad naming.  Both a name change and a comment were warrented.

web/scripts/build-locales.mjs
```diff
@@ -27,17 +34,14 @@ function compareXlfAndSrc(loc) {
...
     }
 
-    // if the xlf is newer (greater) than src, it's out of date.
-    if (xlfStat.mtimeMs > srcStat.mtimeMs) {
-        return false;
-    }
-    return true;
+    // if the xliff file is older (has a lower build time) than the generated translation file, the
+    // translation file is up-to-date
+    return xlfStat.mtimeMs <= srcStat.mtimeMs;
```
kensternberg-authentik, line 39: This was what ESLint whined about.  "Just return the results of the last comparison!"

.github/workflows/api-ts-publish.yml
```diff
@@ -33,12 +33,12 @@ jobs:
       - name: Upgrade /web
         working-directory: web
         run: |
-          export VERSION=`node -e 'console.log(require("../gen-ts-api/package.json").version)'`
+          export VERSION=`node -e 'console.log(require("../../gen-ts-api/package.json").version)'`
```
BeryJu, line 36: I dont think this is correct as the web package json is still in the same directory

BeryJu (approved):

web/LICENSE.txt
```diff
@@ -0,0 +1,18 @@
+The MIT License (MIT)
```
rissson, line 1: Any way we could refer to the "main" license file instead of duplicating one here (and in the sfe package)?

kensternberg-authentik: The licenses for the front-end and the back-end are different.  Jens has also expressed a desire to move to a model of independently shippable apps for verticals (user-settings, user-library, admin, flow, etc), and each would need its own copy of the license.

PR #10454, opened 2024-07-11
policies: add GeoIP policy

### LINKED ISSUE #4433: Dedicated GeoIP/ASN policy
Since the GeoIP DB is now bundled with authentik and is available out of the box, it is logical to assume that even more users will want to take advantage of what GeoIP can offer. 

However, currently the only way to use the GeoIP features is through an expression policy. The problem with that is:
1. Not flexible. To change rules you need to either edit the code in the policy directly or have some kind of a system where it can take settings from environment variables or tenant attributes.
2. Requires programming knowledge

A dedicated GeoIP policy which can be used to set rules for passing/failing. 

Write/update code in an expression policy

**Additional context**
N/A

gergosimonyi:
Closes #4433 

Introduces the new GeoIP policy with the functionality to match client ASN or country (for now).

~Opening a draft of this for a high-level overview. There are still many details to iron out like~

-   [x] Back-end tests
-   [x] Handle disabled geoip
-   [x] Handle IP missing in geoip database
-   [x] Handle empty lists of ASNs and countries
-   [x] Translations
-   [x] Types on `web`
-   [x] Multiselect on `web`

---

## Checklist

If applicable


authentik/policies/geoip/countries.py
```diff
@@ -0,0 +1,253 @@
+"""List of all ISO-3166-1 alpha-2 country codes"""
```
gergosimonyi, line 1: An alternative to this adding a dependency like [iso3166](https://github.com/deactivated/python-iso3166) or [django-countries](https://github.com/SmileyChris/django-countries).

Since ISO 3166 changes _infrequently_, I erred on the side of fewer dependencies.

web/src/admin/policies/geoip/GeoIPPolicyForm.ts
```diff
@@ -0,0 +1,170 @@
...
+                <span slot="header"> ${msg("Policy-specific settings")} </span>
+                <div slot="body" class="pf-c-form">
+                    <ak-form-element-horizontal label=${msg("ASN Mode")} name="asnMode">
+                        <ak-search-select
+                            .fetchObjects=${async () => {
+                                return [
+                                    { name: "allow", label: "Allow" },
+                                    { name: "block", label: "Block" },
+                                ];
+                            }}
```
gergosimonyi, line 100: The use of an `async` function here for a static list of 2 is probably the worst part of this PR. I'll look to improve this.

web/src/admin/policies/geoip/GeoIPPolicyForm.ts
```diff
@@ -0,0 +1,170 @@
...
+
+        if (data.countryMode?.toString() === "") {
+            data.countryMode = "block";
+        }
+
+        if (data.countryList?.toString() === "") {
+            data.countryList = [];
+        } else {
+            data.countryList = data.countryList.split(",");
+        }
```
gergosimonyi, line 42: Defaults should be part of the back-end. Also some botched types here.

kensternberg-authentik: Yeah, I'm looking at this and thinking that I'm seeing a lot of mixed messages compared to the types generated from the schema.  You shouldn't have to be splitting anything; the API should be giving those to you, and the TypeScript definitions in GeoIPPolicy.ts say getting strings where arrays are expected shouldn't be possible).  This function is probably the wrong place to be doing this.

If you're crafting a blob to `send`, start with a default and overwrite it with what you get from the form instead.

web/src/admin/policies/geoip/GeoIPPolicyForm.ts
```diff
@@ -0,0 +1,170 @@
...
+    }
+
+    renderForm(): TemplateResult {
+        return html` <span>
+                ${msg(
+                    "Ensure the user satisfies requirements of geography or network topology, based on IP address.",
+                )}
+            </span>
+            <ak-form-element-horizontal label=${msg("Name")} ?required=${true} name="name">
+                <input
```
kensternberg-authentik, line 63: `?required=${true}` isn't needed here (and I _really_ need to break @beryju of the habit).  `required` is all that's needed.  You only need the `?` syntax for boolean attributes if the attribute can be changed by client code, for example: 
`?hidden=${!this.open}`

web/src/admin/policies/geoip/GeoIPPolicyForm.ts
```diff
@@ -0,0 +1,170 @@
...
+                    class="pf-c-form-control"
+                    required
+                />
+            </ak-form-element-horizontal>
+            <ak-form-element-horizontal name="executionLogging">
+                <label class="pf-c-switch">
+                    <input
+                        class="pf-c-switch__input"
+                        type="checkbox"
+                        ?checked=${first(this.instance?.executionLogging, false)}
```
kensternberg-authentik, line 75: `first` is a leftover from earlier iterations of TypeScript.  Please use the idiom:
```${this.instance?.executionLogging ?? false}```

Although in this case I suspect you could just use the boolean forcing expression: `${!!this.instance?.executionLogging}`

File: web/src/admin/policies/geoip/GeoIPPolicyForm.ts (hunk 2 above)
Comment (kensternberg-authentik) [reply web/src/admin/policies/geoip/GeoIPPolicyForm.ts:100]: Yeah, `search select` is specifically for complex objects that have subtitles and complex interactions.  Consider using a Switch for a boolean choice, or if the ergonomics are bad (it's sometimes hard to word), a Radio.

File: web/src/admin/policies/geoip/GeoIPPolicyForm.ts [hunk 6]
```diff
@@ -0,0 +1,170 @@
...
+                            class="pf-c-form-control"
+                        />
+                        <p class="pf-c-form__helper-text">
+                            ${msg(
+                                "List of autonomous system numbers. Comma separated. E.g. 13335,15169,20940",
+                            )}
+                        </p>
+                    </ak-form-element-horizontal>
+                    <ak-form-element-horizontal label=${msg("Country Mode")} name="countryMode">
+                        <ak-search-select
```
kensternberg-authentik, line 131: ... and given that you've done it twice, break it out into a variable or something so you'll only have to change it once if the rules change.  :-)

web/src/admin/policies/geoip/GeoIPPolicyForm.ts
```diff
@@ -0,0 +1,170 @@
...
+                        <p class="pf-c-form__helper-text">
+                            ${msg(
+                                "Block or only allow connections from IPs from the provided countries.",
+                            )}
+                        </p>
+                    </ak-form-element-horizontal>
+                    <ak-form-element-horizontal label=${msg("Country List")} name="countryList">
+                        <input
+                            type="text"
+                            value="${ifDefined(this.instance?.countryList || "")}"
```
kensternberg-authentik, line 158: Since you've provided a not-`undefined` value for falsity, `ifDefined(this.instance?.countryList || "")` will always be defined.  You can replace that whole expression with `this.instance?.countryList ?? ""`

kensternberg-authentik (commented):
Overall, the Typescript is really promising!  Good work.  There's a bit of rough edges, and it's obviously not done, but it's getting there.

authentik/policies/geoip/models.py
```diff
@@ -0,0 +1,81 @@
...
+    def passes(self, request: PolicyRequest) -> PolicyResult:
+        """TODO"""
+
+        # This is not a single get chain because `request.context` can contain `{ "asn": None }`.
+        asn_data = request.context.get("asn")
+        geoip_data = request.context.get("geoip")
+        asn = asn_data.get("asn") if asn_data else None
+        country = geoip_data.get("country") if geoip_data else None
+
+        if not asn:
```
BeryJu, line 52: both of these cases can also happen if ASN/GeoIP is not configured (pretty rare since we include the DB by default, but it is possible to disable)

authentik/policies/geoip/models.py
```diff
@@ -0,0 +1,81 @@
...
+from rest_framework.serializers import BaseSerializer
+
+from authentik.policies.exceptions import PolicyException
+from authentik.policies.geoip.countries import COUNTRIES
+from authentik.policies.geoip.exceptions import GeoIPNotFoundException
+from authentik.policies.models import Policy
+from authentik.policies.types import PolicyRequest, PolicyResult
+
+
+class GeoIPPolicyMode(models.TextChoices):
```
BeryJu, line 15: Usually we don't have this as a setting directly on the policy but instead use the `negate` flag on the policy binding to inverse the effects. For this case I'm a bit on the fence as you can do two different things in the same policy with it

web/src/admin/policies/geoip/GeoIPPolicyForm.ts
```diff
@@ -0,0 +1,170 @@
...
+@customElement("ak-policy-geoip-form")
+export class GeoIPPolicyForm extends BasePolicyForm<GeoIPPolicy> {
+    loadInstance(pk: string): Promise<GeoIPPolicy> {
+        return new PoliciesApi(DEFAULT_CONFIG).policiesGeoipRetrieve({
+            policyUuid: pk,
+        });
+    }
+
+    async send(data: GeoIPPolicy): Promise<GeoIPPolicy> {
+        if (data.asnMode?.toString() === "") {
```
BeryJu, line 24: This is somewhat of a downside of the generated API client, but this value here should never be undefined so this if shouldn't be required

web/src/admin/policies/geoip/GeoIPPolicyForm.ts
```diff
@@ -0,0 +1,170 @@
...
+                </p>
+            </ak-form-element-horizontal>
+            <ak-form-group .expanded=${true}>
+                <span slot="header"> ${msg("Policy-specific settings")} </span>
+                <div slot="body" class="pf-c-form">
+                    <ak-form-element-horizontal label=${msg("ASN Mode")} name="asnMode">
+                        <ak-search-select
+                            .fetchObjects=${async () => {
+                                return [
+                                    { name: "allow", label: "Allow" },
```
BeryJu, line 97: the value for this (regardless of which UI component is used) should use the enum from the API instead of the raw string

authentik/policies/geoip/countries.py
BeryJu: I don't mind adding a dependency for this, especially since we should add gettext calls to all these humanized names, and some dependency might have full translation for all the country names already

authentik/policies/geoip/models.py
gergosimonyi: Oh, I didn't notice that flag! That makes everything so much simpler. I'm not on the fence about it at all :smile: 

Then there's no need for `*_mode` and this model will just have 2 fields: `asn_list` and `country_list`. Same as Event Matcher Policy for least astonishment:

> If any of the configured values match, the policy passes.

web/src/admin/policies/geoip/GeoIPPolicyForm.ts
gergosimonyi: Combined with the check below (which is still needed because `"".split(",")` is non-empty)

authentik/policies/geoip/models.py
gergosimonyi: (We've had a bit of an offline discussion, but no resolution to this.) I kept the behavior for now, this way we don't leak information on whether a GeoIP database is available.

Let me know if you don't agree.

authentik/policies/geoip/countries.py
gergosimonyi: Added `django-countries`. I'm unfamiliar with `drf-spectacular` and needed a bit of dark magic to make it work. If there's an easier way, let me know.

authentik/policies/geoip/api.py
```diff
@@ -0,0 +1,49 @@
...
+from rest_framework.viewsets import ModelViewSet
+
+from authentik.core.api.used_by import UsedByMixin
+from authentik.policies.api.policies import PolicySerializer
+from authentik.policies.geoip.models import GeoIPPolicy
+
+COUNTRY_SCHEMA = {
+    "code": CountryField(),
+    "name": serializers.CharField(),
+}
```
gergosimonyi, line 15: This is an exact copy of `authentik/policies/geoip/views:11-14`. Weirdly, importing it from there results in `gen-build` silently failing to generate this part of the schema. I'm out of my depth here, so I just duplicated these 4 lines.

authentik/policies/geoip/urls.py
```diff
@@ -0,0 +1,11 @@
+"""API URLs"""
+
+from django.urls import path
+
+from authentik.policies.geoip.api import GeoIPPolicyViewSet
+from authentik.policies.geoip.views import ISO3166View
+
+api_urlpatterns = [
+    ("policies/geoip", GeoIPPolicyViewSet),
+    path("iso-3166/", ISO3166View.as_view(), name="iso-3166-view"),
```
gergosimonyi, line 10: There may or may not be a better place for this.

@BeryJu?

web/src/admin/policies/geoip/GeoIPPolicyForm.ts
```diff
@@ -0,0 +1,125 @@
...
+            policyUuid: pk,
+        });
+    }
+
+    async send(data: GeoIPPolicy): Promise<GeoIPPolicy> {
+        if (data.asns?.toString() === "") {
+            data.asns = [];
+        } else {
+            data.asns = (data.asns as unknown as string).split(",").map(Number);
+        }
```
gergosimonyi, line 28: The actual type of `data` slightly differs from `GeoIPPolicy`.

I haven't found an easy way to reconcile them (the "proper" way is probably building a component on top of `<input>` to run a transforming function when its `value` is read), so I opted for the type assertion.

web/src/admin/policies/geoip/GeoIPPolicyForm.ts
```diff
@@ -0,0 +1,125 @@
...
+                                    })
+                                    .then((results) => {
+                                        return {
+                                            options: results.map((country) => [
+                                                country.code,
+                                                country.name,
+                                            ]),
+                                        };
+                                    });
+                            }}
```
gergosimonyi, line 112: This currently hits the API for no reason every time there's a change in `search`. I'll add an optimization.

authentik/policies/geoip/api.py
BeryJu: That is very odd but also doesn't surprise me

authentik/policies/geoip/urls.py
BeryJu: I'd usually prefer to do it as `policies/geoip/iso3166/`, but obviously that could potentially cause errors due to that path also having the primary key...as a fallback I'd go with `policies/geoip_iso3166/` I guess?

BeryJu (commented):
mostly LGTM

website/docs/policies/index.md
```diff
@@ -10,6 +10,10 @@ This policy is used by the events subsystem. You can use this policy to match ev
 
 See [Expression Policy](expression.mdx).
 
+## GeoIP Policy
```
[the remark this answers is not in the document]
tanberry: Hmmm I see that all of the other headers/titles for policies also capitalize "Policy". Our style is to use sentence style capitalization for headings... do you mind also changing all of the other headings to on this age to use a lower-case "p" on "policies"? :-)

tanberry (changes requested):
Yay we have a GeoIP policy! Thanks so much for updating the docs. A few nits, mostly about capitalization... I see that this whole page does not follow our Style Guide about using sentence-case capitalization for headers/titles. ("GeoIP policy" not "GeoIP Policy".)

website/docs/policies/expression.mdx
```diff
@@ -55,18 +55,26 @@ import Objects from "../expressions/_objects.md";
...
+
     -   `continent`: a two character continent code like `NA` (North America) or `OC` (Oceania).
     -   `country`: the two character [ISO 3166-1](https://en.wikipedia.org/wiki/ISO_3166-1) alpha code for the country.
     -   `lat`: the approximate latitude of the location associated with the IP address.
     -   `long`: the approximate longitude of the location associated with the IP address.
     -   `city`: the name of the city. May be empty.
 
     ```python
-    return context["geoip"]["country"] == "US"
+    return context["geoip"]["continent"] == "EU"
```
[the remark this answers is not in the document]
gergosimonyi: It's possible, e.g. countries in Eurovision famously include Australia :slightly_smiling_face: 

```
    return context["geoip"]["continent"] == "EU" or context["geoip"]["country"] == "AU"
```

but I believe this might be more confusing than just keeping it simple. The params are listed right above, so I don't think "new confusion" arises from changing this.

(I changed this just to suggesting something other that what could be done with a GeoIP policy.)

tanberry (approved):
Thanks so much for updating the Docs!

kensternberg-authentik (approved):
Overall, this looks quite good.  I have a few quibbles about syntax and idioms, but they're more of a "I wouldn't do it this way, but it works and it's not wrong, so let's go with it." If you want to tighten up a few of the wordier expressions, feel free, but otherwise I'm happy.

authentik/policies/geoip/models.py
```diff
@@ -0,0 +1,92 @@
...
+
+        result = PolicyResult(passing, *messages)
+        result.source_results = results
+
+        return result
+
+    def passes_asn(self, request: PolicyRequest) -> PolicyResult:
+        # This is not a single get chain because `request.context` can contain `{ "asn": None }`.
+        asn_data = request.context.get("asn")
+        asn = asn_data.get("asn") if asn_data else None
```
rissson, line 61:
```suggestion
        asn = request.context.get("asn", {}).get("asn", None)
```
should work as well (the same applies for countries below)

authentik/policies/geoip/views.py
```diff
@@ -0,0 +1,20 @@
...
+from rest_framework.generics import ListAPIView
+from rest_framework.permissions import AllowAny
+
+
+class DetailedCountrySerializer(serializers.Serializer):
+    code = CountryField()
+    name = serializers.CharField()
+
+
+class ISO3166View(ListAPIView):
```
rissson, line 13: I think there is an argument for putting this in `api/` (or api.py) as well, for consistency with how we do things elsewhere.

rissson (approved):
Some nitpicking, but otherwise looks quite good!

authentik/policies/geoip/models.py
gergosimonyi: Unfortunately, no! If `request.context == { "asn": None }`, then `request.context.get("asn", {})` will happily return `None` and the second `.get()` will fail with

> AttributeError: 'NoneType' object has no attribute 'get'

authentik/policies/geoip/views.py
gergosimonyi: That makes sense. This used to be much more convoluted, but I should have moved it when I switched it to declarative style. Done!

authentik/policies/geoip/models.py
rissson: Correct, I didn't know you could get none in there

BeryJu (approved):

PR #10476, opened 2024-07-12
flows: remove stage challenge type

BeryJu:
## Details

The `type` field was heavily used during the migration from SSR to SPA/API, however nowadays we use `component` as the main discriminator, and so even when doing anything with flow challenge types, a new component needs to be defined. As a result of that we already have components for `redirect` and `shell`, which are also represented through types

---

## Checklist

If applicable

### DIFF
--- a/authentik/core/views/apps.py
+++ b/authentik/core/views/apps.py
@@ -8,7 +8,6 @@ from django.views import View
 from authentik.core.models import Application
 from authentik.flows.challenge import (
     ChallengeResponse,
-    ChallengeTypes,
     HttpChallengeResponse,
     RedirectChallenge,
 )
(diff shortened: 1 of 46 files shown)


go.mod
```diff
@@ -78,3 +78,5 @@ require (
 	google.golang.org/protobuf v1.33.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
+
+replace goauthentik.io/api/v3 => ./gen-go-api
```
rissson, line 82: `git add .` gone wrong

rissson (changes requested):

go.mod
BeryJu: Ah shit yeah, need to figure out a better way to do this without having to remember to not add this file

PR #10554, opened 2024-07-19
root: Make health checks compatible with cloud platform load balancers

fotinakis:
## Details

Authentik currently returns `HTTP 204` for `/-/health/live` and `/-/health/ready`. This PR changes these to `HTTP 200` responses.

This fixes compatibility with deploying Authentik to multiple cloud platforms, where the cloud load balancers strictly require `HTTP 200` responses for health checks and (strangely) consider `HTTP 204` responses as UNHEALTHY.

There is some nuance here... native Kubernetes pods allow readinessProbe & livenessProbe http checks to return anything between 2XX-3XX to be considered healthy, which is why authentik's Kubernetes setup works currently. However, when deployed in Google Kubernetes Engine for example, the authentik helm chart creates an Ingress controller — on create, the linked load balancer reads the `readinessProbe` of the underlying k8s services, and constructs a [Google Cloud Healthcheck](https://github.com/GoogleCloudPlatform/gke-networking-recipes/tree/main/ingress/single-cluster/ingress-custom-http-health-check) which _strictly_ requires an `HTTP 200` response and is not configurable at all. So even though the pod comes up healthy, the load balancer in front sees it as unhealthy.

**The result being: authentik server is permanently in an UNHEALHTY state** from the cloud load balancer's perspective, with absolutely no mechanism to change this without modifying authentik itself, or wrapping it in a proxy to change the responses.

This is pretty much true across cloud providers — load balancer implementation in most cloud providers require `200 OK` as a standard for health check:

[GCP](https://cloud.google.com/load-balancing/docs/health-checks#optional-flags-hc-protocol-http):

```
If Google Cloud finds [...] and if the HTTP status is 200 (OK),
the probe is considered successful.
```

[AWS](https://docs.aws.amazon.com/elasticloadbalancing/latest/classic/elb-healthchecks.html):

```
If the load balancer receives any response other than "200 OK" within the
response timeout period, the instance is considered unhealthy
```

[DigitalOcean](https://www.digitalocean.com/docs/app-platform/concepts/health-check/):

```
an HTTP health check can be configured to ping the root URL of your site and
ensure that the response is HTTP 200 (OK)
```

Also these precendents in other projects:
* https://github.com/meilisearch/meilisearch/issues/1282
* https://github.com/bitnami/charts/issues/2862 and https://github.com/influxdata/telegraf/issues/4935

I realize this may be considered a potentially backwards-incompatible change. I leave that to you authentik folks to decide what to do about that.

---

(Note: I wasn't able to get the full authentik dev environment up to get these pieces done — if you agree with this PR, can I get some help on this PR to do so from someone with a working dev env?)

## Checklist

If applicable

### DIFF
--- a/website/docs/installation/monitoring.md
+++ b/website/docs/installation/monitoring.md
@@ -6,7 +6,7 @@ authentik can be easily monitored in multiple ways.
 
 ## Server monitoring
 
-Configure your monitoring software to send requests to `/-/health/live/`, which will return a `HTTP 204` response as long as authentik is running. You can also send HTTP requests to `/-/health/ready/`, which will return `HTTP 204` if both PostgreSQL and Redis connections can be/have been established correctly.
+Configure your monitoring software to send requests to `/-/health/live/`, which will return a `HTTP 200` response as long as authentik is running. You can also send HTTP requests to `/-/health/ready/`, which will return `HTTP 200` if both PostgreSQL and Redis connections can be/have been established correctly.
 
 ## Worker monitoring
 
--- a/website/docs/installation/monitoring.md
+++ b/website/docs/installation/monitoring.md
@@ -8,6 +8,8 @@ authentik can be easily monitored in multiple ways.
 
 Configure your monitoring software to send requests to `/-/health/live/`, which will return a `HTTP 204` response as long as authentik is running. You can also send HTTP requests to `/-/health/ready/`, which will return `HTTP 204` if both PostgreSQL and Redis connections can be/have been established correctly.
 
+Updated in version ___: In previous versions, these endpoints returned an `HTTP 204` response.
+
 ## Worker monitoring
 
 The worker container can be monitored by running `ak healthcheck` in the worker container. This will ping the worker and ensure it can communicate with redis as required.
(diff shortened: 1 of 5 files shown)


website/docs/installation/monitoring.md
```diff
@@ -8,6 +8,8 @@ authentik can be easily monitored in multiple ways.
 
 Configure your monitoring software to send requests to `/-/health/live/`, which will return a `HTTP 204` response as long as authentik is running. You can also send HTTP requests to `/-/health/ready/`, which will return `HTTP 204` if both PostgreSQL and Redis connections can be/have been established correctly.
 
+Updated in version ___: In previous versions, these endpoints returned an `HTTP 204` response.
```
BeryJu, line 11: This should be added to the release notes and the docs changed here instead

BeryJu (approved):
It's silly that this is required but LGTM

website/docs/installation/monitoring.md
fotinakis: @BeryJu removed this line from the docs. I think you autogenerate changelog entries, so maybe we just need to update the PR title?

Would you be able to finish out this PR? Could use your help to get this over the line as I don't know what all is required (and I haven't been able to get a fully working dev env). I think regenerating the schema is also needed.

fotinakis:
@BeryJu — would you be able to finish out this PR? I know you'll be 100x faster than I would be — could use your help to get this over the line as I don't know all the authentik-specific dev pieces here (and I haven't been able to get a fully working dev env).

fotinakis:
Thanks so much @BeryJu !

PR #10595, opened 2024-07-23
website/docs: explain guarantees around blueprint ordering

fotinakis:
This PR updates documentation for how blueprint ordering works.

-- 
[previous description for posterity]

## Details

We have experimentally noticed flaky blueprint loading with custom blueprints when loading authentik from a blank slate. This seems to happen quite frequently—sometimes custom blueprints will work correctly, and sometimes they will fail silently without change. This happens especially when using custom blueprints that don't replace /blueprints/default but instead adds a custom directory that patches a default object.

Since there can be many dependencies between blueprints, the order in which each blueprint is discovered and loaded matters.

I believe this is flaky currently because `blueprints_find` is non-deterministic — it uses `Path.rglob` which does not guarantee any specific ordering and is filesystem-dependent.

This PR enforces a deterministic blueprint loading order by sorting the paths before load. (This is an easy and common pattern used in many "initialization" frameworks, like Rails bootup initializers).

Experimentally, this PR 100% eliminates these flakes and makes blueprint loading deterministic.

## Checklist

If applicable

### DIFF
--- a/authentik/blueprints/v1/tasks.py
+++ b/authentik/blueprints/v1/tasks.py
@@ -121,7 +121,7 @@ def blueprints_find() -> list[BlueprintFile]:
     """Find blueprints and return valid ones"""
     blueprints = []
     root = Path(CONFIG.get("blueprints_dir"))
-    for path in root.rglob("**/*.yaml"):
+    for path in sorted(root.rglob("**/*.yaml")):
         rel_path = path.relative_to(root)
         # Check if any part in the path starts with a dot and assume a hidden file
         if any(part for part in path.parts if part.startswith(".")):
--- a/authentik/blueprints/v1/tasks.py
+++ b/authentik/blueprints/v1/tasks.py
@@ -121,7 +121,7 @@ def blueprints_find() -> list[BlueprintFile]:
     """Find blueprints and return valid ones"""
     blueprints = []
     root = Path(CONFIG.get("blueprints_dir"))
-    for path in sorted(root.rglob("**/*.yaml")):
+    for path in root.rglob("**/*.yaml"):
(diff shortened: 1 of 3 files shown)


fotinakis:
@BeryJu a quick PR for your review when you get a moment. Thank you!

BeryJu:
While this will make the discovery of the blueprints deterministic, it will not make the execution order deterministic (it will be in most cases when only one worker container is running, however if there are more they could run at the same time)

fotinakis:
@BeryJu That makes sense. How about this — I've removed the documentation update because it's not strictly true and I understand you may not want to "publicly guarantee" this behavior — but would it work with you to leave the `sorted()` update and have this be a somewhat undocumented behavior? It very much helps the bootstrapping case where there is only 1 worker running and you're trying to create repeatable start-from-scratch authentik deployments with custom blueprints — I can confirm this PR 100% fixes some very hard-to-debug flakes in that case. We are probably going to just replace the entire `/blueprints/default` folder instead of patching the defaults, so this isn't a huge blocker if not.

rissson:
If you have dependencies between blueprints, you should use https://docs.goauthentik.io/developer-docs/blueprints/v1/meta#authentik_blueprintsmetaapplyblueprint to make sure that the required objects are created prior to using them.

fotinakis:
Ok — it sounds like this PR change is not something you all want to support.

I've updated it to be a documentation update PR instead, to document the existing defined behavior and what the best practice is.

rissson (approved):

tanberry (approved):
Thanks so much @fotinakis for updating the docs, and for the improvement on our blueprints loading!

PR #10631, opened 2024-07-25
policies: add unique password policy

### LINKED ISSUE #8307: Checking the new user password to make sure it is different from the previous ones
Can you please tell whether it is possible to configure the password parameters using authentik in such a way that a new password to be different from the previous four passwords? Can authentik check that the new password meets this requirement?
If not, are there plans to add this functionality in new versions?

verkaufer:
## Details

Closes #8307 

Adds a "Password Uniqueness" policy to the set of configurable flow policies.

Documentation PR: https://github.com/goauthentik/authentik/pull/11000

### How it works

After at least one (1) Flow has a UniquePasswordPolicy binding attached & enabled, the system records every user's **hashed** password when the user changes their own password.

When the user submits a new password during a password-change Flow, the system will check if the new password is identical to any of the user's previous passwords. The check will hash the new password against the [Hasher originally used](https://github.com/django/django/blob/main/django/contrib/auth/hashers.py#L167) to hash the old password. This ensures backwards compatiblity if Authentik decides to change hashing algorithsm.

This check only occurs if the Flow Stage has a Unique Password Policy attached.

### Configuration

Admins can configure how many previous passwords the system should evaluate. 

The configured number of passwords to evaluate also defines the number of passwords retained for each user.

Admins may also define which field they wish to use as the "password" field in a Stage. This is similar to the Password Policy configuration.

### Data considerations

The system purges the password history table after the last enabled UniquePasswordPolicy binding is deleted.

Because the number of passwords we save is configurable, it's important admins understand that whatever value they configure will end up saving up to `n^m` passwords, were `n` is the configured policy value & `m` is the number of users.

---

## Checklist

If applicable


authentik/core/tasks.py
```diff
@@ -92,3 +95,56 @@ def clean_temporary_users(self: SystemTask):
...
+
+
+@CELERY_APP.task(bind=True, base=SystemTask)
+@prefill_task
+def purge_password_history_table(self: SystemTask):
+    """Remove all entries from the core.models.UserPasswordHistory table"""
+    messages = []
+    try:
+        # n.b. a performance optimization to execute TRUNCATE
+        # instead of all().delete() would eliminate any FK checks.
```
verkaufer, line 107: We could use a `TRUNCATE` command for performance, but I opted for `.all().delete()` to maintain conventions.

authentik/events/signals.py
```diff
@@ -64,6 +68,26 @@ def on_user_write(sender, request: HttpRequest, user: User, data: dict[str, Any]
...
+        and request.user.pk == user.pk
+        and SESSION_KEY_IMPERSONATE_USER not in request.session
+    )
+    if user_changed_own_password:
+        unique_password_policies = UniquePasswordPolicy.objects.all()
+
+        unique_pwd_policy_binding = PolicyBinding.objects.filter(
+            policy__in=unique_password_policies
+        ).filter(enabled=True)
+
```
verkaufer, line 83: I reuse this query pattern a few times in different contexts. Would it be better to move this into a Manager method on `PolicyBinding`? Something like:

```python
PolicyBinding.objects.all_in_use(type=POLICY_MODEL_HERE)
```

authentik/policies/password/models.py
```diff
@@ -161,3 +162,83 @@ def passes_zxcvbn(self, password: str, request: PolicyRequest) -> PolicyResult:
...
+
+        # If no passwords are found: Policy resolves with “allow”
+        if not password_history:
+            return PolicyResult(True)
+
+        # For each password returned from audit table:
+        for history in password_history:
+            old_password = history.change.get("old_password")
+            if not old_password:
+                # TODO: how do we handle missing password?
```
verkaufer, line 220: Feedback requested: how should the policy behave if the `UserPasswordHistory` record doesn't have an `old_password` key?

authentik/core/models.py
```diff
@@ -887,3 +897,15 @@ def from_request(request: HttpRequest, user: User) -> Optional["AuthenticatedSes
             last_user_agent=request.META.get("HTTP_USER_AGENT", ""),
             expires=request.session.get_expiry_date(),
         )
+
+
+class UserPasswordHistory(models.Model):
+    user = models.ForeignKey("User", on_delete=models.CASCADE, related_name="old_passwords")
+    change = models.JSONField()
```
verkaufer, line 904: I opted for a JSON field over an individual `old_password` field for two reasons:

1. We'll likely never need to query on `old_password` and it's a good candidate for denormalization
2. It allows us to add additional additional data in this audit table without a migration. 

If Authentik prefers to avoid JSON columns I'm happy to switch this out.

authentik/policies/password/models.py
```diff
@@ -161,3 +162,83 @@ def passes_zxcvbn(self, password: str, request: PolicyRequest) -> PolicyResult:
...
+        return PolicyResult(True)
+
+    def _passwords_match(self, *, new_password: str, old_password: str) -> bool:
+        try:
+            hasher = identify_hasher(old_password)
+        except ValueError:
+            LOGGER.warning(
+                "Could not load hash algorithm for old password.",
+            )
+            # TODO: Define behavior if hasher cannot be identified or is unsupported
```
verkaufer, line 237: Should the policy fail if the Password hash algorithm was removed from Django? I know at least 2 are scheduled for deprecation in 5.1.

authentik/core/tasks.py
rissson: Agreed. But it's in a task and we'll probably never have any issue with this regardless, so let's prioritize consistency.

authentik/events/signals.py
rissson: I think it does make sense to do this indeed.

authentik/policies/password/models.py
rissson: I think we should just ignore unknown algos, and maybe schedule a task to remove ones we don't know about?

authentik/events/signals.py
```diff
@@ -64,6 +68,26 @@ def on_user_write(sender, request: HttpRequest, user: User, data: dict[str, Any]
     data["created"] = kwargs.get("created", False)
     Event.new(EventAction.USER_WRITE, **data).from_http(request, user=user)
 
+    """Preserve the user's old password if UniquePasswordPolicy is enabled anywhere"""
```
rissson, line 71: This should really listen for a `password_changed` signal, but we don't have access to the request in there :/

authentik/core/tasks.py
```diff
@@ -92,3 +95,56 @@ def clean_temporary_users(self: SystemTask):
...
+            num_rows_to_preserve = max(
+                num_rows_to_preserve, binding.policy.num_historical_passwords
+            )
+
+    preservable_row_ids = (
+        UserPasswordHistory.objects.filter(user__pk=user_pk)
+        .order_by("-created_at")[:num_rows_to_preserve]
+        .values_list("id", flat=True)
+    )
+    num_deleted, _ = UserPasswordHistory.objects.exclude(pk__in=list(preservable_row_ids)).delete()
```
rissson, line 147: Those two queries can probably be merged into one

authentik/policies/password/models.py
```diff
@@ -161,3 +162,83 @@ def passes_zxcvbn(self, password: str, request: PolicyRequest) -> PolicyResult:
     class Meta(Policy.PolicyMeta):
         verbose_name = _("Password Policy")
         verbose_name_plural = _("Password Policies")
+
+
+class UniquePasswordPolicy(Policy):
```
rissson, line 167: So usually we'd have one app per policy type. This (and API things, etc.) should move to an authentik.policies.unique_password app

authentik/core/models.py
```diff
@@ -887,3 +897,15 @@ def from_request(request: HttpRequest, user: User) -> Optional["AuthenticatedSes
             last_user_agent=request.META.get("HTTP_USER_AGENT", ""),
             expires=request.session.get_expiry_date(),
         )
+
+
+class UserPasswordHistory(models.Model):
```
rissson, line 902: This should probably go alongside the uniquepasswordpolicy model, in its own app (see my comment on that specific model.)

rissson (changes requested):
Some design comments, most of which will need to be checked by @BeryJu anyway

No specific review on the frontend, cause I wouldn't know what I'm saying.

I think we should also provide a way for administrators to clear the password history of a specific user (and all users?) so that a user can be unblocked from changing their password if needed.

All in all, a pretty solid implementation

authentik/core/tasks.py
verkaufer: Yep, agreed. I'll remove this comment

authentik/core/models.py
verkaufer: > Adding migrations is not that much of an issue for existing installs (especially adding a new column should be quite fast).

Wasn't sure if there were concerns around adding migrations or not. Sounds like it's not a problem, though. 

I've run into situations in the past where doing a migration would take a long time if the affected tables had an FK to another table that was frequently taking locks for `UPDATE` 

---
💥 I will change this model to use a field for the previous password instead of JSON and make updates accordingly.

authentik/policies/password/models.py
verkaufer: Ignoring them works for me. Django won't be able to compare against that password, and in this implementation it might as well not exist.

- [ ] Ignore passwords where we can't identify the algo

> maybe schedule a task to remove ones we don't know about
I'm not sure the overhead of running a task (and spending the time to write it) would get us much in return. Passwords are already cleared out in a LIFO-esque way, so eventually the unknown password would age out.

authentik/policies/password/models.py
verkaufer: Not a problem, I'll get that moved.

I think a README at the root of `policies` explaining built-in policy is its own app would help future developers too :+1: I'll throw one in.

- [ ] Move UniquePasswordPolicy to its own app
- [ ] add README.md in the root of `policies` explaining each policy should be an app

authentik/policies/password/models.py
```diff
@@ -161,3 +162,83 @@ def passes_zxcvbn(self, password: str, request: PolicyRequest) -> PolicyResult:
...
+            )
+            return PolicyResult(False, _("Password not set in context"))
+        password = str(password)
+
+        # Query audit table for the last n passwords
+        password_history = UserPasswordHistory.objects.filter(user=request.user)
+
+        # If no passwords are found: Policy resolves with “allow”
+        if not password_history:
+            return PolicyResult(True)
```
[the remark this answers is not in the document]
verkaufer: Good catch, I'll simplify.

authentik/policies/password/models.py
```diff
@@ -161,3 +162,83 @@ def passes_zxcvbn(self, password: str, request: PolicyRequest) -> PolicyResult:
...
+            self.password_field, request.context.get(self.password_field)
+        )
+        if not password:
+            LOGGER.warning(
+                "Password field not set in Password Uniqueness Policy Request",
+                field=self.password_field,
+                fields=request.context.keys(),
+            )
+            return PolicyResult(False, _("Password not set in context"))
+        password = str(password)
```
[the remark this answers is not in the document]
verkaufer: It should be 🤷 The other Password policy applies the same `str` casting and I figured that was done on purpose. 

I can remove if @BeryJu feels the same way

authentik/policies/signals.py
```diff
@@ -42,3 +43,22 @@ def invalidate_policy_cache(sender, instance, **_):
...
+    unique_password_policies = UniquePasswordPolicy.objects.all()
+
+    policy_binding_qs = PolicyBinding.objects.filter(policy__in=unique_password_policies).filter(
+        enabled=True
+    )
+
+    if policy_binding_qs.count() > 1:
+        # No-op; A UniquePasswordPolicy binding other than the one being deleted still exists
+        return
+    purge_password_history_table.delay()
```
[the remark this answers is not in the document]
verkaufer: Makes sense

>we avoid users dispatching the task and emptying the table by mistake.

Can you elaborate? I'm not sure how a user could dispatch this task without first deleting a PolicyBinding 🤔

authentik/core/models.py
verkaufer: I can see both sides; keeping it next to the User model keeps it grouped with User-related concerns, while putting it in with the policy sub-app makes the dependency clear.

I'll wait for @BeryJu to chime in on this one before I move it.

authentik/policies/password/models.py
rissson: I'm not sure how much you've poked around the code, but we do this for pretty much everything. Each provider/source type is its own app for instance. It's not really specific to policies

authentik/core/models.py
rissson: True. Another somewhat related instance where we put the model in the app is with stage_consent, where the user consents are in the app instead of core. Core is quite crowded already, which is why I suggested this initially. Let's wait for what Jens has to say

authentik/policies/signals.py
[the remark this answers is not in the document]
rissson: That task is based on `SystemTask`, which is shown in the UI for the admins to see and interact with, including dispatching it manually.

authentik/core/tasks.py
verkaufer: Are you suggesting merging for performance or to avoid having 2 querysets assigned to two variables? 

imo this is not as readable:
```python

    num_deleted, _ = UserPasswordHistory.objects.exclude(
        pk__in=list(
            UserPasswordHistory.objects.filter(user__pk=user_pk)
            .order_by("-created_at")[:num_rows_to_preserve]
            .values_list("id", flat=True)
        )
    ).delete()
```

authentik/events/signals.py
verkaufer: Moved to the UniquePasswordPolicy app :+1:

authentik/policies/signals.py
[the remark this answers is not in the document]
verkaufer: done in https://github.com/goauthentik/authentik/pull/10631/commits/7d3b5980d91adb5610c665b335fd1be5b648014b

authentik/core/tasks.py
rissson: Oh yeah, I missed the slicing. It should be fine as is then

authentik/core/models.py
verkaufer: Ended up moving UserPasswordHistory into `policies.unique_password` as you suggested :+1:

verkaufer:
@rissson I've applied the suggested changes. Namely:

- move `UniquePasswordPolicy` to its own `policies` sub-app
- move `UserPasswordHistory` out of core and into `policies.unique_password`
- add a Model Manager for finding all instances of in-use PolicyBindings for a Policy type[1]: `PolicyBindings.in_use.for_policy(<Policy>)`

❗ One issue I'm stumped on:
When I moved the "write password on `user_write` signal" code, I'm no longer able to get my `test_stages.py`tests to pass. It seems like `user_write` is not received inside the `policies.unique_password` package. 

When I had the code alongside the `events` listener for `user_write` it worked 🤔 

----
[1] Name of the model manager method is definitely up for discussion

verkaufer:
Also I'll rebase and get the schema conflict resolved before merging

authentik/policies/unique_password/apps.py
```diff
@@ -0,0 +1,9 @@
+"""authentik Unique Password policy app config"""
+
+from django.apps import AppConfig
+
+
+class AuthentikPoliciesUniquePasswordConfig(AppConfig):
```
rissson, line 6:
```suggestion
from authentik.blueprints.apps import ManagedAppConfig


class AuthentikPoliciesUniquePasswordConfig(ManagedAppConfig):
```

> When I moved the "write password on user_write signal" code, I'm no longer able to get my test_stages.pytests to pass. It seems like user_write is not received inside the policies.unique_password package.

By default, Django doesn't load whatever is in `signals.py`. [This wrapper](https://github.com/goauthentik/authentik/blob/main/authentik/blueprints/apps.py#L49) does it for us, and also loads tasks if they exist, and so on.

authentik/policies/models.py
```diff
@@ -142,6 +149,9 @@ def __str__(self) -> str:
             return f"Binding - #{self.order} to {suffix}"
         return ""
 
+    objects = models.Manager()
+    in_use = BoundPolicyQuerySet.as_manager()
```
rissson, line 153:
```suggestion
    objects = BoundPolicyQuerySet.as_manager()
```

Not fully sure about how it's done elsewhere, but I think this is the most common pattern I've seen in the codebase. Just a nitpick though, if you feel it makes sense to have it as a separate manager, fine by me.

authentik/policies/models.py
```diff
@@ -48,6 +48,13 @@ def __str__(self) -> str:
         return f"PolicyBindingModel {self.pbm_uuid}"
 
 
+class BoundPolicyQuerySet(models.QuerySet):
+    """QuerySet for filtering enabled bindings for a Policy type"""
+
+    def for_policy(self, policy: "Policy"):
+        return self.filter(policy__in=policy._default_manager.all()).filter(enabled=True)
```
rissson, line 55: Just to make sure I understand, why are we referring to the default manager here?

authentik/policies/models.py
verkaufer: I prefer keeping this a separate manager for now, given it's our first instance of needing to check actively used policies.

authentik/policies/models.py
verkaufer: I chose `_default_manager` because I wanted to preserve the behavior of the `InheritanceManager()` managers on the different `Policy` subclasses. Is that unnecessary here?

verkaufer:
Added public documentation in a separate PR: https://github.com/goauthentik/authentik/pull/11000

melizeche:
close for #13453

PR #10636, opened 2024-07-25
stages/authenticator: add created, last_updated and last_used metadata

rissson:
## Details

@BeryJu I need your guidance where to update `last_used`.

---

## Checklist

If applicable


authentik/core/api/devices.py
```diff
@@ -20,6 +20,9 @@ class DeviceSerializer(MetaNameSerializer):
     name = CharField()
     type = SerializerMethodField()
     confirmed = BooleanField()
+    created = DateTimeField(read_only=True)
+    last_updated = DateTimeField(read_only=True)
+    last_used = DateTimeField(read_only=True)
```
rissson, line 25: May need a `required=False` or something because that value can be null.

authentik/stages/authenticator/models.py
```diff
@@ -85,6 +86,8 @@ class Device(models.Model):
 
     confirmed = models.BooleanField(default=True, help_text="Is this device ready for use?")
 
+    last_used = models.DateTimeField(null=True)
```
rissson, line 89: We already have `last_t` for some authenticators.

I went with a new common field for all authenticators, that is only populated once the authenticator has been used in other contexts than setup.

kensternberg-authentik (approved):
Looks straightforward enough.

authentik/stages/authenticator_webauthn/migrations/0012_webauthndevice_created_webauthndevice_last_updated_and_more.py
```diff
@@ -0,0 +1,30 @@
...
+    operations = [
+        migrations.AddField(
+            model_name="webauthndevice",
+            name="created",
+            field=models.DateTimeField(auto_now_add=True, default=datetime.datetime(1, 1, 1, 0, 0)),
+            preserve_default=False,
+        ),
+        migrations.AddField(
+            model_name="webauthndevice",
+            name="last_updated",
```
BeryJu, line 22: We _might_ want to backfill these values based on events...? Not sure if we can reliably get the data

BeryJu (approved):

rissson:
@BeryJu we're not updating last_used anywhere

PR #8771, opened 2024-03-01
sources: add property mappings for all oauth and saml sources

rissson:
## Details

Next step of #8750.

Also includes #6427

closes #6184
closes #6622
closes #7713 

--- 

TODO:

- [x] SAML group infos
- [x] flow manager group handling
  - [x] tests
- [x] Oauth tests
  - [x] user pm
  - [x] group pm
- [x] SAML tests
  - [x] user pm
  - [x] group pm

---

## Checklist

-   [ ] Local tests pass (`ak test authentik/`)
-   [ ] The code has been formatted (`make lint-fix`)

If an API change has been made

-   [ ] The API schema has been updated (`make gen-build`)

If changes to the frontend have been made

-   [ ] The code has been formatted (`make web`)

Docs at #10652

PR #10651, opened 2024-07-26
core: add primitives for source property mappings

rissson:
## Details

Extracted from #8771 and #10650 for easier review.

---

## Checklist

If applicable

### DIFF
--- a/authentik/core/api/property_mappings.py
+++ b/authentik/core/api/property_mappings.py
@@ -2,8 +2,15 @@
 
 from json import dumps
 
+from django_filters.filters import AllValuesMultipleFilter, BooleanFilter
+from django_filters.filterset import FilterSet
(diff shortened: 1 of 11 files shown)


authentik/core/api/property_mappings.py
```diff
@@ -67,6 +74,18 @@ class Meta:
         ]
 
 
+class PropertyMappingFilterSet(FilterSet):
+    """Filter for PropertyMapping"""
+
+    managed = extend_schema_field(OpenApiTypes.STR)(AllValuesMultipleFilter(field_name="managed"))
+
+    managed__isnull = BooleanFilter(field_name="managed", lookup_expr="isnull")
```
rissson, line 82: Kept this filter as-is (i.e. no changes to the API), but we can rename it if we want.

BeryJu (approved):

PR #10690, opened 2024-07-30
website/docs: integrations: add what's up docker

dominic-r:
## Details

REPLACE ME
adds documentation to configure authentik as a sso provider for wud

---

## Checklist

If applicable

### DIFF
--- a/website/integrations/services/whats-up-docker/index.md
+++ b/website/integrations/services/whats-up-docker/index.md
@@ -1,0 +1,43 @@
+---
+title: What's Up Docker
+---
+
+<span class="badge badge--secondary">Support level: Community</span>
+
+## What is What's Up Docker
+
+> WUD is an easy-to-use tool that alerts you whenever a new version of your Docker containers is released.
+>
+> -- https://fmartinou.github.io/whats-up-docker/
+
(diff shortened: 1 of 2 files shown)


website/integrations/services/whats-up-docker/index.md
```diff
@@ -0,0 +1,43 @@
...
+## Preparation
+
+The following placeholders will be used:
+
+-   `wud.company` is the FQDN of the WUD install.
+-   `authentik.company` is the FQDN of the authentik install.
+
+## WUD configuration
+
+To configure WUD to use authentik, add the following values to your `.env` file:
```
tanberry, line 22: Which .env file is this? Is it the one authentik reads or one that WUD reads or is that the same file? ;-)

tanberry (approved):
Thanks @4d62 ! I can't recall if it was an earlier review of this PR or a diff PR, but somewhere I mentioned that when the steps for authentik configuration require creating a provider and an application, we can start documenting the process of using the wizard for that. The wizard will soon become the default way to create a provider-app pair, so we might as well start documenting the steps that way. Up to you if you want to do it here or not. :-)

PR #10705, opened 2024-07-31
provider/proxy: allow Cookie Domain with port when using ForwardAuth(domain level)

RMTT:
## Details

Authentik now doesn't support Cookie Domain with port when using ForwardAuth(domain level). For example, when set `Cookie domain` as `localhost:1443`, following error will occur:
```
2024/07/31 10:09:05 net/http: invalid Cookie.Domain "localhost:1443"; dropping domain attribute
```

The root cause of this error is go's `net/http` lib will check whether domain is legal, the `:` is illegal, reference https://github.com/golang/go/blob/eea2e929fae6abbc1422f8ee26de02a4b7e4f8cd/src/net/http/cookie.go#L369

I think we can enable cookie domain with port in application level, it's convenient for some circumstances.

## Checklist

If applicable

### DIFF
--- a/internal/outpost/proxyv2/application/session.go
+++ b/internal/outpost/proxyv2/application/session.go
@@ -77,10 +77,12 @@ func (a *Application) getStore(p api.ProxyOutpostConfig, externalHost *url.URL) 
 		}
 
 		rs.KeyPrefix(RedisKeyPrefix)
+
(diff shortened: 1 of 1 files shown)


rissson:
Closing since that code does not exist anymore. Please see if the issue still exists in 2026.8

PR #10728, opened 2024-08-01
web: search select with focus, autocomplete, and progressive search

kensternberg-authentik:
web: search select with focus and progressive search

### What:

- New component: ak-list-select, which allows you to select from a list of elements, with keyboard control.
- New component: ak-portal, which manages elements by moving “slotted” content into a distant component, usually one attached to the body, and positions it relative to an existing element.
- ak-search-select-view has been revamped to handle focus, change, input, and blur using the browser native event handlers, rather than inventing my own.
- ak-search-select has been turned into a simple driver that manages the view.
- ak-search-select has a new declarative syntax for the most common use case.

### Why:

I seriously over-engineered this thing, leaning too heavily on outdated knowledge or assumptions about how the browser works. The native event handlers attached at the component’s borders works more than fine, and by attaching the event handlers to the portaled component before sending it off to the slots, the correct handlers get the message. This revision leverages the browser a *lot* more, and gets much more effective interaction with much less code.

#### ak-list-select

`<ak-list-select>` is a new component that replaces the ad-hoc menu object of the old SearchSelect. It is a standalone component that just shows a list, allows someone to navigate that list with the keyboard or the mouse. By default, it is limited to half the height of the viewport.

The list does not have an indicator for “selected” at this time. That’s just a side effect of it being developed as an adjunct to search-select. Its design does not preclude extension.

It has a *lot* of CSS components that can be customized. The properties and events are documented, but there is only one event: `change`. Consistent with HTML, the value is not sent with the `change` event; clients are expected to extract it with `change:event.target.value`.

Like all HTML components, it is completely stringly defined; the value is either a string or undefined.

#### ak-portal

`<ak-portal>` is a somewhat specialized “portal” component that places an `ak-list-select` in an object on top of the existing DOM content. It can be generalized to do this with any component, though, and can be extended. It has no events or CSS, since it’s “just” managing the portaling relationship.

#### ak-search-select-view

`<ak-search-select-view>` is the heart of the system. It takes a collection options and behaves like an autocomplete component for them. The only unique event it sends out is `change`, and like `ak-list-select`, it expects the client to retrieve the value.

Like all HTML components, it is completely stringly defined; the value is either a string or undefined.

#### ak-search-select

This is the SearchSelect component we’ve all known to come and love, but with a better pop-up and cleaner keyboard interaction. It emits only one event, `ak-change`, which *does* carry the value with it.

### Storybook & Tests

The Storybooks have been updated to show the current version of Search Select, with a (simulated) API layer as well as more blunt stringly-typed tests for the View layer. A handful of tests have been provided to cover a number of edge cases that I discovered during testing. These run fine with the `npx` command, and I would love to see them integrated into CI/CD.

### Developer Experience Enhancement!

The search select fields `renderElement`, `renderDescription`, and `value` properties of `ak-search-select` have been modified to take a string. For example, the search for the list of user looks like this:

    <ak-search-select
        .fetchObjects=${async (query?: string): Promise<User[]> => {
            const args: CoreUsersListRequest = { ordering: "username" };
            if (query !== undefined) {
                args.search = query;
            }
            const users = await new CoreApi(DEFAULT_CONFIG).coreUsersList(args);
            return users.results;
        }}
        .renderElement=${(user: User): string => {
            return user.username;
        }}
        .renderDescription=${(user: User): TemplateResult => {
            return html`${user.name}`;
        }}
        .value=${(user: User | undefined): string | undefined => {
            return user?.username;
         }}
    ></ak-search-select>

Since the Web Component API for integrating with the authentik TS API is a uniform collection of fields that can't be made into attributes, I've merged them into a single configuration object, which I think is easier to read:

    <ak-search-select-ez
        .config=${{
            fetchObjects: async (query?: string): Promise<Group[]> => {
                const args: CoreGroupsListRequest = {
                    ordering: "name",
                    includeUsers: false,
                    ... query ? { search: query } : {}
                };
                return (await new CoreApi(DEFAULT_CONFIG).coreGroupsList(args)).results;
             },
            renderElement: (group: Group): string => group.name,
            value: (group: Group | undefined): string | undefined => group?.pk,
            selected: (group: Group): boolean => group.pk === this.instance?.group
         }}
        blankable
    >
    </ak-search-select-ez>
    
It's possible to tighten up the `fetch` event further:

            fetchObjects: async (search?: string): Promise<Group[]> =>
                (
                    await new CoreApi(DEFAULT_CONFIG).coreGroupsList({
                        ordering: "name",
                        includeUsers: false,
                        ...(search ? { search } : {}),
                    })
                ).results,

If applicable


authentik/root/settings.py
```diff
@@ -516,7 +516,7 @@ def _update_settings(app_path: str):
     REST_FRAMEWORK["DEFAULT_RENDERER_CLASSES"].append(
         "rest_framework.renderers.BrowsableAPIRenderer"
     )
-    SHARED_APPS.append("daphne")
+    SHARED_APPS.insert(SHARED_APPS.index("django.contrib.staticfiles"),"daphne")
```
kensternberg-authentik: Not sure why this is showing up.  I thought we'd handled this?  @BeryJu?

web/src/elements/ak-dual-select/types.ts
```diff
@@ -2,9 +2,21 @@ import { TemplateResult } from "lit";
 
 import { Pagination } from "@goauthentik/api";
 
-// Key, Label (string or TemplateResult), (optional) string to sort by. If the sort string is
-// missing, it will use the label, which doesn't always work for TemplateResults).
-export type DualSelectPair<T = never> = [string, string | TemplateResult, string?, T?];
+//
+// - key: string
```
kensternberg-authentik, line 6: This is pure "cleaning up the documentation."  Typescript's "named tuples" are lovely for self-documenting code.

web/package.json
```diff
@@ -120,7 +120,7 @@
         "eslint-config-google": "^0.14.0",
         "eslint-plugin-custom-elements": "0.0.8",
         "eslint-plugin-lit": "^1.11.0",
-        "eslint-plugin-sonarjs": "^1.0.3",
+        "eslint-plugin-sonarjs": "^0.25.1",
```
kensternberg-authentik, line 123: How do I tell Dependabot to leave our stuff alone?

web/scripts/eslint-nightmare.mjs
```diff
@@ -27,17 +28,29 @@ const eslintConfig = {
...
         },
         plugins: ["@typescript-eslint", "lit", "custom-elements", "sonarjs"],
+        ignorePatterns: ["authentik-live-tests/**", "./.storybook/**/*.ts"],
         rules: {
             "indent": "off",
             "linebreak-style": ["error", "unix"],
             "quotes": ["error", "double", { avoidEscape: true }],
             "semi": ["error", "always"],
             "@typescript-eslint/ban-ts-comment": "off",
-            "sonarjs/cognitive-complexity": ["error", 9],
```
kensternberg-authentik, line 38: Just some revisions to nightmare mode.  :-)

web/src/admin/common/ak-flow-search/ak-flow-search.stories.ts
```diff
@@ -61,7 +61,7 @@ const mockData = {
 };
 
 const metadata: Meta<AkFlowSearch<Flow>> = {
-    title: "Elements / Select Search / Flow",
+    title: "Elements / Search Select / Flow",
```
kensternberg-authentik, line 64: Embarassing!

web/src/admin/providers/ldap/LDAPProviderForm.ts
```diff
@@ -86,12 +86,8 @@ export class LDAPProviderFormPage extends WithBrandConfig(BaseProviderForm<LDAPP
...
                         return groups.results;
                     }}
-                    .renderElement=${(group: Group): string => {
-                        return group.name;
-                    }}
-                    .value=${(group: Group | undefined): string | undefined => {
-                        return group?.pk;
-                    }}
+                    .renderElement=${"name"}
+                    .value=${"pk"}
```
kensternberg-authentik, line 90: My demo that the new shorter declaration scheme works.  And it does!

web/src/elements/types.ts
```diff
@@ -9,3 +10,66 @@ export type Constructor<T = object> = new (...args: any[]) => T;
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 export type AbstractConstructor<T = object> = abstract new (...args: any[]) => T;
+
```
kensternberg-authentik, line 13: One thing that... irritates... me is that ak-dual-select does *not* adhere to this protocol.  It would be nice if all of our search-and-select tools did.  It's a YAGNI right now, but it's on my mind.

kensternberg-authentik (commented):
Just leaving roadmarks.

web/package.json
BeryJu: I guess https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#ignore?

web/src/admin/providers/ldap/LDAPProviderForm.ts
BeryJu: While I do agree this is much shorter and concise, we loose the compile-time type checking of typescript and I think this could make it very easy to introduce bugs

BeryJu (approved):

PR #10745, opened 2024-08-03
internal: Use loop instead of recursion in NewAPIController

RMTT:
## Details

Use loop instead of recursion in  `NewAPIController`.

Golang has no tail call optimization now, so the recursion `return NewAPIController(akURL, token)` will cause the stack too deep when OutpostsApi failed for a long time.

---

## Checklist

If applicable


BeryJu (commented):
While this should be fine, I'm not 100% sure if this works as intended in all cases as there might be some other state that the current implementation relies on being reset

RMTT:
> While this should be fine, I'm not 100% sure if this works as intended in all cases as there might be some other state that the current implementation relies on being reset

Seems the parameters are same in every recursion, so i think it's probably ok.

PR #10799, opened 2024-08-06
web, core: fix grammatical issue in stage bindings

kensternberg-authentik:
## Details

Global search & replace the phrase "...when the Stage is present to the user." with "...when the Stage is presented to the user."

Because it's bad grammar.  Since the source language phrase has been changed in all cases, it will have the same hash and the resulting translations will still be targeted the same.


tanberry (approved):
Wow, I had no idea that this would involve more than one file, hahaha. Hilarious that I thought I could fix it. Thank you!

rissson (approved):

BeryJu (approved):

PR #10652, opened 2024-07-26
website/docs: add source property mappings, rework provider property mappings

rissson:
## Details

Docs for #8771 and #10650 and #10772

Needs #10810

---

If applicable

-   [x] The documentation has been updated
-   [x] The documentation has been formatted (`make website`)

PR #10810, opened 2024-08-07
web/admin: refactor property mappings forms

rissson:
## Details

REPLACE ME

---

## Checklist

If applicable


web/src/admin/property-mappings/BasePropertyMappingForm.ts
```diff
@@ -1,11 +1,60 @@
...
+interface PropertyMapping {
+    name: string;
+    expression: string;
+}
+
+export abstract class BasePropertyMappingForm<T extends PropertyMapping> extends ModelForm<
+    T,
+    string
+> {
+    docLink(): string {
```
rissson, line 18: This is needed for the changes in #10652. I'll update that PR once this one gets merged, so we can have different docs for provider/source.

BeryJu (commented):
as @kensternberg-authentik pointed out, `?required=${true}` can be replaced with `required` for the same effect, across all these files

BeryJu (approved):

PR #10830, opened 2024-08-08
web: move `common` into its own package

kensternberg-authentik:
web: move `common` into its own package.

### What:

    $ mkdir ./packages/common
    $ git mv ./src/common ./packages/common/src

… and then added all of the boilerplate needed to drive with Wireit, build with ESlint, typecheck with TSC, and then spell check documentation and comments, security checks of package.json and package-lock.json, format.

… and *then* fix all of the minor, nitpicky things ESLint 9 found in the package.

… and *then* wire the whole thing into our build so that we can find it as a package, removing it as an alias from the base package definition and turning it into a workspace. Although it is a workspace package, it’s currently configured to build completely independently.

It could be published as an independent NPM package, although I don’t recommended that at this time.

### Why

I’ve wanted to break the UI up into smaller, more digestible chunks for awhile, but was always reluctant to, since I didn’t want to mess with other teams’ mental models of the code layout. @Beryju, seeing the success of the Simple Flow Executor as an independent package, thought it might be worthwhile to see what effort it took to break the graph of our independent apps (User, Flow, and Admin) and their dependencies (Common \<- Elements \<- Components, Common \<- Locales) into packages.

Turns out, it’s not too bad. It’s going to be fiddly for awhile until things settle down, but overall the experiment has been a success.

### Particulars

The `tsconfig.json` doesn’t refer to the base because we want this to build independently; tooling will be needed to ensure all of our `tsconfig` files in the future will be consistent across all packages.

- We can use the ESLint boilerplate as-is.
- We have to run TSC as a separate (but fortunately parallel) build step, as client code will need the built types. Final builds will be fractionally slower, but Wireit can detect when a monorepo package is unchanged and can skip rebuilding `common` if it’s not needed, so the development loop will be faster.
- The ESBuild boilerplate is different for libraries with UI, libraries without UI (like this one), and apps, and we’ll have to have three different routines for them. Once we are building independent *apps*, getting them into the `dist` folder will be an interesting challenge; we may end up with two different builds, one to bundle it in *in the app*, and another to bundle it *for Django*. That’s mostly an issue of targeting and integration, and shouldn’t take too much time.
- Spelling, formatting, and package checking aren’t affected.
- `Locales` is our biggest challenge, as usual. I have found only [one article on it anywhere](https://medium.com/tech-at-zet/streamlining-localization-in-a-monorepo-using-i18n-js-e7c521ff69d4), and it recommends creating a single package in which to keep all of the localizations and the localization machinery. That seems like a sound approach, but we haven’t (yet) gotten there.

### Commentary

`common` is a bit of a junk drawer: there are global utilities in there, there are app-specific helpers, there are plug-in specific helpers, and so on. Figuring out exactly what does what and making more specific packages may be in our future.


web/packages/common/src/api/config.ts
```diff
@@ -2,12 +2,32 @@ import {
...
-} from "@goauthentik/common/api/middleware";
-import { EVENT_LOCALE_REQUEST, VERSION } from "@goauthentik/common/constants";
-import { globalAK } from "@goauthentik/common/global";
+} from "@goauthentik/common/api/middleware.js";
+import { EVENT_LOCALE_REQUEST, VERSION } from "@goauthentik/common/constants.js";
+import { globalAK } from "@goauthentik/common/global.js";
 
 import { Config, Configuration, CoreApi, CurrentBrand, RootApi } from "@goauthentik/api";
 
+export function getMetaContent(key: string): string {
```
kensternberg-authentik, line 11: The lint continues to be picked.  Here it's "DEFAULT_CONFIG used before definition."

web/packages/common/src/api/middleware.ts
```diff
@@ -17,6 +17,8 @@ export interface RequestInfo {
     status: number;
 }
 
+const HTTP_BAD_REQUEST = 400;
```
kensternberg-authentik, line 20: No magic numbers.

web/packages/common/src/events.ts
```diff
@@ -8,6 +8,13 @@ export interface EventUser {
     is_anonymous?: boolean;
 }
 
+export interface EventModel {
```
kensternberg-authentik, line 11: "EventModel" is used before definition.

web/packages/common/src/helpers/webauthn.ts
```diff
@@ -3,7 +3,7 @@ import * as base64js from "base64-js";
 import { msg } from "@lit/localize";
 
 export function b64enc(buf: Uint8Array): string {
-    return base64js.fromByteArray(buf).replace(/\+/g, "-").replace(/\//g, "_").replace(/=/g, "");
+    return base64js.fromByteArray(buf).replace(/\+/g, "-").replace(/\//g, "_").replace(/[=]/g, "");
```
kensternberg-authentik, line 6: Apparently, `/=/` can be mis-parsed by some tools as an equation, so this is the recommended fix.

web/packages/common/src/sentry.ts
```diff
@@ -15,6 +15,17 @@ import { CapabilitiesEnum, Config, ResponseError } from "@goauthentik/api";
 
 export const TAG_SENTRY_COMPONENT = "authentik.component";
 export const TAG_SENTRY_CAPABILITIES = "authentik.capabilities";
+const MIN_PATH_LENGTH = 2;
+
+// Get the interface name from URL
+export function currentInterface(): string {
```
kensternberg-authentik, line 21: Used-before-defined.
No-magic-numbers.

web/packages/common/src/users.ts
```diff
@@ -61,3 +62,8 @@ export function me(): Promise<SessionUser> {
     }
     return globalMePromise;
 }
+
+export function refreshMe(): Promise<SessionUser> {
+    globalMePromise = undefined;
+    return me();
```
kensternberg-authentik, line 68: `me` used-before-defined
no-magic-numbers

web/packages/common/src/utils.ts
```diff
@@ -9,7 +9,7 @@ export function getCookie(name: string): string {
         for (let i = 0; i < cookies.length; i++) {
             const cookie = cookies[i].trim();
             // Does this cookie string begin with the name we want?
-            if (cookie.substring(0, name.length + 1) === name + "=") {
+            if (cookie.substring(0, name.length + 1) === `${name}=`) {
```
kensternberg-authentik, line 12: unexpected-string-concat

web/packages/common/src/utils.ts
```diff
@@ -151,13 +155,26 @@ export function adaptCSS(sheet: AdaptableStylesheet | AdaptableStylesheet[]): Ad
     return Array.isArray(sheet) ? sheet.map(_adaptCSS) : _adaptCSS(sheet);
 }
 
+const SECONDS_IN_A_MINUTE = 60;
```
kensternberg-authentik, line 158: Yes, this is what the Eslint people actually recommend.

web/src/admin/AdminInterface/AdminInterface.ts
```diff
@@ -1,12 +1,12 @@
 import { ROUTES } from "@goauthentik/admin/Routes";
-import { DEFAULT_CONFIG } from "@goauthentik/common/api/config";
```
kensternberg-authentik, line 2: Everything below here is this: Because "common" is a grab-bag of things, they want '.js' suffixes on everything we pull out of it.  Sad, but true.  Only the last two entries at the end of this page have anything different, and they're the changes to `tsconfig` to understand that `@goauthentik/common` is no longer an alias to a path, it's the name of a package.

web/tsconfig.json
```diff
@@ -3,7 +3,6 @@
     "compilerOptions": {
         "paths": {
             "@goauthentik/admin/*": ["./src/admin/*"],
-            "@goauthentik/common/*": ["./src/common/*"],
```
kensternberg-authentik, line 6: Real work resumes.  Common is now a package, not a path.

BeryJu (approved):

Issue #6586, opened 2023-08-21
Facing SSL certificate issue while attempting to create a generic webhook in Authentik's notification transport. 


chokhadavikas:
**Describe your question/**
Facing SSL certificate issue while attempting to create a generic webhook in Authentik's notification transport to external API. 

**Relevant infos**
Problem Description: Whenever I try to create a generic webhook and configure it to hit an external Spring Boot API, I encounter an SSL error. The error message indicates that there's a certificate verification problem.

"HTTPSConnectionPool(host='example.com', port=443): Max retries exceeded with url: /api/rest/v1 (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self signed certificate in certificate chain (_ssl.c:992)')))"

Steps Taken: I have tried uploading the required SSL certificates through the provided certificates interface in Authentik. However, despite this, the problem persists.
I am able to hit a curl command to my spring boot app from the authetik server console successfully.

**Logs**
**Browser dev console logs** - "HTTPSConnectionPool(host='example.com', port=443): Max retries exceeded with url: /api/rest/v1 (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self signed certificate in certificate chain (_ssl.c:992)')))"
**Server logs** - 
{"event": "Internal Server Error: /api/v3/events/transports/23011638-bfc4-4afc-bb63-7969d92f4812/test/", "level": "error", "logger": "django.request", "timestamp": 1692612295.386091}

**Version and Deployment (please complete the following information):**

-   authentik version: 2023.3.1
-   Deployment: docker, K8s

**Additional context**
Steps Taken: I have tried uploading the required SSL certificates through the provided certificates interface in Authentik. However, despite this, the problem persists.
I am able to hit a curl command to my spring boot app from the authetik server console successfully.

Desired Outcome: I am looking for guidance on how to properly configure the webhook to utilize the uploaded SSL certificates for certificate verification when communicating with the external Spring Boot API.
or is there a way if we can write a python script to bypass the certs.

gap579137:
Any update on this...? Same issue here, fresh install today 2024.2.2

BeryJu:
You can add a custom CA certificate to `/etc/ssl/certs` and is should be picked up by OpenSSL

Vivida1:
[...]

PR #10845, opened 2024-08-09
website/docs: add content to certs page about config'ing a webhook

tanberry:
UPDATE: We are putting this on hold for now because Jens wants to change how the product/UI handles this; users shouldn't have to create the image, etc.

This PR adds a section to the Certificates page about how to configure a webhook to use uploaded certs.

@BeryJu please review carefully; this new content is 50% guessage, based on what I could grok from the Issue.

User need to add a pointer to this image in their `.env` or `values.yaml` file.

Closes Issue #6586 

### DIFF
--- a/website/docs/core/certificates.md
+++ b/website/docs/core/certificates.md
@@ -9,6 +9,7 @@ Certificates in authentik are used for the following use cases:
 -   Connecting to remote docker hosts using the Docker integration
 -   Verifying LDAP Servers' certificates
 -   Encrypting outposts' endpoints
+-   Configure a webhook to use uploaded SSL certificates
 
 ## Default certificate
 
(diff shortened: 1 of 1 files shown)


website/docs/core/certificates.md
```diff
@@ -69,6 +70,24 @@ ak import_certificate --certificate /certs/mycert.pem --private-key /certs/somet
 
 This will import the certificate into authentik under the given name. This command is safe to run as a cron job; authentik will only re-import the certificate if it changes.
 
+## Configure a webhook to use uploaded SSL certificates
+
+When communicating with an external API, you will need to configure a generic webhook to utilize uploaded SSL certificates (CA).
```
kensternberg-authentik, line 75: Question: is the word "generic" there a term of art fulfilling a specific purporse? or is it filler?
Suggestion: The word "utilize" is jargon; the title of this subsection says "use uploaded SSL certificates," and I suggest we stick with it.

website/docs/core/certificates.md
```diff
@@ -69,6 +70,24 @@ ak import_certificate --certificate /certs/mycert.pem --private-key /certs/somet
...
+
+1. Create a custom image with a Docker file that looks something like this:
+
```
kensternberg-authentik, line 78: I know we go for a "friendly, communal" tone, but this feels loose.  Suggestion: "Create a custom image with a Docker file.  For example:"

PR #10995, opened 2024-08-20
core: Refactor

BeryJu:
## Details

Misc refactors

- Move common code from OAuth source/provider, SAML source/provider and LDAP provider into common package
- Remove a lot of statically defined lists of models in blueprints
- Remove static cross app imports where sensible and possible; separate logic for app x that was implemented in app y

...it kinda turned into getting rid of `authentik.lib`, since having authentik.common for cross-provider common things and authentik.lib for general utils or "common" things made it more confusing, move everything to `authentik.common`

This also moves a bunch of things that were in common to places where they make more sense

---

## Checklist

If applicable


BeryJu:
This is never going to be merged as is and will service as reference for future partial refactors

PR #11000, opened 2024-08-21
website/docs: document the Password Uniqueness Policy

verkaufer:
Relates to https://github.com/goauthentik/authentik/pull/10631

## Details

Introduces public documentation for the Unique Password Policy feature added by https://github.com/goauthentik/authentik/pull/10631

Documentation changes were move into a separate PR to keep feedback conversations focused on the specific changes.

👉 This PR should not be merged before https://github.com/goauthentik/authentik/pull/10631.

---

## Checklist

If applicable

### DIFF
--- a/website/docs/policies/index.md
+++ b/website/docs/policies/index.md
@@ -48,15 +48,16 @@ Starting with authentik 2022.11.0, the following checks can also be done with th
 
 ### Password Uniqueness Policy
 
-Prevents users from reusing old passwords when changing their own password. The policy offers an option to limit the number many previous passwords to consider during evaluation.
+This policy allows admins to specify how many previous password hashes should be kept to prevent reuse. By default, the password history depth is 1, requiring a user's new password to be different from their last password.
 
-When this policy is bound and enabled to at least one [User write stage](../flow/stages/user_write.md):
+When this policy is bound and enabled to at least one [user write stage](../flow/stages/user_write.md):
 
-1.  authentik compares the hashes of the new password and the old password for a match. The policy check fails and exits if the hashes match.
-2.  authentik copies the hashed form of the user's old password for future comparison
+1.  authentik compares the hash of the new password in the request with the user's password history. If there is a match, the policy fails.
+2.  When the policy succeeds, the user's current password hash is copied into their password history during the [user write stage](../flow/stages/user_write.md).
+3.  Password hashes are removed, oldest first, from the user's password history if it contains more entries than the configured depth setting.
 
 :::info
-authentik purges the password history for all users when you disable or unbind the all Password Uniquess policy instances in your setup.
+authentik purges the password history for all users when you disable or unbind the all Password Uniqueness policy bindings in your setup.
 :::
 
 Passwords changed by administrators are never subject to this policy.
--- a/website/docs/policies/index.md
+++ b/website/docs/policies/index.md
@@ -46,6 +46,21 @@ Starting with authentik 2022.11.0, the following checks can also be done with th
 -   Check the password hash against the database of [Have I Been Pwned](https://haveibeenpwned.com/). Only the first 5 characters of the hashed password are transmitted, the rest is compared in authentik
 -   Check the password against the password complexity checker [zxcvbn](https://github.com/dropbox/zxcvbn), which detects weak password on various metrics.
 
+### Password Uniqueness Policy
+
+Prevents users from reusing old passwords when changing their own password. The policy offers an option to limit the number many previous passwords to consider during evaluation.
+
+When this policy is bound and enabled to at least one [User write stage](../flow/stages/user_write.md):
+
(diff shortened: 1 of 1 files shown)


website/docs/policies/index.md
```diff
@@ -46,6 +46,21 @@ Starting with authentik 2022.11.0, the following checks can also be done with th
 -   Check the password hash against the database of [Have I Been Pwned](https://haveibeenpwned.com/). Only the first 5 characters of the hashed password are transmitted, the rest is compared in authentik
 -   Check the password against the password complexity checker [zxcvbn](https://github.com/dropbox/zxcvbn), which detects weak password on various metrics.
 
+### Password Uniqueness Policy
+
+Prevents users from reusing old passwords when changing their own password. The policy offers an option to limit the number many previous passwords to consider during evaluation.
```
kensternberg-authentik, line 51: I think this reads awkwardly.  Suggestion: "This policy allows admins to specify how many previous password hashes should be kept to prevent re-use. By default, the password history depth is zero, permitting users to re-use any previous password."

website/docs/policies/index.md
```diff
@@ -46,6 +46,21 @@ Starting with authentik 2022.11.0, the following checks can also be done with th
...
+
+When this policy is bound and enabled to at least one [User write stage](../flow/stages/user_write.md):
+
+1.  authentik compares the hashes of the new password and the old password for a match. The policy check fails and exits if the hashes match.
```
kensternberg-authentik, line 55: Suggestion: "... authentic compares the hash of the requested new password with the user's password history as configured above.  If there is a match, the policy fails."

website/docs/policies/index.md
```diff
@@ -46,6 +46,21 @@ Starting with authentik 2022.11.0, the following checks can also be done with th
...
+2.  authentik copies the hashed form of the user's old password for future comparison
```
kensternberg-authentik, line 56: Suggestion: "When the policy succeeds, the user's current password hash is copied into the password history and the oldest password(s) are removed if the user's password history has more entries than the current depth setting."

kensternberg-authentik (commented):
A few language changes to help people be aware of the mechanics. I feel a solid mental model helps people feel more confident about the reliability of a system.

website/docs/policies/index.md
kensternberg-authentik: (Correct me if the default is wrong. I took that number from https://github.com/goauthentik/authentik/pull/10631/files#diff-cbb45e894de9c1957d2289a8301eaecdcef3ecd580a35d4fefba93f50441b398R30)

website/docs/policies/index.md
```diff
@@ -46,6 +46,21 @@ Starting with authentik 2022.11.0, the following checks can also be done with th
...
+
+:::info
+authentik purges the password history for all users when you disable or unbind the all Password Uniquess policy instances in your setup.
```
tanberry, line 59:
```suggestion
authentik purges the password history for all users when you disable or unbind the all Password Uniqueness policy bindings in your setup.
```

tanberry (approved):
Thanks @verkaufer for the docs! I agree with Ken's rewordings ( I almost always like them more than my own), and then I caught one typo, but looks great otherwise! 

I'm pre-approving just so that I am not the blocker when you/we are ready to merge this.

website/docs/policies/index.md
verkaufer: I like that rephrasing a lot more and I'll get that changed!

The UI displays `1` as the default, which I can change to be `0` (I figured if you're setting up a the policy you want it to actually do something 😅 ). Is there a preference?

website/docs/policies/index.md
verkaufer: 👍 Much cleaner

We could note if more than one Password Uniqueness Policy is bound & active anywhere in authentik, then the system maintains a history depth equal to the greatest configured depth of all Password Uniqueness Policies. 

e.g. I create 2 copies of a Password Uniqueness Policy. I configure one policy with a depth of `1`, and another with a depth of `10`. authentik will maintain `10` old password hashes.

It's a minor detail and probably only important for someone wanting to fully understand how many old hashes are stored. I'll leave it up to you & @tanberry whether we include that.

website/docs/policies/index.md
tanberry: Agree about defaulting to 1, from a usability perspective. Also zeros can be unnerving, lol.

website/docs/policies/index.md
tanberry: Ah, I remember you telling me about this. Yes I think we should indeed include this info in the docs. I predict it will avoid a few GitHub Issues in the future, where people ask what-the-heck.

I think that will also help people understand that there can be multiple instances on a policy.

melizeche:
close for #13686

PR #11028, opened 2024-08-22
web: provide simple tables for API-less displays

kensternberg-authentik:
web: API-free tables for authentik

## What

Provides a pair of table types that mirror @beryju’s Table implementation but are completely independent of the API.

- `ak-simple-table` takes an array of tuples, one for each column, and a tuple for the column headers, and renders a table. It will display sort markers if sort keys are provided, but it does not do sorting itself; it relies on the client supplying the information to receive the `tablesort` event and respond with the content sorted according to the client’s specification.
- `ak-select-table` inherits from and extends `ak-simple-table` in the most obvious way possible: it adds columns, tracks selections, emits a `change` event, and provides a `value` field as well as the `json()` method used by the authentik `Forms` manager. Like its predecessor, it relies on clients to manage sorting.
- Both elements are heavily documented, provide OUIA tags for automated testing, and include comprehensive CSS `part::` definitions.
- Storybook stories have been provided to exercise both types of tables.
- Tests have been provided to assert that the basic functionality of the tables is as specified.
- Both the tests and storybook provide excellent additional documentation on how to use these tables. The Story for `ak-select-table` provides an excellent example of how one would write a client component.
- Test harness has been updated so that devs may include Firefox and Safari. By default, still tests only in Chrome, as that's the only driver that permits headless testing.
- Lints and nits have been picked.

## Why

These tables provide the ability to render complex, rich tables on the client side without the client being a mere thin front-end to the Django back-end. These tables are needed to support a one-to-many relationship with our transactional wizard model, since the information we need to display does not exist on the back-end before the transaction is complete.

-   [X] Local web tests pass (`npm run test`)


web/eslint.config.mjs
```diff
@@ -41,6 +41,10 @@ export default [
         },
         files: ["src/**"],
         rules: {
+            "lit/attribute-names": "error",
```
kensternberg-authentik, line 44: Helpful when creating new elements!

web/package.json
```diff
@@ -296,6 +296,7 @@
                 "lint:types",
                 "lint:components",
                 "lint:spelling",
+                "lint:package",
```
kensternberg-authentik, line 299: The linter for package.json itself was somehow missed.

web/src/elements/forms/SearchSelect/tests/ak-search-select-view.test.ts
```diff
@@ -93,12 +93,14 @@ describe("Search select: Test Input Field", () => {
     });
 
     afterEach(async () => {
-        await document.body.querySelector("#a-separate-component")?.remove();
-        await document.body.querySelector("ak-search-select-view")?.remove();
-        // @ts-expect-error expression of type '"_$litPart$"' is added by Lit
-        if (document.body["_$litPart$"]) {
+        await browser.execute(() => {
```
kensternberg-authentik, line 96: Found in the course of testing: these need to be executed in the context of the browser, which can then be `await()`ed; otherwise, they may not get cleared correctly, which may result in misleading test results.

web/wdio.conf.ts
```diff
@@ -9,6 +9,39 @@ import tsconfigPaths from "vite-tsconfig-paths";
...
 const runHeadless = process.env.CI !== undefined;
+const testSafari = process.env.WDIO_TEST_SAFARI !== undefined;
+const testFirefox = process.env.WDIO_TEST_FIREFOX !== undefined;
+const skipChrome = process.env.WDIO_SKIP_CHROME !== undefined;
+
+const capabilities = [];
+
+const MAX_INSTANCES = 10;
+
+if (!skipChrome) {
```
kensternberg-authentik, line 20: By default, it runs the tests in Chrome.  Environment variables now allow you to control if it should run in Firefox or Safari, and control which on-demand.

web/wdio.conf.ts
```diff
@@ -19,18 +52,16 @@ export const config: Options.Testrunner = {
     runner: [
         "browser",
         {
-            viteConfig: (config: UserConfig = { plugins: [] }) => ({
-                ...config,
+            viteConfig: (userConfig: UserConfig = { plugins: [] }) => ({
+                ...userConfig,
```
kensternberg-authentik, line 56: // eslint no-shadow

web/wdio.conf.ts
```diff
@@ -83,26 +114,13 @@ export const config: Options.Testrunner = {
     // and 30 processes will get spawned. The property handles how many capabilities
     // from the same test should run tests.
     //
-    maxInstances: 10,
+    maxInstances: runHeadless ? MAX_INSTANCES : 1,
```
kensternberg-authentik, line 117: // eslint no-magic-numbers

kazlabagane (approved):

BeryJu (approved):

PR #11087, opened 2024-08-27
stages/authenticator_validate: autoselect last used 2fa device

### LINKED ISSUE #3488: authenticator_validate remember device class
I have configured multiple 2fa devices for backup. Always need to choose my primary 2fa device.

Default 2fa device in flow should be last used or an device configured as primary one.   
2fa flow should not require an click to choose a preset/last used auth device.

gergosimonyi:
## Details

Closes #3488

Users who configure multiple 2fa devices will now land on their last used device's prompt, skipping the device picker. (They can of course go to the device picker from there.)

I've made an exception for static recovery codes as I believe that would cause more confusion than good.

---

## Checklist

If applicable


web/src/flow/stages/authenticator_validate/AuthenticatorValidateStage.ts
```diff
@@ -231,22 +234,40 @@ export class AuthenticatorValidateStage
...
+        // Pick the last used validator, if it's not a recovery key
+        const lastUsedChallenge = this.challenge.deviceChallenges
+            .filter((deviceChallenge) => deviceChallenge.lastUsed)
+            .sort((a, b) => b.lastUsed!.valueOf() - a.lastUsed!.valueOf())[0];
+        if (
+            lastUsedChallenge &&
+            lastUsedChallenge.deviceClass !== DeviceClassesEnum.Static &&
+            !this._initialized
+        ) {
+            this._initialized = true;
```
kensternberg-authentik, line 266: Ah, don't change `property()` or `state()` fields inside `render()`. 

The problem is that I don't quite understand what this code is for.  If you need to set `this.selectedDeviceChallenge`, do it in the `willUpdate()` phase of the render.  [Changes inside `willUpdate()` cannot trigger rerenders.](https://lit.dev/docs/components/lifecycle/#willupdate)

kensternberg-authentik: Also, is there a reason you need to track the `this._initialized`  at all?  What happens if you don't have it?  Could you have the `selectedDeviceChallenge` itself be the sigil you need to know which one to highlight?

kensternberg-authentik (changes requested):
Change required: Do not mutate object state within the `render()` phase.

web/src/flow/stages/authenticator_validate/AuthenticatorValidateStage.ts
gergosimonyi: Let's say the user's last used device is TOTP. Without `this._initialized`, the following could happen:

1. User arrives to Authenticator Validate stage.
2. TOTP is automatically selected and the user is presented with the TOTP validation screen.
3. User decides to use webauthn instead, so clicks "Return to device picker".
4. TOTP is automatically selected again.

Effectively, without some countermeasure, if a device is autoselected the user can't go back to the device picker. `this._initialized` is supposed to be that countermeasure.

I agree that it smells. There should be a better place for the autoselection logic to run exactly once, right after `this.challenge` is loaded. What's that place? It's probably described somewhere in the `lit` documentation, but I haven't found it yet. Do you have any suggestions?

kensternberg-authentik: Ah, okay, that makes a lot of sense.  It _is_ annoying that Lit's lifecycle has a `firstUpdated()` but not a `firstWillupdate()`, which is more or less what you're simulating here.

web/src/flow/stages/authenticator_validate/AuthenticatorValidateStage.ts
```diff
@@ -112,6 +115,42 @@ export class AuthenticatorValidateStage
         `);
     }
 
+    willUpdate(_changed: PropertyValues<this>) {
```
kensternberg-authentik, line 121: This code block chooses which `selectedDeviceChallenge` to pick, if any.  I've moved it here because `willUpdate` is a special method; changes to states and properties within it will not trigger a re-draw, so this can be set without causing Lit to draw this component twice.  I also put in early returns to show that it's a selective, progressive algorithm; if one succeeds, a later one doesn't need to be examined.

web/src/flow/stages/authenticator_validate/AuthenticatorValidateStage.ts
```diff
@@ -33,6 +64,10 @@ export class AuthenticatorValidateStage
     >
     implements StageHost
 {
+    static get styles(): CSSResult[] {
```
kensternberg-authentik, line 67: I extracted the customCSS because it was big enough it distracted from the main purpose of the class, so this makes the entirety a bit more readable.

Also, there's an order to Lit components that we've adopted: Styles, Properties, States, other class fields, constructor, connectedCallback, disconnectedCallback, event handlers, and then the lifecycle methods from `performUpdate()` through `render()` and finally `updated()`.  (Obviously, don't include any that you don't override!) Yes, this means that lifecycle methods that run after `render()` appear after your `render()` method in the class.  Helper methods (sub-renderers, formatters, etc) go _above_ the first lifecycle method in which they're used, in the order in which they appear in the lifecycle method.  We're not perfect about this, but we are getting better.

kensternberg-authentik (approved):

kensternberg-authentik (approved):

authentik/stages/authenticator_validate/stage.py
```diff
@@ -218,6 +218,13 @@ def get_device_challenges(self) -> list[dict]:
                     "device_uid": device.pk,
                     "challenge": get_challenge_for_device(self.request, stage, device),
                 }
+                | (
```
BeryJu, line 221: I think it'd be cleaner to have this in the same dict instead of using the `|` operator

BeryJu (commented):
One more general comment based on the comment i made on the linked issue:
```
Our eventual implementation for this would save the last used authenticator for a pending user, there was the idea of saving this as a cookie like the validation does, however that would only save this on a specific device and not centrally per-user. (Also feel free to open a PR to add this)
```

I think it might be useful to do this too and have that as a higher priority?  There might be a bit of an advantage to that having that too if users use a specific authenticator always on a specific browser? Tbh I'm not 100% sure

BeryJu (approved):

BeryJu (changes requested):
Needs docs otherwise LGTM

website/docs/flow/stages/authenticator_validate/index.md
```diff
@@ -87,3 +87,7 @@ Optionally restrict which WebAuthn device types can be used to authenticate.
 When no restriction is set, all WebAuthn devices a user has registered are allowed.
 
 These restrictions only apply to WebAuthn devices created with authentik 2024.4 or later.
+
+#### Automatic device selection
+
+If the user has more than one device, a device picker screen is shown. Once the user successfully authenticates with a device, that device is marked as "last used". In subsequent challenges, the last used device will be automatically selected for the user. (Should they wish to use another device, the user can return to the device picker screen.)
```
tanberry, line 93:
```suggestion
If the user has more than one device, the user is prompted to select which device they want to use for validation. After the user successfully authenticates with a certain device, that device is marked as "last used". In subsequent logins, the last used device will be automatically selected for the user. Should they wish to use another device, the user can return to the device selection screen.
```

tanberry (approved):
This is a nice  feature! And nice docs. ;-) A few edits, then LGTM!

BeryJu (approved):

PR #11184, opened 2024-09-04
web: fix navigation issues inside wizard

kensternberg-authentik:
- Replace index-based steps with name-based step management
- Rename a _lot_ of components to have shorter, more straightforward names
- Rename files to match those of their components
- Remove default exports; prefer fully named exports
- Create child events rather than custom events; prefer type-checked event handling
- Move step declarations into the same folder with the step renderer
- Steps are now defined in the wizard directly
- Re-arrange some classes to meet with our class definition pattern
- Rename event handler to use the `on...` prefix
- Replace some switch/case with `ts-pattern`
- Migrate `WizardController` functionality into Wizard base class

The first change is the most essential: a wizard is a _state machine_, in which each page is both dialogue with the user and a validation pass on what page to pass control to next (or permit to go back). By using name-based step management we enable the state machine to branch, which in turn permits loops allowing collections to be created inside the wizard. This is an essential step to enabling User/Group/Policy bindings (note the plural) in the Application Wizard.

Using subclasses of `Event` instead of `CustomEvent` permits stronger type checking. It's considered the "correct" way.

Forbidding default exports enforces consistency of names. A rename now requires a deliberate and explicit step.

For single-component files, the filename should always match the component.  For single classes, the same rules applies.

Some of the names were naive, overly long, or could misinform. I've made an effort to fix those based on my knowledge of the authentik service as it now stands.

Moving each step's managerial class into the same folder with the renderer means that the "what to draw" and "how to draw" are in the same place.  No more digging in two seperate folders for information on how a page is managed and rendered.

Using `ts-pattern` enables stricter type checking at compile time.

With the string-based step system, `WizardController` was so small it no longer seemed viable to have the controls and the state in two separate classes. It was originally thought that `WizardController` would have uses outside of the wizard itself, but that hasn't panned out.

The rest is lint and code quality stuff.


web/src/components/ak-wizard-main/types.ts
```diff
@@ -26,22 +28,43 @@ export type WizardButton = [string, WizardNavCommand, boolean?];
...
+    id: string;
     label: string;
-    index: number;
     active: boolean;
     disabled: boolean;
 };
 
 type LitControllerHost = ReactiveControllerHost & LitElement;
 
+export interface WizardStep {
```
kensternberg-authentik, line 39: // lint: `WizardStep` was used before it was defined.

web/src/components/ak-wizard-main/ak-wizard-frame.ts
```diff
@@ -162,47 +177,32 @@ export class AkWizardFrame extends CustomEmitterElement(ModalButton) {
 
     renderFooter() {
         return html`
-            <footer class="pf-c-wizard__footer">${map(this.buttons, this.renderButtons)}</footer>
+            <footer class="pf-c-wizard__footer">${map(this.buttons, this.renderButton)}</footer>
         `;
     }
 
-    renderButtons([label, command]: WizardButton) {
```
kensternberg-authentik, line 169: Didn't need both functions; just a map and a way to tell if the button is disabled.

web/src/components/ak-wizard-main/ak-wizard-frame.ts
```diff
@@ -9,8 +9,23 @@ import { map } from "lit/directives/map.js";
 
 import PFWizard from "@patternfly/patternfly/components/Wizard/wizard.css";
 
+import { WizardNavigationEvent } from "./events.js";
 import { type WizardButton, WizardStepLabel } from "./types";
 
+const WIZARD_BUTTON_LABEL = {
```
kensternberg-authentik, line 15: Default labels; can be overridden.

web/src/components/ak-wizard-main/AkWizard.ts
```diff
@@ -80,36 +98,60 @@ export class AkWizard<D, Step extends WizardStep = WizardStep>
...
      * "is the current step valid?" and "should the Next button be made enabled" are controlled
      * here. (Any step implementing WizardStep can do it anyhow it pleases, putting "is the current
      * form valid" and so forth into the step object itself.)
      */
-    handleUpdate(_detail: D) {
+    onUpdate(_event: WizardUpdateEvent<D>) {
         throw new Error("This function must be overridden in the child class.");
     }
 
+    onNavigation(event: WizardNavigationEvent) {
```
kensternberg-authentik, line 129: ts-pattern allows for some amazing narrowing-by-types, and using it here is a huge win.

web/src/components/ak-wizard-main/AkWizard.ts
```diff
@@ -50,27 +57,38 @@ export class AkWizard<D, Step extends WizardStep = WizardStep>
...
-
     constructor(prompt: string, header: string, description?: string) {
         super();
         this.header = header;
         this.prompt = prompt;
         this.description = description;
-        this.wizard = new AkWizardController(this);
+    }
+
+    connectedCallback() {
```
kensternberg-authentik, line 67: You can quickly see just how little of the controller was left to be moved after replacing all the "find next/prev by index" safety stuff was removed in favor of a constrained state machine.

web/src/admin/providers/saml/SAMLProviderForm.ts
```diff
@@ -193,8 +193,10 @@ export class SAMLProviderFormPage extends BaseProviderForm<SAMLProvider> {
                             .certificate=${this.instance?.signingKp}
                             @input=${(ev: InputEvent) => {
                                 const target = ev.target as AkCryptoCertificateSearch;
-                                if (!target) return;
-                                this.hasSigningKp = !!target.selectedKeypair;
+                                if (!target) {
```
kensternberg-authentik, line 196: // eslint - prefer curly if

web/src/admin/providers/saml/SAMLProviderForm.ts
```diff
@@ -73,11 +73,11 @@ export class SAMLProviderFormPage extends BaseProviderForm<SAMLProvider> {
                 id: this.instance.pk,
                 sAMLProviderRequest: data,
             });
-        } else {
-            return new ProvidersApi(DEFAULT_CONFIG).providersSamlCreate({
-                sAMLProviderRequest: data,
-            });
         }
+
```
kensternberg-authentik, line 77: // eslint - no unnecessary else/return

web/src/admin/applications/wizard/application/ApplicationStep.ts
```diff
@@ -0,0 +1,25 @@
+import { type WizardButton } from "@goauthentik/components/ak-wizard-main/types";
```
kensternberg-authentik, line 1: This is the individual step, broken out into its own file and placed next to the renderer.  The same is true for ProviderChoice, Provider, and Submit.

web/src/admin/applications/wizard/provider/ldap/ak-application-wizard-authentication-by-ldap.ts
```diff
@@ -29,6 +29,9 @@ import {
     uidStartNumberHelp,
 } from "./LDAPOptionsAndHelp";
 
+const UID_START_NUMBER = 2000;
+const GID_START_NUMBER = 4000;
```
kensternberg-authentik, line 33: // eslint - no magic numbers

web/src/admin/applications/wizard/provider/saml/saml-property-mappings-search.ts
```diff
@@ -13,12 +13,12 @@ import {
...
 } from "@goauthentik/api";
 
-async function fetchObjects(query?: string): Promise<SAMLPropertyMapping[]> {
+async function fetchObjects(search?: string): Promise<SAMLPropertyMapping[]> {
     const args: PropertymappingsProviderSamlListRequest = {
         ordering: "saml_name",
     };
     if (query !== undefined) {
-        args.search = query;
+        args.search = search;
```
kensternberg-authentik, line 21: // eslint - no shadow variables; `query` was defined in the upper scope

PR #11271, opened 2024-09-07
web: fix readonly fields appearing white in dark theme

Hicks-99:
## Details

Readonly text fields did not update their background in dark theme. Added the `!important` property to the background style to ensure it overrides the light theme.

---

## Checklist

### DIFF
--- a/web/src/common/styles/theme-dark.css
+++ b/web/src/common/styles/theme-dark.css
@@ -155,7 +155,7 @@ select[multiple] option:checked {
     background-color: var(--ak-dark-background-light);
 }
 .pf-c-form-control[readonly] {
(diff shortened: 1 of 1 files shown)


BeryJu:
thanks @Hicks-99! which page did you see this incorrect design on? Looking on for example the LDAP Provider page which has read only input fields they render with the correct background colour

Hicks-99:
I tested it during the enrollment and authentication flow, and the trigger seems to be the compatibility mode. This causes the CSS selector .pf-c-form-control.ak-stage-prompt[readonly] to take precedence over .pf-c-form-control[readonly], which is supposed to be applied, as shown in the images.
Flow
CSS

BeryJu (approved):

PR #11341, opened 2024-09-11
website/integrations: generate integration splash page dynamically

BeryJu:
## Details

REPLACE ME

---

## Checklist

If applicable


tanberry:
@BeryJu A couple of considerations/questions... are we ok with removing mention of Sources from this splash/landing page? Because Sources used to be under Integrations, before we moved it to the regular docs, I had used this landing page to provide links to both Apps and Sources. But if we think people can easily fins sources now, then I am good with removing it from this page. I'm definitely fine with getting rid of the images, lol.
Other point... I am not a fan of using the generated pages with cards, as you know haha, especially if there are more than 4-6 items. To me it ends up looking chunky and involves a lot of scrolling and staring. Will the cards be in alphabetical order? After get the Deploy Preview fixed let's take a look.

PR #11397, opened 2024-09-16
website/docs: add scaling page

rissson:
## Details

REPLACE ME

---

## Checklist

If applicable

### DIFF
--- a/website/docs/install-config/scaling.mdx
+++ b/website/docs/install-config/scaling.mdx
@@ -1,0 +1,56 @@
+---
+title: Scaling
+---
+
+authentik can be scaled to several thousands of users. This page aims to provide some guidance on installation sizing and tuning for those large installations.
+
+:::note
+The numbers indicated on this page are meant as general guidelines and starting points. Usage patterns vary from one installation to another, and we are constantly optimizing authentik to improve its performance.
+
+The data used to build this page has been collected from benchmarks run on version 2024.6, and from real-world customer feedback.
+:::
+
+## How does authentik scale?
+
+### authentik server
+
+The authentik server is a stateless component that serves HTTP requests. As such, it is completely stateless and scales linearly. That means that if a single authentik server with X resources (CPU and RAM) can serve N requests, two authentik server with each X resources can serve 2N requests.
+
+### authentik worker
+
+The authentik worker is a stateless component that processes background tasks for authentik. This includes:
+
+-   maintenance tasks, such as removing outdated resources
+-   synchronization tasks for sources and providers such as SCIM, Google Workspace, LDAP, etc.
+-   one-off tasks that are triggered by some action by a user or an administrator, such as sending notifications.
+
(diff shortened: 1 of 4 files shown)


website/docs/installation/scaling.mdx
```diff
@@ -0,0 +1,56 @@
...
+-   the number of active users on the instance. This is used to calculate how much database storage will be required, and how many concurrent logins should be expected.
+-   the number of concurrent logins expected. For N thousands users, we don't expect those N thousands users to log in at the exact same time. As such, we estimate that all of the users will log in over a period of 15 minutes, greatly reducing the load on the instance. Furthermore, as login uses [flows](../flow/index.md), a single login process may take several seconds for the user to complete it (entering their password, their MFA method, etc.). We estimate that number at 10 seconds. If you know what usage pattern to expect, you can override that value from the computed one.
+
+import ScalingCalculator from "../../src/components/ScalingCalculator";
+
+<ScalingCalculator></ScalingCalculator>
+
+#### Known issues
+
+The database recommendations given by the calculator are scaled linearly to the number of concurrent logins. This is not the pattern we observe in reality. As such, if the calculator returns ludicrous values for the database sizing, expect those to not be representative of the actual resources needed.
```
fheisler, line 52: > not the pattern we observe in reality

Then what is, roughly? (unclear at that point if it's even better or worse than linear)

website/docs/installation/scaling.mdx
```diff
@@ -0,0 +1,56 @@
...
+The specific usage that authentik makes of PostgreSQL and Redis doesn't scale linearly, but rather logarithmically.
+
+## Calculator
+
+We provide a calculator to size your deployments of the authentik server based on benchmarks and real-world data we have observed.
+
+The calculator takes two inputs:
+
+-   the number of active users on the instance. This is used to calculate how much database storage will be required, and how many concurrent logins should be expected.
+-   the number of concurrent logins expected. For N thousands users, we don't expect those N thousands users to log in at the exact same time. As such, we estimate that all of the users will log in over a period of 15 minutes, greatly reducing the load on the instance. Furthermore, as login uses [flows](../flow/index.md), a single login process may take several seconds for the user to complete it (entering their password, their MFA method, etc.). We estimate that number at 10 seconds. If you know what usage pattern to expect, you can override that value from the computed one.
```
fheisler, line 44: > We estimate that number at 10 seconds. If you know what usage pattern to expect, you can override that value from the computed one.

This makes it sound like I can/should be able to override that value of 10 seconds directly (rather than, I think, leaving the 2nd value blank to be estimated by the calculator versus manually adding my own). It might help to have a filled in default value of i.e. 1000 users and a placeholder grayed value of the interpolated concurrent logins, which I can input a different value into to override.

rissson: The original idea was that one could entirely override the concurrent logins setting, because you might have 10k users, but divided into 3 timezones, which will effectively mean that you only have 3k users that login around the same time.

GirlBossRush:
Rebased and ready for Docusaurus 3.8.

This may still need refinement on the calculations

PR #11413, opened 2024-09-17
web: add missing id attribute for button in ak-flow-input-password

arevindh:
Add missing id attribute to button in ak-flow-input-password 

## Details

Flow component password eye icon was missing id attribute hence the password icon was not changing properly and aria-label also not changing when the icon is clicked, both issues will be fixed in the PR,

## Checklist

If applicable


RahulVadisetty91 (commented):
In web/src/flow/components/ak-flow-password-input.ts script Instead of using setInterval, you might consider using requestAnimationFrame or an updated method for focusing the input field. This will be more efficient and less likely to cause performance issues.

kensternberg-authentik (approved):
Good catch, thanks.

PR #11414, opened 2024-09-17
providers/scim: add comparison with existing group on update and delta update users

BeryJu:
## Details

REPLACE ME

---

## Checklist

If applicable

### DIFF
--- a/authentik/providers/scim/clients/groups.py
+++ b/authentik/providers/scim/clients/groups.py
@@ -2,9 +2,10 @@
 
 from itertools import batched
 
+from django.db import transaction
 from pydantic import ValidationError
 from pydanticscim.group import GroupMember
-from pydanticscim.responses import PatchOp, PatchOperation
+from pydanticscim.responses import PatchOp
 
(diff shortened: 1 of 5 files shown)


authentik/providers/scim/clients/groups.py
```diff
@@ -177,24 +207,75 @@ def _patch(
...
+        )
+        users_to_add = []
+        users_to_remove = []
+        # Check users currently in group and if they shouldn't be in the group and remove them
+        for user in current_group.members:
+            if user.value not in users_should:
+                users_to_remove.append(user.value)
+        # Check users that should be in the group and add them
+        for user in users_should:
+            if len([x.value == user for x in current_group.members]) < 1:
```
ryanpesek, line 245: This doesn't appear to work as written. I think this should maybe be `if len([x for x in current_group.members if x.value == user]) < 1:`

BeryJu: Good catch, thanks!

PR #11462, opened 2024-09-20
web/admin: add application bindings to the application wizard

kensternberg-authentik:
web: wizard for applications, now with bindings!

### Purpose

- Add policy bindings to the application wizard

### The Wizard Base:

- Restructures the Wizard base code.
  - ak-wizard-steps holds the steps and listens for NavigationRequest events to move from one step to the next.
  - WizardStep is a base class (no component registration provided) that provides the *whole frame*, not just the form. It receives the navigation content for the sidebar from ak-wizard-steps, and provides the styling for the header, footer, sidebar, and main form. It has abstractions for `buttons`, `renderMain()`, `handleButton()`, `handleEnable()`, in a section well-marked as “Public API”. Steps inherit from this class.

Conceptually:

- A wizard is a series of pages (“steps”) with a distinct beginning and end, linked in a series, to complete a task.
- Later steps in the series are inaccessible until an earlier steps has granted access to it.
- Access is predicated on the earlier step being complete and valid. The developer is responsible for determining what “complete and valid” means.
- The series is visible, giving the customer a sense of how much effort is needed to complete the task.
- A parent object maintains (and can modify as needed) the list of steps. It *can* maintain the information being collected from the user. Alternatively, that information can be kept in each step.

Details:

- Keeping with the Lit paradigm, “requests to change the system flow up, information changed by valid requests flows down.”
- The information flows up using events: WizardNavigation, WizardUpdate, WizardClose.
- The information flows down using properties.

### The Application Wizard

- ak-application-wizard-main holds the list of steps, providing a unique slot name for each.
  - It maintains the ApplicationWizardState object.
- ApplicationWizardStep inherits from WizardStep and provides:
  - A means of extraction information from forms
  - A convenience method for updating the ApplicationWizardState object, enabling future steps, and navigating to a future step, in the correct order.
  - A method for cleaning error from the error reporting mechanism as the user navigates from an error-handling state.
  - The title, description, and cancelability of the wizard.
- Steps:
  - step: Handles the application. A good starting point for understanding the point of the Wizard. Check the `handleButton()` method to understand how we enable or disable access to future steps.
  - provider-choice: Just a list. Shows validation without the form.
  - provider: Uses a *very* esoteric Lit feature, `unsafeStaticTag`, which enables the display to show anything that conforms to the expectations of ApplicationWizardProviderForm.
    - ApplicationWizardProviderForm repeats some of the base of ApplicationWizardStep, but allows us to provide multiple variants on a single form without having to create separate steps for each form.
    - The forms (`provider-for-ldap`, `provider-for-radius`) are therefore *just* the form and any fetchers needed to populate it.
  - bindings: Shows the table of bindings. Has a custom display for “This table is empty.”
  - edit-binding: Showcase for the `SearchSelectEZ` configuration format. Has an override on the `handleButton` feature to figure out which binding is about to be overridden. Is also a `.hidden` page; it doesn’t show up on the navigation sidebar, as is only navigable-to by buttons not associated with the button bar at the bottom.
  - submit: Has a lot of machinery of state: Reviewing with errors, reviewing without errors, running submission, and success. Uses `ts-pattern` a lot to make sure the state/request pairs make sense.

## Why !?!?!?

The key insight is that, even though a wizard is a series in order, that order can’t be simply maintained in a list. The parent needs various strategies for swapping pages in and out of the sequence, while still maintaining a coherent idea of “flow” and providing the visual cues the user needs to feel confident that the work can be completed and completed quickly. The entire mechanism for using an array and index to navigate, with index numbering, blocked the implementation of the bindings pages.

One thing led to another. *Sigh* Really wish this hadn’t been as much of a mess as it turned out. The end result is pretty good, though. Definitely re-usable.

One important feature to note is that the wizard is *not* tied to the ModalButton object; it’s simply embedded in a modal as-needed. This allows us to use wizards in other places, such as just being in a DIV, or just a page on its own.

## Checklist


kensternberg-authentik (commented):
Yeah, that's legit.  :-)

web/src/admin/applications/wizard/steps/ak-application-wizard-edit-binding-step.ts
```diff
@@ -0,0 +1,235 @@
...
+            <ak-search-select-ez
+                .config=${this.searchSelectConfigs(policyKind)}
+                class="policy-search-select"
+                blankable
+            ></ak-search-select-ez>
+        </ak-form-element-horizontal>`;
+    }
+
+    renderForm(instance?: PolicyBinding) {
+        return html`<ak-wizard-title>${msg("Select A Policy Binding")}</ak-wizard-title>
```
rissson, line 165:
```suggestion
        return html`<ak-wizard-title>${msg("Select A Policy/User/Group Binding")}</ak-wizard-title>
```

web/src/admin/applications/wizard/steps/ak-application-wizard-bindings-step.ts
```diff
@@ -0,0 +1,161 @@
...
+    onDeleteBindings() {
+        const toDelete = this.selectTable
+            .json()
+            .map((i) => (typeof i === "string" ? parseInt(i, 10) : i));
+        const bindings = this.wizard.bindings.filter((binding, index) => !toDelete.includes(index));
+        this.handleUpdate({ bindings }, "bindings");
+    }
+
+    renderEmptyCollection() {
+        return html` <ak-wizard-title>${msg("Configure Policy Bindings")}</ak-wizard-title>
```
rissson, line 100:
```suggestion
        return html` <ak-wizard-title>${msg("Configure Policy/User/Group Bindings")}</ak-wizard-title>
```

rissson (changes requested):

BeryJu (approved):

PR #11482, opened 2024-09-23
stages/authenticator: use RBAC for devices API

rissson:
## Details

REPLACE ME

---

## Checklist

If applicable


authentik/core/api/devices.py
```diff
@@ -52,7 +53,7 @@ class AdminDeviceViewSet(ViewSet):
     """Viewset for authenticator devices"""
 
     serializer_class = DeviceSerializer
-    permission_classes = [IsAdminUser]
+    permission_classes = [IsAuthenticated]
```
BeryJu, line 56: this should probably also use the default `ObjectPermissions`, no?

rissson: I'd love to do that, but there's no way to get a queryset to ObjectPermissions because Device is abstract

rissson: Also that permission class isn't actually used because we override it in the list method

BeryJu (commented):
Please add tests otherwise LGTM

authentik/core/tests/test_devices_api.py
```diff
@@ -0,0 +1,57 @@
...
+
+from json import loads
+
+from django.urls import reverse
+from rest_framework.test import APITestCase
+
+from authentik.core.tests.utils import create_test_admin_user, create_test_user
+
+
+class TestApplicationsAPI(APITestCase):
```
BeryJu, line 11:
```suggestion
class TestDevicesAPI(APITestCase):
```

rissson: I don't know what you mean, i absolutely didn't copy other tests and modify them 😇

BeryJu (approved):

PR #11509, opened 2024-09-25
web: Fix missing integrity fields in package-lock.json

### LINKED ISSUE #196: Validate missing integrity fields
Please describe the problem you are trying to solve.

I have lockfiles that occasionally get their integrity fields removed and npm installs the packages regardless (missing integrity is OK).  Also, I see you have a test that succeeds when there is a missing integrity: https://github.com/lirantal/lockfile-lint/blob/main/packages/lockfile-lint-api/__tests__/validators.integrityHashType.test.js#L68

I want to be able to catch these issues and fix them, to guarantee that the packages we download are the same; this is in-line with this open npm-cli issue: https://github.com/npm/cli/issues/4460#issuecomment-1050025391

I would just write my own script to basically do the same parsing as this library :)

jvanbruegge:
## Details

Same issue as with https://github.com/goauthentik/authentik/pull/9419

There was a CI check for this, however https://github.com/goauthentik/authentik/pull/10157 removed the CI check, which made this issue possible again. `lint-lockfile` does not actually check that all packages are resolved: https://github.com/lirantal/lockfile-lint/issues/196

---

## Checklist

If applicable

### DIFF
--- a/web/package-lock.json
+++ b/web/package-lock.json
@@ -30369,7 +30369,9 @@
                 "chokidar": {
                     "optional": true
                 }
-            }
+            },
+            "resolved": "https://registry.npmjs.org/@swc/cli/-/cli-0.4.0.tgz",
+            "integrity": "sha512-4JdVrPtF/4rCMXp6Q1h5I6YkYZrCCcqod7Wk97ZQq7K8vNGzJUryBv4eHCvqx5sJOJBrbYm9fcswe1B0TygNoA=="
         },
         "packages/sfe/node_modules/chokidar": {
             "version": "3.6.0",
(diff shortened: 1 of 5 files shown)


BeryJu:
https://github.com/goauthentik/authentik/commit/5f08485fff05a665b17389f983d8dcfe5ddf5fbc should already fix this, no?

jvanbruegge:
Ah, I did not see that. And it only solves the issue partially. There are still a few packages in `web/package-lock.json` that are missing their integrity (I was already wondering why only so few) and the whole `website/package-lock.json` is missing the fields.

jvanbruegge:
Found why there were a few packages in `web` without integrity, the check only checked for top level depedencies, however you now also have subpackages. The fix is quite easy

BeryJu:
I'm assuming you're using the same method to get the integrity values by `rm -rf node_modules package-lock.json; npm i`?

jvanbruegge:
No, that might change the version of the package. I used this https://github.com/jeslie0/npm-lockfile-fix to only add the integrity and resolved field

BeryJu (approved):

WilliButz:
@jvanbruegge @BeryJu thank you!

I really don't get why this is not recognized as a problem upstream in npm cli :/

PR #11522, opened 2024-09-26
website: latest migration to new structure

tanberry:
This PR has the new structure for the Docs.

If applicable


Makefile
```diff
@@ -19,14 +19,14 @@ pg_name := $(shell python -m authentik.lib.config postgresql.name 2>/dev/null)
...
-		-S 'website/developer-docs/api/reference/**' \
+		-S 'website/docs/developer-docs/api/reference/**' \
 		authentik \
 		internal \
 		cmd \
 		web/src \
 		website/src \
 		website/blog \
-		website/developer-docs \
+		website/docs/developer-docs \
```
rissson, line 29:
```suggestion
```
covered below

rissson (dismissed):

kensternberg-authentik (approved):
I can't find any build failures, and a code quality check using httrack reveals nothing is 404ing that shouldn't be.  :-). I've asked about one bit of odd behavior, but overall this is reporting solid.

kensternberg-authentik (approved):
These changes are solid.

kensternberg-authentik (approved):
Fine and superseded anyway.

kensternberg-authentik (approved):
This is working.

rissson (approved):

website/migratefile
```diff
@@ -0,0 +1,237 @@
+advanced/tenancy.md -> sys-mgmt/tenancy.md
```
rissson, line 1: Should this be committed?

tanberry: Yes, I think it is OK because it is only moving the file to a new location, but, just like before, it is not in the Table of Contents (`sidebar.js`) so no one knows it is there. This is basically the same as it is in current docs, just moving the location of the Tenancy docs.

gergosimonyi (approved):

BeryJu (approved):

PR #11596, opened 2024-10-02
web: audit and update package.json and associated test harness, with upgrade to WebdriverIO 9

kensternberg-authentik:
This commit:

- Upgrades to WebdriverIO 9.1.2
- Resets our `devDependencies` collection to remove all imports that we either were not using or were duplicates of existing dependencies:
  - *Babel*, of all things
  - Storybook addon css user preferences, now native to Storybook 8
  - SonarJS, *again*, sigh.
  - React
- moves Spotlight to be a build dependency, since it’s a dependency of `./src/common/sentry.ts`
- Fixes a bug where ESLint would report missing features in our build scripts
- Fixes a bug where Wdio might not reach a headless browser before timeout
  - This is still flaky, as the tests seem to have a warmup period.
- Replaces Rollup’s CSSLit with Vite’s CSSLit, which actually works without hacks, for testing.
- Moves the package-lock scanner to its own script, with better reporting and tool verification, which also cleans up the package.lock file a little.
- Removes the redundant sfe build; it’s the `package` where the magic happens.

One side-effect of testing the tests is that that the `pseudolocalize` file has been updated.

-   [X] Local tests pass (`cd web && npm run test`)
  - `Spec Files: 7 passed, 7 total (100% completed) in 00:00:09`


web/.storybook/css-import-maps.ts
```diff
@@ -50,6 +50,7 @@ const rawCssImportMaps = [
     'import PFNotificationDrawer from "@patternfly/patternfly/components/NotificationDrawer/notification-drawer.css";',
```
kensternberg-authentik, line 50: This file is generated from discoveries made during the build process.

web/eslint.config.mjs
```diff
@@ -63,6 +63,7 @@ export default [
             },
             globals: {
                 ...globals.nodeBuiltin,
+                ...globals.node,
```
kensternberg-authentik, line 66: With ESLint 9, these are _different_.

BeryJu (approved):

PR #11617, opened 2024-10-05
Allow cross provider oauth2 token introspection and restrict to confidential clients

### LINKED ISSUE #11618: Allow cross provider oauth2 token introspection and restrict to confidential clients

Authentik's current implementation only allows for the introspection of tokens generated with the provider used to authenticate against the introspection endpoint.

This is an issue because the resource server is then required to use the same provider as the client and it prevents the use of multiple providers, each with their own client_id (let's say Thunderbird, Roundcube and mutt) to access a resource server (a Dovecot server for instance).

See #11617

Remove the provider filter when retrieving tokens.

Additionally to prevent token fishing from public applications, I believe that restricting the introspection endpoint to confidential clients would be beneficial (and one could argue that a public client does not satisfy [RFC7662](https://datatracker.ietf.org/doc/html/rfc7662#section-4) requirements for authentication).

Because of issue #11616 I don't think it is likely that preventing public clients from using the introspection endpoint will cause much harm.

Instead of allowing the introspection of all tokens, a per-provider option or scope (as described [here](https://connect2id.com/products/server/docs/archive/v9/api/token-introspection) for instance) could be an alternative, albeit one requiring more involvement.

bcollet:
Closes: #11618

## Details

Authentik's current implementation only allows for the introspection of tokens generated with the provider used to authenticate against the introspection endpoint.

This is an issue because the resource server is then required to use the same provider as the client and it prevents the use of multiple providers, each with their own client_id (let's say Thunderbird, Roundcube and mutt) to access a resource server (a Dovecot server for instance).

This PR removes the provider filter when retrieving introspection tokens.

Additionally to prevent token fishing from public applications it restricts the introspection endpoint to confidential clients

---

## Checklist

If applicable


BeryJu:
The code in this PR was purposefully changed to _not_ allow this, see https://docs.goauthentik.io/docs/security/cves/CVE-2024-47077

Our current approach being client_credentials and passing in the JWT for provider A into provider B as listed here https://docs.goauthentik.io/docs/add-secure-apps/providers/oauth2/client_credentials (We are also working on updating the docs for this workflow)

I would be open for potentially allowing this again but restricting it to the providers in `jwt_federation_providers` of an OAuth2 provider, however that still wouldn't do policy access checks for the user

bcollet:
> The code in this PR was purposefully changed to _not_ allow this, see https://docs.goauthentik.io/docs/security/cves/CVE-2024-47077

Sorry @BeryJu, I rebased the branch on my fork and it woke up the bots. Closing this PR.

PR #11702, opened 2024-10-16
core: add prompt_data to auth flow

Wouter0100:
## Details

I added the prompt_data and user_path to the auth flow. This allows us to more easily sync users details whenever they're logged in through a Source by using the Write stage, identical to an  Enrolment flow (but we should ensure to disable that the Write stage will create new users).

This makes sure that mappings etc are automatically taken into consideration, and are passed to the Authentication flow. 

While I was at it, I made the code consistent with the `handle_enroll` method. I have this change running in production, and it's part of my effort to implement Social logins for customers, [as mentioned in this write up](https://gist.github.com/Wouter0100/5bd93ab88d5d7e706fef1832c90ff913) (which I have not updated yet, if this is merged I will update it).

---

## Checklist

If applicable

### DIFF
--- a/authentik/core/sources/flow_manager.py
+++ b/authentik/core/sources/flow_manager.py
@@ -328,7 +328,6 @@ class SourceFlowManager:
         connection: UserSourceConnection,
     ) -> HttpResponse:
         """Login user and redirect."""
-        flow_kwargs = {PLAN_CONTEXT_PENDING_USER: connection.user}
         return self._prepare_flow(
             self.source.authentication_flow,
             connection,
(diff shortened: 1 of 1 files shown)


BeryJu (approved):

authentik/core/sources/flow_manager.py
```diff
@@ -342,7 +341,11 @@ def handle_auth(
                     ),
                 )
             ],
-            **flow_kwargs,
+            **{
+                PLAN_CONTEXT_PENDING_USER: connection.user,
+                PLAN_CONTEXT_PROMPT: delete_none_values(self.user_properties),
+                PLAN_CONTEXT_USER_PATH: self.source.get_user_path(),
```
rissson, line 347: This might override what was set by property mappings.

rissson: Actually we set this for the enrolment flow as well, nevermind

rissson (approved):

authentik/core/sources/flow_manager.py
Wouter0100: Thanks for the review. Exactly, I changed the code to be exactly the same for enrollment. The property mappings work perfect here.

PR #11711, opened 2024-10-17
stages/identification: add captcha to identification stage

gergosimonyi:
## Details

Adds an optional captcha to an Identification stage to run in the background while the user inputs their information.

---

## Checklist

If applicable


web/src/flow/stages/captcha/CaptchaStage.ts
```diff
@@ -45,6 +45,16 @@ export class CaptchaStage extends BaseStage<CaptchaChallenge, CaptchaChallengeRe
     @state()
     scriptElement?: HTMLScriptElement;
 
+    @property({ type: Boolean })
+    embedded = false;
+
+    @property()
+    onTokenChange: (token: string) => void = (token) => {
```
kensternberg-authentik, line 52: **This *is* a change request.**

I don't think the default value you set here does what you think it should do.  `onTokenChange` is a way for the client code to pass a mutator, a function that reaches _up_ into the client and sets a value there, as you're doing in the IdentificationStage.ts changes below.

You have two choices.  The first is 

```
onTokenChange: (token: string) => void = (_token: string) => undefined;
```

... which will swallow any errors until you realize that you forgot to supply `onTokenChange` in your client code.

The better choice is:

```
onTokenChange: (token: string) => void = (_token: string) => { 
     throw new Error("Client failed to supply a handling function to CaptchaStage"); 
};
```

... which will crash early and crash informatively if you do forget.  (Regarding my comment about arrow functions, a `throw` is always a statement, so it has to be wrapped in scope delimiters.)  

In either case, the function _declaration's_ argument signature must begin with an underscore (the `_token`), because it is a declaration and eslint will complain that all declared variables are not being used, but it is a convention that eslint should not make that complaint about variable names that begin with an underscore.  (The names in the _definition's_ signature are pure documentation, except when they're keys for a defined record object being passed in. TypeScript does have its little annoyances.)

web/src/flow/stages/identification/IdentificationStage.ts
```diff
@@ -274,6 +278,18 @@ export class IdentificationStage extends BaseStage<
...
             ${this.renderNonFieldErrors()}
+            ${this.challenge.captchaStage
+                ? html`
+                      <input name="token" type="hidden" .value="${this.token}" />
+                      <ak-stage-captcha
+                          embedded
+                          .challenge=${this.challenge.captchaStage}
+                          .onTokenChange=${(token: string) => {
+                              this.token = token;
+                          }}
```
kensternberg-authentik, line 289: **This is not a change request. It is here to clarify a comment above.**

The purpose on `onTokenChange` here is not to call `submit`, it's to send a function down into `ak-stage-captcha` that will *call back*, literally, with a value to set on the client `ak-stage-identification` object.  

If this is the only place where you use `ak-stage-captcha`, you might not think it matters; the default function will always be overwritten, right?  But it's _confusing_, and six months from now some poor fellow may have to make more changes, and reading that they'll have no idea why the default function has that submit clause in it.  And there's a very good chance that poor fellow will be _you_, and you'll have done, learned, and forgotten so many things in that time you won't remember at all why it is the way it is.

web/src/flow/stages/captcha/CaptchaStage.ts
```diff
@@ -157,10 +155,16 @@ export class CaptchaStage extends BaseStage<CaptchaChallenge, CaptchaChallengeRe
         if (this.captchaInteractive) {
             return html`${this.captchaContainer}`;
         }
+        if (this.embedded) {
+            return html``;
```
kensternberg-authentik, line 159: **This *is* a change request.**

Please return `nothing` instead of an empty `html()` call.  Lit will create empty DOM-state tracking tags for an empty `html()`, which in the small is irrelevant, but it's still messy.

You can add `nothing` in the Lit base import:

```
import { CSSResult, PropertyValues, TemplateResult, html, nothing } from "lit";
```

web/src/flow/stages/captcha/CaptchaStage.ts
kensternberg-authentik: **This is not a change request. It is just a suggestion.**

Also?  That type-and-default pairing makes my head hurt.  Maybe you could, somewhere above the `class` declaration, put:

```
type TokenHandler = (token: string) => void;
```

... and then:

``` 
onTokenChange: TokenHandler = (_token: string) => { 
     throw new Error("Client failed to supply a handling function to CaptchaStage");
};
```

web/src/flow/stages/captcha/CaptchaStage.ts
```diff
@@ -122,12 +128,8 @@ export class CaptchaStage extends BaseStage<CaptchaChallenge, CaptchaChallengeRe
         document.body.appendChild(this.captchaContainer);
         const captchaId = hcaptcha.render(this.captchaContainer, {
             sitekey: this.challenge.siteKey,
+            callback: this.onTokenChange,
             size: "invisible",
-            callback: (token) => {
```
kensternberg-authentik, line 126: I really *do* appreciate you taking out all this repetition.

website/docs/add-secure-apps/flows-stages/stages/identification/index.md
```diff
@@ -16,7 +16,11 @@ Select which fields the user can use to identify themselves. Multiple fields can
 
 ## Password stage
 
-To prompt users for their password on the same step as identifying themselves, a password stage can be selected here. If a password stage is selected in the Identification stage, the password stage should not be bound to the flow.
+To prompt users for their password on the same step as identifying themselves, a Password stage can be selected here. If a Password stage is selected in the Identification stage, the Password stage should not be bound to the flow.
+
+## Captcha stage
+
+To run a captcha in the background while the user is inputting their identification, a Captcha stage can be selected here. If a Captcha stage is selected in the Identification stage, the Captcha stage should not be bound to the flow.
```
kensternberg-authentik, line 23: Suggestion (and maybe @tanberry can chime in): "entering" instead of "inputting".  "Inputting" is such an ugly phrase.  Either that or "... while the user inputs their identification, ..."  but that's still mechanizing the human, and I like my humans to stay human.

kensternberg-authentik (changes requested):
I made some comments and requested two changes.  I've marked comment where code change _suggestions_ are just that. I hope that's helpful.  Two of the suggestions are about bad habits, and one is _just_ a pointer to a different way to use SearchSelect.

gergosimonyi:
@kensternberg-authentik

Thank you for the extensive review! I've implemented everything, except

- `ak-search-select-ez`: I'll skip this for now as it doesn't seem to be used anywhere, I'd rather not have its maiden flight be so close to a release

- `nothing` instead of ``` html`` ```: because `nothing` doesn't fit the function's return type, `TemplateResult`, and this is prettier (in my opinion) than changing 2 function signatures like
```
@@ -147,7 +147,7 @@ export class CaptchaStage extends BaseStage<CaptchaChallenge, CaptchaChallengeRe
         return true;
     }
 
-    renderBody(): TemplateResult {
+    renderBody(): TemplateResult | typeof nothing {
         if (this.error) {
             return html`<ak-empty-state icon="fa-times" header=${this.error}> </ak-empty-state>`;
         }
@@ -160,7 +160,7 @@ export class CaptchaStage extends BaseStage<CaptchaChallenge, CaptchaChallengeRe
         return html`<ak-empty-state loading header=${msg("Verifying...")}></ak-empty-state>`;
     }
 
+    render(): TemplateResult | typeof nothing {
         if (this.embedded) {
             return this.renderBody();
         }
```

Let me know if that's acceptable or if you have any other insights.

authentik/stages/identification/stage.py
```diff
@@ -91,14 +93,16 @@ class IdentificationChallengeResponse(ChallengeResponse):
 
     uid_field = CharField()
     password = CharField(required=False, allow_blank=True, allow_null=True)
+    token = CharField(required=False, allow_blank=True, allow_null=True)
```
rissson, line 96:
```suggestion
    captcha_token = CharField(required=False, allow_blank=True, allow_null=True)
```

rissson (changes requested):

rissson (approved):

tanberry (approved):
A couple of edits... and questions... but I'm approving now so that I am not the bottleneck. ;)

kensternberg-authentik (approved):
I approve of the changes you made at my request, but please review @tanberry's suggestion and requests before committing.  At this point, the code should work as expected and is not unmaintainable.

website/docs/add-secure-apps/flows-stages/stages/identification/index.md
```diff
@@ -16,7 +16,11 @@ Select which fields the user can use to identify themselves. Multiple fields can
 
 ## Password stage
 
-To prompt users for their password on the same step as identifying themselves, a password stage can be selected here. If a password stage is selected in the Identification stage, the password stage should not be bound to the flow.
+To prompt users for their password on the same step as identifying themselves, a Password stage can be selected here. If a Password stage is selected in the Identification stage, the Password stage should not be bound to the flow.
+
+## Captcha stage
+
+To run a captcha in the background while the user is entering their identification, a Captcha stage can be selected here. If a Captcha stage is selected in the Identification stage, the Captcha stage should not be bound to the flow.
```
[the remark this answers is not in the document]
tanberry: Thanks @gergosimonyi for explaining about he background.. that CATCHA is being run in the background *if* CATCHA has been configured. Makes sesne, haha. For the second paragraph, though, I do still want to look at how we can better explain this... that can happen later though.

authentik/stages/captcha/stage.py
```diff
@@ -27,6 +27,60 @@ class CaptchaChallenge(WithUserInfoChallenge):
     component = CharField(default="ak-stage-captcha")
 
 
+def get_friendly_captcha_error(error: str) -> str:
```
BeryJu, line 30: Hmm I'm not sure if these error codes are the same for all the Captcha implementations we support. Also things like invalid/missing secret are probably not something we should disclose to end-users
I think we can probably fall back to `Invalid captcha response` as a generic error message

web/src/flow/stages/captcha/CaptchaStage.ts
```diff
@@ -45,6 +46,11 @@ export class CaptchaStage extends BaseStage<CaptchaChallenge, CaptchaChallengeRe
     @state()
     scriptElement?: HTMLScriptElement;
 
+    @property()
+    onTokenChange: TokenHandler = (_token: string) => {
+        throw new Error("Client failed to supply a handling function to CaptchaStage");
```
BeryJu, line 51: Shouldn't the default for this be 
```
this.host?.submit({
    token: token,
});
```

BeryJu (commented):
Just 2 small things this time around

web/src/flow/stages/captcha/CaptchaStage.ts
gergosimonyi: Ken and I don't think so, see some discussion above.

(By the rules of refactoring, you're right. I'm altering component contract here and I changed every callsite to match.)

authentik/stages/captcha/stage.py
gergosimonyi: Yeah, fair enough. I was caught up on how to help admins, but this is not the way, I'll add some logging for that.

I think it's reasonable to keep the ones where we advise end users to try again, and I'll make them even more friendly. I'll include all "recoverable-by-retry" error codes from [reCAPTCHA](https://developers.google.com/recaptcha/docs/verify#error_code_reference), [hCaptcha](https://docs.hcaptcha.com/#siteverify-error-codes-table) and [Turnstile](https://developers.cloudflare.com/turnstile/get-started/server-side-validation/#error-codes).

web/src/flow/stages/captcha/CaptchaStage.ts
BeryJu: I'm mainly asking because when the stage is rendered as default, there's no default handler there, which doesn't seem correct? Or did I miss it

gergosimonyi: This is the point I realize I'm probably missing something. What does "rendered as default" mean here?

My thinking is: this component will always need to be passed `challenge` and `host`. Whoever's supplying that (e.g. the `FlowExecutor`) should also supply `onTokenChange`.

BeryJu: Ah yeah I must've overlooked https://github.com/goauthentik/authentik/pull/11711/files#diff-4ed6c41b430eb15195ace2424e99a073db6c9e588f4627cf0f899ef9fad0f084R311-R312, I'd prefer that from the aspect of the flow executor all stages have the same signature so imo it should be the default callback handler on the stage and then it can be overridden when rendered embedded

BeryJu (approved):

Issue #6476, opened 2023-08-04
 Forward Proxy not working if host_browswer env var does not match host env var in 2023.6.1


jmdilly:
When AUTHENTIK_HOST does not match AUTHENTIK_HOST_BROWSER, forward authentication failed.

Was in issue #4715. Seems still broken.

**To Reproduce**

Create a proxy outpost with different AUTHENTIK_HOST and AUTHENTIK_HOST_BROWSER and use it in forward auth.

docker-compose example : 
```
  outpost_proxy_app:
        image: ghcr.io/goauthentik/proxy:2023.6.1
        environment:
            AUTHENTIK_HOST: https://authentik.local
            AUTHENTIK_INSECURE: "false"
            AUTHENTIK_TOKEN: XXX
            AUTHENTIK_HOST_BROWSER: https://auth.public.tld
```

My provider is forward auth (single application).

**Expected behavior**
Should be able to authenticate.

**Logs**
{"error":"oidc: id token issued by a different provider, expected \"https://authentik.local/application/o/app/\" got \"https://auth.public.tld/application/o/app/\"","event":"failed to redeem code","level":"warning","logger":"authentik.outpost.proxyv2.application","name":"App","timestamp":"2023-08-04T21:12:21Z"}

**Version and Deployment (please complete the following information):**

-   authentik version: 2023.6.1
-   Deployment: docker-compose
-   Reverse proxy: Traefik 2.10

ArjonBu:
I am experiencing the same issue on version 2023.8.1.

hypervtechnics:
We are experiencing the same. Is there any workaround we can follow or ETA for a bugfix?

pestaa:
Also affected. The provider is in proxy mode (not forward auth). The outpost is deployed on Kubernetes manually.

hypervtechnics:
This is still an issue.

HamburgerJungeJr:
I'm experiencing the same issue running version 2023.10.4.

ArjonBu:
I am experiencing the same issue too.

dkruyt:
Also on 2024.2.1

chrootlogin:
Still happening with 2024.8.3, seems to be the same issue like #9622

Issue #4688, opened 2023-02-14
After updating to 2023.2.0 proxy outpost provider fails with HTTP 400


cwildfoerster:
**Describe the bug**
After updating to 2023.2.0, the proxy outpost return HTTP 400, somehow the provider url schemes doesn't match anymore, but i don't know why and if there is a new setting for it. Internal host is connected with HTTP via an internal docker network. Maybe it has to do with https://github.com/goauthentik/authentik/pull/4675 ?

**To Reproduce**
1. Create a proxy provider and application and try to use it.

**Expected behavior**

**Logs**
`error":"oidc: id token issued by a different provider, expected \"https://auth.example.org/application/o/zigbee2mqtt/\" got \"http://auth.example.org/application/o/zigbee2mqtt/\"","event":"failed to redeem code","level":"warning","logger":"authentik.outpost.proxyv2.application","name":"zigbee2mqtt","timestamp":"2023-02-14T15:40:14Z`

**Version and Deployment (please complete the following information):**
 - authentik version: 2023.2.0
 - Deployment: default docker-compose

romancin:
Same problem here, auto updated by Flux and had to restore previous version.

BeryJu:
This is fixed by https://github.com/goauthentik/authentik/commit/58e001c3d5165ee5eed935e4398134eafcf6b7a9, 2023.2.1 will be out shortly

cwildfoerster:
thank you @BeryJu ❤️

weikinhuang:
I'm still seeing this in 2023.2.1

Issue #2284, opened 2022-02-10
Proxy: Host Header Set to Internal Host


eglia:
**Describe the bug**
Since 2022.1.5, the "Host" header of a proxy provider is set to the internal URL. Previously it was set to the external URL (or kept from the original request). This causes issue with some of my proxied applications which then redirect based on this header to an internal URL which is not accessible by the client.

**To Reproduce**

**Expected behavior**

**Screenshots**

**Logs**

**Version and Deployment (please complete the following information):**
 - authentik version: 2022.1.5
 - Deployment: docker-compose
 - Provider: Embeded Proxy Provider (not forward)

**Additional context**

BeryJu:
So this is actually how it should've worked all along, and when I added the fix I didn't consider that it could very well break usecases where another reverse proxy is behind the proxy provider that switches based on the header. I feel bad unitentionally introducing/fixing this in a patch release, I'll add a notice to the release notes.

I'm not sure if it makes sense to cut an entire release just to roll this back in a patch release only to re-introduce it in the next monthly version though

BeryJu:
Actually I decided to add `X-Forwarded-Host`, which can be easily configured in nginx/etc if they're behind the proxy, and those are probably the main usecase that are affected by this

eglia:
Thanks for your reply.
I already suspected that this was an intentional change, and maybe this is the "proper" way to handle it.
However, it leaves me with a bit of a problem with two of my apps, [fava](https://github.com/beancount/fava) and [bazarr ](https://github.com/morpheus65535/bazarr). They both work by sending a 302 redirect when accessing the base url.

My setup looks like this:

|| Nginx Proxy Manager -> | Authentik -> | App |
|--|--------------|-----------|------------|
|URL| app.mydomain.com | authentik.internal | app.internal |

Nginx Proxy Manager sets the Host header to the external URL (app.mydomain.com), which is used by authentik to differentiate between the applications (I assume). Authentik then sets the Host header to the internal URL (app.internal).
When accessing the external URL by a client (app.mydomain.com), the app then responds with a 302 redirect to app.internal/something.
The app is not a reverse proxy, but seems to use the host header for routing.

The easy solution for my usecase would be if I could specify the Host header being sent by authentik (either explicitly or select between internal/external URL), or if authentik would rewrite the location header, similar to the nginx [proxy_redirect](https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_redirect) directive.
[...]

PR #11722, opened 2024-10-18
providers/proxy: fix handling of AUTHENTIK_HOST_BROWSER

### LINKED ISSUE #9622: ID token issued by a different provider - Error 404
**Describe the bug**
Hello dear team, i'm here to report a bug (maybe), but first let me explain my setup.

I do have a central authentik server running inside my home homelab using almost the standard docker compose file to bring it up.

In order to expose some of my internal services on the internet i use a VPS running in the cloud (outsitde my homelab) connected via ZeroTier to my homelab. As reverse proxy i use Nginix Proxy Manager.

So i deployed a Proxy outpost on the VPS referring to the [Manual Outpost deployment in docker-compose](https://docs.goauthentik.io/docs/outposts/manual-deploy-docker-compose) documentation, modifing a bit the compose file to fit my setup. This was made to have a local outpost and avoid unecessary traffic from and to the VPS server.

Here the compose file of the outpost proxy:
```
version: "3.5"
services:
    authentik_proxy:
        image: ghcr.io/goauthentik/proxy
        environment:
            AUTHENTIK_HOST: https://x.x.x.x:9443          <---- Internal IP of the Authentik Server hosted in my homelab
            AUTHENTIK_INSECURE: "true"
            AUTHENTIK_TOKEN: xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
            AUTHENTIK_HOST_BROWSER: https://authentik.mydomain.xxx      <---- as suggested on the docs i put here the external host of my authentik server
        networks:
          dmz-vps:
            ipv4_address: 'x.x.x.x'
networks:
  dmz-vps:
    external:
      name: dmz-vps
```

I can see the Proxy Outpost on the Authentik admin page and i'm able to assign providers to it.

**Screenshots**
image

So i configured Nginix Proxy manager using the template (as always done in my local nginix proxy manager inside my homelab) but instead to point to the authentik server i pointed to the Proxy outpost IP/port.

When i try to access the esposed resoruce and i got this log message in the outpost proxy container:
`warning warning error=oidc: id token issued by a different provider, expected "https://x.x.x.x:9443/application/o/my-app/" got "https://authentik.mydomain.xxx/application/o/my-app/" event=failed to redeem code logger=authentik.outpost.proxyv2.application name=Emby_VPS-proxy timestamp=2024-05-07T08:49:47Z`

And this in the browser happens:

image

Looks like something is not working here...

The only thing that worked for me is to modify the compose file and put hostname (https://authentik.mydomain.xxx) on the AUTHENTIK_HOST instead of the internal IP address. This to me is not an elegant solution because  in this way the container reaches the main authentik server using the public internet (cloudflare tunnel ecc..) instead of using the private ip over the ZeroTier tunnel.

Another route i tested was to create a static entry in my host file and attach that on the container so to force the use of the hostname, but pointing to the private IP.

**Expected behavior**
The expected behavior is to use the private IP to communicate to the main authentik server while being able to use the AUTHENTIK_HOST_BROWSER env to inform the proxy and not have the error 400.

-   authentik version: 2024.4.1
-   Deployment: Docker Compose

chrootlogin:
## Details

This PR fixes the incorrect handling of the AUTHENTIK_HOST_BROWSER environment variable. When this variable is set, the OIDC issuer is now also adjusted to prevent the error: “oidc: id token issued by a different provider”.

This change is necessary because Authentik, as the OIDC issuer, uses the public browser URL for token issuance, but within environments like Kubernetes, the internal cluster URL is typically used. Therefore, the token must be validated against the public URL, not the internal one.

**Rationale**

The current behavior leads to mismatched issuer validation when tokens are issued by the public URL but verified using the internal URL. This PR addresses this by ensuring the issuer is correctly set based on the AUTHENTIK_HOST_BROWSER value.

**Linked Issues**

Closes #9622, #4688, #6476.

---

## Checklist

If applicable

### DIFF
--- a/internal/outpost/proxyv2/application/endpoint.go
+++ b/internal/outpost/proxyv2/application/endpoint.go
@@ -82,6 +82,9 @@ func GetOIDCEndpoint(p api.ProxyOutpostConfig, authentikHost string, embedded bo
 	if embedded {
 		ep.Issuer = updateURL(ep.Issuer, newHost.Scheme, newHost.Host)
 		ep.JwksUri = updateURL(jwksUri, newHost.Scheme, newHost.Host)
+	} else {
+		// Fixes: https://github.com/goauthentik/authentik/issues/9622 / ep.Issuer must be the HostBrowser URL
+		ep.Issuer = updateURL(ep.Issuer, newBrowserHost.Scheme, newBrowserHost.Host)
 	}
 	return ep
 }
--- a/internal/outpost/proxyv2/application/endpoint_test.go
+++ b/internal/outpost/proxyv2/application/endpoint_test.go
@@ -55,7 +55,7 @@ func TestEndpointAuthentikHostBrowser(t *testing.T) {
 	assert.Equal(t, "https://browser.test.goauthentik.io/application/o/authorize/", ep.AuthURL)
 	assert.Equal(t, "https://browser.test.goauthentik.io/application/o/test-app/end-session/", ep.EndSessionEndpoint)
 	assert.Equal(t, "https://test.goauthentik.io/application/o/token/", ep.TokenURL)
-	assert.Equal(t, "https://test.goauthentik.io/application/o/test-app/", ep.Issuer)
+	assert.Equal(t, "https://browser.test.goauthentik.io/application/o/test-app/", ep.Issuer)
 	assert.Equal(t, "https://test.goauthentik.io/application/o/test-app/jwks/", ep.JwksUri)
 	assert.Equal(t, "https://test.goauthentik.io/application/o/introspect/", ep.TokenIntrospection)
 }


BeryJu (approved):

BeryJu:
thanks for the PR @chrootlogin! could you update the tests with this new logic?

chrootlogin:
Yes, i will take a look into it!

luukrijnbende:
I now get the same error but then the other way around.

`authentik_host` is set to `http://authentik-server`
`authentik_host_browser` is set to `https://auth.domain.com`

`oidc: id token issued by a different provider, expected \"https://auth.domain.com/application/o/app/\" got \"http://authentik-server/application/o/app/\"","event":"failed to redeem code","level":"warning","logger":"authentik.outpost.proxyv2.application","name":"app","timestamp":"2024-11-02T17:58:54Z"}`

chrootlogin:
@luukrijnbende Hmm, that actually makes no sense to me. Because your token should be created by the external URL and not with the internal one.
For me it looks like that Authentik issues the token with his internal URL which is wrong.

Do you have a proxy in front of Authentik? And does this proxy forward the Host header? See: [Reverse Proxy Configuration](https://docs.goauthentik.io/docs/install-config/reverse-proxy) Because from what I know authentik issues the token with the hostname taken from Host header. You may also should check the X-Forwarded-Host header: https://docs.goauthentik.io/docs/add-secure-apps/providers/proxy/#x-forwarded-host
Also there may is a related issue: https://github.com/goauthentik/authentik/issues/2284

From my side I can say that Authentik deployed in a Kubernetes cluster with nginx-ingress now does the right thing with 2024.10.0

luukrijnbende:
Yes, I do have a proxy in front of it. I'm running it in Kubernetes as well and I'm using https://haproxy-ingress.github.io/ as my ingress controller. This is sending `X-Forwarded-Host`, `X-Forwarded-Proto`, `X-Forwarded-For`, `Host` and `X-Original-Url`.

My ingress resources are configured as follows to use forward auth;
```
haproxy-ingress.github.io/auth-url: http://ak-outpost-proxy.authentik.svc.cluster.local:9000/outpost.goauthentik.io/auth/nginx
haproxy-ingress.github.io/auth-signin: https://%[hdr(host)]/outpost.goauthentik.io/start?rd=%[url]
```

Some more info:
- I deployed Authentik using the Helm chart.
- I'm using a separate proxy deployment, not the embedded one.
- The separate proxy deployment automatically creates an ingress resource for all apps assigned to it for the `/outpost.goauthentik.io` path.
- If I point the proxy at `https://auth.domain.com` instead of `http://authentik-server` and leave the `authentik_host_browser` empty, it works, but then of course all traffic goes back through the ingress again instead of staying inside the cluster.

chrootlogin:
This really sounds like header issues. Because my setup is mostly the same like yours. The only difference I see is that I use nginx-ingress. Which then is then made available to the interwebs.

I also have separate authentik-outpost deployments with the following env:
```
AUTHENTIK_HOST: "http://authentik-server.security.svc.cluster.local"
AUTHENTIK_HOST_BROWSER: "https://authentik.my-public.domain"
AUTHENTIK_INSECURE: "false"
AUTHENTIK_TOKEN: XXXXXXXXXXXXXXX
```

For Authentik I just use the official helm chart. And on my outpost ingresses I use the `nginx.ingress.kubernetes.io/auth-snippet: proxy_set_header X-Forwarded-Host $http_host;` annotation which resets the X-Forwarded-Host to the actual hostname entered in the browser. I think this setting is very important, as it is what Authentik uses as issuing host.

yodaldevoid:
This also broke my configuration on the upgrade to 2024.10.0 with the same error as @chrootlogin. I am not running in k8s, but instead in Podman. I ended up "fixing" this by setting `AUTHENTIC_HOST` to the external URL for Authentik, eating the cost of the traffic exiting the virtual network.

chrootlogin:
@yodaldevoid Just to make it sure. I don't have any errors anymore after the release of this pull-request.

With the patch I did Authentik now works as excepted. This pull-request which seems to break your setup fixes #9622, #4688, #6476.

So my best guess is that you don't have set the headers Host and X-Forwarded-Host correctly on the Authentik Server (not the Proxy).

Because the issuing works like this:
1) You don't have any token. You get redirected to Authentik Server.
2) Authentik Server receives the Request and does the login flow.
3) When logged in, Authentik issues a token with the **Issuer beeing the Hostname in the Host or X-Forwarded-Host header**.
4) You get redirected to the Outpost.
5) Outpost checks if the Issuer is either the AUTHENTIK_HOST or if set the AUTHENTIK_HOST_BROWSER value.

TL;DR Make sure that the Authentik Server knows it's external Hostname either by setting the Host or X-Forwarded-Host header. You can also check this in the logs. My findings are documented in #9622.

Otherwise if configured wrong. Authentik Server issues tokens for it's internal URL not the external one. And the outpost tries to check with the external one and if they don't match your authentication will be blocked.

And just for clarification. This pull-request only fixes the issue that the AUTHENTIK_HOST_BROWSER was ignored when checking the issuer URL. This broke setups where Authentik issued tokens against the external URL. 

So may you could also check which URL Authentik uses to issue your tokens (iss in your openid claims). When there is the internal url or ip then your setup is definitely wrong.

luukrijnbende:
I did some more digging through the logs and it seems that Authentik uses the URL that the outpost uses to talk to Authentik as the issuer. As far as I can see all requests that are coming through the proxy have the right headers and Authentik can see that the host is `auth.domain.com`. As soon as the proxy does a call it uses the internal URL (bypassing the proxy) and that is then seen as the host.

```
{"auth_via": "oauth_client_secret", "domain_url": "authentik-server", "event": "/application/o/token/", "host": "authentik-server", "level": "info", "logger": "authentik.asgi", "method": "POST", "pid": 78278, "remote": "10.42.1.31", "request_id": "a2df579b2f2c49af8c390e52e78e73ec", "runtime": 43, "schema_name": "public", "scheme": "http", "status": 200, "timestamp": "2024-11-04T15:09:52.328994", "user": "", "user_agent": "goauthentik.io/outpost/2024.8.4 (provider=App)"}
```
Yields a token with issuer: `http://authentik-server/application/o/app/`

If I then change the `authentik_host` to be `http://authentik-server.authentik.svc.cluster.local` it will use that as the host and issue a token with it.
```
{"auth_via": "oauth_client_secret", "domain_url": "authentik-server.authentik.svc.cluster.local", "event": "/application/o/token/", "host": "authentik-server.authentik.svc.cluster.local", "level": "info", "logger": "authentik.asgi", "method": "POST", "pid": 41, "remote": "10.42.1.38", "request_id": "54aefa2e4d6148caa0d43986b13eb233", "runtime": 67, "schema_name": "public", "scheme": "http", "status": 200, "timestamp": "2024-11-04T15:37:41.881911", "user": "", "user_agent": "goauthentik.io/outpost/2024.8.4 (provider=App)"}
```
Yields a token with issuer: `http://authentik-server.authentik.svc.cluster.local/application/o/app/`

If I update the outpost to version `2024.10.0` instead of `2024.8.4` it still issues the token in the same way but then the `id token issued by a different provider` error returns.


I'll see if I can reproduce it using the nginx ingress controller, maybe that will give us some more clues 😄

yodaldevoid:
> @yodaldevoid Just to make it sure. I don't have any errors anymore after the release of this pull-request.

To be clear, I fully expect the problem to be on my end. I am not suggesting that this PR is reverted or that it was implemented wrong. I wanted to leave a comment for anyone else who runs into the same error pointing them to a possibly easy, if not quite perfect, solution.

Thank you for the suggestions, I will look into them.

chrootlogin:
That all makes no sense to me. 😅

Because actually I also use http://authentik-server.security.cluster.local as AUTHENTIK_HOST and this means that Authentik Outpost directly communicates with the Authentik Server which should not include any headers.

chrootlogin:
@yodaldevoid @luukrijnbende Could you please check whats written in the yellow field below the outpost name when opening Authentik Settings -> Applications -> Outposts. There should be the login URL. Do you see there the internal or external URL?

luukrijnbende:
The proxy says, `Logging in via http://authentik-server.authentik.svc.cluster.local.` and that changes whenever I change the `authentik_host` in the configuration. The weird thing is, when I click `View Deployment Info` it does show the external URL `auth.domain.com` in both `authentik_host` and `authentik_host_browser` even though the configuration itself is different.

chrootlogin:
@luukrijnbende Okay, I think we get closer to the root of the problem.

My outpost is deployed manually via a helmchart ([bjw-s universal library Helmchart](https://github.com/bjw-s/helm-charts)). And my Authentik UI shows `Logging in via https://authentik.my-external-domain.com`. It also says my AUTHENTIK_HOST is "https://authentik.my-external-domain.com" even that I have configured AUTHENTIK_HOST to "http://authentik-server.security.cluster.local".

Also from what I can see in the logs of Authentik Server. The outpost definitely communicates through the internal network:
```
{"auth_via": "api_token", "domain_url": "authentik-server.security.svc.cluster.local", "event": "/api/v3/root/config/", "host": "authentik-server.security.svc.cluster.local", "level": "info", "logger": "authentik.asgi", "method": "GET", "pid": xxx, "remote": "10.42.xxx.xxx", "request_id": "xxxxxxxxxx", "runtime": 7, "schema_name": "public", "scheme": "http", "status": 200, "timestamp": "2024-11-05T14:50:07.215403", "user": "ak-outpost-xxxxxxxx", "user_agent": "goauthentik.io/outpost/2024.10.0"}
```

Another question when you watch the logs of **Authentik Server** and open "authentik.my-external-domain.com" in your Browser what do you see?
```
{"auth_via": "session", "domain_url": "authentik.my-external-domain.com", "event": "/api/v3/core/applications/?only_with_launch_url=true&ordering=name&page=1&page_size=100", "host": "authentik.my-external-domain.com", "level": "info", "logger": "authentik.asgi", "method": "GET", "pid": xxxxx, "remote": "xxxxxx", "request_id": "xxxxxxx", "runtime": 44, "schema_name": "public", "scheme": "https", "status": 200, "timestamp": "2024-11-05T15:01:56.608976", "user": "xxxxx", "user_agent": "xxxxxxxxxx"}
```

Under `domain_url` and `host` do you see your external or internal domain? Because when you see your internal domain, then definitely something is wrong.

luukrijnbende:
Hmm, mine is deployed automatically by the integration and the `authentik_host` in the configuration of the outpost is mapped to a secret for that outpost which is used for the environment variables of the deployment. So setting that apparently changes the issuing of the tokens and the URL used for the outpost to communicate with Authentik.

I was thinking about moving it to a separate helm chart to make the deployment of it explicit within my gitops. Which then, using the right environment variables, should fix this issue.

When I open Authentik and login as a user I see my external `auth.domain.com` in all the logs as both `domain_url` and `host`.

chrootlogin:
@luukrijnbende Yes thats really interesting. What I can say from my side is that my pull-request should only modify requests when runned as external deployed Outpost.

See: https://github.com/chrootlogin/authentik/blob/1c474a22ef0561419acd4ec269c8d96a74df08de/internal/outpost/proxyv2/application/endpoint.go
```go
// ...
// For the embedded outpost, we use the configure `authentik_host` for the browser URLs
// and localhost (which is what we've got from the API) for backchannel URLs
//
// For other outposts, when `AUTHENTIK_HOST_BROWSER` is set, we use that for the browser URLs
// and use what we got from the API for backchannel
hostBrowser := config.Get().AuthentikHostBrowser
if !embedded && hostBrowser == "" {
  return ep
}
var newHost *url.URL = aku
var newBrowserHost *url.URL
if embedded {
  if authentikHost == "" {
    log.Warning("Outpost has localhost/blank API Connection but no authentik_host is configured.")
    return ep
  }
  newBrowserHost = aku
} else if hostBrowser != "" {
  browser, err := url.Parse(hostBrowser)
  if err != nil {
    return ep
  }
}
newBrowserHost = browser
// ...
// This is only used in embedded outposts as there we can guarantee that the request
// is routed correctly
if embedded {
  ep.Issuer = updateURL(ep.Issuer, newHost.Scheme, newHost.Host)
  ep.JwksUri = updateURL(jwksUri, newHost.Scheme, newHost.Host)
} else {
  // Fixes: https://github.com/goauthentik/authentik/issues/9622 / ep.Issuer must be the HostBrowser URL
  ep.Issuer = updateURL(ep.Issuer, newBrowserHost.Scheme, newBrowserHost.Host)
}
// ...
```
So my code change ist actually the last `else` which overwrites the Issuer when having a "AUTHENTIK_HOST_BROWSER" and not running as embedded, e.g. deployed manually.

PR #11793, opened 2024-10-24
blueprints: add default Password policy

gergosimonyi:
## Details

REPLACE ME

---

## Checklist

If applicable


blueprints/default/flow-password-change.yaml
```diff
@@ -11,6 +11,15 @@ entries:
     slug: default-password-change
   model: authentik_flows.flow
   id: flow
+- attrs:
+    check_have_i_been_pwned: true
```
BeryJu, line 15: as much as I would like to enable this by default, we probably shouldn't enable this by default for airgapped instance

gergosimonyi: Fair enough. I was going by NIST's

> verifiers SHALL compare the prospective secret against a blocklist that contains known commonly used, expected, or compromised passwords

but on second thought, I agree, defaulting to HIBP is a step too far.

Perhaps we can do something about that later. I'll remove for now and add it to the hardening docs.

BeryJu: we can set `check_zxcvbn` to `true` which includes a blocklist

gergosimonyi: That seems to work okay. Added.

I didn't add `zxcvbn` initially because I felt it was too much restriction (based on NIST). However, a score of 2 seems like it effectively adds the blocklist and some variations, but not much more.

website/docs/customize/policies/index.md
```diff
@@ -32,6 +32,10 @@ This policy can enforce regular password rotation by expiring set passwords afte
 
 ### Password Policy
 
+:::warning
+This policy enables options that violate [NIST's recommendations](https://pages.nist.gov/800-63-4/sp800-63b.html#password) for passwords. To comply with the recommendations, use authentik's default Password policy. See [Hardening authentik](../../security/security-hardening.md#password-policy) for additional hardening.
```
tanberry, line 36:
```suggestion
By default, authentik's Password policy is compliant with [NIST's recommendations](https://pages.nist.gov/800-63-4/sp800-63b.html#password) for passwords. To remain compliant with NIST, be cautious when editing the default values. For additional hardening configuration settings, refer to [Hardening authentik](../../security/security-hardening.md#password-policy).
```

tanberry (approved):
Thanks for your time explaining the compliance, @gergosimonyi

website/docs/security/security-hardening.md
```diff
@@ -4,6 +4,17 @@ title: Hardening authentik
...
 While authentik is secure out of the box, you can take steps to further increase the security of an authentik instance. As everyone knows, there is a consequential tradeoff between security and convenience. All of these hardening practices have an impact on the user experience and should only be applied knowing this tradeoff.
 
+### Password policy
+
+authentik's default Password policy complies with the [NIST SP 800-63 Digital Identity Guidelines](https://pages.nist.gov/800-63-4/sp800-63b.html#password).
+
+However, for further hardening compliant to the NIST Guidelines, consider
+
+-   setting the length of the password to a minimum of 15 characters, and
+-   enabling the "Check haveibeenpwned.com" blocklist comparison
```
BeryJu, line 14:
```suggestion
-   enabling the "Check haveibeenpwned.com" blocklist comparison (note that this cannot be used on Air-gapped instances)
```

PR #11861, opened 2024-10-30
web/admin: provide default invalidation flows for LDAP and Radius

kensternberg-authentik:
These are named Bind/Unbind instead of Authorize/Invalidate, and they seem to have been missed when InvalidationFlows were added to provider options.

-   [X] The tests pass (`npm run test:e2e`)


BeryJu (commented):
Good catch, however for this the default should actually be `default-invalidation-flow` (and same for radius, I think I might've put the incorrect default there too)

BeryJu (approved):
🎉

PR #11913, opened 2024-11-04
sources/ldap: set password_change_date in check_pwd_last_set to avoid loop ending user sessions

### LINKED ISSUE #11436: Users have to login multiple times a day
**Describe the bug**
Some users report that they have to enter there login credentials multiple times a day. I can confirm via the logs the users indeed login multiple times a day with the same browser

**To Reproduce**
Unfortunately I don't really know what leads to this.
I have set a custom Session duration in the user default-authentication-login (24 hours). 
I noticed before accessing the authentik site the access cookie shows the right end time, after visiting the authentik site it changes to 0

**Expected behavior**
A user should only have to enter there credential once a day.

-   authentik version: 2024.6.2

**Additional context**
It might started with the upgrade from 2024.6.1 to 2024.6.2 .

tograss:
## Details

closes #11436
**Explanation:**
In (ms_)check_pwd_last_set the line 
` if created or pwd_last_set > user.password_change_date:`
test if the pw in the ldap store is newer than the authentik one. If so
`
user.set_unusable_password()
user.save()
`
Is executed. Changing the  user password (correctly) ends all user sessions. See https://docs.djangoproject.com/en/5.1/topics/auth/default/#:~:text=Django%20also%20provides%20views%20and,log%20out%20all%20their%20sessions.
However, since user.password_change_date is not set the next ldap_sync will lockout the user again.
I think this wasn't a problem when during login the user pw was always saved to the authentik database. Since this changed a while ago this is now a bug.

The fix is to simply set the password_change_date to the ldap date. One might argue it should be the time the ldap sync run, but I think it is easier to understand for an administrator if the ldap date and the authentik date are the same.

---

## Checklist

If applicable


authentik/sources/ldap/sync/vendor/freeipa.py
```diff
@@ -38,6 +38,7 @@ def check_pwd_last_set(self, attributes: dict[str, Any], user: User, created: bo
                 pwd_last_set=pwd_last_set,
             )
             user.set_unusable_password()
+            user.password_change_date = pwd_last_set
```
BeryJu, line 41: the more correct solution for this would be something like

```diff
diff --git a/authentik/core/models.py b/authentik/core/models.py
index 85e8901ed..470436db0 100644
--- a/authentik/core/models.py
+++ b/authentik/core/models.py
@@ -340,6 +340,10 @@ class User(SerializerModel, GuardianUserMixin, AttributesMixin, AbstractUser):
         self.password_change_date = now()
         return super().set_password(raw_password)

+    def set_unusable_password(self):
+        self.password_change_date = now()
+        return super().set_unusable_password()
+
     def check_password(self, raw_password: str) -> bool:
         """
         Return a boolean of whether the raw_password was correct. Handles
```

instead of setting the date everywhere

tograss: I added a combination of a) and c)

tograss:
@BeryJu 
Hi Bery, can you provide an estimate when you can review / merge this pull request. The related issues will effect more and more users as time progresses.

tograss:
Hi can I help with anything to get this fix merged ?

PR #11926, opened 2024-11-05
website/docs: `2024.10.1` Release Notes

gergosimonyi:
Using `details` may be a bit weird. An alternative is a `h3`.

### DIFF
--- a/website/docs/releases/2024/v2024.10.md
+++ b/website/docs/releases/2024/v2024.10.md
@@ -152,6 +152,8 @@ helm upgrade authentik authentik/authentik -f values.yaml --version ^2024.10
 
 ## API Changes
 
+### API Changes in 2024.10.0
+
 #### What's New
 
 ---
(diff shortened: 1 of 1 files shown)


BeryJu (approved):

tanberry (commented):
Here are my thoughts about the "construction" of the API section.

Even though I think that collapsible sections are fantastic for giving pages a clean look, and improving visual scanabilty, there are two reasons I am not keen to use the `<details open>` element:

1. when collapsed, the content is not searchable on Firefox and Safari (though we could have it open by default, then allow the reader to collapse the section if they want)
2. the background colour of the section, when opened, is kinda jarring with the rest of the page (minor CSS fix probably).

So I prefer Grego's approach of using H3's. (And making those H3s much more specific, for both scanability and for SEO.)

My biggest takeaway from this is that I don't think we should be cramming patch release info into the major version's Release Notes. The 3 or 4 changes in the .1 patch release are hard to find amongst all of the .0 release info, and in general it doesn't seem right to edit/modify existing release notes. Let's just make a new short RN doc for each patch release. That would help our readers quickly determined what is actually in the patch release. ;-)

PR #11945, opened 2024-11-07
website/integrations: mastodon: set correct uid field

dominic-r:
Setting the `OIDC_UID_FIELD` as `sub` created users on Mastodon with random usernames. Once this was changed to `preferred_username`, new users were created with their usernames set in authentik. My assumption is that users would wish to have the same username rather than have a random one generated.

## Details

REPLACE ME

---

## Checklist

If applicable


BeryJu (commented):
👍 but there should be a warning that https://docs.goauthentik.io/docs/sys-mgmt/settings#allow-users-to-change-username should be set to false (which is the default but still) when using preferred_username as the primary identifier

website/integrations/services/mastodon/index.md
```diff
@@ -48,7 +48,7 @@ OIDC_DISCOVERY=true
 OIDC_ISSUER=< OpenID Configuration Issuer>
 OIDC_AUTH_ENDPOINT=https://authentik.company/application/o/authorize/
 OIDC_SCOPE=openid,profile,email
-OIDC_UID_FIELD=sub
+OIDC_UID_FIELD=preferred_username
```
tanberry, line 51: I thin it would be nice to explain bpth options to the readers. Maybe make `preferred_username` be the one in this list, but then below the list you could add a sentence to explain that one can use `OIDC_UID_FIELD=sub` if you want names generated?

tanberry (commented):
a suggestion... instead of removing info, we can add info (by showing both).

tanberry (approved):
Thanks, sorry it took me so long to get to this!

PR #11951, opened 2024-11-07
core: add ability to provide reason for impersonation

rissson:
## Details

For some reason the error message doesn't show up in the frontend, but figuring that out exceeds my frontend abilities.

---

## Checklist

If applicable


authentik/tenants/models.py
```diff
@@ -85,6 +85,10 @@ class Tenant(TenantMixin, SerializerModel):
     impersonation = models.BooleanField(
         help_text=_("Globally enable/disable impersonation."), default=True
     )
+    impersonation_require_reason = models.BooleanField(
```
rissson, line 88: TODO: add this to API and frontend

kensternberg-authentik (approved):
The front-end works is solid.

authentik/tenants/models.py
```diff
@@ -85,6 +85,10 @@ class Tenant(TenantMixin, SerializerModel):
...
+        help_text=_("Require administrators to provide a reason for impersonating a user."),
+        default=True,
```
BeryJu, line 90: if we want to release this in 2024.10.x it should probably default to False to be backwards compatible, otherwise this is fine

web/src/admin/users/UserImpersonateForm.ts
```diff
@@ -0,0 +1,39 @@
...
+    instancePk?: number;
+
+    async send(data: ImpersonationRequest): Promise<void> {
+        return new CoreApi(DEFAULT_CONFIG).coreUsersImpersonateCreate({
+            id: this.instancePk || 0,
+            impersonationRequest: data,
+        });
+        // .then(() => {
+        //     window.location.href = "/";
+        // });
```
BeryJu, line 23: Assuming this is commented out for testing?

authentik/tenants/models.py
rissson: I think it's a good thing to have this default to True as that's what I would expect to be the recommended approach. 

This doesn't have to make it to 2024.10 since it's a new feature

web/src/admin/users/UserImpersonateForm.ts
rissson: yeah, this shouldn't be commented out

rissson: let me just try out again if that works

rissson: yep, all good

BeryJu (approved):

PR #11970, opened 2024-11-08
website/docs: add CSP to hardening

gergosimonyi:


website/docs/security/security-hardening.md
```diff
@@ -36,3 +36,28 @@ To prevent any user from creating/editing CAPTCHA stages block API requests to t
...
+img-src 'https:' 'http:' 'data:';
+object-src 'none';
+```
+
+Your use case may require more origins on various directives, e.g.
+
+-   when using a CAPTCHA service
+-   when using Sentry
+-   when using any custom Javascript in a prompt stage
+-   when using spotlight sidecar for development
```
fheisler, line 63: *JavaScript
*Spotlight Sidecar

website/docs/security/security-hardening.md
```diff
@@ -36,3 +36,28 @@ To prevent any user from creating/editing CAPTCHA stages block API requests to t
 -   `/api/v3/managed/blueprints*`
 
 With these restrictions in place, CAPTCHA stages can only be edited using [Blueprints on the file system](../customize/blueprints/index.md#storage---file).
+
+### Content Security Policy (CSP)
+
+:::caution
+Setting up CSP incorrectly may result in the client not loading necessary third party code.
```
tanberry, line 43:
```suggestion
Setting up CSP incorrectly may result in the client not loading necessary third-party code.
```

website/docs/security/security-hardening.md
```diff
@@ -36,3 +36,28 @@ To prevent any user from creating/editing CAPTCHA stages block API requests to t
...
+
+```
+default-src 'self';
+style-src 'self' 'unsafe-inline';
+script-src 'self' 'unsafe-inline';
+img-src 'https:' 'http:' 'data:';
+object-src 'none';
+```
+
+Your use case may require more origins on various directives, e.g.
```
tanberry, line 58:
```suggestion
Your use case might require more origins on various directives, e.g.
```

website/docs/security/security-hardening.md
tanberry, line 63:
```suggestion
-   when using Spotlight Sidecar for development
```

tanberry (changes requested):
A few editing thangs.

website/docs/security/security-hardening.md
```diff
@@ -36,3 +36,28 @@ To prevent any user from creating/editing CAPTCHA stages block API requests to t
 -   `/api/v3/managed/blueprints*`
 
 With these restrictions in place, CAPTCHA stages can only be edited using [Blueprints on the file system](../customize/blueprints/index.md#storage---file).
+
+### Content Security Policy (CSP)
+
+:::caution
+Setting up CSP incorrectly may result in the client not loading necessary third-party code.
```
tanberry, line 43:
```suggestion
Setting up CSP incorrectly might result in the client not loading necessary third-party code.
```

website/docs/security/security-hardening.md
```diff
@@ -36,3 +36,28 @@ To prevent any user from creating/editing CAPTCHA stages block API requests to t
...
+:::
+
+Content Security Policy (CSP) is a security standard that mitigates the risk of content injection vulnerabilities. authentik doesn't currently support CSP natively, so setting it up depends on your installation. We recommend using a [reverse proxy](../install-config/reverse-proxy.md) to set a CSP header.
+
+authentik will require at least the following origins:
```
tanberry, line 48:
```suggestion
authentik requires at least the following origins:
```

tanberry: Should we explain more about what an origin is? Where to define it? Do we have any links about it to link to?

tanberry (changes requested):
I thought I had edited that use of "may" before, but maybe I missed this one. (We want to use "might", to imply chances, instead of "may" which implies permission. :-)

website/docs/security/security-hardening.md
gergosimonyi: Offline resolution: we should use the word used by the CSP spec: "location"

tanberry (approved):

website/docs/security/security-hardening.md
```diff
@@ -47,3 +47,28 @@ To prevent any user from creating/editing CAPTCHA stages block API requests to t
...
 With these restrictions in place, CAPTCHA stages can only be edited using [Blueprints on the file system](../customize/blueprints/index.md#storage---file).
+
+### Content Security Policy (CSP)
+
+:::caution
+Setting up CSP incorrectly might result in the client not loading necessary third-party code.
+:::
+
+Content Security Policy (CSP) is a security standard that mitigates the risk of content injection vulnerabilities. authentik doesn't currently support CSP natively, so setting it up depends on your installation. We recommend using a [reverse proxy](../install-config/reverse-proxy.md) to set a CSP header.
+
```
rissson, line 58: We probably should cross-reference https://github.com/goauthentik/authentik/pull/12092 so that that behaviour is also documented. People deploying a custom csp might overlook this and override those headers.

website/docs/security/security-hardening.md
```diff
@@ -47,3 +47,28 @@ To prevent any user from creating/editing CAPTCHA stages block API requests to t
...
+authentik requires at least the following allowed locations:
+
+```
+default-src 'self';
+style-src 'self' 'unsafe-inline';
```
BeryJu, line 63: We might want to add comments that explain why we need `unsafe-inline` for styles and scripts based on the comments I had in the notes in notion

gergosimonyi: Fair enough, added.

(A bit weird syntactically because this is a language-agnostic text block, but anyone who's implementing it will understand.)

website/docs/security/security-hardening.md
gergosimonyi: Added a general warning not to overwrite any existing CSP header.

BeryJu (approved):

PR #11981, opened 2024-11-10
web: applications: redirect user on application rename

dominic-r:
closes #11973 

## Details

REPLACE ME

---

## Checklist

If applicable


web/src/admin/applications/ApplicationForm.ts
```diff
@@ -107,6 +107,11 @@ export class ApplicationForm extends WithCapabilitiesConfig(ModelForm<Applicatio
                 },
             });
         }
+
```
BeryJu, line 110: We should be doing both this and the change below in a more generic way for all objects that are identified by a user-configurable identifier cc @kensternberg-authentik

PR #11911, opened 2024-11-04
web/elements: array select

BeryJu:
## Details

Input element which supports dynamically adding new rows

---

## Checklist

-   [ ] Local tests pass (`ak test authentik/`)
-   [ ] The code has been formatted (`make lint-fix`)

If an API change has been made

-   [ ] The API schema has been updated (`make gen-build`)

If changes to the frontend have been made

-   [ ] The code has been formatted (`make web`)

If applicable

-   [ ] The documentation has been updated
-   [ ] The documentation has been formatted (`make website`)

PR #12004, opened 2024-11-12
web/admin: better footer links

kensternberg-authentik:
web/admin/better-footer-links

# What

- A data control that takes two string fields and returns the JSON object for a FooterLink
- A data control that takes a control like the one above and assists the user in entering a collection of such objects.

# Why

We’re trying to move away from CodeMirror for the simple things, like tables of what is essentially data entry. Jens proposed this ArrayInput thing, and I’ve simplified it so you define what “a row” is as a small, lightweight custom Component that returns and validates the datatype for that row, and ArrayInput creates a table of rows, and that’s that.

We’re still working out the details, but the demo is to replace the “Name & URL” table in AdminSettingsForm with this, since it was silly to ask the customer to hand-write JSON or YAML, getting the keys right every time, for an `Array<Record<{ name: string, href: string }>>`. And some client-side validation can’t hurt.

Storybook included. Tests to come.

## Checklist

-   [X] Tests are provided for new functionality.
-   [X] The tests all pass


BeryJu (approved):
🎉 replaces #11911

Only minor thing I've noticed is the tab index not being quite right, when I tab from the first input it looses focus, then tab again and it goes back to the first input and then tab to the second input (I say this is a minor thing cause I'm sure there's quite a few other places where we dont have the tabindex quite right either)

PR #12033, opened 2024-11-14
web: fix bug that prevented error reporting in current wizard.

kensternberg-authentik:
web/bugfix/fix-reporting-in-wizard-submit

# What

- Preserves the errors locally for the Wizard, providing explanation and links to fix the issues
- Addresses [Issue 11977](https://github.com/goauthentik/authentik/issues/11977)
# Why

Just a silly mistake on my part. There shouldn’t be two copies of errors (and there isn’t in the BIG PRs), but this is how it’s designed right now and making the errors show up is an easy fix. In doing so, the “hack” to move the “bad provider name” to the provider page is included.

-   [X] Existing end-to-end tests run (`npm run test:e2e:watch`)


web/src/admin/applications/wizard/commit/ak-application-wizard-commit-application.ts
```diff
@@ -134,10 +138,25 @@ export class ApplicationWizardCommitApplication extends BasePanel {
...
+                    const providerError = errors.app.provider;
+                    errors.provider = errors.provider ?? {};
+                    errors.provider.name = providerError;
+                    delete errors.app.provider;
+                    if (Object.keys(errors.app).length === 0) {
+                        delete errors.app;
+                    }
+                }
+
+                this.errors = errors;
```
kensternberg-authentik, line 155: This is the dumbth that caused the errors not to be preserved locally and so weren't shown locally.

BeryJu (approved):

PR #12080, opened 2024-11-19
core: fix source_flow_manager throwing error when authenticated user attempts to re-authenticate with existing link

BeryJu:
## Details

In previous versions we used to save the connection earlier which caused an Integrity Error which was caused and did a root redirect, however this was refactored to save the connection later and as such we don't redirect and don't catch the exception for this one specific user flow.

This user flow was also not tested, as it was assumed that the user would only authenticate with their source connection when they're not already authenticated, however this is something that might happen accidentally

---

## Checklist

If applicable


rissson (approved):

williamhatcher:
This appears to still be an issue with SAML.

PR #12124, opened 2024-11-21
website/docs: Fix CSP syntax

thegcat:
Scheme sources need to not have quotes https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy#scheme-source


gergosimonyi (approved):
Ah, of course. Thank you!

BeryJu (approved):

PR #12144, opened 2024-11-21
web: table-driven executor for flow

kensternberg-authentik:
web/flow/table-driven-executor

# What

- Replaces the *massive* switch/case block for stages with a lookup table.

# Why

- Because it was a massive switch/case block calling 22 different functions with the *exact same* signatures, and that created both a ton of duplication in the code as well as introduced last-line risks.

## Checklist

-   [X] Existing end-to-end tests passing (`npm run test:e2e`)


kensternberg-authentik:
This has been refined with 19999.

Issue #10983, opened 2024-08-19
Unable to add role to group via UI or API: "Roles can only be used with a single group." 


nate3D:
**Describe the bug**
I have a `ldaplogin` role that I am trying to assign to a second top-level group. When trying to add the role assignment via editing the Group, I get back simply an empty error popup and 400 via the API. The Role assignment is not persisted. Using the API browser, I can see a response of "Roles can only be used with a single group." I can't find any reference to this being the case in the documentation online. 

**To Reproduce**

1. Create a new Role, i.e. `ldaplogin`
2. Create two new groups, i.e. `Group 1`, `Group 2`
3. Assign `ldaplogin` to `Group 1` and save.
4. Assign `ldaplogin` to `Group 2` and save, it will show an empty error popup and the role assignment will not persist.

**Expected behavior**
The group is updated as expected with the new Role assignment, or an error message with corresponding documentation on how to properly structure Role <-> Group assignment hierarchy.

**Screenshots**
<img width="929" alt="AuthentikError_GroupSaveNotPersisting" src="https://github.com/user-attachments/assets/a7f22c94-f80d-4316-8601-a0d969b04f60">

**Logs**
```
[...]

PR #12169, opened 2024-11-24
website/docs: Add note about single group per role

pklaschka:
## Details

This change adds an admonition to document the fact that every role can only ever be assigned to a single group at the same time. Since this is surprising based on a traditional understanding of role-based models, I've decided to make this a `:::warning`.

I'm undecided on the best place for this information, but for now, decided on putting it into the context of the action that can fail: assigning a role to a group.

While this does not close the issue, it documents this behavior to at least address the "needs documentation" aspect of #10983 .

---

## Checklist

If applicable

### DIFF
--- a/website/docs/users-sources/roles/manage_roles.md
+++ b/website/docs/users-sources/roles/manage_roles.md
@@ -37,6 +37,10 @@ To delete a role, follow these steps:
 
 In authentik, roles are assigned to [groups](../groups/index.mdx), not to individual users.
 
+:::warning
+In authentik, each role can only be applied to a single group at the moment.
+:::
+
 1.  To assign the role to a group, navigate to **Directory -> Groups**.
 2.  Click the name of the group to which you want to add a role.
 3.  On the group's detail page, on the Overview tab, click **Edit** in the **Group Info** area.


tanberry:
Hi @pklaschka thanks so much for this PR. It also helped me catch the fact that I need to update the docs to use the new "dual-select" UI component that is now in the UI for selecting roles. I think we introduced it in version 2024.8, but I failed to update the docs about assigning roles to a group. Which version are you running?

On your point, though, about not being able to assign a single role (say Role X) to multiple groups... I don't think that is correct. I was able to assign a role called Multiple Group Role to two different Groups. 

Maybe you mean "at the same time" as in during same action? That is true; one has to go to each Group separately, and assign the role.

pklaschka:
Hi @tanberry , thanks for your quick response!

> On your point, though, about not being able to assign a single role (say Role X) to multiple groups... I don't think that is correct. I was able to assign a role called Multiple Group Role to two different Groups.

If I assign a role (`Role X`) to a group while it is already assigned to a different group, I get a `400 Bad Request` with the message

```json
["Roles can only be used with a single group."]
```

In the UI, it shows a non-descriptive error message:

image

When I unassign the role from its current group assignment and try the same thing again, it works.

This is the case in 2024.10.4, at least… So on my end, the behavior matches the behavior in #10983 (where @BeryJu confirmed that limitation).

pklaschka:
My best guess would be that this limitation stems from this `OneToOneField`:

https://github.com/goauthentik/authentik/blob/5e72ec9c0c9d826c048627180d96a3e94df7bedd/authentik/rbac/models.py#L38-L47

But I have no experience with Django, so this is nothing more than an educated guess by a non-Django dev 😉 .

tanberry:
Aha, I learned a ton here, including the fact that I need to do a search on the `docs_needed` label! :-)

I want to do a bit more research about exactly which version this changed in (I suspect it is above in your notes but I need more coffee), and see if Notes are needed about that. 

But in meantime, let's merge your PR, and then I will open one to update the steps to use the dual-select component.

tanberry (approved):
Thanks again, you are welcome to comb over the docs and help us as much as you want! :-)

PR #12247, opened 2024-12-02
website/docs: add a note about dev environments and  read replicas

tanberry:
This PR adds a Warning about the inability to log back in if you are in a dev environment AND have read replicas configured in your `.env` file.


kensternberg-authentik (approved):
Yep, hit that one myself often enough.

tanberry:
Because we no longer write to the `.env.yaml` file, this will not be an issue going forward; if Blueprints turn into an issue we will open a diff PR.

PR #12260, opened 2024-12-04
web/flows: resize captcha iframes

kensternberg-authentik:
web: streamline CaptchaStage

# What

This commit:

Adds a ResizeObserver to the inner container of the host IFrame, giving us notice when something inside that container (such as a Captcha) requires resizing. I could not get the inner iframe to give up all of its sizing secrets, so I went with 2rem of padding to help prevent scrollsbars or overflowing. Primitive and old-school, but it seems to have worked well!

Re-organizes the Captcha Stage into a pattern that we’re trying to enforce throughout the code base. This makes the DX layout a little more consistent.

Replaces the big trees of conditionals, collections of `if { if { if { } } }` with a state table.

Breaks each Captcha clause into two separate handlers: one for the interactive version, and one for the execute-only version, and replaces the conditionals for each of those with a type:

    type CaptchaHandler = {
        name: string;
        interactive: () => Promise<unknown>;
        execute: () => Promise<unknown>;
    };

Replaces the `foreach()` with `for()`. [You cannot use a `forEach()` with async expressions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach#:~:text=does%20not%20wait%20for%20promises). If you need asynchronous behavior in an ordered loop, `for()` is the safest way to handle it; if you need asynchronous behavior from multiple promises, `Promise.allSettled(handlers.map())` is the way to go.

I tried to tell if this function *meant* to run every handler it found simultaneously, or if it meant to test them in order; I went with the second option, breaking and exiting the loop once a handler had run successfully.

I made the construction of the IFrame and the DocumentContainer lazy and on-demand. There are times when we simply won’t need or use them, and there’s no reason to build them prematurely.

Replaced the `removeEventListener` with a `ListenerController`, now that that’s an option.

# Details

The state table is pretty nifty! For example, in the `render()` function, there was a huge tree of conditionals for all the things that we wanted to have happen, but it turns out that what we really wanted were to check if `this.embedded` is true, if `this.challenge` is not undefined, and, if so, if `this.challenge.interactive` is true, and what to do under all those conditions. I think this makes that clear:

    // [isEmbedded, hasChallenge, isInteractive]
    return match([this.embedded, Boolean(this.challenge), Boolean(this.challenge?.interactive)])
      .with([true,  false, P.any], () => nothing)
      .with([true,  true,  false], () => nothing)
      .with([true,  true,  true],  () => this.renderBody())
      .with([false, false, P.any], () => akEmptyState({ loading: true }))
      .with([false, true,  P.any], () => this.renderMain())
      .exhaustive();

The `.exhaustive()` feature is especially nice: it’ll throw an exception at build time if I haven’t handled every possible combination!

Coverage for using `{ signal: options.signal }` to remove event listeners is [currently excellent](https://caniuse.com/?search=options.signal).


web/src/flow/stages/captcha/CaptchaStage.ts
```diff
@@ -43,19 +60,30 @@ export class CaptchaStage extends BaseStage<CaptchaChallenge, CaptchaChallengeRe
...
+            name: "turnstile",
+            interactive: this.renderTurnstileFrame,
+            execute: this.executeTurnstile,
+        },
+    ];
 
     @state()
     error?: string;
 
-    @state()
```
kensternberg-authentik, line 51: These objects were instantiated at build time and _never changed_.  There's no reason for them to be stateful or reactive.

web/src/flow/stages/captcha/CaptchaStage.ts
```diff
@@ -291,6 +313,63 @@ export class CaptchaStage extends BaseStage<CaptchaChallenge, CaptchaChallengeRe
...
+        // prettier-ignore
+        return match([this.embedded, Boolean(this.challenge), Boolean(this.challenge?.interactive)])
+            .with([true,  false, P.any], () => nothing)
+            .with([true,  true,  false], () => nothing)
+            .with([true,  true,  true],  () => this.renderBody())
+            .with([false, false, P.any], () => akEmptyState({ loading: true }))
+            .with([false, true,  P.any], () => this.renderMain())
+            .exhaustive();
+    }
+
```
kensternberg-authentik, line 328: This happens _after_ the render, so I'm trying to put the post-render lifecycle stuff there.

BeryJu (approved):

PR #12275, opened 2024-12-05
stages/redirect: create redirect stage

gergosimonyi:
Create Redirect stage. With it, you can

- redirect users to another Flow by setting the `target_flow` field of the stage
- redirect users to any URL by setting the `target_static` field of the stage
- if you're redirecting to a Flow, keep the flow context by setting the `keep_context` flag of the stage
- conditionally redirect users by placing a policy on a Redirect stage
- redirect users to dynamic locations by using the new flow context variable `redirect_stage_target`
    - to target a Flow, use the URL format `ak-flow://{flow_slug}`
- use the new flow context variable `is_redirected` in Expression policies, which holds the previous Flow
- use the new authentication requirement: `require_redirect`

---

## Checklist

If applicable


authentik/flows/planner.py
```diff
@@ -229,18 +236,13 @@ def plan(self, request: HttpRequest, default_context: dict[str, Any] | None = No
...
             else:
                 user = request.user
-                # We only need to check the flow authentication if it's planned without a user
-                # in the context, as a user in the context can only be set via the explicit code API
-                # or if a flow is restarted due to `invalid_response_action` being set to
-                # `restart_with_context`, which can only happen if the user was already authorized
-                # to use the flow
-                context.update(self._check_authentication(request))
+
+            context.update(self._check_authentication(request, context))
```
gergosimonyi, line 245: This change seems a bit scary but adding more `if`s seemed too convoluted. Performance loss seems negligible to me.

(If the `Outpost.objects.filter` call within the function is an issue, perhaps preface _that_ with an `if PLAN_CONTEXT_OUTPOST not in context`.)

authentik/flows/stage.py
```diff
@@ -92,7 +92,11 @@ def get_response_instance(self, data: QueryDict) -> ChallengeResponse:
 
     def get(self, request: HttpRequest, *args, **kwargs) -> HttpResponse:
         """Return a challenge for the frontend to solve"""
-        challenge = self._get_challenge(*args, **kwargs)
+        try:
+            challenge = self._get_challenge(*args, **kwargs)
+        except StageInvalidException as exc:
+            self.logger.debug("Got StageInvalidException", exc=exc)
+            return self.executor.stage_invalid()
```
gergosimonyi, line 99: Perhaps this is not even needed? I believe `StageInvalidException` should only be thrown by a validator, in our case, the line below this.

authentik/flows/views/executor.py
```diff
@@ -171,7 +171,8 @@ def dispatch(self, request: HttpRequest, flow_slug: str) -> HttpResponse:
                     # Existing plan is deleted from session and instance
                     self.plan = None
                     self.cancel()
-                self._logger.debug("f(exec): Continuing existing plan")
+                else:
+                    self._logger.debug("f(exec): Continuing existing plan")
```
gergosimonyi, line 175: This seems appropriate, but it changes logging behavior and I wanted to highlight that fact with this comment.

web/src/admin/stages/redirect/RedirectStageForm.ts
```diff
@@ -0,0 +1,153 @@
...
+            </span>
+            <ak-form-element-horizontal label=${msg("Name")} required name="name">
+                <input
+                    type="text"
+                    value="${this.instance?.name ?? ""}"
+                    class="pf-c-form-control"
+                    required
+                />
+            </ak-form-element-horizontal>
+            <ak-form-group .expanded=${true}>
```
kensternberg-authentik, line 60: Suggestion: Just `expanded`, please.  You don't need the whole `.expanded=${true}`.  Really, you don't.

web/src/admin/stages/redirect/RedirectStageForm.ts
```diff
@@ -0,0 +1,153 @@
...
+                <span slot="header"> ${msg("Stage-specific settings")} </span>
+                <div slot="body" class="pf-c-form">
+                    <ak-form-element-horizontal name="keepContext">
```
kensternberg-authentik, line 63: Suggestion: You can replace all of this with the `ak-switch-input` component.  :-).

kensternberg-authentik (approved):
The front-end work looks solid.  No real complaints, just a few suggestions / nitpicks.  I could do many more, but... this is fine.

authentik/stages/redirect/models.py
```diff
@@ -0,0 +1,49 @@
...
+
+    STATIC = "static"
+    FLOW = "flow"
+
+
+class RedirectStage(Stage):
+    """Redirect the user to another flow, potentially with all gathered context"""
+
+    keep_context = models.BooleanField(default=True)
+    mode = models.TextField(choices=RedirectMode.choices, default=RedirectMode.STATIC)
```
rissson, line 22:
```suggestion
    mode = models.TextField(choices=RedirectMode.choices, default=RedirectMode.STATIC)
```
I wouldn't set a default here

rissson:
Otherwise, I've tried out the basics and it works :tada:

authentik/stages/redirect/models.py
gergosimonyi: Fair enough. On the frontend it's fine, but there's no need here. Removed.

website/docs/add-secure-apps/flows-stages/flow/context/index.md
```diff
@@ -68,11 +70,15 @@ When a flow is executed by an Outpost (for example the [LDAP](../../../providers
...
+This key is set to `True` when the flow is executed from an "SSO" context. For example, this is set when a flow is used during the authentication or enrollment via an external source, and if a flow is executed to authorize access to an application.
 
 #### `is_restored` (Token object)
 
-Set when a flow execution is continued from a token. This happens for example when an [Email stage](../../stages/email/index.mdx) is used and the user clicks on the link within the email. The token object contains the key that was used to restore the flow execution.
+This key is set when a flow execution is continued from a token. This happens for example when an [Email stage](../../stages/email/index.mdx) is used and the user clicks on the link within the email. The token object contains the key that was used to restore the flow execution.
+
+#### `is_redirected` (Flow object)
+
+This key is set when the current flow was reached through a [Redirect stage](../../stages/redirect/index.md) in Flow mode.
```
tanberry, line 82: When we use the phrase "This key is set" it implies that the user does nothing, the system does the work of setting the key. That's true for the `is_redirected` key, but is it true for all the others?

website/docs/add-secure-apps/flows-stages/flow/context/index.md
```diff
@@ -189,3 +195,11 @@ Optionally override the email address that the email will be sent to. If not set
 ##### `pending_user_identifier` (string)
 
 If _Show matched user_ is disabled, this key will be set to the user identifier entered by the user in the identification stage.
+
+#### Redirect stage
+
+##### `redirect_stage_target` (string)
+
+[Set this key](../../../../customize/policies/expression/managing_flow_context_keys.md) in an Expression Policy to override [Redirect stage](../../stages/redirect/index.md) to force it to redirect to a certain URL or flow. This is useful when a flow requires that the redirection target be decided dynamically.
```
tanberry, line 203: Yesterday this sentence made perfect sense to me, lol. But tonight I am unsure of this part: "...an Expression Policy to override [Redirect stage](../../stages/redirect/index.md) to force it to redirect to a certain URL or flow."

Are we **overriding** the Redirect stage, or simply utilizing the Redirect stage? The override part is a bit confusing to me. Oh, wait, would an Admin use this Expression policy to target specific users or groups?? Like, all these people get the regular redirect stage, which will tkae them where they need to go. But THESE people, identified in the Expression, should be redirected to this other place...??

website/docs/add-secure-apps/flows-stages/flow/context/index.md
rissson: It's true for all of them

website/docs/add-secure-apps/flows-stages/stages/redirect/index.md
```diff
@@ -0,0 +1,17 @@
...
+
+## Redirect stage modes
+
+### Static mode
+
+When the user reaches this stage, they are redirected to a static URL.
+
+### Flow mode
+
+When the user reaches this stage, they are redirected to a Flow, retaining all [flow context](../../flow/context).
```
tanberry, line 15:
```suggestion
When the user reaches this stage, they are redirected to a specified flow, retaining all [flow context](../../flow/context).
```

website/docs/add-secure-apps/flows-stages/stages/redirect/index.md
```diff
@@ -0,0 +1,17 @@
...
+
+Optionally, untoggle the "Keep flow context" switch. If this is untoggled, all flow context is cleared with the exception of the [is_redirected](../../flow/context#is_redirected-flow-object) key.
```
tanberry, line 17:
```suggestion
Optionally, untoggle the "Keep flow context" switch. If this is untoggled, all flow context is cleared with the exception of the [`is_redirected`](../../flow/context#is_redirected-flow-object) key.
```

website/docs/customize/policies/expression/managing_flow_context_keys.md
```diff
@@ -0,0 +1,17 @@
+---
+title: Managing flow context keys
+---
+
+[Flow context](../../../add-secure-apps/flows-stages/flow/context/index.md) can be managed in [Expression policies](../expression.mdx) via the `context['flow_plan'].context` variable.
+
+Here's an example of setting a key:
```
tanberry, line 7:
```suggestion
Here's an example of setting a key within an Expression policy:
```

tanberry (approved):
A few nits, a few suggestions, a few questions, but overall Hurrah and Thank You. Thanks also for remembering the redirects. ;-)

website/docs/add-secure-apps/flows-stages/flow/context/index.md
gergosimonyi: Setting this key overrides the Redirect stage's `mode` and either `target_static` or `target_flow` (whichever fits the new `mode`). I think "override the Redirect stage" is a good enough wording of that, but I'm happy to adjust it.

---

> would an Admin use this Expression policy to target specific users or groups?

Correct. Let's say you have a Redirect stage with `mode == 'flow'` and `target_flow.slug == 'default-flow'`. Then you can do something like

```
if user_fits_criterion_1:
    context['flow_plan'].context['redirect_stage_target'] = 'ak-flow://custom-flow-1'
elif user_fits_criterion_2:
    context['flow_plan'].context['redirect_stage_target'] = 'ak-flow://custom-flow-2'

# all other users will be redirected according to the Redirect stage's settings
# in this case, to `default-flow`
```

website/docs/add-secure-apps/flows-stages/stages/redirect/index.md
gergosimonyi: I think this is a bit worse, because the link no longer looks like a link (to me).

image

vs

image

Let me know if you'd like it changed anyway, I'm not categorically opposed.

authentik/stages/redirect/models.py
```diff
@@ -0,0 +1,49 @@
...
+
+    STATIC = "static"
+    FLOW = "flow"
+
+
+class RedirectStage(Stage):
+    """Redirect the user to another flow, potentially with all gathered context"""
+
+    keep_context = models.BooleanField(default=True)
+    mode = models.TextField(choices=RedirectMode.choices)
```
BeryJu, line 22: nit: do we even bother saving the mode if we can determine which to redirect as only one of the `target_` variables can be set via API?

BeryJu (approved):

BeryJu (approved):

authentik/stages/redirect/models.py
gergosimonyi: This is one of those forever-questions, right? We have 2 subtypes, how do we represent them and how do we store them in a relational database?

I chose Single Table Inheritance because `RedirectStage` is simple enough. (At least for now. Dear developer reading this in 2027, feel free to change this decision if you think it's gotten out of hand).

PR #12289, opened 2024-12-06
enterprise: allow deletion/modification of users when in read-only mode

BeryJu:
## Details

REPLACE ME

---

## Checklist

If applicable

### DIFF
--- a/website/docs/enterprise/manage-enterprise.md
+++ b/website/docs/enterprise/manage-enterprise.md
@@ -107,7 +107,7 @@ The following events occur when a license expires or the internal/external user 
 
     - Users can authenticate and authorize applications
     - Licenses can be modified
-    - Users can be modified/deleted <span class="badge badge--version">authentik 2024.12+</span>
+    - Users can be modified/deleted <span class="badge badge--version">authentik 2024.10.5+</span>
 
     Once the user count returns to be within the limits of the license, authentik will return to the standard read-write mode and the notification will disappear.
 
--- a/website/docs/enterprise/manage-enterprise.md
+++ b/website/docs/enterprise/manage-enterprise.md
@@ -101,7 +101,15 @@ The following events occur when a license expires or the internal/external user 
 
 - After another 2 weeks, users get a warning banner
 
-- After another 2 weeks, the authentik Enterprise instance becomes “read-only”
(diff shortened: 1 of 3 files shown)


website/docs/enterprise/manage-enterprise.md
```diff
@@ -101,7 +101,15 @@ The following events occur when a license expires or the internal/external user
 
 - After another 2 weeks, users get a warning banner
 
-- After another 2 weeks, the authentik Enterprise instance becomes “read-only”
+- After another 2 weeks, the authentik Enterprise instance becomes "read-only"
```
tanberry, line 104:
```suggestion
- After another 2 weeks (six weeks after the expiry date), the authentik Enterprise instance becomes "read-only".
```

website/docs/enterprise/manage-enterprise.md
```diff
@@ -101,7 +101,15 @@ The following events occur when a license expires or the internal/external user
...
+
+    When an authentik instance has gone into read-only mode, the following actions can still be done:
+
+    - Users can authenticate and authorize applications
+    - Licenses can be modified
+    - Users can be modified/deleted <span class="badge badge--version">authentik 2024.10.5+</span>
+
+    Once the user count returns to be within the limits of the license, authentik will return to the standard read-write mode and the notification will disappear.
```
tanberry, line 112: Oh. I thought the problem was that the license had expired, not the number of users...? This is a it confusing... Do we mean both, that either the license has expired AND/OR the number of users has exceeded the licensed amount?

tanberry, line 112:
```suggestion
    After the user count returns to be within the limits of the license, authentik resets to the standard read-write mode and the notification will disappear.
```

tanberry (approved):
We need some clarification about user count/versus expiry date, but approving so as not to be the bottleneck.

website/docs/enterprise/manage-enterprise.md
tanberry, line 112:
```suggestion
    After the violation is corrected (either the user count returns to be within the limits of the license or the license is renewed), authentik will return to the standard read-write mode and the notification will disappear.
```

PR #12313, opened 2024-12-10
root: Added Minimum Memory and CPU Reservation

kaindlnetwork:
## Details

I've added minimum memory and CPU reservations into the docker compose file to ensure that the containers can start properly and to prevent running into out-of-memory issues by informing Docker about the expected resource consumption.

---

## Checklist

If applicable

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -18,6 +18,11 @@ services:
       POSTGRES_DB: ${PG_DB:-authentik}
     env_file:
       - .env
+    deploy:
+      resources:
+        reservations:
+          cpus: '0.0001'
+          memory: 150M
   redis:
     image: docker.io/library/redis:alpine
     command: --save 60 1 --loglevel warning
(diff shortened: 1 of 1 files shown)


rissson:
I'm not against this change directly, but maybe those should be left commented as it should be up to the user to configure those, especially if they don't have any configured elsewhere on their system. For instance, we don't configure those in the Helm chart, but we do provide examples of how to do so: https://github.com/goauthentik/helm/blob/main/charts/authentik/values.yaml#L352

kaindlnetwork:
@rissson Thank you for your feedback!

I see your point that these settings could be considered user-specific. However, I believe setting minimal memory and CPU reservations in the Docker-Compose file could serve as a best practice for several reasons:

Stability and Error Prevention: Without resource limits, especially new users might encounter hard-to-diagnose issues.
Standardized Configuration: The Docker-Compose file acts as a reference. Clear default settings promote a stable and reproducible setup.
Transparent Documentation: Even if these values were commented out, it would highlight their importance for maintaining a stable environment.
Since Docker-Compose is often used for simpler setups, reasonable default settings could be helpful. Perhaps adding a descriptive comment in the Docker-Compose file would be a good compromise?

What do you think about this approach? I’m open to alternative suggestions.

rissson:
Again, I'd like to see that provided to the user in the same fashion we do it for the Helm chart

PR #12328, opened 2024-12-12
website/docs: add new section about impersonation

tanberry:
This PR adds a new section about Impersonation.

### DIFF
--- a/website/docs/users-sources/user/user_basic_operations.md
+++ b/website/docs/users-sources/user/user_basic_operations.md
@@ -131,7 +131,7 @@ With authentik, an Admin can impersonate a user, meaning that the Admin temporar
 4. At the prompt, provide a reason why you are impersonating this user, and then click **Submit**.
 
 :::info
-An Admin can globally enable or disable impersonation in the [System Settings](../../sys-mgmt/settings.md#impersonation). By default, this option is set to true, meaning all users can be personified.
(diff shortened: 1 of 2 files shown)


rissson:
We also need to add a section about requiring a reason, I'll do that

website/docs/users-sources/user/user_basic_operations.md
```diff
@@ -120,3 +120,18 @@ You may instead deactivate the account to preserve identity data.
...
+
+With authentik, an Admin can impersonate a user, meaning that the Admin temporarily assumes the identity of the user.
+
+1. In the Admin interface, navigate to **Directory > Users** to display all users.
+2. Click the name of the user to display the full User details page.
+3. On the Overview tab, beneath **User Details**, in the **Actions** area, click **Impersonate**.
+4. At the prompt, provide a reason why you are impersonating this user, and then click **Submit**.
+
+:::info
+An Admin can globally enable or disable impersonation in the [System Settings](../../sys-mgmt/settings.md#impersonation). By default, this option is set to true, meaning all users can be personified.
```
rissson, line 134: Not sure `personified` is the correct word here

tanberry: Ah, yeah, we need to use the word "impersonated". Thanks for catching that!

rissson (approved):

PR #12347, opened 2024-12-13
website/docs: remove old version badges

gergosimonyi:
All the ones before 2024


tanberry (approved):
I think this is a good compromise.. it keeps the most recent badges but does a lot of cleanup. Thanks!

gergosimonyi:
There's still plenty of old version references, but not badges.

PR #12386, opened 2024-12-18
blueprints: add AtIndex tag

rissson:
## Details

First step towards #12367. Next we can add something like `!GetObject` which works like `!Find` but returns a mapping instance of only the primary key.

---

## Checklist

If applicable


BeryJu (commented):
Needs a version badge in the docs otherwise lgtm

BeryJu (approved):

PR #12407, opened 2024-12-19
root: redis, make sure tlscacert isn't an empty string

ilpianista:
## Details

I'm provisioning Authentik via the Helm chart and I want to use Redis over TLS without CA validation:

```
authentik:
  redis:
    host: master.XYZ.euw1.cache.amazonaws.com
    tls: true
    tls_reqs: none
```

However, it's failing with:

```
"error":"open : no such file or directory","event":"Failed to append  to RootCAs
```

It looks like it got an empty string as tls ca cert and it tries to append that empty string to the rootcas. I think you can avoid that by ensuring the tls ca cert isn't an empty string.

---

## Checklist

If applicable

### DIFF
--- a/internal/config/struct.go
+++ b/internal/config/struct.go
@@ -26,14 +26,14 @@ type Config struct {
 }
 
 type RedisConfig struct {
-	Host      string  `yaml:"host" env:"HOST, overwrite"`
-	Port      int     `yaml:"port" env:"PORT, overwrite"`
-	DB        int     `yaml:"db" env:"DB, overwrite"`
-	Username  string  `yaml:"username" env:"USERNAME, overwrite"`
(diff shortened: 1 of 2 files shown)


rissson:
Just tried this out, and indeed `TLSCaCert` is filled with an empty string instead of being nil. I'd rather find out why that happens instead of changing how we do the comparison. Otherwise, we can change the type to `string` instead of `*string` and compare to empty string.

@BeryJu thoughts?

ilpianista:
> I'd rather find out why that happens instead of changing how we do the comparison.

I suspect this is caused by the `overwrite` field info when the env var isn't set.

ilpianista:
@BeryJu friendly ping :pray:

BeryJu (approved):

rissson (approved):

PR #12497, opened 2024-12-27
sources/kerberos: authenticate with the user's username instead of the first username in authentik

natural-hair:
## Details

bugfix: kerberos source authentication attempts all passwords against the first username in Authentik.  Now it filters the Authentik user list with the user's username so that more than the first user can authenticate via kerberos source.

How you can tell if it works: check your kerberos server logs to confirm that the user's username is being used; create more than one user and try to log in with additional users.

---

## Checklist

If applicable


rissson:
On PTO but we should also filter to potentially find users where the entered username is their principal (i.e. the identifier in the usersourceconnection)

rissson (approved):

PR #10490, opened 2024-07-13
website/integrations: add MeshCentral

axemann:
## Details

This PR adds documentation for integrating authentik with [MeshCentral](https://meshcentral.com).

No API or frontend changes have been made.

---

## Checklist

-   [x] Local tests pass (`ak test authentik/`)
-   [x] The code has been formatted (`make lint-fix`)

If an API change has been made

-   [ ] The API schema has been updated (`make gen-build`)

If changes to the frontend have been made

-   [ ] The code has been formatted (`make web`)

If applicable

-   [x] The documentation has been updated
-   [x] The documentation has been formatted (`make website`)

PR #12509, opened 2024-12-27
website/integrations: meshcentral: document

dominic-r:
sdko/feat/integration/pickup-10490

closes #10490 

## Details

REPLACE ME

---

## Checklist

If applicable


website/integrations/services/meshcentral/index.md
```diff
@@ -0,0 +1,58 @@
...
+Create an OAuth2/OpenID provider with the following parameters:
+
+- Client Type: `Confidential`
+- Redirect URIs: `https://meshcentral.company/auth-oidc-callback`
+- Scopes: OpenID, Email and Profile
+- Signing Key: Select any available key
+
+Note the Client ID and Client Secret values.
+
+Next, create an application, using the provider you've created above.
```
tanberry, line 32: This makes me realize something... we have the Wizard now, where you can create an app and then a provider (and now even add bindings!), but is it, in a situation like this, easier/better to first create the provider, and *then* the app (using the old, pre-Wizard way)? I'd like to know your opinion, and I'll check with @BeryJu as well.

website/integrations/services/meshcentral/index.md
```diff
@@ -0,0 +1,58 @@
...
+
+## MeshCentral configuration
+
+Edit the `config.json` for your MeshCentral deployment, and add the following code in the `domains:` subsection:
```
tanberry, line 36:
```suggestion
Edit the `config.json` file for your MeshCentral deployment, and add the following code in the `domains:` subsection:
```

website/integrations/services/meshcentral/index.md
```diff
@@ -0,0 +1,58 @@
...
+                "oidc": {
+                    "issuer": "https://authentik.company/application/o/meshcentral/",
+                    "clientid": "<Client ID>",
+                    "clientsecret": "<Client Secret>",
+                    "newAccounts": true
+                }
+            },
+```
+
+To ensure everything is setup correctly, restart your MeshCentral instance and visit the main page. You should be greeted with a new button to allow signing in with OIDC.
```
tanberry, line 58: What a gorgeous sentence!! :-) But why does it show as highlighted in red here? The Deploy Preview looks normal. Or was that your way of showing me you added it, lol?

tanberry (changes requested):
Thanks so much, @4d62 ! I added a few nits, then a question about the Wizard...

website/integrations/services/meshcentral/index.md
tanberry: Thanks for the further info; I wasn't sure if there were certain situations where one really needed to do things the legacy way, and create the provider first. But sounds like the Wizard approach will work no matter what. We hope to make the Wizard the default method soon, so yes I agree would be good to just simplify the steps and link to the Wizard instructions.  And.. then do the same in all the integrations guides, yee-haw.

tanberry (approved):

website/integrations/services/meshcentral/index.md
tanberry: To change it to use the Wizard, let's use [this section](https://docs.goauthentik.io/integrations/services/vmware-vcenter/#authentik-configuration) from the VCenter integration guide as our "template". It provides a section to cover any special notes or info about both the app and the provider. Should do the trick. I guess we should also  add a link to the Wizard docs: https://docs.goauthentik.io/docs/add-secure-apps/applications/manage_apps#instructions (ugh I don't like that title Instructions but oh well).

tanberry: Although... I just saw the section you used for the Komga docs, and I kinda like it too...do as you like for this guide, and then let's collab on creating exactly the block we want for the template.

PR #12526, opened 2025-01-01
web/flows: fix `login` / `log in` inconsistency

sweepies:
## Details
This seems to be the only place in the project where `Login` is used in place of `Log in`. It's incorrect so this fixes that.

## Checklist
Not applicable


tanberry (approved):
Thanks @sweepies !

PR #12549, opened 2025-01-03
website/docs: user-sources: fix FreeIPA docs

eddster2309:
## Details
Was setting up a FreeIPA sync to Authentik and noticed that the location that the docs point to is incorrect, so this is a PR to correct this inline with the [Active Directory page](https://docs.goauthentik.io/docs/users-sources/sources/directory-sync/active-directory/).

## Checklist


dominic-r (approved):
Thanks!

rissson (approved):

PR #12554, opened 2025-01-03
website/integrations: Fix deprecated terraform ressource authentik_scope_mapping in docs

MaBauMeBad:
Just replacing the deprecated Terraform ressource `authentik_scope_mapping` with the new `authentik_property_mapping_provider_scope` in the Grafana Application.

Also fixing in incomplete attempt to do the same with ArgoCD.

### DIFF
--- a/website/integrations/services/argocd/index.md
+++ b/website/integrations/services/argocd/index.md
@@ -110,9 +110,9 @@ resource "authentik_provider_oauth2" "argocd" {
   ]
 
   property_mappings = [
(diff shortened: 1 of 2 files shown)


dominic-r (approved):
Thanks!

As confirmed by https://github.com/goauthentik/terraform-provider-authentik/blob/d66017070e2a71474ef73a313e0729f56b634ee6/internal/provider/provider.go#L135

rissson (approved):

PR #12591, opened 2025-01-08
flow: fix regression in flow planner

### LINKED ISSUE #12445: Regression 2024.12 breaks recovery flow
**Describe the bug**
Since upgrading to 2024.12, I am unable to create recovery links or send recovery emails. The endpoints

```
https://authentik.spacemule.net/api/v3/core/users/XX/recovery/
https://authentik.spacemule.net/api/v3/core/users/XX/recovery_email/?email_stage=XXXXX
```

return 400 with the following error:

```
Recovery flow not applicable to user
```

In the UI, there is a message that no recovery flow is set for the brand *only when using the create recovery link button*.

My brand does have a recovery flow set. I have tried unsetting and resetting it, but the behavior persists.

**To Reproduce**

1. Go to Directory->Users->Specific user
2. Click on "Create Recovery Link" or "Email recovery link" and choose an email stage
3. See error

**Expected behavior**
The email should send, or the link should be copied to the keyboard

**Logs**
I could not find any relevant ones

-   authentik version: 2024.12
-   Deployment: helm

**Additional context**
When reverting to version 2024.10.5, authentik acts as expected.

ht-danielgo:
## Details

I believe that during this commit 
https://github.com/goauthentik/authentik/commit/ff504a3b80e275f39633883025fd6b79e8aa4062#diff-a5c56bb7c60e27dda1b131b3fc2a17e3af6624e7cfaaa2337ec6b077ca489f34R245 the flow planner regressed, meaning that the authentication requirements are evaluated sometimes unnecessarily

This manifested in https://github.com/goauthentik/authentik/issues/12445 which seems to be occurring when the recovery flow has an Unauthenticated requirement and you try to generate a link via the Admin console, which sends a context user. 

closes #12445

---

## Checklist

### DIFF
--- a/authentik/flows/planner.py
+++ b/authentik/flows/planner.py
@@ -241,8 +241,8 @@ class FlowPlanner:
                 user = context[PLAN_CONTEXT_PENDING_USER]
             else:
                 user = request.user
+                context.update(self._check_authentication(request, context))
 
-            context.update(self._check_authentication(request, context))
             # First off, check the flow's direct policy bindings
             # to make sure the user even has access to the flow
             engine = PolicyEngine(self.flow, user, request)


gergosimonyi:
#12445 was fixed via other means.

Thank you for the PR!

PR #12600, opened 2025-01-08
web: fix scrollbar styling

### LINKED ISSUE #11936: Unnecessary scrollbars in Admin pages [Firefox]
**Describe the bug**
I've noticed a minor UI bug in Firefox; the admin pages often contain unnecessary scrollbars. This is only reproducible in Firefox it seems. (Further digging & potential solution in _Additional Context_)

**To Reproduce**

1. Open the Admin dashboard overview page in Firefox browser.
2. Compare against same page in any chromium based browser.

**Expected behavior**
No scrollbars when not necessary.

**Screenshots**
Firefox:
authentik_admin_firefox

Chrome:
authentik_admin_chromium

**Logs**
N/a

-   authentik version: 2024.8.4

**Additional context**
After looking at the CSS for the Admin page, it seems to occur on elements of the `.pf-c-card__body` class which sets `overflow-x: scroll;`.
As Firefox has poor support for `::-webkit-scrollbar` it will not use the custom scrollbar styling defined in [authentik.css](https://github.com/goauthentik/authentik/blob/49577fe333565a57bd86f47c1c80570f1df0cce0/web/src/common/styles/authentik.css#L20).

Not sure if this would be the proper way of fixing this, but setting the overflow to `auto` seems to resolve the issue. Perhaps conditionally by [checking if the browser supports webkit-scrollbar](https://stackoverflow.com/a/65102430).

Alternatively [standardized scrollbar styling](https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-color) could be used as noted by the [webkit-scrollbar dev pages](https://developer.mozilla.org/en-US/docs/Web/CSS/::-webkit-scrollbar).

RoboMagus:
## Details

Closes #11936. 
This PR removes unnecessary scrollbars on the `.pf-c-card__body` class, as well as add scrollbar styling to browsers that do not support the `::-webkit-scrollbar`. (Currently on Firefox thick unstyled grey scrollbars show up whenever a container is scrollable)

Note that this only changes the color scheme, as scrollbar thickness styling is not inherited (see [this issue](
https://github.com/w3c/csswg-drafts/issues/4799)) and therefore would require modifying a significant number of classes to add `scrollbar-width`.

---

## Checklist

If applicable


kensternberg-authentik:
@RoboMagus : Quick question; where would I go in the UI to observe the changes you made?

RoboMagus:
See linked issue #11936 
Its mostly the main admin dashboard that is affected

kensternberg-authentik (approved):
Good work, but what does it mean if the browser doesn't support `@supports` ?

melizeche:
@kensternberg-authentik seems that's pretty well supported in all major browsers, probably many other things also could be broken in a really old browser
 https://caniuse.com/mdn-css_at-rules_supports_selector

PR #12604, opened 2025-01-09
website/integrations: pgadmin: refactor

dominic-r:
sdko/integrations/pgadmin/refactor

@tanberry yeehaw 🤠

## Details

REPLACE ME

---

## Checklist

If applicable


website/integrations/services/pgadmin/index.md
```diff
@@ -13,89 +13,82 @@ sidebar_label: pgAdmin
 >
 > -- https://www.pgadmin.org/
 
-:::note
-This is based on authentik 2022.3.3 and pgAdmin4 6.19
-:::
-
```
tanberry, line 19: Assuming you updated the docs to reflect a more recent version of pgAdmin? If so, we should perhaps state that here.

website/integrations/services/pgadmin/index.md
```diff
@@ -13,89 +13,82 @@ sidebar_label: pgAdmin
...
-	'OAUTH2_ICON' : '<fontawesome-icon>',
-	'OAUTH2_BUTTON_COLOR' : '<button-color>'
-}]
-```
+# authentik configuration
+
+1. From the Admin interface, navigate to **Applications** -> **Applications**.
+2. Use the wizard to create a new application and provider. During this process:
+    - Note the **Client ID**, **Client Secret**, and **slug** values because they will be required later.
+    - Set the redirect URI to `https://pgadmin.company`.
```
tanberry, line 28: We now have a fancy new **Redirect URI** field... one can chose `strict` or `regex`... might be good to specify this is a strict value, if so?

website/integrations/services/pgadmin/index.md
```diff
@@ -13,89 +13,82 @@ sidebar_label: pgAdmin
...
+    - Select any available signing key.
+
+## pgAdmin OAuth Configuration
+
+To configure OAuth in pgAdmin, you can either use the `config_local.py` file or set environment variables if you are deploying pgAdmin in a containerized setup.
```
tanberry, line 33: Do we ned to tell them exactly where to add the environment variables, as in which file? If we can, definitely let's do. Can be quite frustrating for users to be told to do something, but not where to do it.

website/integrations/services/pgadmin/index.md
```diff
@@ -13,89 +13,82 @@ sidebar_label: pgAdmin
...
+
+### Using `config_local.py`
+
+1. Locate or create the `config_local.py` file in the `/pgadmin4/` directory.
+
+    - If the file does not exist, create it manually.
 
-In the code above the following placeholders have been used:
+2. Add the following configuration to `config_local.py`:
```
tanberry, line 41:
```suggestion
2. Add the following configuration settings to `config_local.py`:
```

website/integrations/services/pgadmin/index.md
tanberry: oh, I see you have it right below... ok.

website/integrations/services/pgadmin/index.md
```diff
@@ -13,89 +13,82 @@ sidebar_label: pgAdmin
...
+    Ensure that you promote at least one user to an admin before disabling the internal authentication.
 
-:::note
-pgAdmin needs to be restarted **every** time changes to `config_local.py` are made
-:::
+- To **disable automatic user creation**, set:
+    ```python
+    OAUTH2_AUTO_CREATE_USER = False
+    ```
+    Setting this value to `False` disables automatic user creation. This ensures that only the first signed-in user is registered.
```
tanberry, line 94: Can we get a final section for verification? :-)

website/integrations/services/pgadmin/index.md
tanberry: Good question... yeah true most other integration guides don't have it. But, it is useful data, so let's not leave it out. Anythng useful to our dear readers, let's include. Indeed, maybe one day we could get it added to the others. ;-)

tanberry (approved):
Thank you so much! I rather do like your short-and-sweet instructions for the Wizard. (This integration is a lot less complex than other one, with 4 pairs.)

tanberry (approved):
Thanks @4d62 !

PR #12641, opened 2025-01-11
website/integrations: update uptime-kuma skip path documentation

itskagee:
Modify the skip path regex to be more restrictive without modifying the intended functionality.

## Checklist

If applicable


website/integrations/services/uptime-kuma/index.md
```diff
@@ -40,14 +40,24 @@ Create an application in authentik. Create a Proxy provider with the following p
...
+    ^/assets/.*
+    ^/api/push/.*
+    ^/api/badge/.*
+    ^/api/status-page/heartbeat/.*
     ^/icon.svg
-    ^/api/.*
     ^/upload/.*
-    ^/metrics
+    ```
+    Alternatively, you can get even more specific by analysing the requests for your status pages and modifying the regex rules above accordingly.
```
tanberry, line 51:
```suggestion
    Alternatively, you can get even more specific by analyzing the requests for your status pages and modifying the regex rules above accordingly.    
```

tanberry: We try to use US English, so "z" not "s" but I too struggle with this...

website/integrations/services/uptime-kuma/index.md
```diff
@@ -40,14 +40,24 @@ Create an application in authentik. Create a Proxy provider with the following p
...
+    ^/api/push/.*
+    ^/api/badge/.*
+    ^/api/status-page/heartbeat/.*
     ^/icon.svg
-    ^/api/.*
     ^/upload/.*
-    ^/metrics
+    ```
+    Alternatively, you can get even more specific by analysing the requests for your status pages and modifying the regex rules above accordingly.    
+    For example,
```
tanberry, line 52:
```suggestion
    For example:
```

website/integrations/services/uptime-kuma/index.md
tanberry: Can we explain briefly why adding the splat (or is it a wild card?) is better? How does this make it "more" specific? I am not good with regex, but it seems it would make it less specific, and result in more ... results...?

tanberry (approved):
Thank you @itskagee ! Nothing major, some questions and perhaps some needed clarification, but approving now.

tanberry:
Hi @itskagee the `lint (prettier-check)` build check is failing... can you run `make website` again loclaly to format it, then push to PR?

website/integrations/services/uptime-kuma/index.md
itskagee: Oh whoops, my bad.
I try to keep it US English online, but that UK English schooling slips through sometimes haha.

As for the regex, I added the splats because the previous regex was essentially working the same way. e.g. `^/status` was being matched like how `^/status/.*` is now, so I changed it to make it more informative at the first glace itself.

And as for the specificity, I could only get a few to be specific enough without causing problems. So those now end with `$` to force the end of a match. There are still some entries that end with `.*`, but those have to be that way, unfortunately.

So essentially, the changes proposed here either keep the regex the same, or make some of it more specific in use.

I also removed a few entries like `^/metrics` which REALLY shouldn't have been public without authentication.

itskagee:
Apologies.

I thought a linting check might have been unnecessary for such a small change, so I did it from the github interface itself.

I've since updated the PR to add your suggestions and the linting changes!

PR #12656, opened 2025-01-13
flows/inspector: add button to open flow inspector

BeryJu:
## Details

Change flow inspector from {open, close} to a tri-state of {open, close, available}

When a user has the permission to inspect the flow the URL will automatically include `inspector=available` which renders a button to open the inspector. When clicking the button to execute the flow with inspector, the URL will be set to use `inspector=open` which always opens the inspector.

---

## Checklist

If applicable

### DIFF
--- a/website/docs/add-secure-apps/flows-stages/flow/inspector.md
+++ b/website/docs/add-secure-apps/flows-stages/flow/inspector.md
@@ -14,13 +14,13 @@ As shown in the screenshot below, the flow inspector displays next to the select
 Be aware that when running a flow with the inspector enabled, the flow is still executed normally. This means that for example, a [User write](../stages/user_write.md) stage _will_ write user data.
 :::
 
-### Permissions and debug mode
-
-By default, the inspector is only enabled when the currently authenticated user is a superuser, OR if a user has been granted the [permission](../../../users-sources/access-control/permissions.md) **Can inspect a Flow's execution** (or is a user assigned to role with the permission).
+The inspector is accessible to users that have been granted the [permission](../../../users-sources/access-control/permissions.md) **Can inspect a Flow's execution**, either directly or through a role. Superuser can always inspect flow executions.
 
 When developing authentik with the debug mode enabled, the inspector is enabled by default and can be accessed by both unauthenticated users and standard users. However the debug mode should only be used for the development of authentik. So unless you are a developer and need the more verbose error information, the best practice for using the flow inspector is to assign the permission, not use debug mode.
 
-### Open the Flow Inspector
+Starting with authentik 2025.2, when the user can access the inspector, a button is shown in the top right of the [default flow executor](./executors/if-flow.md) which opens the flow inspector.
+
+### Manually running a flow with the inspector
 
 1. To access the inspector, open the Admin interface and navigate to **Flows and Stages -> Flows**.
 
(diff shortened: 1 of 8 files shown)


website/docs/add-secure-apps/flows-stages/flow/inspector.md
```diff
@@ -14,13 +14,13 @@ As shown in the screenshot below, the flow inspector displays next to the select
 Be aware that when running a flow with the inspector enabled, the flow is still executed normally. This means that for example, a [User write](../stages/user_write.md) stage _will_ write user data.
 :::
 
-### Permissions and debug mode
-
-By default, the inspector is only enabled when the currently authenticated user is a superuser, OR if a user has been granted the [permission](../../../users-sources/access-control/permissions.md) **Can inspect a Flow's execution** (or is a user assigned to role with the permission).
+The inspector is accessible to users that have been granted the [permission](../../../users-sources/access-control/permissions.md) **Can inspect a Flow's execution**, either directly or through a role. Superuser can always inspect flow executions.
```
tanberry, line 17:
```suggestion
The inspector is accessible to users that have been granted the [permission](../../../users-sources/access-control/permissions.md) **Can inspect a Flow's execution**, either directly or through a role. Superusers can always inspect flow executions.
```

tanberry: And we still want to add more, right, about how they also need View permissions on that specific flow?

website/docs/add-secure-apps/flows-stages/flow/inspector.md
```diff
@@ -14,13 +14,13 @@ As shown in the screenshot below, the flow inspector displays next to the select
...
 
 When developing authentik with the debug mode enabled, the inspector is enabled by default and can be accessed by both unauthenticated users and standard users. However the debug mode should only be used for the development of authentik. So unless you are a developer and need the more verbose error information, the best practice for using the flow inspector is to assign the permission, not use debug mode.
 
-### Open the Flow Inspector
+Starting with authentik 2025.2, when the user can access the inspector, a button is shown in the top right of the [default flow executor](./executors/if-flow.md) which opens the flow inspector.
```
tanberry, line 21:
```suggestion
Starting with authentik 2025.2, for users with appropriate permissions to access the inspector a button is shown in the top right of the [default flow executor](./executors/if-flow.md) which opens the flow inspector.
```

website/docs/add-secure-apps/flows-stages/flow/inspector.md
BeryJu: I think you're mixing that up with app entitlements, in this case they dont need any other permission

website/docs/add-secure-apps/flows-stages/flow/inspector.md
```diff
@@ -35,10 +35,8 @@ Alternatively, a user with the correct permission can launch the inspector by ad
 :::info
 Troubleshooting:
 
-- If the flow inspector does not launch and a "Bad request" error displays, this is likely either because you selected a flow that is not defined in your instance or the flow has a policy bound directly to it that prevents access, so the inspector won't open because the flow can't run results.
-- If the flow inspector launches but is empty, you can refresh the browser or advance the flow to load the inspector. This can occur when a race condition happens (the inspector tries to fetch the data before the flow plan is fully planned and as such the panel just shows blank).
-
-:::
+- If the flow inspector does not launch and a "Bad request" error displays, this is likely either because you selected a flow that has a policy bound directly to it that prevents access, so the inspector won't open because the flow can't be executed.
```
tanberry, line 38: You say "either" but then only give one possible reason. Is the other reason that the user does not have view permissions on the flow?

If yes, then maybe this wording:

- If the flow inspector does not launch and a "Bad request" error displays, this is likely either because you selected a flow that has a policy bound directly to it that prevents access (so the inspector won't open because the flow can't be executed) or because you do not have view permission on that specific flow.

PR #12663, opened 2025-01-14
website/integrations: cloudflare-access: refactor

dominic-r:
sdko/integrations/cfa/refactor

## Details

REPLACE ME

---

## Checklist

If applicable


website/integrations/services/cloudflare-access/index.md
```diff
@@ -0,0 +1,44 @@
...
+
+> Cloudflare Access is a secure, cloud-based zero-trust solution for managing and authenticating user access to internal applications and resources.
+>
+> -- https://www.cloudflare.com/zero-trust/products/access/
+
+## Preparation
+
+The following placeholders are used in this guide:
+
+- `company.cloudflareaccess.com` is the FQDN of your Cloudflare Access subdomain.
```
tanberry, line 20:
```suggestion
- `company.cloudflareaccess.com` is the FQDN of your Cloudflare Access subdomain
```

website/integrations/services/cloudflare-access/index.md
```diff
@@ -0,0 +1,44 @@
...
+1. From the Admin interface, navigate to **Applications** -> **Applications**.
+2. Use the wizard to create a new application and provider. During this process:
+    - Note the **Client ID**, **Client Secret**, and **slug** values because they will be required later.
+    - Set a `Strict` redirect URI to `https://company.cloudflareaccess.com/cdn-cgi/access/callback`.
+    - Select any available signing key.
+
+## Cloudflare Access configuration
+
+1. Open the [Cloudflare Access dashboard](https://one.dash.cloudflare.com) and navigate to **Settings** -> **Authentication**.
+2. Click on **Login methods**, then select **Add** -> **OpenID Connect**.
```
tanberry, line 36:
```suggestion
2. Click **Login methods**, and then select **Add** -> **OpenID Connect**.
```

website/integrations/services/cloudflare-access/index.md
```diff
@@ -0,0 +1,44 @@
...
+
+- `company.cloudflareaccess.com` is the FQDN of your Cloudflare Access subdomain.
+- `authentik.company` is the FQDN of the authentik install.
+
+To proceed, you need to register for a free Cloudflare Access account and have both a Cloudflare account and a publicly accessible authentik instance with a trusted SSL certificate.
+
+## authentik configuration
+
+1. From the Admin interface, navigate to **Applications** -> **Applications**.
+2. Use the wizard to create a new application and provider. During this process:
```
tanberry, line 28: For CF-access we can;t use the generic templte here; there are too many details about the configuration settings that we need to provide. Let's use the more robust structure, with space to call out specific values needed for both the app and the provider. For example, for the app, they need to need to set the **Policy engine mode** is set to **ANY**. There are a few specific config settings for providers as well (client type to **Confidential**). Take a very careful look at each of the steps, each of the values, called out in the original guide, and make sure they don't get lost, please. ;-)

tanberry (changes requested):
A few easy changes, and then let's rework the authentik configuration section to separate app form provider and call out the important config settings for each.

website/integrations/services/cloudflare-access/index.md
tanberry: Agree that if they are defaults (and I like your sentence telling them they can leave defaults) AND if they are not specifically called out in the original guide, then we can skip them. 

But any of the values that are explicitly stated or called out in the CF guide need to be called out explicitly in this new Guide too.. even if it is the default. This is one that Jens and I stepped through, so I am confident that we need to mention those specific values. Let's try it see how it look, please.

tanberry (approved):
Thanks @4d62 for updating this doc to use the wizard.

Issue #12332, opened 2024-12-12
In Dropdown List select item twice


cheggerdev:
**Describe the bug**
In a dropdown list anywhere in authentik an item must be selected twice to be accepted.
Condition to reproduce is that the list is long enough that you have to scroll down.
If the list is not long enough or the list item to chose does not require to scroll then it works fine.

**To Reproduce**

1. Go to 'Create' anywhere
2. Click on 'flow, policy, user, group'
3. Scroll down to 'your item to select'
4. List closes and item clicked on is not selected
5. repeat steps 2-4 again, the second time step 4 works

**Expected behavior**
I can select the item on first time no matter if I have to scroll down or not.

**Version and Deployment (please complete the following information):**

-   authentik version: [2024.10.5]
-   Deployment: [docker]

cheggerdev:
I can still reproduce this with 2024.12.2.

I use the firefox browser.

To reproduce the dropdown list must have more than 12 items.
The first twelve items work as expected: Just click once and the item is selected.

The thirteens and beyond item I have to scroll down and upon the **second** click the item is selected.

cheggerdev:
This issue still persists with 2025.2.0

cheggerdev:
This issue still persists with 2025.2.4

cheggerdev:
This issue still persists with 2025.4.0

cheggerdev:
This issue still persists in 2025.6.0

cheggerdev:
This issue still persists in 2025.8.0
There is #16278 on the way.

cheggerdev:
I can no longer reproduce this issue with 2025.8.1. Thank you @GirlBossRush

PR #12690, opened 2025-01-16
web: clean up dual-select event management

kensternberg-authentik:
web: restructure ak-dual-select to avoid using CustomEvent

# What

- Replace all `CustomEvent<T>` handlers with a child class Event.
- Use the standard `dispatchEvent()` and `addEventListener()` clauses.
- Move common methods of the two panes into an abstract parent class.

# Why

CustomEvent is a bit of a side-show in JavaScript; it’s perfectly acceptable to inherit from Event, and there’s much better support for type management while using it, plus deconstructing the received event object is cleaner with child Events than unpacking the `details` object.

This codebase is now more open to change, and is closer to our still-evolving style.

Doing it this way mean that, once you have an event defined, you don’t have to remember if you’re sending a custom event or a normal event, and you don’t need all that infrastructure for making your Lit objects sensitive to custom event handling. There’s enough mixin clutter already.

And just like I hate clutter, I hate duplication. The two panes had a lot in common with ARIA handling and storing, clearing, and assigning selected items to the “pending move” table. Moving all that into a parent class exposed the differences: one is a source and the other a sink; one reflects changes made, the other possible changes to be made.

# Testing

The Storybook tests have all been updated to match the codebase, and there are standalone tests for the various components (pagination, pane control, search) that can be exercised before deployment.


web/src/elements/ak-dual-select/ak-dual-select.ts
```diff
@@ -123,47 +106,17 @@ export class AkDualSelect extends CustomEmitterElement(CustomListenerElement(AKE
...
+    handleMove(event: DualSelectMoveRequestEvent) {
+        match(event.move)
+            .with("add-all", () => this.addAllVisible())
+            .with("add-one", () => this.addOne(event.key))
+            .with("add-selected", () => this.addSelected())
+            .with("delete-all", () => this.removeAll())
+            .with("remove-all", () => this.removeAllVisible())
+            .with("remove-one", () => this.removeOne(event.key))
+            .with("remove-selected", () => this.removeSelected())
+            .exhaustive();
```
kensternberg-authentik, line 118: My favorite new thing. If you miss one of the possible sub-types, this will not pass lint or build.

cheggerdev:
Is there a possibility that this addresses https://github.com/goauthentik/authentik/issues/12332 ?

kensternberg-authentik:
@cheggerdev : Thank you for pointing that out!  I don't think so... but that means I'm going to have to look closer.  :-)

BeryJu (approved):
LGTM I think

PR #12700, opened 2025-01-16
core: Tidy contributor onboarding, fix typos.

GirlBossRush:
## Details

This PR contains a _subset_ of fixes discovered when progressing through the authentik contributor guide. Highlights include:

- Merging of disparate code spell dictionaries
- Optional Poetry env for hygiene commands (e.g. lint, build, etc)
- Fix for stale links, typos.

---

## Checklist

If applicable


pyproject.toml
```diff
@@ -4,6 +4,32 @@ version = "2024.12.2"
...
+    "go.sum",
+    "cmd",
+    "internal",
+    "web/src",
+    "locale",
+    "**/web/xliff/**",
+    "./website/blog/**",
+    "./website/docs/**",
+    "./website/integrations/**",
+    "./website/src/**",
```
rissson, line 27:
```suggestion
    "locale",
    "**/web/xliff/**",
    "./website/src/**",
```

Many of those we definitely don't want to skip. It's nice to catch spelling mistakes in the code. You can also remove `authentik` from that list.

On top of that, I would say that this might cause more problems down the line with generated files from the tools we use (like `__pycache__`, `gen-go-api`, `gen-ts-api`, `unittest.xml` to name a few`). I would rather go with the previous approach of including what needs to be included and leaving the rest alone.

rissson: But then again, we were missing a few things like `scripts`. If codespell could ignore the same things as .gitignore, plus a few generated committed things that'd be great, if we can reuse the .gitignore file, let's not duplicate that list

rissson (changes requested):

website/package.json
```diff
@@ -1,12 +1,17 @@
...
     "scripts": {
         "aws-cfn": "cd docs/install-config/install/aws && cdk synth > template.yaml && ./fix_template.py",
-        "build": "cp ../docker-compose.yml static/docker-compose.yml && cp ../schema.yml static/schema.yaml && docusaurus gen-api-docs all && cross-env NODE_OPTIONS='--max_old_space_size=65536' docusaurus build",
-        "build-bundled": "cp ../schema.yml static/schema.yaml && docusaurus gen-api-docs all && cross-env NODE_OPTIONS='--max_old_space_size=65536' docusaurus build",
+        "build": "run-s build:docker-compose build-bundled",
+        "build-bundled": "run-s build:schema build:api-docs build:docusaurus",
+        "build:api-docs": "docusaurus gen-api-docs all",
+        "build:docker-compose": "cp ../docker-compose.yml static/docker-compose.yml",
+        "build:docusaurus": "cross-env NODE_OPTIONS='--max_old_space_size=65536' docusaurus build",
+        "build:schema": "cp ../schema.yml static/schema.yml",
```
GirlBossRush, line 14: @rissson Found the source of the mystery `.yaml` file 😊

I've tidied up the build sub-scripts to aid in readability.

Makefile
```diff
@@ -55,23 +41,22 @@ test-docker:  ## Run all tests in a docker-compose
 	rm -f .env
 
 test: ## Run the server tests and produce a coverage report (locally)
-	coverage run manage.py test --keepdb authentik
-	coverage html
-	coverage report
+	poetry run coverage run manage.py test --keepdb authentik
```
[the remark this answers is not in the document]
GirlBossRush: Hi @rissson!

Thanks for reviewing my PR. I'm open to either approach, but it may help to keep the prefix.

When running `make test` without a Poetry shell activated, `make` doesn't know where to find the `coverage` command and emits an error without a hint at the solution.

```sh
❯ make test
coverage run manage.py test --keepdb authentik
make: coverage: No such file or directory
make: *** [test] Error 1
```

The virtual environment instructions appear in the [full-stack setup docs](https://docs.goauthentik.io/docs/developer-docs/setup/full-dev-environment#backend-setup), but my concern is that contributors who start with the [front-end](https://docs.goauthentik.io/docs/developer-docs/setup/frontend-dev-environment) or [docs](https://docs.goauthentik.io/docs/developer-docs/setup/website-dev-environment) environments might miss the Python requirements when switching contexts.

Since ci-main.yml uses the poetry prefix consistently, would it make sense to keep it for now? We could look at removing it from both places as a separate task if you think that would be valuable.

Thanks for your feedback!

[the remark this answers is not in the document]
rissson: So if you add `poetry run` invocations to the Makefile, then it makes sense to remove it from the CI.

Both approaches (having it in CI or having it in the Makefile) are valid, but I think we should only do one of them. I agree that having it in the Makefile is easier for new contributors

Makefile
```diff
@@ -55,23 +41,21 @@ test-docker:  ## Run all tests in a docker-compose
...
+	poetry run black $(PY_SOURCES)
+	poetry run ruff check --fix $(PY_SOURCES)
 
 lint-codespell:  ## Reports spelling errors.
-	codespell -w $(CODESPELL_ARGS)
+	poetry run codespell -w
 
 lint: ## Lint the python and golang sources
-	bandit -r $(PY_SOURCES) -x web/node_modules -x tests/wdio/node_modules -x website/node_modules
-	golangci-lint run -v
```
rissson, line 71: Not sure why those got removed

Makefile
```diff
@@ -55,23 +41,21 @@ test-docker:  ## Run all tests in a docker-compose
...
 
 core-install:
-	poetry install
+	poetry run bandit -c pyproject.toml -r $(PY_SOURCES)
```
rissson, line 58:
```suggestion
	poetry install
```

Makefile
```diff
@@ -274,16 +258,17 @@ ci--meta-debug:
...
+	poetry run ruff check $(PY_SOURCES)
 
 ci-codespell: ci--meta-debug
-	codespell $(CODESPELL_ARGS) -s
+	poetry run codespell -s
 
 ci-bandit: ci--meta-debug
-	bandit -r $(PY_SOURCES)
+	poetry run bandit -r $(PY_SOURCES)
+	golangci-lint run -v internal/...
```
rissson, line 271:
```suggestion
	poetry run bandit -r $(PY_SOURCES)
```

manage.py
```diff
@@ -4,11 +4,11 @@
 import sys
 import warnings
 
-from authentik.lib.config import CONFIG
 from cryptography.hazmat.backends.openssl.backend import backend
 from defusedxml import defuse_stdlib
 from django.utils.autoreload import DJANGO_AUTORELOAD_ENV
 
+from authentik.lib.config import CONFIG
```
rissson, line 11:
```suggestion
```

pyproject.toml
```diff
@@ -12,7 +32,7 @@ exclude = 'node_modules'
 [tool.ruff]
 line-length = 100
 target-version = "py312"
-exclude = ["**/migrations/**", "**/node_modules/**"]
+exclude = ["web/tools/**", "**/migrations/**", "**/node_modules/**"]
```
rissson, line 35: Why are we excluding those scripts from being linted and formatted?

Makefile
```diff
@@ -4,7 +4,7 @@
 PWD = $(shell pwd)
 UID = $(shell id -u)
 GID = $(shell id -g)
-NPM_VERSION = $(shell python -m scripts.npm_version)
+NPM_VERSION = $(shell python -m scripts.generate_semver)
```
rissson, line 7: We could also do this

```suggestion
NPM_VERSION = $(shell poetry run scripts/generate_semver.py)
```

rissson (changes requested):

pyproject.toml
```diff
@@ -4,6 +4,26 @@ version = "2024.12.2"
...
+[tool.codespell]
+skip = [
+    "**/node_modules",
+    "**/package-lock.json",
+    "schema.yml",
+    "./blueprints/schema.json",
+    "go.sum",
+    "cmd",
+    "internal",
+    "web/src",
```
rissson, line 19:
```suggestion
```

These are source code directories that we want to ignore as well

tanberry (commented):
I can't say one way or the other on the code, but thanks for catching the misspellings!!

website/docusaurus.config.ts
```diff
@@ -134,7 +134,7 @@ module.exports = async function (): Promise<Config> {
                     docsPluginId: "docs",
                     config: {
                         authentik: {
-                            specPath: "static/schema.yaml",
+                            specPath: "static/schema.yml",
```
rissson, line 137: We should add a redirect for this in the netlify.toml

rissson (changes requested):
One more little thing I didn't catch previously, mb

rissson (approved):

rissson (approved):

GirlBossRush:
@BeryJu Rebased to latest main and ready for review 😁

rissson (approved):

BeryJu (commented):
Need to verify if we need a redirect for schema.yml on the website repo
Need to check if we need to update any docs that directly run the `ak` executable to add a `poetry run` prefix

website/docusaurus.config.ts
```diff
@@ -152,7 +152,7 @@ const createConfig = (): Config => {
...
```
GirlBossRush, line 155: @BeryJu https://github.com/goauthentik/website/pull/216

BeryJu (approved):

PR #12704, opened 2025-01-16
providers/oauth2: support token revocation for public clients

### LINKED ISSUE #12053: Token revoke endpoint & public clients
**Describe the bug**
It's not possible to revoke a token via `/application/o/revoke/` without providing client_secret for public clients.

**To Reproduce**

1. Create an oauth2 provider with client type=public.
2. Authorize using the created provider.
3. Generate tokens.
4. Try revoking the access/refresh token via `/application/o/revoke/`.

**Expected behavior**
Public clients can revoke a token via `/application/o/revoke/` without providing client_secret.

**Screenshots**

**Logs**

-   authentik version: 2024.10.2

**Additional context**
The [RFC](https://datatracker.ietf.org/doc/html/rfc7009#section-5) states that a client's request must contain a valid client_id, in the case of a public client, or valid client credentials, in the case of a confidential client.

I believe the culprit is this function: https://github.com/goauthentik/authentik/blob/main/authentik/providers/oauth2/utils.py#L181.

rissson:
## Details

Closes #12053

From the RFC:

>    A malicious client may attempt to guess valid tokens on this endpoint
   by making revocation requests against potential token strings.
   According to this specification, a client's request must contain a
   valid client_id, in the case of a public client, or valid client
   credentials, in the case of a confidential client.  The token being
   revoked must also belong to the requesting client.  If an attacker is
   able to successfully guess a public client's client_id and one of
   their tokens, or a private client's credentials and one of their
   tokens, they could do much worse damage by using the token elsewhere
   than by revoking it.  If they chose to revoke the token, the
   legitimate client will lose its authorization grant and will need to
   prompt the user again.  No further damage is done and the guessed
   token is now worthless.

---

## Checklist

If applicable

### DIFF
--- a/authentik/providers/oauth2/views/token_revoke.py
+++ b/authentik/providers/oauth2/views/token_revoke.py
@@ -9,8 +9,12 @@ from django.views.decorators.csrf import csrf_exempt
 from structlog.stdlib import get_logger
 
 from authentik.providers.oauth2.errors import TokenRevocationError
-from authentik.providers.oauth2.models import AccessToken, OAuth2Provider, RefreshToken
-from authentik.providers.oauth2.utils import TokenResponse, authenticate_provider
+from authentik.providers.oauth2.models import AccessToken, ClientTypes, OAuth2Provider, RefreshToken
+from authentik.providers.oauth2.utils import (
+    TokenResponse,
+    authenticate_provider,
+    provider_from_request,
+)
 
 LOGGER = get_logger()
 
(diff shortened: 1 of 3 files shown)


authentik/providers/oauth2/views/token_revoke.py
```diff
@@ -27,7 +31,9 @@ def from_request(request: HttpRequest) -> "TokenRevocationParams":
         """Extract required Parameters from HTTP Request"""
         raw_token = request.POST.get("token")
 
-        provider = authenticate_provider(request)
+        provider, _, _ = provider_from_request(request)
+        if provider and provider.client_type == ClientTypes.CONFIDENTIAL:
+            provider = authenticate_provider(request)
```
rissson, line 36: We may want to further split authenticate_provider to avoid parsing the request multiple times.

BeryJu (approved):

PR #12804, opened 2025-01-24
lifecycle: much improved debugging experience

BeryJu:
## Details

REPLACE ME

---

## Checklist

If applicable

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -2,6 +2,7 @@
     "recommendations": [
         "bashmish.es6-string-css",
         "bpruitt-goddard.mermaid-markdown-syntax-highlighting",
+        "charliermarsh.ruff",
         "dbaeumer.vscode-eslint",
         "EditorConfig.EditorConfig",
         "esbenp.prettier-vscode",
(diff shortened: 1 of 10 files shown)


sirux88:
@BeryJu:
Is there a specific reason for the changes the last commit (6b48ee5).
If you keep the dot instead of the slash local debugging (in current working directory) and debugging in container should work (on the container working directory is '/')

BeryJu:
> @BeryJu: Is there a specific reason for the changes the last commit ([6b48ee5](https://github.com/goauthentik/authentik/commit/6b48ee5eabb0c45aedaee78ab92a9b0bd77f233a)). If you keep the dot instead of the slash local debugging (in current working directory) and debugging in container should work (on the container working directory is '/')

I did not work for me when testing this with a container running in compose, however this is a point that will be revisited in the future anyways as we want to move the root directory in the container to no longer copy a bunch of stuff directly into `/`

sirux88:
> > @BeryJu: Is there a specific reason for the changes the last commit ([6b48ee5](https://github.com/goauthentik/authentik/commit/6b48ee5eabb0c45aedaee78ab92a9b0bd77f233a)). If you keep the dot instead of the slash local debugging (in current working directory) and debugging in container should work (on the container working directory is '/')
> 
> I did not work for me when testing this with a container running in compose, however this is a point that will be revisited in the future anyways as we want to move the root directory in the container to no longer copy a bunch of stuff directly into `/`

Hm strange this doesn't make sense with the docker files current state. I'll investigate your issue with the docker container and see if I can figure out what the problem is. If I can fix it I'll open a new PR and reference it here resp. there. 

This needs to be done anyway if you want to move the files (somewhen) from `/`

PR #12806, opened 2025-01-24
lifecycle: better pre release test

BeryJu:
## Details

Add script to run pre-release docker tests locally
also fixes a bug that tries to install wheels that no longer exist

---

## Checklist

If applicable

### DIFF
--- a/lifecycle/ak
+++ b/lifecycle/ak
@@ -88,7 +88,6 @@ elif [[ "$1" == "bash" ]]; then
 elif [[ "$1" == "test-all" ]]; then
     prepare_debug
     chmod 777 /root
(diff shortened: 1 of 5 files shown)


lifecycle/ak
```diff
@@ -1,4 +1,5 @@
-#!/usr/bin/env -S bash -e
+#!/usr/bin/env -S bash
```
rissson, line 1: -S is no longer required here

scripts/test_docker.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
```
rissson, line 1: `#!/usr/bin/env bash` :eyes:

PR #12859, opened 2025-01-28
providers/oauth2: cleanup tokens when user is deactivated

### LINKED ISSUE #12858: user refresh tokens still active after user deactivation.
**Describe the bug**

When disabling a user in Authentik, the refresh tokens associated with the users are not deactivated. 
It also seems like the user is able to refresh its token. 

**To Reproduce**

Create a user -> create refresh tokens -> deactivate the user -> Try to use the token. 

**Expected behavior**

I expect all tokens/permissions associated with the user to be deleted/rewoked. 

**Screenshots**

**Logs**

[2024.10.4]
helm (https://charts.goauthentik.io)

More information: 

I "solved" this by deleting the user and lowering the refresh token validity period.
Only after user deletion, the user cannot refresh its token after it expired.

rissson:
## Details

Closes #12858

---

## Checklist

If applicable

### DIFF
--- a/authentik/core/api/users.py
+++ b/authentik/core/api/users.py
@@ -78,13 +78,13 @@ from authentik.core.models import (
     User,
     UserTypes,
 )
+from authentik.core.signals import user_deactivated
 from authentik.events.models import Event, EventAction
 from authentik.flows.exceptions import FlowNonApplicableException
 from authentik.flows.models import FlowToken
 from authentik.flows.planner import PLAN_CONTEXT_PENDING_USER, FlowPlanner
 from authentik.flows.views.executor import QS_KEY_TOKEN
 from authentik.lib.avatars import get_avatar
-from authentik.providers.oauth2.models import DeviceToken, RefreshToken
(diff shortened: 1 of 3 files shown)


authentik/core/api/users.py
```diff
@@ -765,6 +766,8 @@ def partial_update(self, request: Request, *args, **kwargs) -> Response:
         response = super().partial_update(request, *args, **kwargs)
         instance: User = self.get_object()
         if not instance.is_active:
+            RefreshToken.objects.filter(session__user=instance).delete()
```
BeryJu, line 769: This should be in a signal in the oauth2 provider app, as a user might get deactivated through other means (and we shouldn't introduce new direct app dependencies if possible)

BeryJu (changes requested):

authentik/core/api/users.py
rissson: any other places where this signal should be sent from?

BeryJu (approved):

PR #12893, opened 2025-01-30
website: Use Docusaurus Frontmatter for badges

GirlBossRush:
## Details

This PR is composed of several small fixes that focus on long-term goal to reduce boilerplate that appears in doc pages, such as...

- Version badges indicating the minimum release needed to use certain features
- Enterprise badges indicating a feature is limited to a specific plan
- Preview badges indicating that a feature is still in development

Also included are a few fragile anchor links, missing syntax highlights on code blocks, and typo fixes.

---

## Checklist

-   [ ] Local tests pass (`ak test authentik/`)
-   [ ] The code has been formatted (`make lint-fix`)

If an API change has been made

-   [ ] The API schema has been updated (`make gen-build`)

If changes to the frontend have been made

-   [ ] The code has been formatted (`make web`)

If applicable

-   [x] The documentation has been updated
-   [x] The documentation has been formatted (`make website`)

PR #12897, opened 2025-01-30
website: remove images from integrations index page

dominic-r:
## What?

Removes the images from the integrations index page.

## Details

REPLACE ME

---

## Checklist

If applicable


BeryJu (commented):
I love the idea for this (I think I experimented with something similar some time ago), especially with https://github.com/goauthentik/authentik/pull/12893 moving the support badge to the frontmatter making the first line in the card be an excerpt of the service.

One thing we should do in the future is add icons to these integrations, however we'd a) need to backport that for all existing services and b) figure out a better place to store them (I'm not a fan of putting them in this repo too) 

LGTM from me, cc @tanberry for adding some copy on top of the integrations page and @kensternberg-authentik / @GirlBossRush for reviewing the react components

tanberry:
> Removed component following a discussion with Tana. Sorry Jens... cc @tanberry It should be ready now

Just for now, so we can get through the release. Then let's pick up combo about using the react components...

Or if you want to do it all in one PR, then can the Deploy preview get fixed, so I can see how it will look?

tanberry (approved):
We can merge this one now, and then do the react components in a diff one?

PR #12914, opened 2025-02-01
sources: prevent deletion of built-in source

dominic-r:
sdko/sources/no-delete-built-in

## What?

This PR modifies `renderToolbarSelected` to disable the "Delete" button for the built-in authentik login source. While the source will reappear after a server restart, the change is intended as a precaution to prevent potential issues by blocking deletion. Furthermore, it adds a function to the `SourceViewSet` class to return an error when attmpting to delete the source using the API url.

## Testing:

* Attempted to delete the built-in source
* Added sources normally and attempted to remove them to ensure no functionality was lost
* If the built-in source is in a multi-select with a valid source, the deletion won't proceed

## Details

REPLACE ME

---

## Checklist

If applicable


authentik/core/api/sources.py
```diff
@@ -154,6 +155,17 @@ def user_settings(self, request: Request) -> Response:
             matching_sources.append(source_settings.validated_data)
         return Response(matching_sources)
 
+    def destroy(self, request: Request, *args, **kwargs):
+        """Prevent deletion of built-in sources"""
+        instance: Source = self.get_object()
+
+        if instance.managed == "goauthentik.io/sources/inbuilt":
```
rissson, line 162: This should be a constant somewhere :eyes: (It's also used in `authentik/core/apps.py`)

rissson: For the managed outpost, we define it in `authentik/outposts/apps.py`, but I quite like this approach of having it directly attached to the model where it makes sense to use it.

I'll let @BeryJu be the judge of whether we should continue this way to start putting stuff on models directly.

rissson (approved):

dominic-r:
@rissson / @BeryJu , just checking in to see if there's an update regarding the above ^

BeryJu (approved):

PR #12943, opened 2025-02-06
website/docs: sys mgmt: document authentik backups/restoration

### LINKED ISSUE #8411: docs: How to backup and restore Authentik
**Describe your question/**

How to create a back up of all the configuration of the Authentik, including username/authorization method, etc...?
How to restore this back up in the new setup Authentik?

I've read about blueprint but not sure where I can run the command ak export_blueprint.
I can see the export for the Flow but there is no export features for Stages or other areas...

-   authentik version: [e.g. 2023.10.7]
-   Deployment: docker-compose,

dominic-r:
## What?

This PR adds general documentation to backup authentik: the postgres database, the redis instance, and static directories. 

## Closes
Closes https://github.com/goauthentik/authentik/issues/8411


rissson:
> Depends on #12962 for the Docker container names.

You don't need to, we assume in the documentation that a bunch of commands are run from the docker-compose directory.

website/docs/sys-mgmt/backup.md
```diff
@@ -0,0 +1,82 @@
+---
+title: Backup authentik
+---
+
+This guide provides steps to back up authentik, including its PostgreSQL database, Redis data, and some important directories. These instructions cover both Kubernetes and Docker environments.
+
```
rissson, line 6: Let's add explanations about what is stored where, and why it is important to back it up. For instance, some users might not want to go through the trouble of backing up redis.

rissson (changes requested):

rissson:
One thing I forgot to mention, I don't want us to provide specific instructions for backing things up, but instead provide references to our dependencies respective documentation.

rissson:
> For example, linking the Postgres backup page instead of providing a command?

Exactly

> I'm curious, is there any specific reasoning behind this?

There is.

First, there is the added maintenance for us. Providing specific instructions would mean that we have to maintain those, possibly for multiple versions of postgres (for instance the helm chart and the docker-compose ones don't currently match). It also means that we would need to provide some level of support for issues and questions coming in. We currently don't have that kind of bandwidth.

Backing things up (whether it's postgres, redis, files, or whatever else) is never easy and straightforward. Providing commands would make it look like it is, but it isn't. There are many other things to consider, like the fact that we also need to backup users and their password so that on restore.

While on the topic of restoration, we would also need to provide additional instructions for that.

We also need to consider that using the provided postgres/redis setup with the docker-compose and helm chart is not the way everyone runs authentik.

I would much rather have explanations of what needs to be backed up, why it needs to be backed up, and solid pointers to resources explaining how to do it, instead of us trying to "make things up" even though it isn't our expertise.

> There’s a risk they could end up relying on commands from unreliable third-party documentation or their own improvised solutions

Not our problem. Seriously. I know it sounds harsh, but if anyone is running an application (authentik or else) and rely on that data without having 1. learned how to do backups and 2. proper setup and testing of said backups, they shouldn't be running that application.

Which is why I want us to point people in the right direction, with links to references of the upstream backup documentation for example.

website/docs/sys-mgmt/backup-restore/backup.md
```diff
@@ -0,0 +1,36 @@
...
+- **Backup Guidance:**
+    - Use PostgreSQL's native tools (e.g., [`pg_dump`](https://www.postgresql.org/docs/current/app-pgdump.html), [`pg_dumpall`](https://www.postgresql.org/docs/current/app-pg-dumpall.html), or [continuous archiving](https://www.postgresql.org/docs/current/continuous-archiving.html)).
+    - Exclude system databases (`postgres`, `template0`, `template1`).
+- **Official Documentation:** [PostgreSQL Backup and Restore](https://www.postgresql.org/docs/current/backup.html)
+
+## Redis database
+
+- **Role:** Manages temporary data:
+    - User sessions (lost data = users must reauthenticate).
+    - Reputation scores (saved hourly to PostgreSQL).
```
rissson, line 20:
```suggestion
```
This is no longer the case

website/docs/sys-mgmt/backup-restore/backup.md
```diff
@@ -0,0 +1,36 @@
...
+    - Use Redis' native tools (e.g., [`SAVE`](https://redis.io/commands/save) or [`BGSAVE`](https://redis.io/commands/bgsave)).
+- **Official Documentation:** [Redis Persistence](https://redis.io/docs/management/persistence/)
+
+## Static directories
+
+These directories are mounted as volumes in containerized installations:
+
+| Directory               | Purpose                                                         | Backup Notes                                             |
+| ----------------------- | --------------------------------------------------------------- | -------------------------------------------------------- |
+| **`/media`**            | Stores application icons, flow backgrounds, and uploaded files. | Only required if not using external storage (e.g., S3).  |
```
rissson, line 33: External storage should ideally be backed up as well.

website/docs/sys-mgmt/backup-restore/backup.md
```diff
@@ -0,0 +1,36 @@
...
+| **`/certs`**            | Custom TLS certificates (if provided externally).               | Backup if you manage certificates through authentik.     |
```
rissson, line 34: So certificates managed by authentik get always stored in the database. Thus, this directory doesn't strictly need to be backed up. I would say something like "Backup if you rely on those certificates existing on the file system. authentik doesn't need them to persist after they've been imported"

website/docs/sys-mgmt/backup-restore/restore.md
```diff
@@ -0,0 +1,33 @@
+---
+title: Restore authentik
```
rissson, line 2: I feel like the restore document is just a copy from the backup one. I would argue we can merge the two together.

website/sidebars.js
```diff
@@ -579,6 +579,15 @@ export default {
                         "sys-mgmt/ops/geoip",
                     ],
                 },
+                {
+                    type: "category",
```
rissson, line 583: This should definitely go under "System Management" > "Operations"

rissson (changes requested):
Overall a pretty good document. I very much like the way things are laid out and organized

rissson (approved):

website/docs/sys-mgmt/ops/backup-restore.md
```diff
@@ -0,0 +1,51 @@
+---
+title: Backup & Restore
```
tanberry, line 2: Can we make this title a bit more descriptive and SEO-rich? Maybe "Backup and restore your authentik instance" or some such?

dominic-r: see priv

website/docs/sys-mgmt/ops/backup-restore.md
```diff
@@ -0,0 +1,51 @@
...
+---
+
+This guide outlines the critical components to back up and restore in authentik.
```
tanberry, line 5: Ohhhh.. so it is only about backing up specific "components" (we might need a diff word), not the entire instance?

dominic-r: Well... If you backup the db, the cache, and static folders you technically backup everything. As you wish

website/docs/sys-mgmt/ops/backup-restore.md
dominic-r:
```suggestion
title: Backup and restore your authentik instance
sidebar_label: Backup & Restore
```

website/docs/sys-mgmt/ops/backup-restore.md
dominic-r: I suppose we could/should leave this specific wording to make it clear it's not a more in dept backup of say snapshotting the RAM or anything

tanberry (approved):
Thank you so much for adding this important bit of How To documentation, @dominic-r !!

website/docs/sys-mgmt/ops/backup-restore.md
```diff
@@ -0,0 +1,52 @@
...
+
+## PostgreSQL database
+
+### Backup
+
+- **Role:** Stores all persistent data (users, policies, configurations, etc.).
+- **Impact of Loss:** Complete data loss, requiring full restoration to recover functionality.
+- **Backup Guidance:**
+    - Use PostgreSQL's native tools (`pg_dump`, `pg_dumpall`, or [continuous archiving](https://www.postgresql.org/docs/current/continuous-archiving.html)).
```
dominic-r, line 15:
```suggestion
    - Use PostgreSQL's native tools ([`pg_dump`](https://www.postgresql.org/docs/current/app-pgdump.html), [`pg_dumpall`](https://www.postgresql.org/docs/current/app-pg-dumpall.html), or [continuous archiving](https://www.postgresql.org/docs/current/continuous-archiving.html)).
```

dominic-r:
Comme toujours, c'est un plaisir! @tanberry

septatrix:
Does losing the Redis data mean that pending emails are lost? For invite email that might be considered permanent data loss

rissson:
> Does losing the Redis data mean that pending emails are lost? For invite email that might be considered permanent data loss

That's correct. We're working on solutions to avoid those issues

septatrix:
> > Does losing the Redis data mean that pending emails are lost? For invite email that might be considered permanent data loss
> 
> That's correct. We're working on solutions to avoid those issues

Depending on far away these solutions are it would be best to correct the documentation to state that losing redis *does* mean losing irrecoverable state

PR #12950, opened 2025-02-06
*: remove outdated preview badges

gergosimonyi:


tanberry (approved):
I definitely approve but i think Ken will need to as well...?

BeryJu (approved):

PR #12809, opened 2025-01-25
website/docs: updated debugging docs

BeryJu:
## Details

docs for https://github.com/goauthentik/authentik/pull/12804

---

## Checklist

-   [ ] Local tests pass (`ak test authentik/`)
-   [ ] The code has been formatted (`make lint-fix`)

If an API change has been made

-   [ ] The API schema has been updated (`make gen-build`)

If changes to the frontend have been made

-   [ ] The code has been formatted (`make web`)

If applicable

-   [ ] The documentation has been updated
-   [ ] The documentation has been formatted (`make website`)

PR #12961, opened 2025-02-08
root: Improve debugging experience

sirux88:
## Details

This PR is ment to improve the debugging experiance:
1) It sets the vscode debugging python configurations to use relative paths instead of absolute paths. This reverts the last commit of #12804 
2) Since usage of `docker-compose.override.yml` is necessary due to #12809 it adds `docker-compose.override.yml` to gitignore to remove noice when comitting

Further thoughts:
1) Remove `Debug: Attach Worker` configuration since it only duplicates `Debug: Attach Server Core` in its curretns state. Renaming `Debug: Attach Server Core` would be useful for clarification then
2) If debugging `worker` and `server` should be able at the same time it would be necessary distinguish between both when setting up the debugging ports on startup or via port forwarding in `docker-compose.override.yml`.
Both needs to be documented somewhere
3) The [dev-server](https://github.com/goauthentik/authentik/pkgs/container/dev-server) containers don't include the dev-dependencies. They need to be installed in runtime but the container needs root rights to do that. Might be ok for developing but  the dependencies could be included aswell by making changes to `Dockerfile` (Sidemark: Image size will increase ~30%)

---

## Checklist

If applicable


sirux88:
@BeryJu:
As mentioned in #12804 I reviewed what you stated there. I could not reproduce your issue. After properly setting up everything (pulling `dev-server` container, creating `docker-compose.override.yml`, creating `.env` file and spinning up docker compose) I could debug the python part of the application. 

For testing purpose I added some minor python changes. After that the breaking points around the changes didn't get hit anymore (which makes sense since the changes aren't visible in the container itself).
I could resolve this by adding:
```yml
volumes:
          - ./authentik:/authentik
```
to `docker-compose.override.yml`

Maybe you could reevaluate your problem with this PR/comment

PR #12962, opened 2025-02-08
docker: provide default container names

dominic-r:
## What?

This PR adds default container names to the Postgres database, Redis, server, and worker.

The majority of Docker Compose files (for authentik and other services out there in the internet) already hard-code container names, so this change aligns with common best practices.  

By setting explicit names, we allow users to reference containers directly with commands like:  
```sh
docker exec -it authentik-worker ak
```
instead of relying on `docker compose [...]`, which fails with  
```
no configuration file provided: not found
```
if the user is not in the directory containing the Compose file.  
  
This improves the experience of the user (and mine too as I need to reference these containers in some doc pages) when following documentation steps and makes the output of `docker ps` less of a mess.

## Details

REPLACE ME

---

## Checklist

If applicable

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -3,6 +3,7 @@
 services:
   postgresql:
     image: docker.io/library/postgres:16-alpine
+    container_name: authentik-postgresql
     restart: unless-stopped
     healthcheck:
       test: ["CMD-SHELL", "pg_isready -d $${POSTGRES_DB} -U $${POSTGRES_USER}"]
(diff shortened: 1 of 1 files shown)


rissson:
While I'm not necessarily against this change, this might (?) be a breaking change. I'm not sure overriding the way docker-compose manages container names is something we want to do. What were the specific reasons for this change?

rissson:
> Won't really be a breaking change. Users will rebuild containers and use the --remove-orphans option to remove old containers

That's exactly where the break happens. For one thing, users with automated deploy processes (even though I highly doubt they just download the docker-compose in that case) might not remove orphan containers by default. Also, we don't include that option in our upgrading docs.

rissson:
Yeah I don't think we need/want this change for now

PR #13167, opened 2025-02-21
website/docs: Add Passkeys reference where WebAuthn is mentioned

melizeche:
## Details

Add Passkeys reference in several parts where WebAuthn is mentioned for better docs UX and SEO)

Rationale:

Passkeys is the de-facto term when people think about passwordless login and biometric 2fa and there's nothing on the website or docs that mention that term

image

image

---

## Checklist

If applicable


dominic-r:
Technically, WebAuthn != Passkey. If i'm not mistaken, WebAuthn is the web standard for passwordless authentication and passkeys are a type of credential which implement that standard. If we want to ignore technicalities, LGTM.

Here's a source which confirms it: https://www.corbado.com/faq/how-does-webauthn-differ-from-passkeys

melizeche:
@dominic-r Yup, I agree they are not the same thing, but is the term that most users are familiar with (at least more than WebAuthn). As I said, this is more about UX and SEO because if a user doesn't know that WebAuthn is the protocol, they might think that authentik doesn't support passkeys

but because this is a docs issue I'd like to get @tanberry approval before merging

dominic-r:
Just throwing this out there, Teffen added tags to the Google docs: https://github.com/goauthentik/authentik/blob/main/website/docs/users-sources/sources/social-logins/google/workspace/index.md?plain=1#L4 . Does that show up in search? If so, that could be a decent idea as well? I agree about SEO and UX and yea let's get tana's opinion.

tanberry:
Oooof this is  a tough one to phrase. Words are indeed hard.

We agree they are not exactly the same thing. Yet we agree it is good for SEO and UX to be clear that we support Passkeys. Having a WebAuthn-based device means that we do support the use of Passkeys. But we need to be careful not to imply that they are the same thing... and using phrases like "WebAuthn/Passkey" or "WebAuthn (Passkeys)" does equalize them.

Maybe this type of phrasing:

> Passwordless authentication currently only supports WebAuthn-based devices (which provides for the use of passkeys), like security keys and biometrics. 

> ... WebAuthn (whch enables passkey-based authentication)...

website/docs/add-secure-apps/flows-stages/stages/authenticator_webauthn/index.mdx
```diff
@@ -1,7 +1,9 @@
 ---
-title: WebAuthn authenticator setup stage
+title: WebAuthn / Passkeys authenticator setup stage
```
melizeche, line 2: @tanberry I’ve updated the other references to prevent confusion, but I believe we should include the term “passkey” in the title of the authenticator documentation since that’s what will appear in a Google search

Do you think that we can leave it this way or do you have any suggestions?

tanberry (approved):
Thanks for doing this, and for taking the time to phrase it just right.

PR #13272, opened 2025-02-26
outposts: add support for gateway API

### LINKED ISSUE #14070: FR: HTTPRoute support for Kubernetes integration
I'm frustrated when things I want to use, don't support HTTPRoute.

I would like to be able to use the included Kubernetes support to create proxied HTTPRoutes where it goes though Authentik proxy and then to app.

PR it myself, but now entirely sure how it should be approached.

**Additional context**

The idea is basically similar functionality as is currently possible with the included Kubernetes support for creating Ingresses automatically and then proxying the app though after auth and all.

rissson:
## Details

Closes #14070

---

## Checklist

If applicable


rissson:
Tested and is working fine. Integration tests not really possible due to using custom CRDs.

authentik/providers/proxy/controllers/k8s/httproute.py
```diff
@@ -0,0 +1,234 @@
...
+from authentik.outposts.controllers.k8s.triggers import NeedsUpdate
+from authentik.outposts.controllers.kubernetes import KubernetesController
+from authentik.providers.proxy.models import ProxyMode, ProxyProvider
+
+if TYPE_CHECKING:
+    from authentik.outposts.controllers.kubernetes import KubernetesController
+
+
+@dataclass(slots=True)
+class RouteBackendRef:
```
BeryJu, line 19: Assuming none of these have classes in the Kubernetes API client?

BeryJu (approved):
Does this need some logic so only either this or ingress runs?

rissson:
things tested manually:
- does not change an existing outpost that has a k8s integration (i.e. no additional resources are created without user intervention)
- once parentRefs are set in the configuration (which is the manual change mentioned above), the httproute is indeed created on the cluster with the correct configuration (tested as well)
- adding/removing providers from the outpost correctly changes the httproute
- deleting the outpost deletes the httproute

authentik/providers/proxy/controllers/k8s/httproute.py
rissson: They unfortunately don't, or I would've loved to use them. In the future, they'll probably get added and we can use them at that point

rissson:
> Does this need some logic so only either this or ingress runs?

No, because this resource needs `kubernetes_httproute_parent_refs` to be set to a non-empty list for the httproute to be created. At which point, we can specify in the documentation (upcoming PR) that when using HTTPRoute, `ingress` should be added to `kubernetes_disabled_components`

PR #13332, opened 2025-03-01
core, web: translations - improve es

jalonsoa:
Improve ES translations


rissson:
Thank you for your contribution. Translations are handled through transifex, please see https://docs.goauthentik.io/docs/developer-docs/translation for instructions

PR #13394, opened 2025-03-04
stages/email: token_expiry format

melizeche:
## Changes

- Changed `token_expiry` field from `IntegerField` to `TextField` with string format
- Added data migration to convert existing integer values to `"minutes={value}"`
- Updated UI form to accept text input with format explanation
- Modified token generation logic to parse the string format
---

## Checklist

If applicable


authentik/stages/email/migrations/0005_alter_emailstage_token_expiry.py
```diff
@@ -0,0 +1,52 @@
...
+import authentik.lib.utils.time
+from authentik.lib.utils.time import timedelta_from_string
+from django.db import migrations, models
+
+
+def convert_integer_to_string_format(apps, schema_editor):
+    EmailStage = apps.get_model("authentik_stages_email", "EmailStage")
+    for stage in EmailStage.objects.all():
+        stage.token_expiry = f"minutes={stage.token_expiry}"
+        stage.save()
```
rissson, line 12:
```suggestion
    db_alias = schema_editor.connection.alias
    EmailStage = apps.get_model("authentik_stages_email", "EmailStage")
    for stage in EmailStage.objects.using(db_alias).all():
        stage.token_expiry = f"minutes={stage.token_expiry}"
        stage.save(using=db_alias)
```

rissson: Also, would it be possible to do something like `EmailStage.objects.using(db_alias).update(token_expiry="minutes="+F("token_expiry"))`? (It might need a Cast in there somewhere)

authentik/stages/email/migrations/0005_alter_emailstage_token_expiry.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+def convert_string_to_integer_format(apps, schema_editor):
+    EmailStage = apps.get_model("authentik_stages_email", "EmailStage")
+    for stage in EmailStage.objects.all():
```
rissson, line 17: Again, you need to `.using()`

authentik/stages/email/migrations/0005_alter_emailstage_token_expiry.py
```diff
@@ -0,0 +1,52 @@
...
+            name="token_expiry",
+            field=models.TextField(
+                default="minutes=30",
+                help_text="Time the token sent is valid (Format: hours=3,minutes=17,seconds=300).",
+                validators=[authentik.lib.utils.time.timedelta_string_validator],
+            ),
+        ),
+        migrations.RunPython(
+            convert_integer_to_string_format,
+            convert_string_to_integer_format,
```
rissson, line 50: We don't really provide backwards migrations. No need to bother with them.

authentik/stages/email/migrations/0005_alter_emailstage_token_expiry.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("authentik_stages_email", "0004_emailstage_activate_user_on_success"),
+    ]
+
+    operations = [
+        migrations.AlterField(
```
rissson, line 39: I wonder if it's better to add a new field, do the data migration, and then remove the old one.

web/src/admin/stages/email/EmailStageForm.ts
```diff
@@ -202,18 +202,20 @@ export class EmailStageForm extends BaseStageForm<EmailStage> {
...
-                        label=${msg("Token expiry")}
+                        label=${msg("Token expiration")}
                         ?required=${true}
                         name="tokenExpiry"
                     >
                         <input
-                            type="number"
-                            value="${first(this.instance?.tokenExpiry, 30)}"
+                            type="text"
+                            value="${first(this.instance?.tokenExpiry, "minutes=15")}"
```
rissson, line 211:
```suggestion
                            value="${first(this.instance?.tokenExpiry, "minutes=30")}"
```
To be consistent with the backend default

web/src/admin/stages/email/EmailStageForm.ts
```diff
@@ -202,18 +202,20 @@ export class EmailStageForm extends BaseStageForm<EmailStage> {
...
                             class="pf-c-form-control"
                             required
                         />
                         <p class="pf-c-form__helper-text">
-                            ${msg("Time in minutes the token sent is valid.")}
+                            ${msg(
+                                "Time the token sent is valid (Format: hours=3,minutes=17,seconds=300).",
```
rissson, line 217: I'm pretty sure we have a helper somewhere you can import to avoid having to rewrite the `(Format: hours=...)` thing

rissson (changes requested):

authentik/stages/email/migrations/0005_alter_emailstage_token_expiry.py
melizeche: for testing is good and doesn't hurt, I tested a lot and it works fine

authentik/stages/email/migrations/0005_alter_emailstage_token_expiry.py
melizeche: in this case it doesn't have an issue because int is converted to str without problems and the field cannot be null

web/src/admin/stages/email/EmailStageForm.ts
BeryJu: Indeed we have `<ak-utils-time-delta-help>`

authentik/stages/email/stage.py
```diff
@@ -74,7 +75,7 @@ def get_token(self) -> FlowToken:
         pending_user = self.get_pending_user()
         current_stage: EmailStage = self.executor.current_stage
         valid_delta = timedelta(
```
BeryJu, line 77:
```suggestion
        valid_delta = timedelta_from_string(current_stage.token_expiry) + timedelta(minutes=1)  # + 1 because django timesince always rounds down
```

I think thats a bit easier to read

BeryJu (approved):

rissson (approved):

Issue #13316, opened 2025-02-28
User interface displaying outdated cached data


ThowZzy:
**Describe the bug**
The icons, name etc of the apps in the user interface are not updating after changing something.

**To Reproduce**

1. Create an app
2. Change icon or something else
3. Reload the user interface page
4. Not updated

**Expected behavior**
Always get the latest version of an app in user interface

**Screenshots**

![Image](https://github.com/user-attachments/assets/4dd1931a-ab00-4ea3-8a8e-f080557eb12b)

![Image](https://github.com/user-attachments/assets/90281361-c2b4-4345-86a9-ad702dde0014)

**Logs**
Output of docker-compose logs or kubectl logs respectively

**Version and Deployment (please complete the following information):**

-   authentik version: 2025.2.1
-   Deployment:docker-compose

**Additional context**
It might be related to the language or something. The only way to update these is to log out and log back in, OR change the language of the user interface. 
And by the way, my browser is FRENCH, I selected the user interface to ENGLISH, but when i refresh it keeps reverting to french, and i need to manually go in the settings in "Update details" and hit "save" which fixes everything.

rissson:
Those are cached. You can clear that cache from the admin interface Flows > Flows and click clear cache

ThowZzy:
> Those are cached. You can clear that cache from the admin interface Flows > Flows and click clear cache

I saw you mentioning this solution somewhere else, and I forgot to mention that this doesn't fix it for me.

rattencreep:
Having users log out and back in again fixes this for me.

kensternberg-authentik:
The "List of Applications a User can Access" is managed as a _Policy_, not a Flow.  Try going to "Customization -> Policies" and clicking **Clear Cache** there.  I've checked and that does work.

ThowZzy:
> The "List of Applications a User can Access" is managed as a _Policy_, not a Flow. Try going to "Customization -> Policies" and clicking **Clear Cache** there. I've checked and that does work.

Thanks. This worked.

I would still consider this behavior a bug, as I don't see a reason to keep outdated cached data of the apps and displaying it to all users after editing them. The most annoying part is that I would need to clear this cache everytime I assign a user to a group that gives access to specific apps. The newly available apps also don't show for users unless I clear the cache. Maybe this cache could be automatically cleared everytime an admin modifies an app or gives access to an app?

PR #13399, opened 2025-03-05
web/admin: add button to clear application cache

rissson:
## Details

REPLACE ME

---

## Checklist

If applicable

### DIFF
--- a/web/src/admin/applications/ApplicationListPage.ts
+++ b/web/src/admin/applications/ApplicationListPage.ts
@@ -20,7 +20,7 @@ import { ifDefined } from "lit/directives/if-defined.js";
 
 import PFCard from "@patternfly/patternfly/components/Card/card.css";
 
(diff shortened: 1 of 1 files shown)


melizeche (approved):

kensternberg-authentik (approved):
This feels very much like a work-around.  A more comprehensive answer would be to give the User Applications cache its own location (not under `goauthentik.io/policies` in the cache), and clear it automagically whenever there's a change. The bug OP is [right about the extra work we make admins do here](https://github.com/goauthentik/authentik/issues/13316#issuecomment-2701384724).

rissson:
As discussed, we are agreed. This is meant as an easy way for admin to clear that cache without having to go to another page that isn't obvious to find. Further work is required to split the caches and only clear the necessary ones when that button is pushed (along with validating the necessary permissions).

Issue #12653, opened 2025-01-13
Hide disabled users when adding users to an entitlement


erlandan:
We have a high number of disabled users. That makes the the lists for adding users to groups or entitlements  unmanageable long. It would make it a lot more user friendly to hide disabled users as default.

In addition, for entitlement we would  like the  possibility to add more than just one user or group each time. 

Version: 2024.12.2

kensternberg-authentik:
@erlandan The PR that addresses this only addresses the "adding users to groups," as that was fairly straightforward; the second feature would take a significant architectural change we're not quite ready to make at this time. As for the third, if you have a group of people you want to apply to an entitlement, is the Group mechanism not sufficient to your need (especially now that it can filter out inactive users)?

erlandan:
Hiding disabled users when adding users to group is sufficient us. We always add group(s) to an entitlement.

PR #13403, opened 2025-03-05
web/admin: allow user lists to show active only

kensternberg-authentik:
web/admin: allow admins to show only active users in Group assignments

## What

Adds a flag and a visible control to the “Add users to groups” dialog to limit the users shown to only those marked as “active.”

## Why

Requested, it was small, it made sense, and it was fairly trivial to implement. All the infrastructure already existed.

## Testing

- Ensure you have both “active” and “inactive” users in your sample group.
- Visit Groups -\> (One Group) -\> Users -\>. Click “Add existing user.” Click the `+` symbol.
- A new toggle control, “Show inactive users,” should now be visible.
- Click it and note whether or not the visible display corresponds to the stote of the control.

image

image

## Note

As a design decision, showing inactive users is now **off** by default.

## Note

This commit does not address the second half of the request, “… the ability to add more than one user to an entitlement.” We recommend that if you have a group of people who correspond to a given entitlement that you create a named group for them.

## Related Issue:

- [Hide disabled users when adding users to a group or entitlement \#12653](https://github.com/goauthentik/authentik/issues/12653)


web/src/admin/groups/MemberSelectModal.ts
```diff
@@ -24,12 +35,18 @@ export class MemberSelectTable extends TableModal<User> {
...
+    // The `showDisabledUsers` clause below is necessary because the back-end for searches is
+    // tri-state: `isActive: true` will only show active users, `isActive: false` will show only
+    // inactive users; only when it's _missing_ will you get all users.
     async apiEndpoint(): Promise<PaginatedResponse<User>> {
         return new CoreApi(DEFAULT_CONFIG).coreUsersList({
             ...(await this.defaultEndpointConfig()),
             includeGroups: false,
+            ...(this.showDisabledUsers ? {} : { isActive: true }),
         });
     }
```
GirlBossRush, line 51: I'm hesitant to speak confidently about this change because it's a part of the app I haven't yet explored. IMO your intuition is correct – The API's behavior needs clarification.   If the filter logic does operate with three states, it feels appropriate that the UI also expose the same level of control.

Nits:

- The positive (show) + negative (disabled) of `showDisabledUsers` makes cavewoman brain hurt.
- I'm starting to get your pain with the whole `new SomeRandomApi(DEFAULT_CONFIG).doThing({...}` 😅 

```suggestion
    visibleUsersFilter: "active" | "inactive" | "all" = "all";

    async apiEndpoint(): Promise<PaginatedResponse<User>> {
        /**
         *  Three-state to match the API's filter.
         * - `true`: Only active users
         * - `false`: Only inactive users
         * - `undefined`: All users
         */
        const isActive =
            this.visibleUsersFilter === "active"
                ? true
                : this.visibleUsersFilter === "inactive"
                  ? false
                  : undefined;

        return new CoreApi(DEFAULT_CONFIG).coreUsersList({
            ...(await this.defaultEndpointConfig()),
            includeGroups: false,
            isActive,
        });
    }
```

web/src/admin/groups/MemberSelectModal.ts
```diff
@@ -41,6 +58,36 @@ export class MemberSelectTable extends TableModal<User> {
...
+                                    <i class="fas fa-check" aria-hidden="true"></i>
+                                </span>
+                            </span>
+                            <span class="pf-c-switch__label">${msg("Show inactive users")}</span>
+                        </label>
+                    </div>
+                </div>
+            </div>`;
+    }
+
```
GirlBossRush, line 90: Echoing sentiments above, this may need more than a checkbox, possibly a drop-down select for "All", "Active", "Inactive"

```suggestion

    renderToolbarAfter() {
        const toggleInactiveUsersVisibility = () => {
            this.visibleUsersFilter = this.visibleUsersFilter === "all" ? "active" : "all";
            this.page = 1;
            this.fetch();
        };

        return html`&nbsp;
            <div class="pf-c-toolbar__group pf-m-filter-group">
                <div class="pf-c-toolbar__item pf-m-search-filter">
                    <div class="pf-c-input-group show-disabled-toggle-group">
                        <label class="pf-c-switch">
                            <input
                                class="pf-c-switch__input"
                                type="checkbox"
                                ?checked=${this.visibleUsersFilter === "all"}
                                @change=${toggleInactiveUsersVisibility}
                            />
                            <span class="pf-c-switch__toggle">
                                <span class="pf-c-switch__toggle-icon">
                                    <i class="fas fa-check" aria-hidden="true"></i>
                                </span>
                            </span>
                            <span class="pf-c-switch__label">${msg("Show inactive users")}</span>
                        </label>
                    </div>
                </div>
            </div>`;
    }
    
```

web/src/admin/groups/MemberSelectModal.ts
```diff
@@ -7,13 +7,24 @@ import { TableColumn } from "@goauthentik/elements/table/Table";
...
 import { CoreApi, User } from "@goauthentik/api";
 
 @customElement("ak-group-member-select-table")
 export class MemberSelectTable extends TableModal<User> {
+    static get styles() {
+        return [
+            ...super.styles,
+            css`
+                .show-disabled-toggle-group {
+                    margin-left: 0.5rem;
```
GirlBossRush, line 22: Nit: Makes mobile overrides easier to adjust but this is super minor.
```suggestion
                    margin-inline-start: 0.5rem;
```

GirlBossRush (changes requested):
This is definitely an improvement. Happy to approve if we want to get a quick fix, but I've added some notes on the stateful nature of the UI.

BeryJu (approved):

web/src/admin/groups/MemberSelectModal.ts
kensternberg-authentik:
```
type UserListFilter = "active" | "all" | "inactive";
type UserFilter = Partial<Pick<CoreUsersListRequest, "isActive">>;

// ...

        const userFilter: UserFilter = match(this.userListFilter)
            .with("all", () => ({}))
            .with("active", () => ({ isActive: true }))
            .with("inactive", () => ({ isActive: false }))
            .exhaustive();

        return new CoreApi(DEFAULT_CONFIG).coreUsersList({
            ...(await this.defaultEndpointConfig()),
            ...userFilter,
            includeGroups: false,
        });
```

😈

GirlBossRush (approved):

PR #13453, opened 2025-03-10
enterprise/policies: Add Password Uniqueness History Policy

### LINKED ISSUE #8307: Checking the new user password to make sure it is different from the previous ones

melizeche:
## Details

Continue work done in #10631
Closes https://github.com/goauthentik/authentik/issues/8307
Documentation PR: #13686

Steps to test:
1. Create and setup a new "Password Uniqueness Policy"
2. Edit default-password-change flow->stage bindings-> edit stage -> default-password-change-prompt and add the policy
image

---

## Checklist

If applicable


authentik/policies/unique_password/models.py
```diff
@@ -0,0 +1,136 @@
...
+
+    password_field = models.TextField(
+        default="password",
+        help_text=_("Field key to check, field keys defined in Prompt stages are available."),
+    )
+
+    # Limit on the number of previous passwords the policy evaluates
+    # Also controls number of old passwords the system stores.
+    num_historical_passwords = models.PositiveIntegerField(
+        default=0,
```
rissson, line 30: What does 0 mean in that case? Also it defaults to 1 in the frontend. If 0 is not valid, we should error on this.

authentik/policies/unique_password/models.py
```diff
@@ -0,0 +1,136 @@
...
+    user = models.ForeignKey(User, on_delete=models.CASCADE, related_name="old_passwords")
+    # Mimic's column type of AbstractBaseUser.password
+    old_password = models.CharField(max_length=128)
+    created_at = models.DateTimeField(auto_now_add=True)
+
+    class Meta:
+        verbose_name = _("User Password History")
+
+    def __str__(self) -> str:
+        if self.created_at:
```
rissson, line 131: This should never be null.

authentik/policies/unique_password/signals.py
```diff
@@ -0,0 +1,49 @@
...
+from authentik.policies.unique_password.tasks import (
+    purge_password_history_table,
+    trim_user_password_history,
+)
+from authentik.stages.user_write.signals import user_write
+
+
+@receiver(post_delete, sender=PolicyBinding)
+def purge_password_history(sender, instance, **_):
+    from authentik.policies.unique_password.models import UniquePasswordPolicy
```
rissson, line 21: Do we have to have this as a JIT import?

authentik/policies/unique_password/signals.py
```diff
@@ -0,0 +1,49 @@
...
+
+    if not isinstance(instance.policy, UniquePasswordPolicy):
+        return
+    purge_password_history_table.delay()
+
+
+@receiver(user_write)
```
rissson, line 28: There is a `password_changed` signal you can listen on. It's sent from `authentik/core/models.py::User::set_password`

authentik/policies/unique_password/signals.py
```diff
@@ -0,0 +1,49 @@
...
+def copy_password_to_password_history(
+    sender, request: HttpRequest, user: User, data: dict[str, Any], **kwargs
+):
+    """Preserve the user's old password if UniquePasswordPolicy is enabled anywhere"""
+    from authentik.policies.unique_password.models import UniquePasswordPolicy, UserPasswordHistory
+
+    user_changed_own_password = (
+        any("password" in x for x in data.keys())
+        and request.user.pk == user.pk
+        and SESSION_KEY_IMPERSONATE_USER not in request.session
```
rissson, line 38: I don't think we should have a special case for impersonation here. In my mind, this is happening in a password change flow, and should run the same way on impersonation than on normal runs. If the admin wants to change a password for a user without going through all the checks, they can still do so from the admin interface directly.

authentik/policies/unique_password/signals.py
rissson: It will also allow you to ignore password changes in case `request.user != user`, in which case the user's password was changed by an admin from the admin interface, and should probably (?) be ignored

authentik/policies/unique_password/signals.py
```diff
@@ -0,0 +1,49 @@
...
+    if user_changed_own_password:
+        # Check if any UniquePasswordPolicy is in use
+        unique_pwd_policy_in_use = UniquePasswordPolicy.is_in_use()
+
+        if unique_pwd_policy_in_use:
+            """NOTE: Because we run this in a signal after saving the user,
+            we are not atomically guaranteed to save password history.
+            """
+            UserPasswordHistory.objects.create(user=user, old_password=user.password)
+            trim_user_password_history.delay(user.pk)
```
rissson, line 49: We probably want this to run on a schedule rather than triggering it here.

authentik/policies/unique_password/signals.py
```diff
@@ -0,0 +1,49 @@
...
+from authentik.stages.user_write.signals import user_write
+
+
+@receiver(post_delete, sender=PolicyBinding)
+def purge_password_history(sender, instance, **_):
+    from authentik.policies.unique_password.models import UniquePasswordPolicy
+
+    if not isinstance(instance.policy, UniquePasswordPolicy):
+        return
+    purge_password_history_table.delay()
```
rissson, line 25: We may (?) want this to run on a schedule rather than triggering it here.

authentik/policies/unique_password/tasks.py
```diff
@@ -0,0 +1,68 @@
...
+        return
+
+    try:
+        # n.b. a performance optimization to execute TRUNCATE
+        # instead of all().delete() would eliminate any FK checks.
+        UserPasswordHistory.objects.all().delete()
+    except Exception as err:
+        LOGGER.debug("Failed to purge UserPasswordHistory table.")
+        self.set_error(err)
+        return
```
rissson, line 33:
```suggestion
    UserPasswordHistory.objects.all().delete()
```

Tasks are not auto-retried unless specified. It's fine to not retry this, it doesn't really matter if there are some unused hashes until the next run, especially if this task runs on a schedule.

authentik/policies/unique_password/tasks.py
```diff
@@ -0,0 +1,68 @@
...
+            num_rows_to_preserve = max(
+                num_rows_to_preserve, binding.policy.num_historical_passwords
+            )
+
+    preservable_row_ids = (
+        UserPasswordHistory.objects.filter(user__pk=user_pk)
+        .order_by("-created_at")[:num_rows_to_preserve]
+        .values_list("id", flat=True)
+    )
+    num_deleted, _ = UserPasswordHistory.objects.exclude(pk__in=list(preservable_row_ids)).delete()
```
rissson, line 64: This can probably be a single query with .difference and avoiding the call to `list`

rissson (changes requested):

authentik/policies/unique_password/tasks.py
```diff
@@ -0,0 +1,58 @@
...
+    num_rows_to_preserve = 0
+    for binding in enabled_bindings:
+        if hasattr(binding.policy, "num_historical_passwords"):
+            num_rows_to_preserve = max(
+                num_rows_to_preserve, binding.policy.num_historical_passwords
+            )
+
+    entries = UserPasswordHistory.objects.filter(user__pk=user_pk)
+    to_keep = entries.order_by("-created_at")[:num_rows_to_preserve]
+    num_deleted, _ = entries.difference(to_keep).delete()
```
melizeche, line 54: @rissson  This is how I see to use `.difference()` here
Is this what you meant?

authentik/policies/unique_password/tasks.py
```diff
@@ -0,0 +1,95 @@
...
+    entries = UserPasswordHistory.objects.filter(user__pk=user_pk)
+    count = entries.count()
+
+    # Only delete if we have more entries than we need to preserve
+    if count > num_rows_to_preserve:
+        # Keep newest records, delete the rest
+        to_keep_ids = entries.order_by("-created_at")[:num_rows_to_preserve].values_list(
+            "id", flat=True
+        )
+        num_deleted, _ = entries.exclude(id__in=to_keep_ids).delete()
```
melizeche, line 61: @rissson I went with this approach in the end because I got `django.db.utils.NotSupportedError: Calling QuerySet.delete() after difference() is not supported.` when I try to use `.difference()`
What do you think?

authentik/policies/unique_password/models.py
```diff
@@ -0,0 +1,132 @@
...
+    # Mimic's column type of AbstractBaseUser.password
+    old_password = models.CharField(max_length=128)
+    created_at = models.DateTimeField(auto_now_add=True)
+
+    class Meta:
+        verbose_name = _("User Password History")
+
+    def __str__(self) -> str:
+        timestamp = f"{self.created_at:%Y/%m/%d %X}" if self.created_at else "N/A"
+        return f"Previous Password (user: {self.user_id}, recorded: {timestamp})"
```
melizeche, line 132: @rissson I had to add back the null check because of our model test runner
https://github.com/goauthentik/authentik/actions/runs/14046806875/job/39329299494?pr=13453

melizeche: Another approach would be to not put the timestamp in __str__

rissson: ah yeah I forgot about that. We can not put the timestamp as well yeah

authentik/policies/unique_password/tasks.py
rissson: Ah yeah, I see. Removing the `list` call should be enough then

authentik/enterprise/policies/unique_password/models.py
```diff
@@ -0,0 +1,151 @@
...
+
+    def __str__(self) -> str:
+        timestamp = f"{self.created_at:%Y/%m/%d %X}" if self.created_at else "N/A"
+        return f"Previous Password (user: {self.user_id}, recorded: {timestamp})"
+
+    @classmethod
+    def create_for_user(cls, user: User, password: str):
+        # To check users' passwords against Have I been Pwned, we need the first 5 chars
+        # of the password hashed with SHA1 without a salt...
+        pw_hash_sha1 = sha1(password.encode("utf-8")).hexdigest()  # nosec
```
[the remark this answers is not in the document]
BeryJu: It's almost as if there's a comment explaining why this code does what it does...

rissson (approved):

authentik/enterprise/policies/unique_password/api.py
```diff
@@ -0,0 +1,26 @@
+from rest_framework.viewsets import ModelViewSet
+
+from authentik.core.api.used_by import UsedByMixin
+from authentik.enterprise.policies.unique_password.models import UniquePasswordPolicy
+from authentik.policies.api.policies import PolicySerializer
+
+
+class UniquePasswordPolicySerializer(PolicySerializer):
```
BeryJu, line 8: missing `EnterpriseRequiredMixin`

authentik/enterprise/policies/unique_password/tasks.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+@CELERY_APP.task(bind=True, base=SystemTask)
+@prefill_task
+def purge_password_history_table(self: SystemTask):
+    """Remove all entries from the UserPasswordHistory table"""
+    unique_pwd_policy_bindings = PolicyBinding.in_use.for_policy(UniquePasswordPolicy)
+
+    if unique_pwd_policy_bindings.count() > 1:
+        # No-op; A UniquePasswordPolicy binding other than the one being deleted still exists
```
melizeche, line 21: note: not delete on no binding, should be on no UniquePasswordPolicy

authentik/enterprise/policies/unique_password/signals.py
```diff
@@ -0,0 +1,23 @@
...
+)
+
+
+@receiver(password_changed)
+def copy_password_to_password_history(sender, user: User, *args, **kwargs):
+    """Preserve the user's old password if UniquePasswordPolicy is enabled anywhere"""
+    # Check if any UniquePasswordPolicy is in use
+    unique_pwd_policy_in_use = UniquePasswordPolicy.is_in_use()
+
+    if unique_pwd_policy_in_use:
```
rissson, line 19: Should we also save as long as the policy exists?

authentik/enterprise/policies/unique_password/tasks.py
```diff
@@ -0,0 +1,98 @@
...
+            A Unique Password Policy instance still exists.""",
+        )
+        return
+
+    UserPasswordHistory.objects.all().delete()
+    self.set_status(TaskStatus.SUCCESSFUL, "Successfully purged UserPasswordHistory")
+
+
+@CELERY_APP.task()
+def trim_user_password_history(user_pk: int):
```
rissson, line 34: As this is only called from `trim_all_password_histories`, I think we can just have one task deleting for all users

rissson: And it can also call `check_and_purge_password_history` as well.

I fail to see the difference between `purge_password_history_table` and `check_and_purge_password_history` though

authentik/enterprise/policies/unique_password/signals.py
melizeche: I think that we should only save when the user explicitly wants to do that

authentik/enterprise/policies/unique_password/tasks.py
melizeche: You're right, I refactored  check_and_purge_password_history and purge_password_history_table into one function 👍

melizeche: as for `trim_all_password_histories` this approach allows us to have better error isolation(like if there's a problem with a particular user) and would scale better in huge user base

rissson: In other tasks we don't separate like this. I still want both of those tasks to be merged. If there are several thousands of users with password histories, I don't want several thousands of tasks to be scheduled.

rissson (changes requested):

rissson (approved):

GirlBossRush (approved):

PR #13506, opened 2025-03-13
website/docs: refactor pip to pipx on linux and improve styling

ImmanuelVonNeumann:
## Details

When setting up a local authentik-install on Ubuntu 24.04 using the documentation provided command ``pip install poetry poetry-plugin-shell`` will fail with error ``error: externally-managed-environment`` due to [PEP 668](https://peps.python.org/pep-0668/).

It is generally understood that pipx should be used in place of pip due to security concerns.
To keep the documentation in line with modern practices, I changed the documentation to use pipx in place of pip for linux.

Maybe the same should be done for the mac tab-item.
If so, please let me know.

---

## Checklist

If applicable

### DIFF
--- a/website/docs/developer-docs/setup/full-dev-environment.mdx
+++ b/website/docs/developer-docs/setup/full-dev-environment.mdx
@@ -58,7 +58,7 @@ values={[
     To install native dependencies on Debian or Ubuntu, run:
 
     ```sh
-    $ pip install poetry poetry-plugin-shell
+    $ pipx install poetry poetry-plugin-shell
     $ sudo apt-get install  libgss-dev krb5-config libkrb5-dev postgresql-server-dev-all
     $ sudo apt-get install postresql redis
     ```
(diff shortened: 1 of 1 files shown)


tanberry:
Hi @ImmanuelVonNeumann thank you for this PR... we have actually recently made a change to replace Poetry with [uv](https://docs.astral.sh/uv/). We do have a Docs PR to address this; your review and comments on it would be welcome. But I think we can close this PR of yours. Let me know if you have questions, and apologies for the delayed reply.

ImmanuelVonNeumann:
> Hi @ImmanuelVonNeumann thank you for this PR... we have actually recently made a change to replace Poetry with [uv](https://docs.astral.sh/uv/). We do have a Docs PR to address this; your review and comments on it would be welcome. But I think we can close this PR of yours. Let me know if you have questions, and apologies for the delayed reply.

Thanks for letting me know and no worries about the delay. 
I am closing the PR.

PR #13550, opened 2025-03-17
providers/scim: modify filtergroup(s) behavior to allow (multi-)group filtering

### LINKED ISSUE #6065: SCIM Group Filtering
I am using Authentik to provision users and groups into a Netskope lab environment with required property mappings.  While I can control which users are provisioned, it also sends over ALL groups associated to those users which flood the Netskope tenant with unnecessary groups unrelated to Netskope and other SCIM targets.  

I would like to see a section for the SCIM provider configuration called Group Filtering placed below User Filtering and above Attribute mapping.  This would allow an administrator to specify specific groups that would be provisioned.  It could be similiar to AzureAD SCIM or Okta's group rules.  

Limiting the groups imported into Authentik via the AD sync, but that would limit the instance of Authentik to only handling Netskope SCIM, decreasing the value for Authentik.

**Additional context**
None to remark.

ImmanuelVonNeumann:
## Details

This PR aims to change add group filtering to the SCIMProvider by incorporating the following changes:
- Rename and modify ``filter_group`` to ``filter_groups`` (including django migrations to keep a previously selected filter_group present)
- Multiple groups can be selected as ``filter_groups``
- Only groups within ``filter_groups`` are synced to the SCIM Server
- Only users which are members of at least one group within ``filter_groups`` are synced to the SCIM Server

Reason:
Most SCIM Clients allow groups to be filtered, however authentik currently does not.
This issue has been mentioned in #6065 which has been given the ``enhancement/confirmed``-label.

Among the many use-cases one would be a multi-tenant setup where groups of one tenant should not be visible to another tenant.

> [!NOTE]  
> Since this does change group-syncing behavior (i.e. not syncing previously synced groups anymore) I suggest this change to be implemented in a major release.
> Additionally, blueprints for scim_providers need to adapt to the new name ``filter_groups``.

Fixes #6065

---

## Checklist

If applicable


BeryJu:
Hi, thanks for the PR!

When I added the `filter_group` it was actually more so thought of as a "temporary" solution with the intended solution being that the SCIM provider would use the policies bound to the application the provider is a backchannel provider of to figure out which users to sync. However this would still leave room for either always syncing all groups or syncing all groups the synced users in.

There is also an already existing option for doing this more dynamically which is having a custom user/group mapping that dynamically raises `SkipObject`, which allows for more fine-grained rules for skipping certain objects

ImmanuelVonNeumann:
> Hi, thanks for the PR!
> 
> When I added the `filter_group` it was actually more so thought of as a "temporary" solution with the intended solution being that the SCIM provider would use the policies bound to the application the provider is a backchannel provider of to figure out which users to sync. However this would still leave room for either always syncing all groups or syncing all groups the synced users in.
> 
> There is also an already existing option for doing this more dynamically which is having a custom user/group mapping that dynamically raises `SkipObject`, which allows for more fine-grained rules for skipping certain objects

Thanks for your fast response, @BeryJu!

From my understanding the final solution could be that:
- Users who have access to the connected application are being synced
- Groups who are in filter_groups are being synced
- Users gain memberships for groups they are in if the group is also within filter_groups (since others are not synced)

Is this correct?

---

One further question: I could not find documentation on the implications of the field "filterset_fields" within the model.
Is there a documentation for it?

BeryJu:
> > Hi, thanks for the PR!
> > When I added the `filter_group` it was actually more so thought of as a "temporary" solution with the intended solution being that the SCIM provider would use the policies bound to the application the provider is a backchannel provider of to figure out which users to sync. However this would still leave room for either always syncing all groups or syncing all groups the synced users in.
> > There is also an already existing option for doing this more dynamically which is having a custom user/group mapping that dynamically raises `SkipObject`, which allows for more fine-grained rules for skipping certain objects
> 
> Thanks for your fast response, @BeryJu!
> 
> From my understanding the final solution could be that:
> 
> * Users who have access to the connected application are being synced
> * Groups who are in filter_groups are being synced
> * Users gain memberships for groups they are in if the group is also within filter_groups (since others are not synced)

Correct, I think having `SkipObject` in a property mapping as an advanced option makes sense but having a static list of groups to sync is a more accessible way. We can add the policy-based user-filtering at a later stage.

One potential issue that could arise when this is merged is what happens with groups that were previously provisioned and are now outside of the group filter, they would still exist in authentik and the remote system but would not get further updates (and currently would not be correctly deletable iirc)

> 
> Is this correct?
> 
> One further question: I could not find documentation on the implications of the field "filterset_fields" within the model. Is there a documentation for it?

`filterset_fields` comes from django-filters, see https://django-filter.readthedocs.io/en/latest/guide/rest_framework.html

ImmanuelVonNeumann:
> One potential issue that could arise when this is merged is what happens with groups that were previously provisioned and are now outside of the group filter, they would still exist in authentik and the remote system but would not get further updates (and currently would not be correctly deletable iirc)

I believe so, if the previously sent state is not currently being tracked.
Also group-membership would only be synced for filtered groups.

To solve this, the sent state would need to be tracked so that differences to it can be identified and remediated on the next sync.

> `filterset_fields` comes from django-filters, see https://django-filter.readthedocs.io/en/latest/guide/rest_framework.html

Thanks!

ImmanuelVonNeumann:
I believe the checks failed due to issues in uv, can we do a rerun of failed jobs?

rissson:
@ImmanuelVonNeumann you might need to `git merge main` for the issues to go away

tanberry (approved):
Thank you @ImmanuelVonNeumann for this PR (and for the detailed description, much appreciated!). From a Docs perspective, the only comment/question is whether we should tell users *how* to configure the group filters, but maybe that is outside of this PR's scope, and/or super-simple and doesn't need further explanation? Otherwise it looks great to me, and thanks for remembering to update the docs!

ImmanuelVonNeumann:
> From a Docs perspective, the only comment/question is whether we should tell users _how_ to configure the group filters, but maybe that is outside of this PR's scope, and/or super-simple and doesn't need further explanation?

I agree.
Additionally, I believe this sentence from the docs change might be confusing:

> Additionally, optional group filters can be configured to only synchronize the users that are members of the selected groups and the groups themselves.

I can work on an updated version where:
- Documentation about group filters is moved into a sub-header within "Filtering"
- That section contains an explanation about where group filters can be configured
- Impact of the selected group filters is better outlined

authentik/providers/scim/models.py
```diff
@@ -121,12 +124,16 @@ def get_object_qs(self, type: type[User | Group]) -> QuerySet[User | Group]:
...
             return base.order_by("pk")
         if type == Group:
-            # Get queryset of all groups with consistent ordering
-            return Group.objects.all().order_by("pk")
+            # Get a queryset of all roups with consistent ordering
+            # according to the provider's settings
+            base = Group.objects.all()
+            if self.filter_groups.exists():
+                base = base.filter(pk__in=self.filter_groups.all())
+            return base.order_by("pk")
```
rissson, line 136: Additional feedback from existing users of the feature, this is actually a breaking change, and would prevent all groups from being synced, even if filtering users.

ImmanuelVonNeumann: > Additional feedback from existing users of the feature, this is actually a breaking change, and would prevent all groups from being synced, even if filtering users.

I am not sure if I understand you correctly.

If you are saying that previously synced groups will not be synced when using a filter group, then that is indeed correct.

While the same users will be synced, with this change, the groups fill also be filtered by the selected filter groups.
However, from my understanding this is more in line with the filtering logic of most IdP solutions.

I have also noted this in the PR:
> Since this does change group-syncing behavior (i.e. not syncing previously synced groups anymore) I suggest this change to be implemented in a major release. Additionally, blueprints for scim_providers need to adapt to the new name `filter_groups`.

Some of the benefits I see here:
- Only groups that are designed for a specific scim sync are being evaluated (currently syncing can fail due to the scim-server. For example if the server expects a specific group attribute)
- Currently, groups cannot be isolated in the scim sync which can lead to information exposure when authentik is used with multi-tenancy (group from clientA will be synced to the scim-server of clientB)
- More in line with other IdP solutions (as noted above)

Alternatively, another attribute could be used for group-filtering.
However, I chose not to do this as it would be confusing to have an attribute named "filter group" and an additional attribute used to filter groups.

ImmanuelVonNeumann: @rissson I think I understand your point better now (please correct me if wrong).

As long as user-filtering is tied to filter groups there are use-cases (especially when groups shared among multiple "filter groups" are involved) which cannot be implemented with this solution.

Example:
There are two SCIM servers for it- and non-it personnel. However there is a group "first responder" which intersects both groups and should be available on both servers. 
The only solution is to add this group to both filter groups, which in turn now syncs all members of "first responder", (even non-it/it personnel) to both servers.

I think the best solution here would be as mentioned by @BeryJu 

> When I added the `filter_group` it was actually more so thought of as a "temporary" solution with the intended solution being that the SCIM provider would use the policies bound to the application the provider is a backchannel provider of to figure out which users to sync. However this would still leave room for either always syncing all groups or syncing all groups the synced users in.

Where the sync of users is determined by their permissions to "view" the application of the backchannel provider and sync of groups is determined by the filter groups attribute.

ImmanuelVonNeumann:
Closing this in favor of #13947

PR #13564, opened 2025-03-18
web/common: utils: fix infinite value handling in getRelativeTime function

### LINKED ISSUE #13562: User view page does not render
**Describe the bug**
Since upgrading to `2025.2.1`, I am unable to navigate and browse a user's view page. The page loads for any given user, but nothing is shown there, and when I look at the console, I can see that a call to `getRelativeTime` failed because a user's `passwordChangeDate` doesn't seem to meet the method's signature. _More about the exception below._

**To Reproduce**

1. Go to Directory > Users
2. Browse any user
3. Check browser's console

**Expected behavior**
The user's view page renders and I can see their details, password reset, history of changes, ...

**Screenshots**
A users view page showing no information and the exception in the browser's console.
Image

**Logs**
The browser's console call stack and exception.
```
Uncaught (in promise) RangeError: Value need to be finite number for Intl.RelativeTimeFormat.prototype.format()
    at RelativeTimeFormat.format (<anonymous>)
    at KT (utils.ts:176:24)
    at W.renderUserCard (UserViewPage.ts:162:31)
    at W.renderBody (UserViewPage.ts:389:32)
    at W.render (UserViewPage.ts:142:20)
```

**Version and Deployment:**

-   authentik version: 2025.2.2

dominic-r:
Closes #13562

## Details

REPLACE ME

---

## Checklist

If applicable


dominic-r:
Uhh I don't think all these run fails are caused by this change so I'll ignore them for now

GirlBossRush (approved):

web/src/common/utils.ts
```diff
@@ -157,23 +157,26 @@ export function adaptCSS(sheet: AdaptableStylesheet | AdaptableStylesheet[]): Ad
...
+        ["second", 1000],
+    ];
 
-    // "Math.abs" accounts for both "past" & "future" scenarios
     for (const [key, value] of _timeUnits) {
-        if (Math.abs(elapsed) > value || key == "second") {
-            return rtf.format(Math.round(elapsed / value), key);
+        if (Math.abs(elapsed) > value || key === "second") {
+            let rounded = Math.round(elapsed / value);
+            if (!isFinite(rounded)) {
```
kensternberg-authentik, line 176: "Because coercion inside the isNaN() function [can be surprising,](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isNaN#description:~:text=Because%20coercion%20inside%20the%20isNaN()%20function%20can%20be%20surprising%2C%20you%20may%20prefer%20to%20use%20Number.isNaN()) you may prefer to use Number.isNaN()."

kensternberg-authentik (approved):
Approved, but I made a suggestion.

web/src/common/utils.ts
dominic-r: My code doesn't seem to use isNaN at the moment. Sorry if I'd not get what you mean, but do you want me to add a check for NaN or replace part of my patch with a NaN check?

kensternberg-authentik: `isFinite()` uses `isNan()` under the covers, as the link documents.  Although the `isFinite()` page has the same quote:

"Because coercion inside the isFinite() function can be [surprising](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isNaN#description), you may prefer to use [Number.isFinite()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isFinite)."

PR #13655, opened 2025-03-24
root: fix dependency install due to description-file

BeryJu:
## Details

See https://github.com/vsoch/oci-python/pull/23
See https://github.com/pypa/setuptools/issues/4910

also had to force bump durationpy to 0.9 which has this fixed

---

## Checklist

-   [ ] Local tests pass (`ak test authentik/`)
-   [ ] The code has been formatted (`make lint-fix`)

If an API change has been made

-   [ ] The API schema has been updated (`make gen-build`)

If changes to the frontend have been made

-   [ ] The code has been formatted (`make web`)

If applicable

-   [ ] The documentation has been updated
-   [ ] The documentation has been formatted (`make website`)

PR #13656, opened 2025-03-24
infrastructure: use uv instead of poetry in test-migrations-from-stable

melizeche:
## Details

Fix broke CI 

---

## Checklist

If applicable


rissson (changes requested):
This is incorrect. This `poetry run` is run from the 2025.2.2 tag, which is still poetry, and thus correct. See #13538

rissson:
Ah, seems like #13655 has a fix for this as well

melizeche:
@rissson gotcha! Thanks!

PR #13761, opened 2025-04-03
website/docs: Updated redirect URI setup for Synology DSM

Wr0ngName:
Based on the feedback from Synology's developers, and testing: the redirect URI should not contain #/signup as it breaks authentication if multiple redirect URIs are defined.

In DSM 7.2's code itself, Host and HTTPS headers are used internally to match the corresponding entry in the list.

I've lost a few days on that, figured that could help some users. Also, I've read the guidelines and hope it suits your requirements.


dominic-r (approved):
Thanks!

tanberry (approved):
Thanks so much @Wr0ngName for this important fix!

PR #13781, opened 2025-04-05
outpost: proxy: handle nil HTTP response in attemptBasicAuth function

dominic-r:
Fixes a nil pointer dereference that occurs when an HTTP request fails in the attemptBasicAuth function. Added additional checks to safely handle cases where the HTTP response or its body is nil.

## Details

REPLACE ME

---

## Checklist

If applicable


internal/outpost/proxyv2/application/auth_basic.go
```diff
@@ -37,12 +37,23 @@ func (a *Application) attemptBasicAuth(username, password string) *Claims {
...
-			b = []byte(err.Error())
+	if res != nil && res.Body != nil {
+		defer func() {
+			if err := res.Body.Close(); err != nil {
+				a.log.WithError(err).Warning("failed to close response body")
+			}
+		}()
+	}
+	if err != nil || res == nil || res.StatusCode > 200 {
+		if res != nil && res.Body != nil {
```
BeryJu, line 48: These checks are way too convoluted, we should only need to check for err being nil and not for res being nil, and things can be re-ordered to require less duplicate if conditions

dominic-r: Is this better? https://github.com/goauthentik/authentik/pull/13781/commits/61a4dd4a9123fd394f0e413aa274418204083c6c

BeryJu (approved):

PR #13795, opened 2025-04-07
rbac: add `InitialPermissions`

### LINKED ISSUE #12711: Set default RBAC permissions for new objects
When running Authentik in "least privilege" scenarios, a user might be allowed a global permission to create objects, but without permission to change, or delete other objects. In this use-case the user will not be able to modify the object they created. 

The user should automatically be added to the "User Object Permissions" table with all permissions granted for objects created. 

N/A

**Additional context**
N/A

gergosimonyi:
This is a powerful construct between Permission and Role to set initial permissions for newly created objects.

Closes #12711


authentik/core/api/utils.py
```diff
@@ -29,6 +33,26 @@ def is_dict(value: Any):
 
 
 class ModelSerializer(BaseModelSerializer):
+    def create(self, validated_data):
```
rissson, line 36: Some objects get created by authentik, like tokens. Should this instead be a `post_save` signal?

gergosimonyi: Very good question.

I've explored both `post_save` and custom signals, as well as overwriting the `save` method, but ultimately every implementation has its downsides. Specifically: there's no good, universal way of getting the information of who is actually creating the object, apart from capturing it at the API level. At least as far as I can tell. So for non-API object creations, we're fucked.

(Sidenote: thread safety comes into this. I haven't seen any guarantees that just getting "current user" from a global scope will always be correct, and giving the wrong user access is worse than sometimes not giving the right user access.)

So my plan, at least for now, is to mark this feature as experimental. It will not break things, but it will only work _most_ of the time. I expect to fix others on a case-by-case basis when an Issue is opened.

gergosimonyi: (Sidenote 2: another approach might be to architect our system in such a way to give every object an `owner`, but we deemed that too big of a commitment for now. The current PR's implementation is _easy to delete_, as Jens likes to say.)

rissson: Fair enough, thank you for the explanations

authentik/core/api/utils.py
```diff
@@ -29,6 +33,26 @@ def is_dict(value: Any):
...
+        instance = super().create(validated_data)
+        user = self.context["request"].user
```
rissson, line 38: I'm very surprised that this doesn't actually break anything. We use serializers in a bunch of places where we don't have a request. It might also be useful to bail out early if `not user.is_authenticated()`

authentik/core/api/utils.py
```diff
@@ -29,6 +33,26 @@ def is_dict(value: Any):
 
 
 class ModelSerializer(BaseModelSerializer):
```
rissson, line 35: Some APIs don't use our overlay. The list is, from a very very quick grep:
- `authentik/enterprise/stages/authenticator_endpoint_gdtc/api.py`
- `authentik/tenants/api/tenants.py` <- this one actually makes sense not to override I think
- `authentik/enterprise/providers/ssf/views/stream.py`
- `authentik/stages/prompt/api.py`
- `authentik/blueprints/api.py`
Might be worth looking at `authentik.lib.utils.reflection.all_subclasses(rest_framework.serializers.ModelSerializer)` and make sure that they all also depend on our own. (and perhaps adding that as a test!)

rissson (commented):
I need to checkout this and try it for myself to actually fully understand what's going on, but here is a second first impression

authentik/core/api/utils.py
gergosimonyi: Replaced this with a safer version

authentik/core/api/utils.py
gergosimonyi: I'll take a look at these, thank you. (Maybe not today.)

gergosimonyi: So our own `ModelSerializer` is from #10139 and it makes perfect sense. I believe there's no downside to switching these all (including `TenantSerializer`) to ours. I will add that test too, thank you for your recommendation.

rissson: Yep perfect, thanks!

gergosimonyi:
(Note: my force push only fixed a typo in a commit message.)

GirlBossRush (approved):

GirlBossRush (approved):

rissson (approved):

Issue #13441, opened 2025-03-08
SCIM works successfully but logs Exceptions when syncing with GitHub organization


jorhett:
**Describe the bug**

I've set up SCIM for GitHub Organization.  When I look at the SCIM provider I see a long Exception error.  However, when I query the API or look in the GitHub console, SCIM has worked correctly!

**To Reproduce**

1. Setup GitHub Organization SAML as documented at https://docs.goauthentik.io/integrations/services/github-organization/
2. Create a fine-grained token with Write access to Organization Members 
3. Setup a SCIM provider with URLs as documented by GitHub https://docs.github.com/en/enterprise-cloud@latest/rest/scim/scim?apiVersion=2022-11-28#scim-provisioning-for-organizations
4. Run a sync 
5. View SCIM mappings for each user by querying https://api.github.com/scim/v2/organizations/ORGNAME/Users confirms that the users are synced
6. View invitations for users provisioned in Authentik but not active in GitHub in the console or https://api.github.com/scim/v2/organizations/ORGNAME/invitations

...

7. Be utterly confused why Authentik logs only a NotFound exception in Sync Status

**Logs**

<details>
<summary>Expand to see JSON logs...</summary>

```
{"domain_url": null, "event": "Task started", "level": "info", "logger": "authentik.root.celery", "pid": 5521, "schema_name": "public", "task_id": "9c78603a-f0bb-4a99-a90b-6b75962ce2b1", "task_name": "scim_sync_objects", "timestamp": "2025-03-08T18:34:27.202935"}
{"domain_url": null, "event": "failed to get ServiceProviderConfig", "exc": "NotFoundSyncException(<Response [404]>)", "level": "warning", "logger": "authentik.lib.sync.outgoing.base", "pid": 5521, "provider": "GitHub SCIM", "schema_name": "public", "task_id": "task-9c78603af0bb4a99a90b6b75962ce2b1", "timestamp": "2025-03-08T18:34:27.417693"}
{"domain_url": null, "event": "Task finished", "level": "info", "logger": "authentik.root.celery", "pid": 5521, "schema_name": "public", "state": "SUCCESS", "task_id": "9c78603af0bb4a99a90b6b75962ce2b1", "task_name": "scim_sync_objects", "timestamp": "2025-03-08T18:34:31.940549"}
[...]

PR #13947, opened 2025-04-09
providers/scim: modify user- and group syncing behavior

### LINKED ISSUE #6065: SCIM Group Filtering

ImmanuelVonNeumann:
## Details

This is the successor to https://github.com/goauthentik/authentik/pull/13550 with two main changes:
- filter_groups have been renamed to group_filters (reasoning further below)
- users are filtered by the scim-provider's application policies as suggested [here](https://github.com/goauthentik/authentik/pull/13550#issuecomment-2738062003)

### Full Details:

This PR aims to change user and group filtering of the SCIMProvider by incorporating the following changes:
- Group syncing
  - Rename and modify ``filter_group`` to ``group_filters`` (including django migrations to keep a previously selected filter_group present)
  - Multiple groups can be selected as ``group_filters``
  - Only groups within ``group_filters`` are synced to the SCIM Server
- User syncing
  - Only users which are able to view the SCIMProvider's Application are being synced

### Reasoning:
Most SCIM Clients allow groups to be filtered, however authentik currently does not.
This issue has been mentioned in #6065 which has been given the ``enhancement/confirmed``-label.

Among the many use-cases one would be a multi-tenant setup where groups of one tenant should not be visible to another tenant.

I have chosen to rename the attribute to "group_filters" as I believe it to be more explanatory. 
My understanding of a "filter_group" is a group that has filtering logic (not limited to groups but rather defined by groups).
Whereas a "group_filter" filters groups.
If this change is not welcome, I am happy to undo it.

> [!NOTE]  
> Since this does change user- and group-syncing behavior (f.e. not syncing previously synced groups anymore) as well as the schema, I suggest this change to be implemented in a major release.

Fixes #6065

---

## Checklist

If applicable


ImmanuelVonNeumann:
Another design decision to be had is:  
Should you be able to sync no groups?  
If so, how?

My initial thought is to change the syncing behavior when no groups are selected from syncing all groups to none.  
In my opinion this would be more intuitive.

related to: #13441

BeryJu (commented):
Code looks good, we're discussing internally how to best deal with the migration path and to not break any existing setups. 

I think if during the migration we detect the SCIM provider has one filter group selected, we enabled dry-run in the migration to effectively disable the provider until the user has reviewed the config. (Also I can't think of a scenario where you'd have a different set of users provisioned in SCIM than users who can access the application but it might exist somewhere).

The other thing that we want to do before shipping this is optimize the PolicyEngine more, especially when using user or group bindings as without optimizations this would have quite a performance impact especially with a lot of providers setup.

ImmanuelVonNeumann:
> I think if during the migration we detect the SCIM provider has one filter group selected, we enabled dry-run in the migration to effectively disable the provider until the user has reviewed the config.

Sounds like a good idea.
It might be even better to reset the selection (``filter_group``) as the behavior has changed. 

Especially in regards to how an empty selection for ``group_filters`` is handled (syncs all groups vs syncs no groups).
I believe that "sync no groups" would enable more use-cases and be more intuitive.
However, this would mean that if all groups are selected, new groups would not automatically be synced.
To solve this, we could have a button "sync all" which disables the selection and syncs all groups to the scim server.

BeryJu:
> > I think if during the migration we detect the SCIM provider has one filter group selected, we enabled dry-run in the migration to effectively disable the provider until the user has reviewed the config.
> 
> Sounds like a good idea. It might be even better to reset the selection (`filter_group`) as the behavior has changed.
> 
> Especially in regards to how an empty selection for `group_filters` is handled (syncs all groups vs syncs no groups). I believe that "sync no groups" would enable more use-cases and be more intuitive. However, this would mean that if all groups are selected, new groups would not automatically be synced. To solve this, we could have a button "sync all" which disables the selection and syncs all groups to the scim server.

Sorry for the once again delayed response, we've since merged a big improvement to the PolicyEngine that uses a lot less queries and time for static user/group bindings so from that aspect we can go ahead with this one.

I've merged the migrations and set the dry_run to true in the migration if a filter group has been set and merged main into that.

for the `group_filters` behaviour I would be tempted to say to add a 3-state enum (sync no groups, sync filtered groups, sync all groups) instead of trying to be smart about it? Making it an imperative action isn't really the best way to go about that imo

cc @goauthentik/backend for thoughts

BeryJu:
One thing we should also do for this is post-sync review to compare the SCIM users in our database with the users that have access and deactivate/remove any users that were synced but no longer have access

ImmanuelVonNeumann:
I have gone ahead and fixed the linting issues.

---

> for the `group_filters` behaviour I would be tempted to say to add a 3-state enum (sync no groups, sync filtered groups, sync all groups) instead of trying to be smart about it? Making it an imperative action isn't really the best way to go about that imo

This solution definitely makes sense in the current authentik "eco-system", however I believe that this might be a symptom of an underlying issue. I have created [Issue 15816](https://github.com/goauthentik/authentik/issues/15816) where I explain this further since it doesn't affect this PR directly.

If the backend-team agrees with the 3-state enum approach I can go ahead and see if it is feasible for me to implement this feature.

---

> One thing we should also do for this is post-sync review to compare the SCIM users in our database with the users that have access and deactivate/remove any users that were synced but no longer have access

I agree that this feature should be introduced. 
However, I am currently uncertain whether I can take this on.

BeryJu:
@ImmanuelVonNeumann sorry for keeping this waiting for so long, I still want to get this merged  but I haven't gotten around to re-reviewing this and figuring out solutions for the issues in this thread

ImmanuelVonNeumann:
@BeryJu I have gone ahead and rebased the PR.

In my opinion, the current implementation is the most simplistic and implicit usage:
- If you want to sync all groups you leave it as is (empty). This is the default state
- If you want to sync specific groups you select those

If you would want to sync no groups you would have to make an empty "placeholder group" you select.
In the case that this has to be implemented the easiest solution would be a checkbox "disable group synchronization". This should be more implicitly understandable and easy to implement than a radio-selection (enum approach).

BeryJu (approved):

PR #14066, opened 2025-04-15
website/integrations: fix for actual-budget wiki guide

jmarmstrong1207:
## Details

This is a simple fix: remove the front slash at the end of the redirect uri. Will not work if this slash exists in the redirect uri.
---


dominic-r (approved):
Thanks!

dewi-tik (approved):

PR #14092, opened 2025-04-15
policies: fix send deadlock in 'PolicyProcess'

### LINKED ISSUE #10801: SAML endpoints hang when a user has big attributes
**Describe the bug**
On 2024.6.3 my SAML redirects just hang indefinitely when a file (350kB, a profile picture) is set in a user's attributes.

**To Reproduce**

1. Add a very large attribute to your user
1. Set up a SAML provider + application
3. Set up a application that uses SAML (Sentry for example)
4. Use the application to get to the `/application/saml/<provider>/sso/binding/redirect` endpoint
5. See load indefinitely

**Expected behavior**

These endpoints should not hang forever when a user has a large attribute set.

**Logs**

```
authentik-1   | {"event":"tracing request to backend","headers":{/* redacted */},"level":"trace","logger":"authentik.router","timestamp":"2024-08-06T23:57:12Z","url":"http://localhost:8000/application/saml/devminer-sentry/sso/binding/redirect/?SAMLRequest=..."}
authentik-1   | {"auth_via": "session", "binding": "<PolicyBinding: Binding from Sentry [DevMiner] #0 to User Anna DevMiner>", "domain_url": "auth.devminer.xyz", "event": "P_ENG: Evaluating policy", "host": "auth.devminer.xyz", "level": "debug", "logger": "authentik.policies.engine", "pid": 127, "request": "<PolicyRequest user=devminer obj=Sentry [DevMiner] http_request=<ASGIRequest: GET '/application/saml/devminer-sentry/sso/binding/redirect/?SAMLRequest=...'>>", "request_id": "47db9e226d8f48518e9253a445489a99", "schema_name": "public", "timestamp": "2024-08-06T23:57:12.707939"}
authentik-1   | {"auth_via": "session", "binding": "<PolicyBinding: Binding from Sentry [DevMiner] #0 to User Anna DevMiner>", "domain_url": "auth.devminer.xyz", "event": "P_ENG: Starting Process", "host": "auth.devminer.xyz", "level": "debug", "logger": "authentik.policies.engine", "pid": 127, "request": "<PolicyRequest user=devminer obj=Sentry [DevMiner] http_request=<ASGIRequest: GET '/application/saml/devminer-sentry/sso/binding/redirect/?SAMLRequest=...'>>", "request_id": "47db9e226d8f48518e9253a445489a99", "schema_name": "public", "timestamp": "2024-08-06T23:57:12.751424"}
authentik-1   | {"auth_via": "session", "domain_url": "auth.devminer.xyz", "event": "P_ENG(proc): Running policy", "host": "auth.devminer.xyz", "level": "debug", "logger": "authentik.policies.process", "pid": 127, "policy": null, "process": "PolicyProcess", "request_id": "47db9e226d8f48518e9253a445489a99", "schema_name": "public", "timestamp": "2024-08-06T23:57:12.753039", "user": "devminer"}
authentik-1   | {"auth_via": "session", "cached": true, "domain_url": "auth.devminer.xyz", "event": "P_ENG(proc): finished", "host": "auth.devminer.xyz", "level": "debug", "logger": "authentik.policies.process", "passing": true, "pid": 127, "policy": null, "process": "PolicyProcess", "request_id": "47db9e226d8f48518e9253a445489a99", "result": "<PolicyResult passing=True>", "schema_name": "public", "timestamp": "2024-08-06T23:57:12.755303", "user": "devminer"}
```

-   authentik version: 2024.6.3 (also happens with 2024.6.1)

### LINKED ISSUE #20120: PolicyEngine deadlock when User.attributes contains large data (e.g. avatar)
### Describe the bug

## Summary                                                                       
                                                                                                                                                                                                                   
  The `PolicyEngine` deadlocks when evaluating policy bindings that involve a `User` whose `attributes` field contains large data (e.g., a base64-encoded avatar image). The `multiprocessing.Pipe` used for IPC   
  between the policy evaluation subprocess and the parent process has a fixed OS buffer size (64KB on Linux). When the serialized `PolicyResult` exceeds this buffer, `Pipe.send()` blocks indefinitely because the
   reader hasn't started consuming yet — creating an unrecoverable deadlock.

  This causes the `/application/o/authorize/` endpoint to hang forever for authenticated users, completely breaking OIDC login flows.

### How to reproduce

## Steps to Reproduce

  1. Upload an avatar image for any user (e.g., via the Authentik admin UI or API). A normal JPEG photo (~2MB) is sufficient.
  2. Create an Application with at least one policy binding — either a **user binding** referencing that user, or a **group binding** where that user is a member (since the `PolicyRequest` also contains the
  user).
  3. Authenticate as that user and access the `/application/o/authorize/` endpoint (i.e., perform a normal OIDC login flow).
  4. The request hangs indefinitely. No log entry is produced. No error is returned.

### Expected behavior

## Expected Behavior

  The authorize endpoint should process the request and return a redirect (302) to the authorization flow, regardless of the size of the user's `attributes` field.

## Actual Behavior

  The authorize endpoint hangs forever. The gunicorn worker thread is blocked on `Pipe.send()` and never recovers. The request produces no ASGI access log entry (since logging happens after the response).

  ## Root Cause Analysis

  The deadlock occurs in `PolicyEngine.build()` (`authentik/policies/engine.py`):

  ```python
  # Phase 1: Evaluate all bindings sequentially
  for binding in self.iterate_bindings():
      our_end, task_end = Pipe(False)
      task = PolicyProcess(binding, self.request, task_end)
      task.daemon = False
      if not CURRENT_PROCESS._config.get("daemon"):
          task.run()          # <-- synchronous: calls connection.send(result)
      else:
          task.start()        # <-- async: spawns subprocess
      self.__processes.append(PolicyProcessInfo(process=task, connection=our_end, binding=binding))

  # Phase 2: Read results (only reached AFTER all bindings are processed)
  for proc_info in self.__processes:
      if proc_info.process.is_alive():
          proc_info.process.join(proc_info.binding.timeout)
      if not proc_info.result:
          proc_info.result = proc_info.connection.recv()

  The issue chain:

  1. PolicyProcess.execute() sets policy_result.source_binding = self.binding on the result
  2. PolicyProcess.run() calls self.connection.send(result) which pickles the PolicyResult
  3. The pickle includes source_binding → PolicyBinding → User → User.attributes (containing the avatar)
  4. pickle.dumps() on a User with a 2MB avatar produces a ~2.7MB payload
  5. multiprocessing.Pipe has a 64KB OS buffer on Linux
  6. send() writes 64KB, then blocks waiting for the reader to drain the buffer
  7. The reader (recv()) is in Phase 2, which only runs after Phase 1 completes
  8. Phase 1 is blocked on send() → deadlock

  Verification

  # In ak shell:
  import pickle

  user = User.objects.get(username='django_admin')
  print(len(pickle.dumps(user)))          # 2,709,191 bytes (has avatar)
  print(len(pickle.dumps(user.attributes)))  # 2,708,497 bytes (the avatar)

  # The pipe buffer is 64KB — this will never fit

  Environment

  - authentik version: 2025.6.4
  - Platform: Linux (Docker, kernel 5.15)
  - Python: 3.12
  - Trigger: Uploading a standard JPEG avatar (~2MB) for a user referenced in application policy bindings

  Impact

  - Severity: High — completely breaks OIDC login for any application with policy bindings when any referenced user has a non-trivial avatar
  - Affected endpoints: /application/o/authorize/ (and potentially any endpoint that triggers PolicyEngine evaluation)
  - User-facing: Browser shows infinite loading spinner on the Authentik authorization page after entering credentials
  - Recovery: Only workaround is removing all policy bindings from the application or removing the avatar from the user

  Suggested Fixes

  Several approaches could resolve this:

  1. Don't include source_binding in the serialized result — source_binding is set in PolicyProcess.execute() for debugging/logging, but it forces the entire binding (and its related User/Group) through the
  pipe. The binding reference could be passed separately or looked up by ID after recv().
  2. Exclude heavy fields from User pickling — Implement __getstate__/__reduce__ on the User model to exclude attributes (or at least large blob values) during pickle serialization.
  3. Use a different IPC mechanism — Replace multiprocessing.Pipe with a mechanism that doesn't have fixed buffer limits (e.g., multiprocessing.Queue, shared memory, or temporary files for large payloads).
  4. Read results eagerly — Instead of the two-phase approach (write all, then read all), read each result immediately after the corresponding task completes to prevent buffer accumulation.
  5. Limit avatar storage size — Enforce a maximum size for data stored in User.attributes, or store avatars as file references rather than inline base64 blobs.

### Screenshots

_No response_

### Additional context

_No response_

### Deployment Method

Docker

### Version

2025.6.4

### Relevant log output

```shell

```

lhjt:
## Details

This refactor updates the policy engine and evaluator components to use a multiprocessing Queue instead of a Connection for inter-process communication.

The reason for this is if the payload size for a `send()` is too large, it will block. This caused SAML responses with large payloads to hang forever.

Fixes #10801.
Closes #20120

---

## Checklist


rissson:
I've started the tests for this, you should be able to see there if anything is failing.

lhjt:
I have deployed this locally and tested the SAML issue outlined in #10801 (which was the same issue I ran into). I can confirm it is working as expected and it is now ready to review and merge.

rissson (approved):

authentik/policies/engine.py
```diff
@@ -130,24 +130,26 @@ def build(self) -> "PolicyEngine":
...
+                    PolicyProcessInfo(process=task, result_queue=result_queue, binding=binding)
                 )
             # If all policies are cached, we have an empty list here.
             for proc_info in self.__processes:
                 if proc_info.process.is_alive():
                     proc_info.process.join(proc_info.binding.timeout)
                 # Only call .recv() if no result is saved, otherwise we just deadlock here
                 if not proc_info.result:
-                    proc_info.result = proc_info.connection.recv()
+                    proc_info.result = proc_info.result_queue.get()
```
rissson, line 152: Having played with this a bit, I think that this call might block forever if a task times out.

rissson (changes requested):
Thank you for the contribution! I've played around with this a bit, and it definitely feels like using a queue is the correct approach.

However, I'm not keen about the following things:
- creating a queue for each fork seems overkill, one queue for everything should do the trick. However, due to sync issues, retrieving the results is not as easy as it seems.
- using multiprocessing.Queue for non-forking tasks is also overkill, we should just use the standard queue.Queue in that case

There are multiple improvements to do on this, like #14130, so don't treat this like an actual review, but more as notes for myself and others.

authentik/policies/engine.py
lhjt: Yes though it's possible for me to change this to also have its own timeout. Shall it just raise an exception in this case? And how long should the timeout be if so?

lhjt:
Hey @rissson would you be able to run https://github.com/goauthentik/authentik/pull/14133#issuecomment-2813399091 on this PR as well? It'll make testing it in my setup a lot faster since I can't build `amd64` images on my arm mac. Thanks 🙂

authentik/policies/engine.py
lhjt: I've gone ahead and made changes to this. Please let me know if the exception raised is appropriate for this circumstance.

BeryJu (approved):

rissson (approved):

PR #14113, opened 2025-04-16
website: integrations: apache guacamole: remove redirect URI comments

dominic-r:
## Details

REPLACE ME

---

## Checklist

If applicable

### DIFF
--- a/website/integrations/services/apache-guacamole/index.mdx
+++ b/website/integrations/services/apache-guacamole/index.mdx
@@ -89,7 +89,7 @@ Additionally, ensure your `guacamole.properties` file (typically located in `/et
     openid-client-id=<Client ID from authentik>
     openid-issuer=https://authentik.company/application/o/<your-slug>/
     openid-jwks-endpoint=https://authentik.company/application/o/<your-slug>/jwks/
-    openid-redirect-uri=https://guacamole.company/.
+    openid-redirect-uri=https://guacamole.company/
     openid-username-claim-type=preferred_username
     ```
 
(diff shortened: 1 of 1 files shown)


dewi-tik (approved):

website/integrations/services/apache-guacamole/index.mdx
```diff
@@ -89,7 +89,7 @@ Additionally, ensure your `guacamole.properties` file (typically located in `/et
     openid-client-id=<Client ID from authentik>
     openid-issuer=https://authentik.company/application/o/<your-slug>/
     openid-jwks-endpoint=https://authentik.company/application/o/<your-slug>/jwks/
-    openid-redirect-uri=https://guacamole.company/ # This must match the Redirect URI set in authentik (Including trailing slash).
+    openid-redirect-uri=https://guacamole.company/.
```
BeryJu, line 92:
```suggestion
    openid-redirect-uri=https://guacamole.company/
```

tanberry: Oooof good catch on that left-over punctuation, thanks @BeryJu !

I think we should leave the helpful Notes about making sure that the value matches the Redirect URI, and that the trailing slash is important. Why would we NOT want to include that helpful info? (Though we do need to make them match.)

tanberry (approved):
Thanks @dominic-r for this PR. As you suggested, let;'s find a common place to document that theRedirect URI needs to match, and the trailing slash is necessary.

PR #14140, opened 2025-04-17
web/admin: Fix Safari out-of-memory crash

GirlBossRush:
## Details

This PR addresses an issue where Safari crashes when viewing the Admin interface. The fix is a combination of state handling and reworking of the layout to avoid cyclical changes during page resizing.

---

## Checklist

If applicable

### DIFF
--- a/web/src/elements/PageHeader.ts
+++ b/web/src/elements/PageHeader.ts
@@ -10,15 +10,18 @@ import { me } from "@goauthentik/common/users";
 import "@goauthentik/components/ak-nav-buttons";
 import { AKElement } from "@goauthentik/elements/Base";
 import { WithBrandConfig } from "@goauthentik/elements/Interface/brandProvider";
+import { DefaultBrand } from "@goauthentik/elements/sidebar/SidebarBrand";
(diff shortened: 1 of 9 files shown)


web/src/elements/PageHeader.ts
```diff
@@ -63,69 +84,212 @@ export class PageHeader extends WithBrandConfig(AKElement) {
...
+                        padding-inline-end: var(--pf-global--spacer--xl);
+                    }
                 }
-                .pf-c-page__main-section {
-                    flex-grow: 1;
-                    flex-shrink: 1;
+
+                .brand {
+                    grid-area: brand;
+                    background-color: var(--pf-c-page__sidebar--BackgroundColor);
```
BeryJu, line 168:
```suggestion
                    background-color: var(--pf-c-page__sidebar--m-light--BackgroundColor);
```

BeryJu: even with that the logo has a gray border it shouldn't have 
image

BeryJu (changes requested):
With the new navbar there's layout shifts depending on if the page header has a subtitle or not

also breaks the layout with this not being aligned 
image

PR #14135, opened 2025-04-17
web: Fix issue where references to Lit SSR break page styles.

GirlBossRush:
## Details

This PR addresses an issue that stems from use of Lit's Node.js exclusive SSR package on the front-end, causing page styles not to load. This occurs on the Captcha stage, or if a flow footer includes links.

⚠️  **Requires testing of Captcha flow before approving**

Screenshot 2025-04-17 at 19 48 40

---

## Checklist

-   [ ] Local tests pass (`ak test authentik/`)
-   [ ] The code has been formatted (`make lint-fix`)

If an API change has been made

-   [ ] The API schema has been updated (`make gen-build`)

If changes to the frontend have been made

-   [ ] The code has been formatted (`make web`)

If applicable

-   [ ] The documentation has been updated
-   [ ] The documentation has been formatted (`make website`)

PR #14002, opened 2025-04-11
Monorepo Part 2: Putting it all together

GirlBossRush:
## Details

This PR concludes the majority of #1368, allowing us to take advantage of NPMs workspaces, TypeScript's project-wide compiler, and our packaged linter configurations.

Overall, the changes in this PR should keep existing code behaviors as they were, sometimes at the expense of turning off ESLint's more aggressive "nightmare mode" rules. A follow up PR will address disabled rules and avoid intermixing too much linter noise.

---

## Checklist

-   [ ] Local tests pass (`ak test authentik/`)
-   [ ] The code has been formatted (`make lint-fix`)

If an API change has been made

-   [ ] The API schema has been updated (`make gen-build`)

If changes to the frontend have been made

-   [ ] The code has been formatted (`make web`)

If applicable

-   [ ] The documentation has been updated
-   [ ] The documentation has been formatted (`make website`)

PR #14159, opened 2025-04-21
web: Use monorepo package utilities to build packages

GirlBossRush:
## Details

This PR enables much of the monorepo utilities drafted in #14002. Some highlights include:

- ESBuild uses a single build context with multiple entry points, improving build time and correcting chunk-ordering issues when two entry points reference the same file (Partial fix in #14135 and #14089)
- Static files copied before build now run in parallel.

---

## Checklist

If applicable


kensternberg-authentik (approved):
I like these changes, and the tests are passing and it runs nicely.

PR #14186, opened 2025-04-22
website/docs: updated user count info

tanberry:
This PR improves the wording about how user counts are defined. Thanks @fheisler !

### DIFF
--- a/website/docs/enterprise/manage-enterprise.mdx
+++ b/website/docs/enterprise/manage-enterprise.mdx
@@ -115,7 +115,11 @@ The following events occur when a license expires or the internal/external user 
 
 ### About users and licenses
 
-License usage is calculated based on total user counts that authentik regularly captures. This data is checked against all valid licenses, and the sum total of all users. Internal and external users are counted based on the number of active users of the respective type saved in authentik. Service account users are not counted towards the license.
+License usage is calculated based on total user counts that authentik regularly captures. This data is checked against all valid licenses, and the sum total of all users. Internal and external users are counted based on the total number of users of the respective type saved in authentik.
+
+:::info
+Disabled accounts and service account users are not counted towards the user count of the license.
+:::
 
 An **internal** user is typically a team member, such as a company employee, who has access to the full Enterprise feature set. An **external** user might be an external consultant, a volunteer in a charitable site, or a B2C customer who logged onto your website to shop. External users don't get access to Enterprise features, nor to the **My applications** page in authentik. Instead, external users are authenticated and then redirected to log directly into their [default application](../customize/brands.md#external-user-settings).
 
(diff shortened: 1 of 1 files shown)


website/docs/enterprise/manage-enterprise.mdx
```diff
@@ -115,7 +115,11 @@ The following events occur when a license expires or the internal/external user
 
 ### About users and licenses
 
-License usage is calculated based on total user counts that authentik regularly captures. This data is checked against all valid licenses, and the sum total of all users. Internal and external users are counted based on the number of active users of the respective type saved in authentik. Service account users are not counted towards the license.
+License usage is calculated based on total user counts that authentik regularly captures. This data is checked against all valid licenses, and the sum total of all users. Internal and external users are counted based on the total number of users of the respective type saved in authentik.
+
+:::info
+Disabled accounts and service account users are not counted towards the user count of the license.
```
dominic-r, line 121:
```suggestion
Accounts that are disabled, as well as service accounts, are excluded from the license user count.
```

website/docs/enterprise/manage-enterprise.mdx
```diff
@@ -115,7 +115,11 @@ The following events occur when a license expires or the internal/external user
 
 ### About users and licenses
 
-License usage is calculated based on total user counts that authentik regularly captures. This data is checked against all valid licenses, and the sum total of all users. Internal and external users are counted based on the number of active users of the respective type saved in authentik. Service account users are not counted towards the license.
+License usage is calculated based on total user counts that authentik regularly captures. This data is checked against all valid licenses, and the sum total of all users. Internal and external users are counted based on the total number of users of the respective type saved in authentik.
```
dominic-r, line 118:
```suggestion
License usage is determined by the total number of users that authentik regularly records. This information is compared to all active licenses and the overall user count. Both internal and external users are included, based on the total number of each type stored in authentik.
```

dominic-r (changes requested):
Thanks! Little nits:

dominic-r:
Could we also update the title to make it clear these are changes to ent/license docs?

dewi-tik (approved):
I like dominic's suggestions, particularly the first, but they're also not strictly required.

website/docs/enterprise/manage-enterprise.mdx
tanberry: I almost didn't like it because it separates the words "disabled" and "accounts" but I like the rest of it so much so lgtm, thx. ;-)

website/docs/enterprise/manage-enterprise.mdx
tanberry: thanks, I like that last bit, am taking it... :-)

PR #14258, opened 2025-04-26
stages/captcha: Make stage more managed with provider-specific defaults [reopened]

dominic-r:
Better UX

### DIFF
--- a/web/src/admin/stages/captcha/CaptchaStageForm.ts
+++ b/web/src/admin/stages/captcha/CaptchaStageForm.ts
@@ -46,7 +46,9 @@ export class CaptchaStageForm extends BaseStageForm<CaptchaStage> {
     protected updated(_changedProperties: PropertyValues<this>): void {
         if (!this.hasInitializedInstance && this.instance !== undefined) {
             this.hasInitializedInstance = true;
-            this.revealPrivateKey = false;
+            if ((this.instance as Partial<CaptchaStage>).pk) {
+                this.revealPrivateKey = false;
+            }
         }
     }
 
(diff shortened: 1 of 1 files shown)


gergosimonyi:
As someone who sets up Captcha stages often enough to be unhealthy, this would save me so many clicks!

Functionality-wise we must not break support for a drop-in replacement for these captcha services (or even just a proxy someone might have set up for them), so the ability to configure the endpoints must be an option.

---

If the goal of this PR is to make it easier for an admin to set up Captcha stages, I'll propose a frontend-only alternative solution. Have a button (3 buttons?) on the Captcha stage form which populate `jsUrl` and `apiUrl` with the URLs for reCaptcha/hCaptcha/Turnstile. What do you think?

dominic-r:
@gergosimonyi , I agree: Custom endpoints is next on my list. However, there are several reasons which make me prefer keeping this the way it is (other than the fact that I had fun making it not entirely in the frontend): 

In audit logs, it's easier to show the creation of a stage with provider recaptcha_enterprise rather than a stage with js url/api url of https://...

Admin ease of use is increased (and my sake when coding this). Think of Recaptcha Enterprise users: You'd click the button you suggested, but what would you do with the required project id? Would it be left as a placeholder in the autofill you suggested, would the URL not be generated until the user inputs the project id? With this PR, all you need is to enter the project ID and we take care of everything.

Yes, user experience can also be improved with your suggestion, but I'd say 90% (I might be entirely wrong) of captcha users don't need to change the URLs, so the suggestion would just add extra fields which would, IMO, look a little strange in a dedicated section, collapsible or not. The stage configuration is long enough as is, especially for a CAPTCHA.

Also, think of user error: Do we block JS/API url edits unless the user clicks a specific edit button? If not, you might accidentally add a character, accidentally remove one and not even notice it. If you require additional user action to edit the urls, that's friction which could be avoided by hiding those URLs under a "Custom" option of the dropdown.

In the event where a URL is updated, probably never, we can patch it in the release without adding a migration or requiring users to update their stages.

LMK what you think.

dominic-r:
(Need to fix an issue where the secret key is marked as already filled before even putting naything, will try to fix that)

dominic-r:
Also, the captcha options might need little fixing, TBD

dominic-r:
See: https://github.com/goauthentik/authentik/pull/16129

PR #14279, opened 2025-04-29
internal/outpost: fix incorrect usage of golang SHA API

authentik-db-cooper:
## Details

Move away from 512 and 256 SHA hashes in golang and use checksums.

---

## Checklist

If applicable

### DIFF
--- a/internal/outpost/proxyv2/application/application.go
+++ b/internal/outpost/proxyv2/application/application.go
@@ -118,8 +118,8 @@ func NewApplication(p api.ProxyOutpostConfig, c *http.Client, server Server, old
 	mux := mux.NewRouter()
 
 	// Save cookie name, based on hashed client ID
-	h := sha256.New()
-	bs := string(h.Sum([]byte(*p.ClientId)))
+	h := sha256.Sum256([]byte(*p.ClientId))
+	bs := string(h[:])
 	sessionName := fmt.Sprintf("authentik_proxy_%s", bs[:8])
 
 	// When HOST_BROWSER is set, use that as Host header for token requests to make the issuer match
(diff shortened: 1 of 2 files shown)


internal/outpost/proxyv2/application/application.go
```diff
@@ -118,8 +118,8 @@ func NewApplication(p api.ProxyOutpostConfig, c *http.Client, server Server, old
 	mux := mux.NewRouter()
 
 	// Save cookie name, based on hashed client ID
-	h := sha256.New()
-	bs := string(h.Sum([]byte(*p.ClientId)))
+	h := sha256.Sum256([]byte(*p.ClientId))
+	bs := string(h[:])
```
rissson, line 122: Does this break existing sessions? Or is the resulting hash the same?

internal/outpost/radius/handler.go
```diff
@@ -68,7 +68,8 @@ func (rs *RadiusServer) ServeRADIUS(w radius.ResponseWriter, r *radius.Request)
 		}
 	}
 	if pi == nil {
-		nr.Log().WithField("hashed_secret", string(sha512.New().Sum(r.Secret))).Warning("No provider found")
+		logValue := sha512.Sum512(r.Secret)
+		nr.Log().WithField("hashed_secret", string(logValue[:])).Warning("No provider found")
```
BeryJu, line 72:
```suggestion
		hashedSecret := sha512.Sum512(r.Secret)
		nr.Log().WithField("hashed_secret", string(hashedSecret[:])).Warning("No provider found")
```

internal/outpost/proxyv2/application/application.go
BeryJu: it does change the name of the cookie as with the previous invocation we were basically not hashing the value at all @rissson

PR #14288, opened 2025-04-29
web: cleanup/loading attribute always true

kensternberg-authentik:
web: remove Lit syntax from always true attributes

## What

Replaces instances of `?loading=${true}` and `?loading="${true}"` with `loading`

## Why

The Lit syntax is completely unnecessary when the attribute’s state is constant, and it’s a few (just a few) extra CPU cycles for Lit to process that.

More to the point, it annoys me.

## How

    $ perl -pi.bak -e 's/\?loading=\$\{true\}/loading/' $(rg -l '\?loading=\$\{true\}')
    $ find . -name '*.bak' -exec rm {} \;
    $ perl -pi.bak -e 's/\?loading="\$\{true\}"/loading/' $(rg -l '\?loading="\$\{true\}"')
    $ find . -name '*.bak' -exec rm {} \;


GirlBossRush (approved):

web/src/elements/charts/Chart.ts
```diff
@@ -231,9 +231,7 @@ export abstract class AKChart<T> extends AKElement {
                               <p slot="body">${pluckErrorDetail(this.error)}</p>
                           </ak-empty-state>
                       `
-                    : html`${this.chart
-                          ? html``
-                          : html`<ak-empty-state ?loading="${true}"></ak-empty-state>`}`}
+                    : html`${this.chart ? html`` : html`<ak-empty-state loading></ak-empty-state>`}`}
```
BeryJu, line 234: Could replace that html`` with nothing too but also makes sense if you want to do that in another cleanup pr

BeryJu (approved):

PR #14289, opened 2025-04-29
Web/cleanup/empty state better slot handling

kensternberg-authentik:
web: move optional textual information out of attributes and into slots

## What

Replaces instances of:

    <ak-empty-state header=${msg(...)}></ak-empty-state>

with

    <ak-empty-state><span slot="header">${msg(...)}</span></ak-empty-state>

## Why

1.  It’s correct.
2.  It lets us elide the decorations for any slots we aren’t using.
3.  It’s preparation for moving to Patternfly 5
4.  It annoyed me.

## How

Since we already have Patternfly Elements installed, we have access to the PFE-Core, which has the unbelievable useful `SlotsController`. Using it, I created a conditional render template that will only put in the header, body, and primary slots if there is something in the lightDOM requesting those slots. The conditional template will still put the spinner in if the header is not provided but the loading state is true.

I then had to edit all the places where this is used. For about 30 of them, this script sufficed:

    perl -pi.bak -e 's/header="?(\$\{msg\([^)]+\)\})"?>/><span slot="header">\1<\/span>/' \
         $(rg -l `<ak-empty-state[^>]header')

The other six had to be done by hand. I have tested a handful of the automatic ones, and all of the ones that were edited manually. I’m pleasantly surprised that the textual rules [are inherited by the slots as expected](https://htmlwithsuperpowers.netlify.app/styling/inheritable.htm).


web/src/elements/EmptyState.ts
```diff
@@ -59,15 +68,17 @@ export class EmptyState extends AKElement implements IEmptyState {
...
-                <h1 class="pf-c-title pf-m-lg">
-                    ${this.loading && this.header === undefined ? msg("Loading") : this.header}
-                </h1>
-                <div class="pf-c-empty-state__body">
-                    <slot name="body"></slot>
-                </div>
-                <div class="pf-c-empty-state__primary">
-                    <slot name="primary"></slot>
-                </div>
+                ${showHeader ? html` <h1 class="pf-c-title pf-m-lg">${header()}</h1>` : nothing}
```
kensternberg-authentik, line 71: This way, we don't fill the space with empty elements, and Lit knows not to track them.

GirlBossRush (approved):

PR #14291, opened 2025-04-29
website/docs: clean up oauth redirect paragraph

fheisler:

### DIFF
--- a/website/docs/add-secure-apps/providers/oauth2/index.mdx
+++ b/website/docs/add-secure-apps/providers/oauth2/index.mdx
@@ -64,7 +64,7 @@ When using an OAuth 2.0 provider in authentik, the OP must validate the provided
 
 When you create a new OAuth 2.0 provider and app in authentik and you leave the **Redirect URI** field empty, then the first time a user opens that app, authentik uses that URL as the saved redirect URL.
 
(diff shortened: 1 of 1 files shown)


tanberry (approved):
whah thanks @fheisler !

website/docs/add-secure-apps/providers/oauth2/index.mdx
```diff
@@ -64,7 +64,7 @@ When using an OAuth 2.0 provider in authentik, the OP must validate the provided
 
 When you create a new OAuth 2.0 provider and app in authentik and you leave the **Redirect URI** field empty, then the first time a user opens that app, authentik uses that URL as the saved redirect URL.
 
-For advanced use cases, an authentik admin can use regular expressions (regex) instead of a redirect URL. For example, if you want to list 10 diff applications, instead of listing all ten you can create an expression with wildcards. Be aware, when using regex, that authetnik uses a dot as a separator in the URL, but in regex a dot means "one of any character", a wildcard. So you should escape the dot to prevent its interpration as a wildcard.
+For advanced use cases, an authentik admin can use regular expressions (regex) instead of a redirect URL. For example, if you want to list ten different applications, instead of listing them all individually, you can create an expression with wildcards. When using regex, be aware that authentik uses a dot as a separator in the URL, but in regex a dot means "one of any character," a wildcard. You should therefore escape the dot with `\.` to prevent its interpration as a wildcard.
```
dominic-r, line 67:
```suggestion
For advanced use cases, an authentik admin can use regular expressions (regex) instead of a redirect URL. For example, if you want to list ten different applications, instead of listing them all individually, you can create an expression with wildcards. When using regex, be aware that authentik uses a dot as a separator in the URL, but in regex a dot means "one of any character:, a wildcard. You should therefore escape the dot with `\.` to prevent its interpration as a wildcard.
```

PR #14325, opened 2025-05-01
stages/email: fix email scanner voiding token

BeryJu:
## Details

Some email scanners open links in emails in a full browser, which will invalidate the flow token making the email unusable for the user (and depending on the flow finishing the flow)

closes #13805

---

## Checklist

If applicable


gergosimonyi (approved):
What this fails to enable is something like a

1. TOTP check       <---- User doesn't have their TOTP right now, so they want to try again later. This should not require a new link/token/email
2. Prompt for new password    <---- Consume the token here

But this does address the more immediate issue, so approved.

PR #14330, opened 2025-05-02
web: Clean up browser-only module imports that crash WebDriverIO.

GirlBossRush:
## Details

This PR separates a small part of an issue in #14292 -- where our E2E test suite bundles in our browser-only `globalAK` context, as well as removing code paths which can inadvertently lead back to it.

The changes in this PR touch a lot of files but are limited to removing the `first` utility in favor of JavaScript's `??` [nullish-coalescing](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing syntax).  

---

## Checklist

If applicable

### DIFF
--- a/web/src/common/utils.ts
+++ b/web/src/common/utils.ts
@@ -1,5 +1,3 @@
-import { kebabCase } from "change-case";
-
 import { CSSResult, css } from "lit";
 
 export function getCookie(name: string): string {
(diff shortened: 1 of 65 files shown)


web/src/common/utils.ts
```diff
@@ -19,16 +19,11 @@ export function getCookie(name: string): string {
 }
 
 export function convertToSlug(text: string): string {
-    return text
-        .toLowerCase()
-        .replace(/ /g, "-")
-        .replace(/[^\w-]+/g, "");
+    return kebabCase(text);
```
BeryJu, line 22: I dont think `kebabCase` does the second replace of removing anything but `\w` from the string (technically that should actually be \w\d)

kensternberg-authentik (approved):
And there is much rejoicing!

PR #14274, opened 2025-04-29
web: NPM workspaces

GirlBossRush:
**Looking for the related OpenAPI generator changes? See #14332** 

## Details

This PR prepares NPM for use with [Workspaces](https://docs.npmjs.com/cli/v11/using-npm/workspaces).

---

## Checklist

-   [ ] Local tests pass (`ak test authentik/`)
-   [ ] The code has been formatted (`make lint-fix`)

If an API change has been made

-   [ ] The API schema has been updated (`make gen-build`)

If changes to the frontend have been made

-   [ ] The code has been formatted (`make web`)

If applicable

-   [ ] The documentation has been updated
-   [ ] The documentation has been formatted (`make website`)

PR #14332, opened 2025-05-02
root: Prep OpenAPI generators for NPM Workspaces.

GirlBossRush:
## Details

**Depends on #14274** 

This PR prepares the OpenAPI generator scripts for use with [NPM Workspaces](https://docs.npmjs.com/cli/v11/using-npm/workspaces).

- Updates TypeScript API package to use NPM's `link` command.
- - Allows `@goauthentik/web` to use a shared `node_modules` directory at the repo root when Workspaces are enabled.
- - Allows linking via package name instead of directory name.
- Defines a repo-level `tsconfig.json`
- - Allows TypeScript to perform type-checking against repo scripts. 
- - Preps for project references, allowing the type-checker to work across packages that reference each other.
- Partial work-around for issue affecting [Docker Desktop on Mac](https://www.docker.com/blog/incident-update-docker-desktop-for-mac/), allowing front-end only setup to generate API clients.  

---

## Checklist

If applicable


scripts/openapi-generator.mjs
```diff
@@ -0,0 +1,100 @@
...
+    inputSpec = resolve(cwd, "schema.yml"),
+    commandArgs = [],
+}) {
+    if (existsSync(outputDirectory)) {
+        console.log(`Removing existing generated API client from ${outputDirectory}`);
+
+        rmSync(outputDirectory, { recursive: true, force: true });
+    }
+
+    const containerEngine = commandExists("docker") || commandExists("podman");
```
GirlBossRush, line 67: I ran into [this issue](https://www.docker.com/blog/incident-update-docker-desktop-for-mac/) while testing macOS Sequoia. If we want to avoid looking for Podman, I can remove this and set an alias 😅

scripts/openapi-generator.mjs
```diff
@@ -0,0 +1,100 @@
...
+
+    if (!containerEngine) {
+        throw new Error("Container engine not found. Is Docker or Podman available in the PATH?");
```
GirlBossRush, line 70: Similar to above, this error should help the front-end only setup in case the developer is affected, or forgot to install an engine all-together.

Makefile
```diff
@@ -116,49 +115,21 @@ gen-diff:  ## (Release) generate the changelog diff between the current schema a
...
+gen-client-ts:  ## Build and install the authentik API for Typescript into the authentik UI Application
+	./scripts/gen-client-ts.mjs
 
-gen-clean-go:  ## Remove generated API client for Go
-	rm -rf ./${GEN_API_GO}/
+	npm i --prefix ${GEN_API_TS}
 
-gen-clean-py:  ## Remove generated API client for Python
-	rm -rf ./${GEN_API_PY}/
+	cd ./${GEN_API_TS} && npm link
```
GirlBossRush, line 123: Fixes issue where locally-generated API package can become out of sync when adding or removing packages in `@goauthentik/web`

gergosimonyi (commented):
If I cared about this enough, I'd say that I'm not a fan of this shell-to-node-to-shell-to-docker call. If a project forces me to use `make`, the least it can do is let me see exactly what's happening in the `Makefile`.

But since I don't care too much, I'll just throw this to @BeryJu. Maybe he does.

GirlBossRush:
Closing until #14370 settles

PR #14370, opened 2025-05-03
web: NPM Workspaces

GirlBossRush:
## Details

**Depends on #14555**

This PR updates our ESLint and Prettier scripts for use at the repo root. Also included is a collection of ignored "nightmare mode" rules that can be addressed in follow up PRs.

- Fewer package-lock.json fixes, rebases, collisions
- A fix for the generated API client de-synchronizing when installing dependencies
- A single shared TypeScript context for checking types
  - Cached type checking across all packages while developing.
  - Faster type information when developing, and smaller type reconciliations when importing packages within the mono repo.

---

## Checklist

If applicable


gergosimonyi (approved):
Backend (which here is nothing) ✅

Dockerfile
```diff
@@ -1,49 +1,41 @@
...
+# Stage 1: Build Node packages
+FROM --platform=${BUILDPLATFORM} docker.io/library/node:24-slim AS node-packages
 
-ENV NODE_ENV=production
-
-WORKDIR /work/website
+ARG GIT_BUILD_HASH
+ENV GIT_BUILD_HASH=$GIT_BUILD_HASH
 
-RUN --mount=type=bind,target=/work/website/package.json,src=./website/package.json \
```
rissson, line 10: So the goal here is to only re-run this step when `package.json` or `package-lock.json` changed, and thus improve build speeds when only the code changes. I don't know how that would work with workspaces, probably need to mount a few more files, but in any case, we should preserve the current behavior

.github/workflows/translation-extract-compile.yml
```diff
@@ -32,15 +32,25 @@ jobs:
...
       - name: Setup authentik env
         uses: ./.github/actions/setup
-      - name: Generate API
+      - uses: actions/setup-node@v4
+        with:
+          node-version-file: package.json
+          cache: "npm"
+          cache-dependency-path: package-lock.json
+      - name: Prepare Dependencies
+        run: npm ci
```
rissson, line 41: Isn't this already done by .github/actions/setup?

Dockerfile
rissson: I've only added this comment here, but it also goes for other changed dockerfiles

rissson (changes requested):

PR #14476, opened 2025-05-12
website/docs: Update Azure AD identity provider doc page

skalahonza:
## Details
Describe how to add Azure AD source to login screen.

## Checklist

If applicable

### DIFF
--- a/website/docs/users-sources/sources/social-logins/azure-ad/index.mdx
+++ b/website/docs/users-sources/sources/social-logins/azure-ad/index.mdx
@@ -111,6 +111,23 @@ return True
 
 Try to login with a **_new_** user. You should see no prompts and the user should have the correct information.
 
+### Azure AD source on login screen
+
+By default, after creating an Azure AD OAuth Source in Authentik, it is not automatically available on the login screen.
+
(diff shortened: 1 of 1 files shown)


website/docs/users-sources/sources/social-logins/azure-ad/index.mdx
```diff
@@ -111,6 +111,23 @@ return True
 
 Try to login with a **_new_** user. You should see no prompts and the user should have the correct information.
 
+### Azure AD source on login screen
```
dominic-r, line 114: This goes for every social login source and I don't think we include this step-by-step on other pages.

Tho, if we decide to go ahead and include this:
* authentik should always be written in lowercase
* Please use a numbered list for this
* You can't assume flows/stages (e.g. default-authentication-flow) as users may use different flows or they were renamed
* You can't assume the name either as some might choose a non-standard name

dominic-r (changes requested):

website/docs/users-sources/sources/social-logins/azure-ad/index.mdx
skalahonza: I’ve updated the code to address your points

PR #14510, opened 2025-05-14
core: fix unable to create group if no enable_group_superuser permission is given

BeryJu:
## Details

A restricted user should be able create a permission without having the permission to the enable/disable superuser flag. The permission needs to be checked when a group is updated or the superuser flag is set

---

## Checklist

If applicable


gergosimonyi (approved):
This is the better approach for a patch release.

PR #14511, opened 2025-05-14
core: move group superuser permission check

gergosimonyi:
An alternative approach to #14510, because I believe this is the correct place for it.

The downside to this is the message is not tied to the field. Maybe that can be remedied?

### DIFF
--- a/authentik/core/tests/test_groups_api.py
+++ b/authentik/core/tests/test_groups_api.py
@@ -118,10 +118,10 @@ class TestGroupsAPI(APITestCase):
             reverse("authentik_api:group-list"),
             data={"name": generate_id(), "is_superuser": True},
         )
-        self.assertEqual(res.status_code, 400)
+        self.assertEqual(res.status_code, 403)
         self.assertJSONEqual(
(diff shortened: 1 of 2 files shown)


authentik/core/tests/test_groups_api.py
```diff
@@ -144,10 +157,10 @@ def test_superuser_update_no_perm(self):
             reverse("authentik_api:group-detail", kwargs={"pk": group.pk}),
             data={"is_superuser": False},
         )
-        self.assertEqual(res.status_code, 400)
+        self.assertEqual(res.status_code, 403)
         self.assertJSONEqual(
             res.content,
-            {"is_superuser": ["User does not have permission to set superuser status to False."]},
+            {"detail": "User does not have permission to set the given superuser status."},
```
BeryJu, line 163: The only thing I dislike about this change is that we no longer get the error tied to the specific `is_superuser` field but rather just a general error

BeryJu (commented):
One nit-ish, otherwise LGTM

PR #14522, opened 2025-05-15
website/integrations: fix azure AD source docs

woutervanelten:
- Location in Authentik updated to Directory -> Federation and Social login.
- Consumer secret clarified.
- Profile URL updated to the Graph API that is the advised method.


dewi-tik (approved):

dewi-tik:
Thank you for this @woutervanelten. I've actually re-written the entire azure ad doc to an Entra ID doc and this will be merged along with 2025.8 which should be releases in the next couple of weeks. There will also be updates to the authentik UI itself to update the source to Entra ID along with changes to the profile URL. So I'm going to close this PR in favor of that one. Regardless, thank you for the contribution.

PR #14541, opened 2025-05-17
web/NPM Workspaces: ESbuild version cleanup

GirlBossRush:
## Details

**Split off from #14370**

This PR cleans up usage of the `VERSION` constant in the web bundle:

- More consistently uses shared Node and Browser [`import.meta.env`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import.meta)
  - Aligns with Vite's approach to [limiting `process.env` to Node.js's runtime](https://vite.dev/guide/env-and-mode).
- Allows for build-time version to injected as a constant, while still supporting the fetch for new versions at runtime.   

---

## Checklist

If applicable


web/packages/monorepo/environment.js
```diff
@@ -14,19 +28,26 @@
...
- * @param {Prefix} [prefix='process.env.']
+ * @param {Prefix} [prefix='import.meta.env.']
  *
  * @returns {{[K in keyof EnvRecord as `${Prefix}${K}`]: JSONify<EnvRecord[K]>}}
  */
-export function serializeEnvironmentVars(input, prefix = /** @type {Prefix} */ ("process.env.")) {
+export function serializeEnvironmentVars(
+    input,
+    prefix = /** @type {Prefix} */ ("import.meta.env."),
+) {
```
kensternberg-authentik, line 50: This is so clearly a mapping, so, you know, you could make it a mapping:

```
 Object.fromEntries(
        Object.entries(input).map(([key, value]) => [prefix + key, JSON.stringify(value ?? "")]),
    );
```

kensternberg-authentik (approved):

PR #14616, opened 2025-05-21
Web: Controllers cleanup

GirlBossRush:
## Details

This PR addresses a related category of issues surrounding Lit's context controllers.

- Controllers are now consistently mounted via `this.addController`, ensuring their host callbacks are invoked when changes occur.
- Initial values provided by the server are synchronously to the controller's host,
resulting in immediate access the context and fewer re-renders.
- Changes applied to context such as branding propagate now consistently propagate to the host.
- Controllers now abort requests when the host is disconnected
---

## Checklist

If applicable


kensternberg-authentik (approved):
This is actually very good.  I would have Jens double check the FlowExecutor, just because I'm less familiar with it than I should be.

PR #14649, opened 2025-05-22
web/maintenance: remove `writeOnly` hacks from Form and HorizontalFormElement

kensternberg-authentik:
web/maintenance: remove `writeOnly` hacks from Form and HorizontalFormElement

## What

The `writeOnly` hack substituted an obscuring, read-only field for secret keys and passwords that an admin should never be able to see/read, only *write*, but allowed the user to click on and replace the key or password. The hack performed this substitution within `HorizontalFormElement` and dispersed a flag throughout the code to enforce it. Another hack within `Form` directed the API to not update / write changes to that field if the field had never been activated.

This commit replaces the `writeOnly` hack with a pair of purpose-built components, `ak-private-text-input` and `ak-private-textarea-input`, that perform the exact same functionality but without having to involve the HorizontalFormElement, which really should just be layout and generic functionality. It also replaces all the `writeOnly` hackery in Form with a simple `doNotProcess` flag, which extends and genericizes this capability to any and all input fields.

The only major protocol change is that `?writeOnly` was a *positive* flag; you controlled it by saying `this.instance !== undefined`; `?revealed` is a *positive* flog; you reveal the working input field when `this.instance === undefined`.

It is not necessary to specify the monospace, autocomplete, and spell-check features; those are enabled or disabled automatically when the `input-hint="code"` flag is passed.

## Why

Removing special cases from processing code is an important step toward the Authentik Elements NPM package, as well as the Schema-Driven Forms update.

## Note

This is actually a very significant change; this is important functionality that I have hand-tested quite a bit, but could wish for automated testing that also checks the database back-end to ensure the fixes made write the keys and passwords as required. Checking the back-end directly is important since these fields are never re-sent to the front-end after being saved!

Things like `placeholder`, `required`, and getting the `name`, `label` or `help` are all issues very subject to Last-Line Effect, so give this the hairiest eyeball you’ve got, please.


web/src/components/ak-private-text-input.ts
```diff
@@ -0,0 +1,73 @@
...
+    @property({ type: Boolean, reflect: true })
+    public revealed = false;
+
+    @property({ type: String })
+    placeholder = "";
+
+    #onReveal() {
+        this.revealed = true;
+    }
+
```
kensternberg-authentik, line 25: The next function is the heart of this whole shebang: it moves the functionality for "hide value but show it exists / allow input to change value" _out_ of the forms management code and directly into a component that does the correct thing without needing hacks in both HorizontalFormElement and Form.

web/src/elements/forms/Form.ts
```diff
@@ -33,6 +33,8 @@ type HTMLNamedElement = Pick<HTMLInputElement, "name">;
 
 export type AkControlElement<T = string | string[]> = HTMLInputElement & { json: () => T };
 
+const doNotProcess = <T extends HTMLElement>(element: T) => element.dataset.formIgnore === "true";
```
kensternberg-authentik, line 36: While this effectively _adds_ one more line of code, it's much more generic and much less special-case.  It just says "Here are the rules for input elements you shouldn't process: they'll tell you when you shouldn't process them." Simple, straightforward.

web/src/elements/forms/HorizontalFormElement.ts
```diff
@@ -29,11 +29,6 @@ import PFBase from "@patternfly/patternfly/patternfly-base.css";
  * 3. Updated() pushes the `name` field down to the children, as if that were necessary; why isn't
  *    it being written on-demand when the child is written? Because it's slotted... despite there
  *    being very few unique uses.
- * 4. There is some very specific use-case around the `writeOnly` boolean; this seems to be a case
```
kensternberg-authentik, line 32: Welp, that's no longer necessary, now is it? 😀
Next: Kill the other three comments. 😈

web/src/admin/crypto/CertificateKeyPairForm.ts
```diff
@@ -45,37 +46,24 @@ export class CertificateKeyPairForm extends ModelForm<CertificateKeyPair, string
...
-                    class="pf-c-form-control pf-m-monospace"
-                    placeholder="-----BEGIN CERTIFICATE-----"
-                    required
-                ></textarea>
-                <p class="pf-c-form__helper-text">${msg("PEM-encoded Certificate data.")}</p>
-            </ak-form-element-horizontal>
-            <ak-form-element-horizontal
-                name="keyData"
-                ?writeOnly=${this.instance !== undefined}
+                input-hint="code"
```
BeryJu, line 52: where did this input-hint come from?

kensternberg-authentik: `input-hint="code"` activates these three features in `ak-private-textarea-input`:

```
                    autocomplete="off"
                    spellcheck="false"
                    class="pf-c-form-control pf-m-monospace"
```

... which you can clearly see in the textarea we are replacing.

Since those are the common settings you, Jens, use when making a textarea work as a code space.  Having an attribute instead of remembering the same three lines a dozen or so times seemed parsimonious.

GirlBossRush (approved):

web/src/admin/crypto/CertificateKeyPairForm.ts
kensternberg-authentik: (Sigh... that comment had been "pending" for 2 weeks.)

PR #14696, opened 2025-05-26
web: Import organization

GirlBossRush:
## Details

This PR is huge but the changes are the result of a reproducible formatter script. The changes here aim to solve three architectural constraints:

**1. Modules (and their sub-exports) are resolvable regardless if the runtime is Node.js or a bundler**

**2. Modules (and their sub-exports) can be bundled by tools such as ESBuild, Vite, SWC, or any combination of configuration defined by a tools such as Storybook and WebDriverIO.**

**3. Modules (and their sub-exports) can interchangeably use JavaScript and TypeScript – while maintaining project-wide type checking.**

These constraints are made possible via...

### Consistent use of the package.json `exports` map.

This change makes use of Node's built-in ["subpath exports" within the web package](https://nodejs.org/api/packages.html#subpath-exports).

For JavaScript packages that are publishable (such as our ESBuild plugin, Prettier, etc), the sub-exports create a module-API to specific files or paths that can be imported. While the `"import"` key refers to a JavaScript file, the `"types"` key refers to a generated type definition inferred from JSDoc comments.

### Consistent use of the package.json `imports` map.

This change makes use of Node's built-in ["private-mapping" within the web package](https://nodejs.org/api/packages.html#subpath-imports). This removes a dependency on TypeScript's "paths" alias, allowing for non-TypeScript code (such as build scripts) to make use of aliases over deeply nested paths.

Private mappings are also paired with identical public mappings for use in package-like contexts, such in Storybook.

### Clean up of side-effect imports

Imports which apply side-effects (such as registering web components, polyfills, etc) are now configured to appear before other modules, ensuring that they preemptively apply their effects. In most cases this hasn't cause trouble, but the new alignment should remove this class of bugs and reduce diffs when addressing merge conflicts.

## How to review this PR

This PR is made possible in stages.

### Clean up locale imports.

`@goauthentik/locales/*` => `#locales/*`.

### Clean ambiguous CSS imports.

`path.css.js` => `path.styles.js`.

### Clean up entrypoint imports.

`./index.entrypoint` => `#entrypointName/index.entrypoint`.

### Format imports, extensions.

The big one. This is performed via:

```sh
AK_FIX_LEGACY_IMPORTS=true npx prettier --config ../packages/prettier-config/index.js --write -u .
```

### Remove TS aliases.

After the above, we can remove the TS aliases and any related packages.

---

## Checklist

If applicable


kensternberg-authentik (approved):
Fully mechanical changes; no change of behavior observed in the UI after deployment, no notably baroque decisions.  Looks great.

PR #14779, opened 2025-05-29
website/docs: update user ref doc with parent group example

dewi-tik:
## Details

Adds an extra example to the doc. How to obtain user groups and parent groups.

---

## Checklist

If applicable

### DIFF
--- a/website/docs/users-sources/user/user_ref.mdx
+++ b/website/docs/users-sources/user/user_ref.mdx
@@ -42,7 +42,7 @@ user.ak_groups.filter(name__startswith='test')
 
 ### List a user's group memberships including parent groups
 
-Use the following example to list all groups that a User object is a member of, including parent groups:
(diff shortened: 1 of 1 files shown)


website/docs/users-sources/user/user_ref.mdx
```diff
@@ -40,6 +40,14 @@ Use the following example to list groups that a User object is a member of, but
 user.ak_groups.filter(name__startswith='test')
 ```
 
+### List a user's group memberships including parent groups
+
+Use the following example to list all groups that a User object is a member of, including parent groups:
```
tanberry, line 45:
```suggestion
Use the following example to list all groups that a user object is a member of, including parent groups:
```

tanberry (approved):
a question and one edit but approving so that I am not blocking. Thanks @dewi-tik !

tanberry (approved):
Thanks @dewi-tik !

PR #14783, opened 2025-05-29
web: Fix issue where dual select type is not specific.

GirlBossRush:
## Details

Fixes runtime error introduced in #14296.

This stemmed from the dual select types surfacing only within the element -- Rather than in the parent html template.

---

## Checklist

If applicable

### DIFF
--- a/web/src/elements/ak-dual-select/ak-dual-select-dynamic-selected-provider.ts
+++ b/web/src/elements/ak-dual-select/ak-dual-select-dynamic-selected-provider.ts
@@ -4,7 +4,7 @@ import { ref } from "lit/directives/ref.js";
 
 import { AkDualSelectProvider } from "./ak-dual-select-provider.js";
 import "./ak-dual-select.js";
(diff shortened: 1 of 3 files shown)


web/src/elements/ak-dual-select/ak-dual-select-dynamic-selected-provider.ts
```diff
@@ -37,7 +37,7 @@ export class AkDualSelectDynamic extends AkDualSelectProvider {
 
         this.#didFirstUpdate = true;
 
-        this.selector(this.options).then((selected) => {
+        this.selector?.(this.options).then((selected) => {
```
kensternberg-authentik, line 40: I has a sad that the `?` is necessary here.

kensternberg-authentik (approved):
Better!  Sad to lose the defaulting help, tho.  Can we use a utility type to derive the `T` in DualSelectPairSource when returning the default value?

PR #14788, opened 2025-05-30
core: Migrate permissions before deleteing OldAuthenticatedSession

melizeche:


gergosimonyi:
The current state is:
- `2025.4.0` (state: `authentik_core 0047`): `AuthenticatedSession` permissions don't work
- `2025.4.1` (state: `authentik_core 0048`): `AuthenticatedSession` permissions have to be recovered

What this is supposed to fix is people upgrading from `2025.4.0` or earlier by the path
- `2025.4.0` [--> `2025.4.2`] --> `2025.6.0` or
- `2025.2.x` --> `2025.4.2` --> `2025.6.0`

The approach by this PR works for the second case, but doesn't work for the first case. Whoever starts from `2025.4.0` will already have `authentik_core 0047`.

Instead of this approach, this code should only change `authentik_core 0048`.

I'll work on this and commit to this PR.

BeryJu (approved):

PR #14807, opened 2025-06-02
website/integrations: update cloudflare access callback url

terafirmanz:
Very simple Doc change.

The callback URL example for Cloudflare Access had a trailing / that breaks the callback URL being matched by a strict policy.


dewi-tik (approved):
Thanks for catching this @terafirmanz

PR #14867, opened 2025-06-04
.

Gaspa79:
.

### DIFF
--- a/authentik/core/templates/base/skeleton.html
+++ b/authentik/core/templates/base/skeleton.html
@@ -16,7 +16,7 @@
         {% block head_before %}
         {% endblock %}
         <link rel="stylesheet" type="text/css" href="{% static 'dist/authentik.css' %}">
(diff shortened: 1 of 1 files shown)


BeryJu:
The `branding_custom_css` is supposed to contain CSS rules and not links to CSS files

Gaspa79:
.

BeryJu:
> > The `branding_custom_css` is supposed to contain CSS rules and not links to CSS files
> 
> Oh okay, thanks for commenting. Then I wonder why it gets resolved to the file path/name. Currently, all custom styling is broken (since moving the customizations by brand) because the variable resolves to the path

`branding_custom_css` simply refers to the same field on the brand so whatever is saved in the brand will be returned here. The previous migration bug caused the data to not be saved but I can't see a possibility of the migration setting the CSS value to the path of the css file

Gaspa79:
.

PR #14905, opened 2025-06-04
root: Better version bump

BeryJu:
## Details

REPLACE ME

---

## Checklist

If applicable


.github/actions/setup/action.yml
```diff
@@ -10,42 +13,51 @@ runs:
   using: "composite"
   steps:
     - name: Install apt deps
+      if: ${{ contains(inputs.dependencies, 'system') }}
```
BeryJu: Might be a bit cleaner to break this up into individual sub-actions for each type

Makefile
```diff
@@ -93,6 +93,15 @@ update-test-mmdb:  ## Update test GeoIP and ASN Databases
 	curl -L https://raw.githubusercontent.com/maxmind/MaxMind-DB/refs/heads/main/test-data/GeoLite2-ASN-Test.mmdb -o ${PWD}/tests/GeoLite2-ASN-Test.mmdb
 	curl -L https://raw.githubusercontent.com/maxmind/MaxMind-DB/refs/heads/main/test-data/GeoLite2-City-Test.mmdb -o ${PWD}/tests/GeoLite2-City-Test.mmdb
 
+bump:  ## Bump authentik version. Usage: version=20xx.xx.xx make bump
+	uv version $(version)
```
BeryJu, line 101: Should probably add an error check for missing `version=`

Makefile
```diff
@@ -93,6 +93,15 @@ update-test-mmdb:  ## Update test GeoIP and ASN Databases
...
+	$(MAKE) gen-build
```
```suggestion
	$(MAKE) gen-build gen-compose aws-cfn
```

docker-compose.yml
```diff
@@ -1,90 +1,85 @@
...
+    env_file:
+    - .env
     environment:
-      AUTHENTIK_SECRET_KEY: ${AUTHENTIK_SECRET_KEY:?secret key required}
-      AUTHENTIK_REDIS__HOST: redis
       AUTHENTIK_POSTGRESQL__HOST: postgresql
-      AUTHENTIK_POSTGRESQL__USER: ${PG_USER:-authentik}
       AUTHENTIK_POSTGRESQL__NAME: ${PG_DB:-authentik}
       AUTHENTIK_POSTGRESQL__PASSWORD: ${PG_PASS}
-    # `user: root` and the docker socket volume are optional.
```
BeryJu, line 68: I still wish there was a better programatical way of adding these comments back in

rissson: We can just do f-string templating if we really want to keep those. Pretty ugly though :/

.github/actions/setup/action.yml
rissson: True, but at some point we might want to force `system` deps to be installed if `python` deps are installed for example

BeryJu: yeah I meant 1 action for each and in the main one include them conditionally but its also a bit of a hassle

rissson: It is yeah, let's keep it like this for now

rissson (approved):

GirlBossRush (approved):

PR #14927, opened 2025-06-05
providers/proxy: add option to override host header with property mappings

BeryJu:
## Details

REPLACE ME

---

## Checklist

If applicable

### DIFF
--- a/website/docs/add-secure-apps/providers/proxy/index.md
+++ b/website/docs/add-secure-apps/providers/proxy/index.md
@@ -152,3 +152,17 @@ return {
 ```
 
 Afterwards, edit the _Proxy provider_ and add this new mapping. The expression is only evaluated when the user logs into the application.
+
+## Host header:ak-version[2025.6.1]
+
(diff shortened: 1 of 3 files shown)


website/docs/add-secure-apps/providers/proxy/index.md
```diff
@@ -152,3 +152,17 @@ return {
 ```
 
 Afterwards, edit the _Proxy provider_ and add this new mapping. The expression is only evaluated when the user logs into the application.
+
+## Host header:ak-version[2025.6.1]
+
+By default, the proxy provider will use forward the Host header used by the client to the upstream application. Starting with authentik 2025.6.1, it is possible to dynamically adjust the Host header with a property mapping.
```
rissson: there is either one word too many, or one too few

BeryJu: how does one english again

rissson (approved):

PR #14929, opened 2025-06-05
root: remove /if/help

rissson:
## Details

size goes brr

---

## Checklist

If applicable

### DIFF
--- a/.github/workflows/ci-website.yml
+++ b/.github/workflows/ci-website.yml
@@ -84,6 +84,8 @@ jobs:
       attestations: write
     steps:
       - uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
       - name: Set up QEMU
         uses: docker/setup-qemu-action@v3.6.0
       - name: Set up Docker Buildx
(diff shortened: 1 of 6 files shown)


dominic-r:
Wouldn't this affect pages like https://authentik.company/if/admin/#/core/applications ? The help box seems to pull content from docs at first glance

dominic-r:
https://github.com/goauthentik/authentik/blob/main/web/src/admin/applications/ApplicationListPage.ts seems to import  `import MDApplication from "~docs/add-secure-apps/applications/index.md";` but it might work without build docs? idk

rissson:
Yes the build works without the docs assets included in the image

BeryJu (approved):
LGTM, cc @rissson for the docs container build/image pipeline

.github/workflows/ci-website.yml
```diff
@@ -72,3 +72,53 @@ jobs:
       - uses: re-actors/alls-green@release/v1
         with:
           jobs: ${{ toJSON(needs) }}
+  build-container:
```
rissson, line 44: Should we also add this before the ci-website-mark and require it?

BeryJu: could do, but it runs the same build script as what we already require so eh...although I guess we could replace that build script with the container build

PR #14957, opened 2025-06-08
policies: Optimize policy checking for static bindings

BeryJu:
## Details

Currently we spawn a sub process to check policies for every type of binding, regardless whether it's bound to a user, group or policy. Realistically we only need to do this for bindings bound to policies as user and group bindings are static.

This PR does that, checking any bindings bound to a user or group on a database level, skipping the rest of the policy engine

#### Computing policy result for a PBM with 1000 bindings all bound to groups:

`main`: 3000~ queries
Now: 2 queries

~~TODO: Needs some more test to ensure the behaviour with `Group.parent` is identical to the old behaviour~~

closes #14130

---

## Checklist

If applicable

### DIFF
--- a/authentik/core/api/applications.py
+++ b/authentik/core/api/applications.py
@@ -149,10 +149,10 @@ class ApplicationViewSet(UsedByMixin, ModelViewSet):
         return applications
 
     def _filter_applications_with_launch_url(
-        self, pagined_apps: Iterator[Application]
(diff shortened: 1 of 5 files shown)


rissson:
A bit out of scope but still related, could we add a help text to the order property mentioning that it only is the order in the the result of policies is considered, and not the order in which they are run?

PR #14983, opened 2025-06-09
web/admin: remove all special cases of slug handling, replace with a "smart slug" component

kensternberg-authentik:
web/components: Remove all special cases of slug handling, replace with a “smart slug” component

This commit removes all special handling for the `slug` attribute in our text. A variant of the text input control that can handle formatting-as-slugs has replaced all the slugificiation code; simply drop it onto a page and tell it the (must be unique) selector from which to get the data to be slugified. 

**Note**: The slug component only looks up one tier of the DOM, so be careful that both the text input and its slug accessory occupy the same DOM context.  It also expects them to be part of the same render pass, so that the two fields can find each other after `firstUpdated()`.

### The Component

Now that we know a (lot) more about Lit, this component has been slightly updated to meet our current standards.

- web/src/components/ak-slug-input.ts

Changes made:

- The “listen for the source object” has been moved to the `firstUpdated`, so that it no longer has to wait for the end of a render.  
- The `dirtyFlag` handler now uses the `@input` syntax.
- Updated the slug formatter to permit trailing dashes.
- Uses the `@bound` decorator, eliminating the need to do binding in the constructor (and so eliminating the constructor completely).

### Component uses:

The following components were revised to use `ak-slug-input` instead of a plain text input with the slug-handling added by our forms manager.

- applications/ApplicationForm.ts
- flows/FlowForm.ts
- sources/kerberos/KerberosSourceForm.ts
- sources/ldap/LDAPSourceForm.ts
- sources/oauth/OAuthSourceForm.ts
- sources/plex/PlexSourceForm.ts
- sources/saml/SAMLSourceForm.ts
- sources/scim/SCIMSourceForm.ts

### Remove the redundant special slug handling code

- web/src/elements/forms/Form.ts
- web/src/elements/forms/HorizontalFormElement.ts

### A special case among special cases

- web/src/admin/stages/invitation/InvitationForm.ts

This form is our one case where we have a slug input field with no corresponding text source. Adding a simple event handler to validate the value whenever it changed and write back a “clean” slug was the most straightforward solution. I added a help line; it seemed “surprising” to ask someone for a name and not follow the same rules as “names” everywhere else in our UI without explanation. @tanberry, I'd appreciate if you'd [validate that change](https://github.com/goauthentik/authentik/pull/14983/files#diff-c72d13fd459adc51b2beff0dd65ad0705944ff0d18bc906c6889c03dda5cdbb7) (line 62).


web/src/components/ak-slug-input.ts
```diff
@@ -1,42 +1,63 @@
...
+ * source.
+ *
+ * For the same reason, both the source text field and the slug field must share the same immediate
+ * parent DOM object.
+ *
+ * Since we expect the source text field and the slug to be part of the same form and rendered not
+ * just in the same form but in the same form group, these are not considered burdensome
+ * restrictions.
+ */
+
```
GirlBossRush: Nit: remove newline between JSDoc comment to allow VSCode’s not-so-Intellisense to associate it with the class.

web/src/components/ak-slug-input.ts
```diff
@@ -47,6 +68,7 @@ export class AkSlugInput extends HorizontalLightComponent<string> {
         }
     }
 
+    @bound
```
GirlBossRush, line 72: Nit: Use arrow method to bind. The debugger will have a shorter stack trace. Also use a private method prefix or mark the method with `public` or `protected`

web/src/components/ak-slug-input.ts
```diff
@@ -1,42 +1,63 @@
...
+    /**
+     * A selector indicating the (must be unique) source text input control.
+     */
     @property({ type: String })
     source = "";
 
+    @property({ type: String, reflect: true })
+    value = "";
+
     origin?: HTMLInputElement | null;
```
GirlBossRush: Nit: mark with private property prefix `#` or with `public` if we want to expose it

GirlBossRush (approved):
A much needed improvement on one of our thorniest input elements 👏 

I’ve made a few suggestions but nothing which should prevent this from going out. 😊

web/src/components/ak-slug-input.ts
kensternberg-authentik: I disagree, for the simple reason that arrow functions are a block on inheritability, since they're desugared to something that is not part of the prototype, but hidden (made private) inside the desugaring constructor.

PR #15020, opened 2025-06-12
website/docs: updates nginx doc

dewi-tik:
## Details

Updates the nginx proxy config based on problem reports.

---

## Checklist

If applicable

### DIFF
--- a/website/docs/add-secure-apps/providers/proxy/_nginx_standalone.md
+++ b/website/docs/add-secure-apps/providers/proxy/_nginx_standalone.md
@@ -1,4 +1,4 @@
-Create a `http_top.conf` file in your niginx custom directory (`data/nginx/custom/`) with the following content:
+Create a `http_top.conf` file in your nginx `conf.d` directory (`/nginx/conf.d/`) with the following content:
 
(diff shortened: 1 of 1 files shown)


rissson:
What exactly was the bug report for this change?

rissson (approved):

website/docs/add-secure-apps/providers/proxy/_nginx_standalone.md
```diff
@@ -1,10 +1,19 @@
+Create a `http_top.conf` file in your niginx custom directory (`data/nginx/custom/`) with the following content:
```
dominic-r:
```suggestion
Create a `http_top.conf` file in your nginx custom directory (`data/nginx/custom/`) with the following content:
```

dominic-r: Also, I might be missing something, but where do we get `data/nginx/` from instead of using the standard base directory of `/etc/nginx/`?

dewi-tik:
Further testing has shown an issue with this. Changing to draft for now

PR #15035, opened 2025-06-13
core: move files in database

rissson:
## Details

REPLACE ME

---

## Checklist

If applicable


rissson:
- doing things with a separate model is good
- put it in a new app
- for database storage, needs test/benchmarking for performance

existing and potential use cases:
- icons
- backgrounds
- csv reports
- pdf reports
- avatars
- rac session recording -> probably needs to go into s3 directly from the outpost whatever happens because of size

PR #15041, opened 2025-06-13
stages/authenticator_webauthn: add option to configure max attempts

BeryJu:
## Details

Some cleanup and add the option to limit the max webauthn registration attempts

---

## Checklist

If applicable


authentik/stages/authenticator_validate/challenge.py
```diff
@@ -36,10 +37,12 @@
 from authentik.stages.authenticator_sms.models import SMSDevice
 from authentik.stages.authenticator_validate.models import AuthenticatorValidateStage, DeviceClasses
 from authentik.stages.authenticator_webauthn.models import UserVerification, WebAuthnDevice
-from authentik.stages.authenticator_webauthn.stage import SESSION_KEY_WEBAUTHN_CHALLENGE
+from authentik.stages.authenticator_webauthn.stage import PLAN_CONTEXT_WEBAUTHN_CHALLENGE
 from authentik.stages.authenticator_webauthn.utils import get_origin, get_rp_id
 
 LOGGER = get_logger()
+if TYPE_CHECKING:
```
gergosimonyi, line 44: Not to lower the morale but can authentik switch to a real language please? :)))

authentik/stages/authenticator_webauthn/models.py
```diff
@@ -84,6 +84,8 @@ class AuthenticatorWebAuthnStage(ConfigurableStage, FriendlyNamedStage, Stage):
 
     device_type_restrictions = models.ManyToManyField("WebAuthnDeviceType", blank=True)
 
+    max_attempts = models.PositiveIntegerField(default=0)
```
gergosimonyi, line 87: `null=True` instead of treating 0 as Infinity?

(Then this can be a real `_Positive_IntegerField`)

gergosimonyi (commented):
Apart from webauthn registration currently being broken by #14460

screenshot

and `EmptyState` currently being broken by #14289 (resulting in no error being shown)

screenshot

this seems good 😄

authentik/stages/authenticator_webauthn/models.py
BeryJu: I would do that but the API schema then turns it into `number | undefined | null` which is annoying to deal with in the UI

PR #15046, opened 2025-06-14
fix(scim source): Make SCIM group and user handling fully SCIM 2.0 compliant

amritbrartech:
## Details

This PR enhances SCIM source functionality in Authentik to ensure full compliance with the SCIM 2.0 specification for both users and groups. Key changes include:

**SCIM 2.0 Compliance Enhancements:**
	•	Updated PATCH handling for groups and users to support SCIM 2.0 operations (add, replace, remove).
	•	Corrected schemas usage in SCIM payloads for both users and groups.
	•	Added support for PUT operations on group and user endpoints.
	•	Fixed error handling to return proper SCIM error responses with appropriate scimType.

**Group Handling Improvements:**
	•	Replaced incorrect schema in group SCIM response.
	•	Ensured members field is handled correctly during PATCH operations.
	•	Guarded against invalid UUID inputs for group IDs.

**User Handling Improvements:**
	•	Added PATCH endpoint with proper SCIM operation support (e.g., update userName).
	•	Normalized startIndex parsing and improved pagination response compliance.

**General Enhancements:**
	•	Introduced SCIMValidationError exception for standardized error messages.
	•	Included JSONParser alongside SCIMParser to support broader content types.
	•	Improved robustness of group_to_scim and user_to_scim by validating existence of internal objects.

---

## Checklist

If applicable

### DIFF
--- a/authentik/sources/scim/views/v2/groups.py
+++ b/authentik/sources/scim/views/v2/groups.py
@@ -135,6 +135,49 @@ class GroupsView(SCIMObjectView):
         return Response(self.group_to_scim(connection), status=200)
 
     @atomic
+    def patch(self, request: Request, group_id: str, **kwargs) -> Response:
+        """Patch group handler"""
+        connection = SCIMSourceGroup.objects.filter(
+            source=self.source, group__group_uuid=group_id
+        ).first()
+        if not connection:
+            raise Http404
+
+        # Start with current attributes
+        data = connection.attributes.copy() if hasattr(connection, "attributes") else {}
+
+        # SCIM PATCH expects a list of operations
+        operations = request.data.get("Operations", [])
+        for op in operations:
+            op_type = op.get("op", "").lower()
+            path = op.get("path", "")
+            value = op.get("value")
+
+            # Only supporting 'replace' for displayName and members for now
+            if op_type == "replace":
+                if path.lower() == "displayname":
+                    data["name"] = value
+                elif path.lower() == "members":
+                    data["members"] = value
+            elif op_type == "add" and path.lower() == "members":
+                # Add members to the current list
+                current = data.get("members", [])
+                if not isinstance(current, list):
+                    current = []
+                current.extend(value)
+                data["members"] = current
+            elif op_type == "remove" and path.lower() == "members":
+                # Remove members by value
+                current = data.get("members", [])
+                if not isinstance(current, list):
+                    current = []
+                remove_ids = set(m.get("value") for m in value)
+                data["members"] = [m for m in current if m.get("value") not in remove_ids]
+
+        connection = self.update_group(connection, data)
+        return Response(self.group_to_scim(connection), status=200)
+
+    @atomic
     def delete(self, request: Request, group_id: str, **kwargs) -> Response:
         """Delete group handler"""
         connection = SCIMSourceGroup.objects.filter(
--- a/authentik/sources/scim/views/v2/groups.py
+++ b/authentik/sources/scim/views/v2/groups.py
@@ -143,36 +143,46 @@ class GroupsView(SCIMObjectView):
         if not connection:
             raise Http404
 
-        # Start with current attributes
         data = connection.attributes.copy() if hasattr(connection, "attributes") else {}
+        current_members = {m.get("value") for m in data.get("members", []) if m.get("value")}
 
-        # SCIM PATCH expects a list of operations
         operations = request.data.get("Operations", [])
         for op in operations:
             op_type = op.get("op", "").lower()
             path = op.get("path", "")
             value = op.get("value")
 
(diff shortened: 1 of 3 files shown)


authentik/sources/scim/views/v2/groups.py
```diff
@@ -97,8 +126,7 @@ def update_group(self, connection: SCIMSourceGroup | None, data: QueryDict):
                     self.logger.warning("Invalid group member", exc=exc)
                     continue
                 query |= Q(uuid=member.value)
-            if query:
-                group.users.set(User.objects.filter(query))
+            group.users.set(User.objects.filter(query))
```
dominic-r, line 129: This now always runs a filter(query) even if query is an empty Q() (which matches everything). It could unintentionally set all users if members are empty. I'd add back `if query:`

authentik/sources/scim/views/v2/groups.py
```diff
@@ -133,6 +161,61 @@ def put(self, request: Request, group_id: str, **kwargs) -> Response:
...
+                if op_type == "replace":
+                    # Replace all members
+                    data["members"] = value if value is not None else []
+                    current_members = {m.get("value") for m in data["members"] if m.get("value")}
+                elif op_type == "add":
+                    # Add only new members
+                    if not isinstance(value, list):
+                        value = [value]
+                    for member in value:
+                        member_id = member.get("value")
```
dominic-r, line 200: This member is always a dict, but malformed input might send strings or nested formats. I'd suggest adding validation such as `if not isinstance(member, dict)` and  raise a  SCIMValidationError.

authentik/sources/scim/views/v2/groups.py
```diff
@@ -133,6 +161,61 @@ def put(self, request: Request, group_id: str, **kwargs) -> Response:
...
+        data = connection.attributes.copy() if hasattr(connection, "attributes") else {}
+        current_members = {m.get("value") for m in data.get("members", []) if m.get("value")}
+
+        operations = request.data.get("Operations", [])
+        for op in operations:
+            op_type = op.get("op", "").lower()
+            path = op.get("path", "")
+            value = op.get("value")
+
+            if path.lower() == "displayname":
```
dominic-r, line 182: Only lowercase path, but SCIM allows json path-like notation (`members[value eq "abc"]`) or nested fields.

authentik/sources/scim/views/v2/users.py
```diff
@@ -76,11 +93,18 @@ def get(self, request: Request, user_id: str | None = None, **kwargs) -> Respons
...
+            if start_index < 1:
+                start_index = 1
+        except (ValueError, TypeError):
+            start_index = 1
         return Response(
             {
                 "totalResults": page.paginator.count,
                 "itemsPerPage": page.paginator.per_page,
-                "startIndex": page.start_index(),
+                "startIndex": start_index,
```
dominic-r, line 107: uses user-supplied startIndex, not page.start_index() which reflects actual pagination result. I'd use page.start_index() to indicate the real start.

authentik/sources/scim/views/v2/users.py
```diff
@@ -128,15 +152,51 @@ def post(self, request: Request, **kwargs) -> Response:
...
+            raise Http404
+
+        data = connection.attributes.copy() if hasattr(connection, "attributes") else {}
+        operations = request.data.get("Operations", [])
+        for op in operations:
+            op_type = op.get("op", "").lower()
+            path = op.get("path", "")
+            value = op.get("value")
+
+            if path.lower() == "username" or path.lower() == "userName":
```
dominic-r, line 183: only userName is handled. SCIM clients may send emails, name.familyName, etc.

authentik/sources/scim/views/v2/users.py
```diff
@@ -128,15 +152,51 @@ def post(self, request: Request, **kwargs) -> Response:
...
+                if op_type in ["replace", "add"]:
+                    data["userName"] = value
+                elif op_type == "remove":
+                    data.pop("userName", None)
+            # Add more SCIM attribute handling as needed
```
dominic-r, line 188: hmm

authentik/sources/scim/views/v2/users.py
BeryJu: For this filter parsing we should really be using something like https://github.com/15five/scim2-filter-parser as it does a correct token -> AST parsing (that library seems to not have been updated in quite a while so we might want to vendor it)

authentik/sources/scim/views/v2/groups.py
BeryJu: For this filter parsing we should really be using something like https://github.com/15five/scim2-filter-parser as it does a correct token -> AST parsing (that library seems to not have been updated in quite a while so we might want to vendor it)

PR #15050, opened 2025-06-15
website/integrations: change nextcloud scope name to avoid confusion

Keksmo:
This PR changes the Nextcloud Docs by replacing the proposed scope name 'profile' to 'nextcloud'. The reason for this is, that the name 'profile' is very close to the default scope 'profiles', which had lead me (and probably many others) to hours of debugging even though the issue is very simple and can easily be avoided by renaming the scope, which eliminates this confusion.

## Checklist

If applicable

### DIFF
--- a/website/integrations/services/nextcloud/index.mdx
+++ b/website/integrations/services/nextcloud/index.mdx
@@ -81,7 +81,7 @@ If you want to control user storage and designate Nextcloud administrators, you 
     - **Create Scope Mapping**:
 
         - **Name**: `Nextcloud Profile`
-        - **Scope name**: `profile`
+        - **Scope name**: `nextcloud`
         - **Expression**:
 
             ```python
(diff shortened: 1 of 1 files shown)


tanberry:
Thanks @Keksmo for this contribution! Although authentik doesn't really show "profiles" scope as a default (we don't suggest one at all), "profiles" is a common OpenID Connect scope ... so I can understand why it's potentially confusing for us to have suggested "profile" as the scope name! Thank you for clarifying this point in our docs.

tanberry:
Thanks @Keksmo for this contribution! Although authentik doesn't really show "profiles" scope as a default (we don't suggest one at all), "profiles" is a common OpenID Connect scope ... so I can understand why it's potentially confusing for us to have suggested "profile" as the scope name! Thank you for clarifying this point in our docs.

tanberry:
Thanks @Keksmo for this contribution! Although authentik doesn't really show "profiles" scope as a default (we don't suggest one at all), "profiles" is a common OpenID Connect scope ... so I can understand why it's potentially confusing for us to have suggested "profile" as the scope name! Thank you for clarifying this point in our docs.

tanberry (approved):
Thanks again @Keksmo and sorry about the lost hours debugging... :-(

PR #15076, opened 2025-06-16
website: split integrations partially

BeryJu:
## Details

Partial split of the integrations build:
- Same repo, same package
- Different docusaurus config
- Allows us to always deploy this from the main branch and deploy docs from version-xxxx branch

---

## Checklist

If applicable


dominic-r:
Should a `netlify.toml` rule also get added to redirect old links from https://docs.goauthentik.io/integrations/services/service_name to https://integrations.goauthentik.io/whatever-path-it-would-be/service_name ?

tanberry:
Thanks for your work on this... of course a few questions:

-- It looks like we would have to start using full/absolute paths when linking between integrations and regular docs (both ways). What are the downsides of that? Dominic raises a good question: does that mean our link-checker would not work?
-- what other changes might there be for both our contributors and us? 
-- Will this impact people who have bookmarked certain URLs?

BeryJu:
@dominic-r Indeed it does, added!

> Thanks for your work on this... of course a few questions:
> 
> -- It looks like we would have to start using full/absolute paths when linking between integrations and regular docs (both ways). What are the downsides of that? Dominic raises a good question: does that mean our link-checker would not work? -- what other changes might there be for both our contributors and us? -- Will this impact people who have bookmarked certain URLs?

@tanberry 
- Link checker: it wouldn't change, it already doesn't work great since even without this PR they are two docs plugin instances that can't check links between each other
- No changes for contributors aside from having to use absolute links
- Nope, will be redirected

website/netlify.toml
```diff
@@ -51,6 +51,11 @@
   to = "/docs/sources/:splat"
   status = 302
 
+[[redirects]]
+  from = "/integrations*"
+  to = "http://integrations.goauthentik.io:splat"
```
dominic-r:
```suggestion
  to = "https://integrations.goauthentik.io:splat"
```

BeryJu: d'oh

tanberry (approved):
Whoot!!

dominic-r:
Out of curiosity, how would the deploy previews work now for integrations? Will there just be another comment/job by the Netlify bot?

BeryJu:
> Out of curiosity, how would the deploy previews work now for integrations? Will there just be another comment/job by the Netlify bot?

@dominic-r yeah pretty much, different URL to click on but same workflow

rissson (approved):

Issue #15141, opened 2025-06-19
Error messages for "no such objects" cases no longer appear in the ui


dominic-r:
Was probably introduced in https://github.com/goauthentik/authentik/pull/14289/files#diff-d7e2cfdbffbee3580701b2c4f0ee7fb5c6288a338923058ef1708f1517d4e28c

When visiting pages where you don't have any objects configured, the usual message underneath the icon is gone. For example, when visiting the initial permissions page and you have no initial permissions, you'll only see the icon.

Repro:
1. Launch a dev build of authentik ATM, I'm on SHA 2d480bffb493f036286d60ceb4f337c02f6eda85
2. Navigate to the admin interface and go to Directory > Initial permissions.
3. If you have no initial permissions, notice how the help text is missing.

Some examples: https://files.slack.com/files-tmb/T047V3272M9-F091Y4BR7RU-998628b874/k__p_720.png https://files.slack.com/files-pri/T047V3272M9-F092H43UA57/k__p.png

![Image](https://github.com/user-attachments/assets/72fef09a-2959-4f15-9b3d-8ee872e41163)

![Image](https://github.com/user-attachments/assets/a9a32497-3dc3-46b8-b310-2f7db271d138)

**Describe the bug**

**To Reproduce**

**Expected behavior**

**Screenshots**

**Logs**
Output of docker-compose logs or kubectl logs respectively

**Version and Deployment (please complete the following information):**

-   authentik version: [e.g. 2025.2.0]
-   Deployment: [e.g. docker-compose, helm]

**Additional context**

PR #15152, opened 2025-06-19
web/elements/empty-state: Fix issues with EmptyState and Loading Overlay

kensternberg-authentik:
web/bug/empty-state: Fix issues with EmptyState and Loading Overlay

## What

- Add a method, `hasSlotted()`, to the Base component.
- Revise `EmptyState` to use `hasSlotted()`.
- Revise `LoadingOverlay` to use `hasSlotted()`.
- Provide (hopefully complete) Storybook stories for both
- Revise use of these components throughout the codebase.

### Adding `hasSlotted()` to the Base component.

The essential problem here was mine: I misunderstood what the Patternfly `SlotController` does (and, yikes, how it does it). Slots aren’t magical; they’re just named containers, in which lightDOM elements that appear between the opening and closing tags of a web component can be strategically placed, shown or hidden, and to some extent styled, within the rendered and visible results of the shadowDOM component that will fill the browser’s RECT allocated to that component.

SlotController tries to associate the template with slots by creating the shadowDOM *first*, then working backwards to see if there are lightDOM components to put into those slots. That’s not what we want; we want to see if there are lightDOM components that meet our slot requirements and, if there are, create corresponding slots for them.

That’s what `hasSlotted()` does: it returns true or false to the question, “Is there currently in the lightDOM for this component an entry requesting a known slot name?” Components are free to do what they want with that knowledge.

### Revise `EmptyState` to use `hasSlotted()`

`<ak-empty-state>` now has several modes, all well-documented in the Storybook story. But in short, the Title is now a default slot; any HTML Elements not sent to one of the named slots are put into the Title. The two named slots are `body` and `primary`. The header is bold and large; body is just text, and primary is boxed to indicate that one or more buttons should be placed there, to allow interaction.

The extra modes are controlled by boolean attributes:

- `loading`: Shows the loading spinner, overriding the `icon` attribute
- `default`: Shows the loading spinner *and* the word “Loading” (i18n-aware).

The priority for all of these is:

- Has something in the default (header) slot: That text will be shown. Overrides both
- `default` overrides `loading`
- `loading`

### Revise `LoadingOverlay` to use `hasSlotted()`

`<ak-loading-overlay>` is a specialized variant of `<ak-empty-state>` over what will become `<ak-backdrop>`, but for now is just internal. It allows only for the heading and primary slots, forwarding them `<ak-empty-state>`. Since this is literally the *Loading*Overlay, showing the `loading` spinner is the default; to prevent it, pass `no-spinner` as an attribute.

If applicable


web/src/admin/admin-overview/cards/RecentEventsCard.ts
```diff
@@ -89,7 +89,7 @@ export class RecentEventsCard extends Table<Event> {
```
kensternberg-authentik, line 89: Only a few files are significant changes.  Most are like this: either revising the slot name, revising the attribute collection, or `prettier` having new opinions using the new configuration.

web/src/admin/applications/entitlements/ApplicationEntitlementPage.ts
```diff
@@ -118,13 +118,12 @@ export class ApplicationEntitlementsPage extends Table<ApplicationEntitlement> {
...
-            html`<ak-empty-state
-                header=${msg("No app entitlements created.")}
-                icon="pf-icon-module"
-            >
+            html`<ak-empty-state icon="pf-icon-module"
+                ><span>${msg("No app entitlements created.")}</span>
+
                 <div slot="body">
                     ${msg(
-                        "This application does currently not have any application entitlement defined.",
```
kensternberg-authentik, line 127: Zut alors! A minor fix found while reviewing the work.

web/src/elements/Base.ts
```diff
@@ -148,5 +148,31 @@ export class AKElement extends LitElement implements AKElementProps {
         return this.#styleRoot;
     }
 
+    protected hasSlotted(name: string | null) {
```
kensternberg-authentik, line 151: First big change: `hasSlotted()` added to base.  This is primitive; it doesn't do much with detecting changes or notifying the component if the lightDOM is fiddled with.  Right now, there are no use cases that call for such detection, so I'm going with YAGNI until then. I don't think this implementation results in YCGTFH (You Can't Get There From Here) if we need to do so in the future.

web/src/elements/EmptyState.ts
```diff
@@ -50,32 +75,53 @@ export class EmptyState extends AKElement implements IEmptyState {
         ];
     }
 
+    willUpdate() {
```
kensternberg-authentik, line 78: Magical default slot content!

web/src/elements/LoadingOverlay.ts
```diff
@@ -5,30 +5,59 @@ import { spread } from "@open-wc/lit-helpers";
...
+ * @slot body - Descriptive text explaining the loading state
+ *
+ */
 @customElement("ak-loading-overlay")
 export class LoadingOverlay extends AKElement implements ILoadingOverlay {
     // Do not camelize: https://www.merriam-webster.com/dictionary/topmost
     @property({ type: Boolean, attribute: "topmost" })
     topmost = false;
 
-    @property({ type: Boolean })
```
kensternberg-authentik, line 21: _Sigh_ Boolean properties [should never be `true` by default](https://www.w3.org/TR/2008/WD-html5-20080610/semantics.html#boolean); doing so make it impossible to declaratively set them to `false`. Besides, `no-spinner` is the _much more infrequent_ use of this component.

web/src/elements/stories/EmptyState.stories.ts
```diff
@@ -1,108 +1,254 @@
 import type { Meta, StoryObj } from "@storybook/web-components";
```
kensternberg-authentik, line 1: Look! Look! _Lots_ of documentation!

web/src/elements/stories/LoadingOverlay.stories.ts
```diff
@@ -1,74 +1,154 @@
 import type { Meta, StoryObj } from "@storybook/web-components";
```
kensternberg-authentik, line 1: Even more documentation!  (The `mdx` files were rolled into this using the Storybook `autodocs` feature.)

kensternberg-authentik (commented):
Added guideposts!

dominic-r:
Closes https://github.com/goauthentik/authentik/issues/15141 and https://github.com/goauthentik/authentik/pull/15142

BeryJu (approved):

PR #15235, opened 2025-06-25
blueprints: add JSON tag to parse JSON from string

BeryJu:
## Details

Yep another one

---

## Checklist

If applicable


rissson:
Should we call it ParseJSON instead? So we can then add a RenderJSON tag in the future 👀

rissson (approved):

PR #15247, opened 2025-06-25
sources/oauth: change azure ad to entra ID

dewi-tik:
## Details

Updates the UI facing term from azure ad to entra ID. Also removes the prepopulated URL from the well known OIDC field because it isn't required and can override the profile URL which causes issues.

---

## Checklist


authentik/sources/oauth/types/azure_ad.py
```diff
@@ -73,9 +73,7 @@ class AzureADType(SourceType):
     authorization_url = "https://login.microsoftonline.com/common/oauth2/v2.0/authorize"
     access_token_url = "https://login.microsoftonline.com/common/oauth2/v2.0/token"  # nosec
     profile_url = "https://graph.microsoft.com/v1.0/me"
-    oidc_well_known_url = (
-        "https://login.microsoftonline.com/common/.well-known/openid-configuration"
-    )
+    oidc_well_known_url = " "
```
dominic-r:
```suggestion
    oidc_well_known_url = None
```

or

```suggestion
    oidc_well_known_url = ""
```

would probably be more appropriate, but have you verified that any tests/code logic is not broken by this change? Also, if it's not too much work we could consider updating function and file names to Entra for consistency. I can open a followup PR if you wish

dewi-tik: I'll try your first suggestion later today.

The second suggestion causes an issue where the field disappears from the UI. 

Tests fail on this at the moment. I'll have a look later today or potentially someone from backend team can pick that up. 

Jens suggested that updating all references to Entra ID was too much work for the time being.

PR #15252, opened 2025-06-26
website/docs: add hint that flows need cookies

lemoer:
The executor itself does not set a session cookie, but requires one to be set before it works. This took me days to figure out, so maybe this will be helpful to somebody in the future.

---

## Checklist

If applicable

### DIFF
--- a/website/docs/developer-docs/api/flow-executor.md
+++ b/website/docs/developer-docs/api/flow-executor.md
@@ -11,7 +11,7 @@ Because the flow executor stores its state in the HTTP Session, so you need to e
 :::
 
 :::info
-Note that the HTTP Session must be obtained as cookie before GET /api/v3/flows/executor/:slug can be called. If you are using a JWT for authentication, you first have to obtain a session cookie via GET /api/v3/flows/instances/:slug/execute/ before requesting GET /api/v3/flows/executor/:slug.
+Note that the HTTP session must be obtained as a cookie before `GET /api/v3/flows/executor/:slug` can be called. If you are using a JWT for authentication, you first have to obtain a session cookie via `GET /api/v3/flows/instances/:slug/execute/` before requesting `GET /api/v3/flows/executor/:slug`.
 :::
 
 The main endpoint for flow execution is `/api/v3/flows/executor/:slug`.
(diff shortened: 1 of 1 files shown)


website/docs/developer-docs/api/flow-executor.md
```diff
@@ -10,6 +10,10 @@ However, any flow can be executed via an API from anywhere, in fact that is what
 Because the flow executor stores its state in the HTTP Session, so you need to ensure that cookies between flow executor requests are persisted.
 :::
 
+:::info
+Note that the HTTP Session must be obtained as cookie before GET /api/v3/flows/executor/:slug can be called. If you are using a JWT for authentication, you first have to obtain a session cookie via GET /api/v3/flows/instances/:slug/execute/ before requesting GET /api/v3/flows/executor/:slug.
```
dewi-tik:
```suggestion
Note that the HTTP session must be obtained as a cookie before `GET /api/v3/flows/executor/:slug` can be called. If you are using a JWT for authentication, you first have to obtain a session cookie via `GET /api/v3/flows/instances/:slug/execute/` before requesting `GET /api/v3/flows/executor/:slug`.
```

dewi-tik (dismissed):
Thanks for the suggestion @lemoer

rissson (approved):

tanberry (approved):
Thanks so much @lemoer for adding this important detail!

dewi-tik (approved):

lemoer:
It turns out that this only seems to work if the user is an admin user. Does someone have another idea how I can exchange a JWT into an http session cookie?

tanberry:
> It turns out that this only seems to work if the user is an admin user. Does someone have another idea how I can exchange a JWT into an http session cookie?

@lemoer do we need to add a note to this doc to state that, that it only applies to Admins? As for your questions about converting a JWT, you could also ask that over on Discord.

lemoer:
@tanberry Thank you. I asked in the Discord and was advised to create a bug report, which I did in #15747. If this really turns out to be a bug and the bug is fixed, I think, we can remove the documentation hint again, which we added in this PR. Then it would have turned out, that this was only a work around for the bug.

Issue #2023, opened 2021-12-27
Logging out from authentik should invalidate sessions of proxied applications


rperpe:
**Is your feature request related to a problem? Please describe.**
If users logout from authentik they expect that they're logged out from proxied applications (using embedded outpost and integrated proxy provider) too - without GETting `/akprox/sign_out` on every single application. It's a security risk if users don't consider this.

**Describe the solution you'd like**
As soon as `https://authentik.tld/flows/-/default/invalidation/` is called, all sessions on proxied applications must be invalidated immediately (probably via OIDC back-channel logout?).
Same is expected if forward auth proxy provider is used (assuming there exists the same issue).

**Describe alternatives you've considered**
Are there any alternatives/workarounds (instructing users to manually logout from every single application isn't considered as a workaround)?

**Additional context**
N/A

PS: Many thanks to @BeryJu for his hard work and sharing authentik on GitHub!

xpufx:
I do agree that it's a security issue when the user logs out of one app and calls it a day not realizing all it would take to log back in is to go to the login page again. Same for other apps they haven't even considered the fate of.

However it would also confuse a lot of users when they log out of their webmail account and now their kanban software is logged out too.

The other thing is, the applications need to play along with this as well. Not all apps allow you to change what happens when you hit that log out button. Authentik needs to know when it happens for it to be able to do something, no?

This doesn't look like as straight forward as it might seem at first. If it were to be implemented somehow, I'd like the behavior to be configurable.

BeryJu:
I agree that this should at least be configurable, if not default. There are two reasons why it isnt currently possible:

- authentik doesnt support OAuth Backchannel Logout, which is how authentik would tell an outpost that a user logged out.

This isn't the case because relatively few applications and libraries support it, so I haven't gotten around to implementing it

- The authentik Outpost can't remove sessions in the background

Because the way sessions work with gorilla/sessions, the outpost can only delete a session in the context of an HTTP request, and can only delete the session of that specific HTTP request. I briefly looked for an alternative library but didnt find anything with the required features (ability to remove sessions without HTTP request, Redis and FS Storage)

Ornias1993:
I see some confusion about SLO (Single Log Out)
Lets seperate a few things:
1. Proxied applications and apps using forwardAuth with multiple authentik applications defined
2. SAML/OIDC
3. LDAP

1. It should be possible to invalidate other authentik applications, when one of them is logged out. forwardAuth definately  supports this.

[...]

PR #15401, opened 2025-07-04
providers/oauth2: backchannel logout

### LINKED ISSUE #12710: OIDC backchannel logout
There is currently no support for OIDC backchannel logout in Authentik. This causes application sessions to be kept active even though the user has signed out from Authentik.  

Support for the [OpenID Connect back-channel logout](https://openid.net/specs/openid-connect-backchannel-1_0.html) spec. 
This should be incorporated as a optional configuration parameter per application.  

For applications without a backchannel logout specified, or if the backchannel request failed, it would be nice if Authentik could notify the user that they might still be logged in. 

A temporary workaround here is configuring the token timeout low. 

**Additional context**
There also exists a [OIDC front-channel logout](https://openid.net/specs/openid-connect-frontchannel-1_0.html) method, however this is not used as often.

melizeche:
Implements OpenID Connect Back-Channel Logout 1.0 (https://openid.net/specs/openid-connect-backchannel-1_0.html) for automatic session termination across federated OAuth2 applications

TODO:

- [x] add tests
- [x] UI
- [x] session termination logic
- [x] remove view and URL, that could go in the Oauth2 source
- [x] testing with Keycloak as RP
- [x] testing with real applications (Matrix, etc)
- [x] ~manage multiple logout URIs~
- [x] use static single logout URI
- [x] Refactor to use Dramatiq instead of celery

closes #12710
docs #15845
related #2023

 
## UI
image
image
image

---

## Checklist

If applicable


authentik/flows/stage.py
```diff
@@ -301,6 +302,28 @@ def get_challenge(self, *args, **kwargs) -> Challenge:
...
+
+                # Send back-channel logout notifications
+                if session:
+                    send_backchannel_logout_notification(session=session)
+                else:
+                    send_backchannel_logout_notification(user=self.request.user)
+            except Exception as exc:
+                # Log the error but don't fail the logout process
+                self.logger.warning(
+                    "Failed to send backchannel logout notifications",
```
dewi-tik:
```suggestion
                    "Failed to send back-channel logout notifications",
```

authentik/providers/oauth2/tasks.py
```diff
@@ -0,0 +1,169 @@
...
+
+    # Generate the logout token
+    try:
+        logout_token = create_logout_token(provider, session_id, sub)
+    except Exception as exc:
+        LOGGER.warning("Failed to create logout token", exc=exc)
+        return False
+
+    # Get the back-channel logout URI from the provider's dedicated backchannel_logout_uris field
+    # Backchannel logout requires explicit configuration - no fallback to redirect URIs
```
dewi-tik:
```suggestion
    # Back-channel logout requires explicit configuration - no fallback to redirect URIs
```

web/src/admin/providers/oauth2/OAuth2ProviderFormForm.ts
```diff
@@ -192,6 +209,23 @@ export function renderForm(
                     </ak-array-input>
                     ${redirectUriHelp}
                 </ak-form-element-horizontal>
+                <ak-form-element-horizontal
+                    label=${msg("Backchannel Logout URIs")}
```
dewi-tik:
```suggestion
                    label=${msg("Back-Channel Logout URIs")}
```

web/src/admin/providers/oauth2/OAuth2ProviderViewPage.ts
```diff
@@ -244,6 +244,24 @@ export class OAuth2ProviderViewPage extends AKElement {
                                     </div>
                                 </dd>
                             </div>
+                            <div class="pf-c-description-list__group">
+                                <dt class="pf-c-description-list__term">
+                                    <span class="pf-c-description-list__text"
+                                        >${msg("Backchannel Logout URIs")}</span
```
dewi-tik:
```suggestion
                                        >${msg("Back-Channel Logout URIs")}</span
```

web/src/admin/providers/oauth2/OAuth2ProviderFormForm.ts
```diff
@@ -115,6 +119,19 @@ export const redirectUriHelp = html`${redirectUriHelpMessages.map(
     (m) => html`<p class="pf-c-form__helper-text">${m}</p>`,
 )}`;
 
+const backchannelLogoutUriHelpMessages = [
+    msg(
+        "URIs to send backchannel logout notifications to when users log out. Required for OpenID Connect Back-Channel Logout functionality.",
```
dewi-tik:
```suggestion
        "URIs to send back-channel logout notifications to when users log out. Required for OpenID Connect Back-Channel Logout functionality.",
```

dewi-tik (commented):
Minor suggestions. Just hyphenating backchannel in user facing locations so that it matches the OIDC specs. Not essential.

authentik/flows/stage.py
```diff
@@ -301,6 +302,28 @@ def get_challenge(self, *args, **kwargs) -> Challenge:
                     "flow_slug": self.request.brand.flow_invalidation.slug,
                 },
             )
+
+        # Trigger back-channel logout notifications if the user is authenticated
```
rissson: This should be done via a signal like https://github.com/goauthentik/authentik/blob/main/authentik/providers/oauth2/signals.py#L8

authentik/providers/oauth2/models.py
```diff
@@ -193,9 +196,13 @@ class OAuth2Provider(WebfingerProvider, Provider):
         default=generate_client_secret,
     )
     _redirect_uris = models.JSONField(
-        default=dict,
```
rissson, line 196: This was changed by @BeryJu recently for api schema consistency? I'm not entirely sure about the reason

authentik/providers/oauth2/tasks.py
```diff
@@ -0,0 +1,169 @@
...
+    from authentik.providers.oauth2.models import AccessToken, RefreshToken
+
+    # Get all access tokens (including expired) and refresh tokens for this user or session
+    provider_pks = set()
+
+    if session:
+        # Get providers from access tokens (covers all OAuth2 flows)
+        access_tokens = AccessToken.objects.filter(session=session)
+        for token in access_tokens:
+            provider_pks.add(token.provider.pk)
```
rissson:
```suggestion
        access_tokens = AccessToken.objects.select_related("provider").filter(session=session)
        for token in access_tokens:
            provider_pks.add(token.provider.pk)
```
or 
```suggestion
        access_tokens = AccessToken.objects.filter(session=session)
        for token in access_tokens:
            provider_pks.add(token.provider_id)
```
I prefer the first one, even if less efficient, it's more readable.

Also applies to the other ones below.

authentik/providers/oauth2/urls.py
```diff
@@ -48,6 +49,11 @@
         EndSessionView.as_view(),
         name="end-session",
     ),
+    path(
+        "<slug:application_slug>/backchannel_logout/",
+        BackChannelLogoutView.as_view(),
+        name="backchannel-logout",
+    ),
```
rissson: Maybe I'm misunderstanding, but why do we need an endpoint here? Isn't it us sending the requests to the applications?

rissson (changes requested):

authentik/providers/oauth2/urls.py
melizeche: From my understanding of the specification there are 3 scenarios for the backchannel logout:
* User logs out from the dashboard: we notify all OAuth2 apps
* Session expires/terminates: we notify all OAuth2 apps
* User logs out from an OAuth2 app → App sends backchannel logout to authentik: we notify all other apps

We need this endpoint for the 3rd scenario

https://openid.net/specs/openid-connect-backchannel-1_0.html#BCRegistration
https://openid.net/specs/openid-connect-backchannel-1_0.html#BCRequest

rissson: Both requests you linked are made from authentik to the RP.

We also don't want a "random" app to be able to log out users in authentik. This endpoint would however be useful for oidc sources, but that's out of scope for this PR.

authentik/providers/oauth2/api/providers.py
```diff
@@ -43,6 +43,9 @@ class OAuth2ProviderSerializer(ProviderSerializer):
     """OAuth2Provider Serializer"""
 
     redirect_uris = RedirectURISerializer(many=True, source="_redirect_uris")
+    backchannel_logout_uris = RedirectURISerializer(
+        many=True, source="_backchannel_logout_uris", required=False
+    )
```
rissson: Why would those need to be the same as redirect URIs? If I understand correctly, they should only be static URLs

authentik/providers/oauth2/tasks.py
```diff
@@ -0,0 +1,143 @@
...
+
+    # Get providers from access tokens (covers all OAuth2 flows)
+    access_tokens = AccessToken.objects.select_related("provider").filter(session=session)
+    for token in access_tokens:
+        # Send back-channel logout notifications to all tokens
+        # for provider_pk in provider_pks:
+        send_backchannel_logout_request.delay(
+            provider_pk=token.provider.pk,
+            iss=token.id_token.iss,
+            session_id=session.session.session_key,
```
rissson: Let's avoid passing the session key as arguments to the task.

Also, this causes an extra query to happen.

authentik/providers/oauth2/views/backchannel_logout.py
```diff
@@ -0,0 +1,228 @@
+"""OAuth2 Provider Back-Channel Logout Views"""
```
rissson: As discussed, this view should not be here.

authentik/providers/oauth2/views/provider.py
```diff
@@ -72,6 +72,14 @@ def get_info(self, provider: OAuth2Provider) -> dict[str, Any]:
             "device_authorization_endpoint": self.request.build_absolute_uri(
                 reverse("authentik_providers_oauth2:device")
             ),
+            "backchannel_logout_uri": self.request.build_absolute_uri(
+                reverse(
+                    "authentik_providers_oauth2:backchannel-logout",
+                    kwargs={"application_slug": provider.application.slug},
+                )
+            ),
```
rissson: As discussed, this should not be here.

rissson (changes requested):

authentik/providers/oauth2/api/providers.py
melizeche: Initially we brainstormed this with Jens, but now it seems unnecessary. I'll change it to a static URL 👍

authentik/providers/oauth2/signals.py
```diff
@@ -1,13 +1,29 @@
...
 
 from authentik.core.models import AuthenticatedSession, User
 from authentik.providers.oauth2.models import AccessToken, DeviceToken, RefreshToken
+from authentik.providers.oauth2.tasks import send_backchannel_logout_notification
+
+LOGGER = get_logger()
 
 
 @receiver(pre_delete, sender=AuthenticatedSession)
 def user_session_deleted_oauth_tokens_removal(sender, instance: AuthenticatedSession, **_):
```
rissson: Maybe update this signal name

authentik/providers/oauth2/signals.py
```diff
@@ -1,13 +1,29 @@
...
     """Revoke tokens upon user logout"""
+    LOGGER.debug("Sending back-channel logout notifications signal!", session=instance)
+    try:
+        send_backchannel_logout_notification(session=instance)
```
rissson: This should be a task, no? I'm not sure we want to delay the logout process by doing a bunch of db queries and iterating on those results to create tasks synchronously.

rissson (commented):
I have a bunch more comments about tasks, but they mainly apply to dramatiq, so I'll hold off until we convert this to use dramatiq.

authentik/providers/oauth2/models.py
BeryJu: should be `default=list`

rissson: Perfect, thanks!

authentik/providers/oauth2/signals.py
melizeche: the problem is that we need the sessions to exist, so this needs to be run before the deletion in the following lines

authentik/providers/oauth2/signals.py
```diff
@@ -1,17 +1,35 @@
...
+    LOGGER.debug("Sending back-channel logout notifications signal!", session=instance)
+
+    access_tokens = AccessToken.objects.filter(
         user=instance.user,
         session__session__session_key=instance.session.session_key,
-    ).delete()
+    )
+
+    backchannel_logout_notification_dispatch.send(
+        sender=None,
```
rissson:
```suggestion
```

authentik/providers/oauth2/tasks.py
```diff
@@ -0,0 +1,126 @@
...
+    if not backchannel_logout_uri:
+        self.info(
+            "No back-channel logout URI found for provider",
+            provider=provider.name,
+            client_id=provider.client_id,
+        )
+        return
+
+    # Send the back-channel logout request
+    response = get_http_session().post(
```
rissson, line 47: Let's make sure this follows redirects.

authentik/providers/oauth2/tasks.py
```diff
@@ -0,0 +1,126 @@
...
+    for revocation in revocations:
+        provider_pk, iss, sub = revocation
+        provider = OAuth2Provider.objects.filter(pk=provider_pk).first()
+        send_backchannel_logout_request.send_with_options(
+            args=(provider_pk, iss, sub),
+            rel_obj=provider,
+        )
+
+
+def send_backchannel_logout_notification(session: AuthenticatedSession = None) -> None:
```
rissson: This function should be removed, no?

authentik/providers/oauth2/tasks.py
```diff
@@ -0,0 +1,126 @@
...
+    response.raise_for_status()
+
+    self.info(
+        "Back-channel logout successful",
+        provider=provider.name,
+        client_id=provider.client_id,
+        sub=sub,
+    )
+    Event.new(
+        "backchannel_logout",
```
rissson: This is supposed to be an `authentik.events.models.EventAction` I think. Do we really need to create an event for all of these?

authentik/providers/oauth2/tasks.py
```diff
@@ -0,0 +1,126 @@
...
+        data={"logout_token": logout_token},
+        headers={"Content-Type": "application/x-www-form-urlencoded"},
+    )
+    response.raise_for_status()
+
+    self.info(
+        "Back-channel logout successful",
+        provider=provider.name,
+        client_id=provider.client_id,
+        sub=sub,
```
rissson:
```suggestion
        sub=sub,
```
The provider is attached to the task and shown in the UI, no need to add it here.

authentik/providers/oauth2/tasks.py
```diff
@@ -0,0 +1,126 @@
...
+
+    # Get the back-channel logout URI from the provider's dedicated backchannel_logout_uri field
+    # Back-channel logout requires explicit configuration - no fallback to redirect URIs
+
+    backchannel_logout_uri = provider.backchannel_logout_uri
+    if not backchannel_logout_uri:
+        self.info(
+            "No back-channel logout URI found for provider",
+            provider=provider.name,
+            client_id=provider.client_id,
```
rissson:
```suggestion
```
The provider is attached to the task and shown in the UI, no need to add it here.

authentik/providers/oauth2/tests/test_backchannel_logout.py
```diff
@@ -0,0 +1,286 @@
...
+        # Test case 3: With both session_id and sub
+        token3 = self._create_logout_token(session_id=session_id, sub=sub)
+        decoded3 = self._decode_token(token3)
+
+        self.assertEqual(decoded3["sid"], session_id)
+        self.assertEqual(decoded3["sub"], sub)
+        self.assertIn("events", decoded3)
+
+    @patch("authentik.providers.oauth2.tasks.get_http_session")
+    @patch("authentik.providers.oauth2.tasks.CurrentTask.get_task")
```
rissson: this shouldn't be needed. You can do `send_backchannel_logout_request.send(*args, **kwargs)` and the task will run properly in the correct environment for CurrentTask.get_task to work

authentik/providers/oauth2/utils.py
```diff
@@ -211,3 +211,37 @@ def __init__(
...
+    iss: str, provider: OAuth2Provider, session_key: str = None, sub: str = None
+) -> str:
+    """Create a logout token for Back-Channel Logout
+
+    As per https://openid.net/specs/openid-connect-backchannel-1_0.html
+    """
+    import uuid
+    from time import time
+
+    from authentik.providers.oauth2.id_token import hash_session_key
```
rissson: Let's avoid runtime imports.

web/src/admin/providers/oauth2/OAuth2ProviderViewPage.ts
```diff
@@ -355,6 +369,18 @@ export class OAuth2ProviderViewPage extends AKElement {
                         </form>
                     </div>
                 </div>
+                <div
+                    class="pf-c-card pf-l-grid__item pf-m-12-col pf-m-12-col-on-xl pf-m-12-col-on-2xl"
+                >
+                    <div class="pf-c-card pf-l-grid__item pf-m-12-col-on-2xl">
+                        <div class="pf-c-card__title">${msg("Tasks")}</div>
+                        <ak-task-list
```
rissson, line 378: You need to `import "#elements/tasks/TaskList";`

rissson (changes requested):

rissson (approved):
Merge at will!

GirlBossRush (approved):

PR #15431, opened 2025-07-07
root: add npm install to set up authentik Typescript API

PeshekDotDev:
## Details

When following through [the steps to set up local development](https://docs.goauthentik.io/docs/developer-docs/setup/full-dev-environment), the developer is asked to run:

```sh
make install
make gen-dev-config
make migrate
make gen
```

`make gen` tries to run `npm link`, but the `make install` step never handles running `npm install` in the root of the directory, so `npm link` does not work.

This PR adds `npm install` to the `core-install` step ran by `make install`, fixing the issue

---

## Checklist

If applicable

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -151,7 +151,7 @@ gen-client-ts: gen-clean-ts  ## Build and install the authentik API for Typescri
 		--git-repo-id authentik \
 		--git-user-id goauthentik
 
-	cd ${PWD}/${GEN_API_TS} && npm link
+	cd ${PWD}/${GEN_API_TS} && npm install && npm link
 	cd ${PWD}/web && npm link @goauthentik/api
 
 gen-client-py: gen-clean-py ## Build and install the authentik API for Python


rissson:
We probably should add `make web-install` to the install docs, or make it a dependency of `core-install`

rissson:
Ah no yeah, it's the one in the root folder. cc. @GirlBossRush

Makefile
```diff
@@ -49,6 +49,7 @@ lint: ## Lint the python and golang sources
 	golangci-lint run -v
 
 core-install:
+	npm install
```
```suggestion
core-install: web-install
```

PeshekDotDev:
Apologies, by running `npm install` in the root of the project, we we're giving `gen-ts-api` access to the `tsc` node_module that it's missing during the `gen-client-ts` step. I am making an update to the PR that runs `npm install` in the `gen-ts-api` folder instead of in the root

Makefile
PeshekDotDev: The `web-install` is ran as part of the `make install` command already, but I get the intent here as I had the same confusion debugging this. I put the `npm install` in a spot that made it seem like the web was part of it, I now realize the best spot to put the `npm install` is during the `gen-client-ts` step

PR #15465, opened 2025-07-08
outposts: use sqlite or postgres for sessions

dominic-r:
Completly untested. Just a test to see how it might look + aligns with Redis removal vision

## Details

REPLACE ME

---

## Checklist

If applicable


authentik/outposts/consumer.py
```diff
@@ -150,6 +150,11 @@ def event_update(self, event):  # pragma: no cover
 
     def event_session_end(self, event):
         """Event handler which is called when a session is ended"""
+        self.logger.info(
+            "Sending session end event to outpost",
+            session_id=event.get("session_id"),
+            outpost=self.outpost.name if self.outpost else "unknown",
```
rissson:
```suggestion
        self.logger.debug(
            "Sending session end event to outpost",
            session_id=event.get("session_id"),
            outpost=self.outpost,
```

authentik/outposts/migrations/0023_remove_proxysession_authentik_o_session_5897a7_idx_and_more.py
```diff
@@ -0,0 +1,73 @@
+# Generated by Django 5.1.11 on 2025-07-09 23:58
```
rissson: Just as a reminder, there really needs to be only one migration

authentik/outposts/signals.py
```diff
@@ -83,4 +83,11 @@ def pre_delete_cleanup(sender, instance: Outpost, **_):
...
 def logout_revoke(sender: type[AuthenticatedSession], instance: AuthenticatedSession, **_):
     """Catch logout by expiring sessions being deleted"""
-    outpost_session_end.delay(instance.session.session_key)
+    session_key = instance.session.session_key
+    LOGGER.info(
+        "Session deletion detected - sending outpost session end",
+        session_key=session_key,
+        user=instance.user.username,
+        session_pk=instance.session.pk,
+    )
```
rissson, line 92: We never want to log `session_key`.

internal/config/struct.go
```diff
@@ -2,11 +2,13 @@ package config
...
-	Storage        StorageConfig        `yaml:"storage"`
-	LogLevel       string               `yaml:"log_level" env:"AUTHENTIK_LOG_LEVEL, overwrite"`
-	ErrorReporting ErrorReportingConfig `yaml:"error_reporting" env:", prefix=AUTHENTIK_ERROR_REPORTING__"`
-	Redis          RedisConfig          `yaml:"redis" env:", prefix=AUTHENTIK_REDIS__"`
-	Outposts       OutpostConfig        `yaml:"outposts" env:", prefix=AUTHENTIK_OUTPOSTS__"`
+	Storage        StorageConfig           `yaml:"storage"`
+	LogLevel       string                  `yaml:"log_level" env:"AUTHENTIK_LOG_LEVEL, overwrite"`
+	ErrorReporting ErrorReportingConfig    `yaml:"error_reporting" env:", prefix=AUTHENTIK_ERROR_REPORTING__"`
+	SQLite         SQLiteConfig            `yaml:"sqlite" env:", prefix=AUTHENTIK_SQLITE__"`
+	PostgreSQL     StoragePostgreSQLConfig `yaml:"postgresql" env:", prefix=AUTHENTIK_POSTGRESQL__"`
```
rissson:
```suggestion
	PostgreSQL     PostgreSQLConfig `yaml:"postgresql" env:", prefix=AUTHENTIK_POSTGRESQL__"`
```

internal/outpost/ak/api.go
```diff
@@ -86,7 +87,7 @@ func NewAPIController(akURL url.URL, token string) *APIController {
 			outposts, _, err := apiClient.OutpostsApi.OutpostsInstancesList(context.Background()).Execute()
 			return outposts, err
 		},
-		retry.Attempts(0),
+		retry.Attempts(config.Get().Retries.Attempts),
```
rissson: This is out of scope I think. I also don't like the config name, it's way too generic.

internal/outpost/proxyv2/pgstore/connect.go
```diff
@@ -0,0 +1,213 @@
...
+		"database": config.Database,
+		"user":     config.User,
+		"schema":   schema,
+		"sslmode":  config.SSLMode,
+		"has_cert": config.SSLCert != "",
+	}).Debug("Connecting to PostgreSQL")
+
+	// Create GORM DB instance directly
+	db, err := gorm.Open(postgres.New(postgres.Config{
+		DSN: connStr,
```
rissson, line 185: How would this interact with #15577 ?

internal/outpost/proxyv2/sqlitestore/connect.go
```diff
@@ -0,0 +1,334 @@
...
+	// 1. Explicit user configuration
+	// 2. /dev/shm (if available and has sufficient space)
+	// 3. /tmp/authentik-sessions
+	// 4. OS temp directory fallback
+
+	if c.BasePath != "" {
+		basePath = c.BasePath
+		pathType = "user-configured"
+	} else if canUseShm(logger) {
+		basePath = "/dev/shm/authentik-sessions"
```
rissson: I don't like that we're hardcoding this. The only reason /dev/shm is used currently is that it's `$TMPDIR`. We should either use a user-defined path or use `$TMPDIR`, not rely on hardcoded paths.

rissson: And I'm not even sure allowing the user to define a path is necessary, since they can simply change $TMPDIR

authentik/outposts/models.py
```diff
@@ -488,3 +489,60 @@ def delete(self):
...
+        return super().get_queryset().filter(deleted_at__isnull=True)
+
+    def cleanup_expired(self) -> int:
+        """Delete expired sessions and return count of deleted sessions"""
+        from django.utils import timezone
+        result = self.filter(expires__lt=timezone.now()).delete()
+        return result[0] if result else 0
+
+
+class ProxySession(ExpiringModel):
```
rissson: I think this might be better suited to go in `authentik.providers.proxy`?

authentik/outposts/migrations/0023_remove_proxysession_authentik_o_session_5897a7_idx_and_more.py
dominic-r: True. Merged as one.

internal/outpost/proxyv2/pgstore/connect.go
dominic-r: I'll push some code which *should* make it work, but I'll wait until it's in main to really test it out

dominic-r:
more urgently, and the rest of the idea will be revisited after https://github.com/goauthentik/authentik/pull/16628

PR #15483, opened 2025-07-09
website: Docusaurus 3.8 prep integrations

GirlBossRush:
## Details

This PR prepares our Integrations docs for Docusaurus 3.8.

### Use auto-generated category items

Docusaurus 3.8 can cache files a more aggressively when auto-generated categories. However, we still manually define each sidebar category by hand to retain our ability to define their order.

### Updated package.json to not use `devDependencies`

This is a minor bit of house cleaning but something which helps us avoid confusion when dependencies are upgraded. Since Docusaurus can only publish static files, all `dependencies` are implicitly needed for development.

### Clean up of links

Previously not caught by Docusaurus's current link checker.

---

## Checklist

If applicable


tanberry (approved):
Thanks! Let's just think through when which PR gets merged when, so that links about the diff categories of the Integration Guides don't get broken...

PR #15538, opened 2025-07-14
sources/oauth: add entra ID source and move logic over

BeryJu:
## Details

closes #15247

leave Azure AD in there for a future removal

adds Deprecated text in UI for existing AAD sources

---

## Checklist

If applicable

### DIFF
--- a/authentik/sources/oauth/apps.py
+++ b/authentik/sources/oauth/apps.py
@@ -10,6 +10,7 @@ AUTHENTIK_SOURCES_OAUTH_TYPES = [
     "authentik.sources.oauth.types.apple",
     "authentik.sources.oauth.types.azure_ad",
     "authentik.sources.oauth.types.discord",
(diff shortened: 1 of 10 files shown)


BeryJu:
fyi @dominic-r @dewi-tik for the azure AD and twitter change this is the better way to do it not break things during upgrades

dominic-r:
That's smart! Can we also include the twitter change in this pr?

dominic-r:
Also, shouldn't azuread be removed from the ProviderType enum to prevent creation with the deprecated source?

BeryJu:
> Also, shouldn't azuread be removed from the ProviderType enum to prevent creation with the deprecated source?

nah that is generated based on models or source types (can't remember) but we can't remove neither yet as they'll break existing azure AD sources

> That's smart! Can we also include the twitter change in this pr?

probably a separate PR

PR #15571, opened 2025-07-15
core: add updated_at field to user

### LINKED ISSUE #15570: Filtering users by created at and/or last updated at

In a similar situation as in #15569, when interfacing authentik with other systems we want to limit the scope of syncing information out of authentik to only the users modified/created since a specific timestamp (ideally the last sync).

However the API does not expose such filter, upon further inspection there's no timestamps in the database table either :/ 

Two fields:
- `created_at = DateTimeField(auto_now_add=True)`
- `updated_at = DateTimeField(auto_now=True)`

And the corresponding filters in the API that enable consumers to do:

1. `GET /api/v3/core/users/?updated_at__gt=%utc-timestamp%`
2. `GET /api/v3/core/users/?created_at__gt=%utc-timestamp%`

And returns (respectively)

1. Users modified since `%utc-timestamp%`
2. Users created since `%utc-timestamp%`

Again using an intermediate storage to keep track of the dates, but that's an extra moving part we don't enjoy having around.

**Additional context**
We have an impl (pending tests) here: https://github.com/josegomezr/authentik/tree/feat/user_timestamps

josegomezr:
## Details

Brings `updated_at` and `created_at` fields to the core user model.

Additionally filters are also provided for exact value, greater than, less than against the new timestamp dates.

Closes #15570

---

## Checklist

If applicable (not sure if it is)


authentik/core/models.py
```diff
@@ -274,6 +274,9 @@ class User(SerializerModel, GuardianUserMixin, AttributesMixin, AbstractUser):
     ak_groups = models.ManyToManyField("Group", related_name="users")
     password_change_date = models.DateTimeField(auto_now_add=True)
 
+    created_at = models.DateTimeField(auto_now_add=True)
```
rissson: `AbtractUser` already has this a `date_joined`

josegomezr: Ah nice! I'll remove then this field then.

josegomezr: On that matter: should I align names with `date_joined` [`<type>_<purpose>`]? Something like `date_updated`? or is `updated_at` a good enough name?

rissson: We have a `CreatedUpdatedModel` which uses `created` and `last_updated` as field names. Let's align on `last_updated`, and leave the other one as is to not introduce a breaking change in the API.

authentik/core/models.py
```diff
@@ -293,6 +295,7 @@ class Meta:
             models.Index(fields=["uuid"]),
             models.Index(fields=["path"]),
             models.Index(fields=["type"]),
+            models.Index(fields=["last_updated"]),
```
BeryJu, line 299: Not sure if we want an index for this, `date_joined` doesn't have one

BeryJu (commented):
q about index, otherwise LGTM

authentik/core/models.py
josegomezr: True, I personally would index both as the `UsersFilter` allows querying against them.

But I'm open to remove it if that aligns with y'all.

rissson: Let's index both

rissson (approved):

PR #15572, opened 2025-07-15
core: Expose password hashes in the user list API Endpoint

### LINKED ISSUE #15569: Show password hashes on User List API

We're running authentik in conjunction with another subordinated identification system that wants to validate user password on their side.

However the password is not shown in the API endpoint /core/users/ response.

A switch like `include_groups` that will signal authentik to return the password hashes stored in the DB out of the API. Probably also a designated permission so the passwords are only exposed to a user with the specific permission.

We have an implementation hosted in a fork here: https://github.com/josegomezr/authentik/tree/feat/show_password_api

Besides implementing it ourselves we were (ab-)using LDAP as a "bus" for transmitting password information across systems, due to our design we want to avoid (ab-)using LDAP for this purpose.

**Additional context**

Firefox screenshot of the API Response:

Image

josegomezr:
## Details

Return the stored password hash in the user list API endpoint.

The field will only appear if:
- The query string `include_password=true` is provided in the request
- The current user has the `authentik_core.view_password_hashes` permission.

Closes #15569

---

## Checklist

If applicable (not sure if it applies)


BeryJu:
I'd strongly prefer to not include something like this, even with the additional permission required, this still makes it quite easy to shoot yourself in the foot. We could add an endpoint that requires admin auth to validate an individual user's password instead?

tacerus:
> We could add an endpoint that requires admin auth to validate an individual user's password instead?

Whilst interesting too, the goal was rather for the other IDP to synchronize user data completely in order for it to operate independently of Authentik and merely connect occasionally to pull/push changed data.

josegomezr:
How about shielding this "feature" behind a very explicit config:

- `AUTHENTIK_I_KNOW_WHAT_IM_DOING_IS_GONNA_BE_DANGEROUS_YES_SHOW_ME_THE_PW_HASH=1`

By default is deactivated, so it's harder to shoot yourself in the foot, and if you get to that point you must know what you're doing.

fheisler:
> the goal was rather for the other IDP to synchronize user data completely in order for it to operate independently of Authentik and merely connect occasionally to pull/push changed data.

Generally when talking about two-way synchronization across multiple IdPs, this would already be supported by various enterprise features (e.g. source stage) rather than as a typical homelab use case.

tacerus:
We read https://docs.goauthentik.io/docs/add-secure-apps/flows-stages/stages/source/ but it is by our understanding not suitable to sync complete user information (including passphrase hashes) _out of_ Authentik to a secondary IDP.

PR #15626, opened 2025-07-17
sources/saml: add location selection for Signature node

### LINKED ISSUE #7873: SAML Signature Validation shortcommings
**Describe the bug**
When connected to a Shibboleth IdP (not sure yet if applicable for all Shibboleth IdPs or only this specific one), the signature validation fails.

After debugging this was traced back to this issue:

The verification code expects the signature to be located at a specific xpath:

https://github.com/goauthentik/authentik/blob/026e80bd10be43df1e9bad96a63e85271b28c4c9/authentik/sources/saml/processors/response.py#L82-L86

The SAML specification allows this signature to be located at a higher anchor, which is the case for the SAML Response we get from our IdP. (`/samlp:Response/ds:Signature` instead of `/samlp:Response/saml:Assertion/ds:Signature`)

Further Reference: https://docs.oasis-open.org/security/saml/v2.0/saml-core-2.0-os.pdf , Section 5.3 Signature Inheritance

> A SAML assertion may be embedded within another SAML element, such as an enclosing <Assertion> or a request or response, which may be signed. When a SAML assertion does not contain a <ds:Signature> element, but is contained in an enclosing SAML element that contains a <ds:Signature> element, and the signature applies to the <Assertion> element and all its children, then he assertion can be considered to inherit the signature from the enclosing element. The resulting interpretation should be equivalent to the case where the assertion itself was signed with the same key and signature options.

### LINKED ISSUE #12035: InvalidSignature raised for SAML2 Response Assertions
I have found a bug in the current version 2024.10.2 when the SAML2 signature is not correctly verified against the certificate stored in Authentik. I have found the issue to be with the below line, however I am unsure as to whether the `NS_MAP` would also need to change to support the`saml2p:Response`. When a response comes in with a signature this automatically raises a InvalidSignature exception and causes the request to fail.

https://github.com/goauthentik/authentik/blob/def0a42bf1422668095638f221d47484f2b2fc9e/authentik/sources/saml/processors/response.py#L83

I have been using Google Workspace as the SAML IDP, which generates the following response:

```xml
<?xml version="1.0" encoding="UTF-8" standalone="no"?>
<saml2p:Response xmlns:saml2p="urn:oasis:names:tc:SAML:2.0:protocol" Destination="https://auth.com/source/saml/google-auth/acs/" ID="_9f8ea230f008ef523f6bb4069dd0487c" InResponseTo="_bca72c89d9b440d589e009a713528fd8" IssueInstant="2024-11-14T22:17:08.682Z" Version="2.0">
 <saml2:Issuer xmlns:saml2="urn:oasis:names:tc:SAML:2.0:assertion">https://accounts.google.com/o/saml2?idpid=xxxx</saml2:Issuer>
 <ds:Signature xmlns:ds="http://www.w3.org/2000/09/xmldsig#">
  <ds:SignedInfo>
   <ds:CanonicalizationMethod Algorithm="http://www.w3.org/2001/10/xml-exc-c14n#"></ds:CanonicalizationMethod>
   <ds:SignatureMethod Algorithm="http://www.w3.org/2001/04/xmldsig-more#rsa-sha256"></ds:SignatureMethod>
   <ds:Reference URI="#_9f8ea230f008ef523f6bb4069dd0487c">
    <ds:Transforms>
     <ds:Transform Algorithm="http://www.w3.org/2000/09/xmldsig#enveloped-signature"></ds:Transform>
     <ds:Transform Algorithm="http://www.w3.org/2001/10/xml-exc-c14n#"></ds:Transform>
    </ds:Transforms>
    <ds:DigestMethod Algorithm="http://www.w3.org/2001/04/xmlenc#sha256"></ds:DigestMethod>
    <ds:DigestValue>NAPl1OID+yV0fu8Optx54yRJ9veTGOv9bUrYOteK9zE=</ds:DigestValue>
   </ds:Reference>
  </ds:SignedInfo>
  <ds:SignatureValue>ic/d8DOP8sxuv5Zg4gCfvhz2C2T9AMmmu8zl1DrPqZ3HxP65dX97wCkz5BOZT+S+YSSFz/lFO123
mRrq6rm5bqUDt2UcpWdnyIyTec6NSuUgoTkjioQhMb90QFiCocsEkqWj7l46+DzGNePVGo0iwpRE
rcH/3Mw7BOHWYXP+DLcJ1zkVH5stEJ5WU6Tr5uDio0xuQwcb73S45o2R7W2It1RY7Sys3cteWScI
TpY8hJyooBdxgxSYouxqjMsjB9UvHiQvBiD9ccqh/2G0fqm2MSIWPhobLKwvnZ1Ohpa4vqBwT5hg
WhIyu9T1IvswcQmckj03HtSVANNIQmPIjVlQOw==</ds:SignatureValue>
  <ds:KeyInfo>
   <ds:X509Data>
    <ds:X509SubjectName>ST=California,C=US,OU=Google For Work,CN=Google,L=Mountain View,O=Google Inc.</ds:X509SubjectName>
   ...
   </ds:X509Data>
  </ds:KeyInfo>
 </ds:Signature>
 <saml2p:Status>
  <saml2p:StatusCode Value="urn:oasis:names:tc:SAML:2.0:status:Success"></saml2p:StatusCode>
 </saml2p:Status>
 <saml2:Assertion xmlns:saml2="urn:oasis:names:tc:SAML:2.0:assertion" ID="_a947162d5f95e377bb39afc2dc4f2813" IssueInstant="2024-11-14T22:17:08.682Z" Version="2.0">
 ...
 </saml2:Assertion>
</saml2p:Response>
```

PeshekDotDev:
## Details

Some IdP's sign the Response node, others sign the Assertion node. With this feature, you can tell authentik where to look for the Signature in the incoming SAML assertion, including both locations if needed.

closes #12035 
closes #7873

---

## Checklist

If applicable


ikob:
I proposed a minimal set of changes to address the signature node issue as : https://github.com/ant-r-ih/authentik/commit/993ad891e2f5d893b251719dfa7aec33c2b63a9d

- Removed the fixes related to explicitly selecting the Signature node, as I don’t think a dedicated API is necessary. With xmlsec, verification works correctly in either case.
- More test fixtures to cover more scenarios.

web/src/admin/sources/saml/SAMLSourceForm.ts
```diff
@@ -81,6 +91,49 @@ export class SAMLSourceForm extends WithCapabilitiesConfig(BaseSourceForm<SAMLSo
...
+                    <span class="pf-c-switch__toggle">
+                        <span class="pf-c-switch__toggle-icon">
+                            <i class="fas fa-check" aria-hidden="true"></i>
+                        </span>
+                    </span>
+                    <span class="pf-c-switch__label"> ${msg("Verify Assertion Signature")} </span>
+                </label>
+                <p class="pf-c-form__helper-text">
+                    ${msg(
+                        "When enabled, Authentik will look for a Signature inside of the Assertion element.",
```
dominic-r:
```suggestion
                        "When enabled, authentik will look for a Signature inside of the Assertion element.",
```

:)

web/src/admin/sources/saml/SAMLSourceForm.ts
```diff
@@ -81,6 +91,49 @@ export class SAMLSourceForm extends WithCapabilitiesConfig(BaseSourceForm<SAMLSo
...
+                    <span class="pf-c-switch__toggle">
+                        <span class="pf-c-switch__toggle-icon">
+                            <i class="fas fa-check" aria-hidden="true"></i>
+                        </span>
+                    </span>
+                    <span class="pf-c-switch__label"> ${msg("Verify Response Signature")} </span>
+                </label>
+                <p class="pf-c-form__helper-text">
+                    ${msg(
+                        "When enabled, Authentik will look for a Signature inside of the Response element.",
```
dominic-r:
```suggestion
                        "When enabled, authentik will look for a Signature inside of the Response element.",
```

:)

ikob: Thanks! The typo is from the original PR — my patch reverts that part, so it’s already fixed.  
Would you prefer a separate PR for the typo, or should we wait for the PR author to handle it?

PeshekDotDev:
Hey @ikob, the implementation of selecting the location of the signature is on purpose, as IDP's can choose to put it in multiple locations and we want to match the customizability that we have for providers. But it's still greatly appreciated that you wanted to help improve authentik

ikob:
@PeshekDotDev, 
First of all, I think the PR discussed is an important fix for the SAML source and should be merged into soon.

In the original PR, the signature node needs to be selected explicitly. In my update, the code works without such explicit selection and supports signatures for either the Response or Assertion (or both). 
Whether the operator should configure this explicitly or the code should handle both is a design choice, but personally I prefer the more tolerant approach (in line with Postel’s Law).

My update also adds fixtures for validation errors, so applying it to the original PR would improve test coverage as well.

PeshekDotDev:
@ikob I think it's an important PR too. I will get it merged soon. I added the tests you made as well. We decided to go with the explicit choosing of location in order to match the experience of applying a signature location in providers, although I understand why automatic section finding makes for a good user experience too

GirlBossRush (approved):

authentik/sources/saml/processors/response.py
```diff
@@ -116,27 +116,44 @@ def _decrypt_response(self):
...
+            signatures.append(signature_nodes)
+
+        if self._source.signed_assertion:
+            signature_nodes = self._root.xpath(
+                "/samlp:Response/saml:Assertion/ds:Signature", namespaces=NS_MAP
+            )
+
+            if len(signature_nodes) != 1:
+                raise InvalidSignature("No Signature exists in the Assertion element.")
+            signatures.append(signature_nodes)
```
BeryJu: .extend instead of .append

authentik/sources/saml/processors/response.py
```diff
@@ -116,27 +116,44 @@ def _decrypt_response(self):
...
-        xmlsec.tree.add_ids(self._root, ["ID"])
 
-        ctx = xmlsec.SignatureContext()
-        key = xmlsec.Key.from_memory(
-            self._source.verification_kp.certificate_data,
-            xmlsec.constants.KeyDataFormatCertPem,
-        )
-        ctx.key = key
+        for signature in signatures:
+            signature_node = signature[0]
```
BeryJu:
```suggestion
```

authentik/sources/saml/models.py
```diff
@@ -200,6 +200,9 @@ class SAMLSource(Source):
         default=RSA_SHA256,
     )
 
+    signed_assertion = models.BooleanField(default=False)
```
BeryJu:
```suggestion
    signed_assertion = models.BooleanField(default=True)
```

for backwards compat, this is the current behaviour (only checked when verification keypair is selected anyways)

BeryJu (commented):
A little bit of stuff

authentik/sources/saml/processors/response.py
PeshekDotDev: done

authentik/sources/saml/processors/response.py
PeshekDotDev: done

authentik/sources/saml/models.py
PeshekDotDev: done

BeryJu (approved):

PR #15636, opened 2025-07-17
admin: add support bundle

BeryJu:
## Details

Primarily for enterprise support

---

## Checklist

If applicable

### DIFF
--- a/lifecycle/support.py
+++ b/lifecycle/support.py
@@ -1,5 +1,8 @@
+import platform
 from hashlib import sha512
+from pprint import pprint
 from ssl import OPENSSL_VERSION
+from sys import version as python_version
 
 from cryptography.exceptions import InternalError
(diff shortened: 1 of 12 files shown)


GirlBossRush (approved):

lifecycle/support.py
```diff
@@ -0,0 +1,111 @@
...
+            "platform": platform.platform(),
+            "uname": " ".join(platform.uname()),
+        },
+        "config": get_limited_config(),
+    }
+    return payload
+
+
+def encrypt(raw):
+    with open("authentik/enterprise/public.pem", "rb") as _key:
```
rissson, line 92: I'm not sure we're able to decrypt this. In any case, we probably want to use a different key from the licensing one, to give access to support people so they can decrypt the bundle w/o infra intervention

PR #15659, opened 2025-07-18
blueprints: add File tag

### LINKED ISSUE #15526: Read value from file in blueprint
I'm deploying both Authentik and an application (Forgejo) in Kubernetes. I want to provision OAuth authentication automatically, using a secret containing the OAuth client key and secret that will be referenced by both the application and Authentik.
It's currently difficult to achieve this with blueprints securely.

The best solution IMO would be to mount the secret as files, and then read the files using a YAML tag similar to `!Env`, like `!File /path/to/file`.

Mounting secrets as environment variables is a bad security practice, so this excludes `!Env`. 
My current solution is to create a separate secret for the entire blueprint, duplicating the OAuth client key and secret values. Using context values does not change that as it also requires a blueprint, and risks exposing the secret in the GUI.

**Additional context**

guillaumejchauveau:
## Details

Add `!File` YAML tag to blueprints.

This tag is basically a clone of the `!Env` tag with the code of `ConfigLoader.parse_uri` for `file://` config values. Tests are also based on `ConfigLoader`.

Closes #15526 

---

## Checklist

If applicable

### DIFF
--- a/web/package-lock.json
+++ b/web/package-lock.json
@@ -25,7 +25,7 @@
                 "@goauthentik/api": "^2025.6.4-1753191510",
                 "@lit/context": "^1.1.2",
                 "@lit/localize": "^0.12.2",
-                "@lit/reactive-element": "^2.0.4",
+                "@lit/reactive-element": "^2.1.1",
                 "@lit/task": "^1.0.3",
                 "@mdx-js/mdx": "^3.1.0",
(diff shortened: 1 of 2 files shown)


guillaumejchauveau:
I was unable to setup a development environment my machine, so I could not complete the checklist. Since the code is pretty simple and entirely based on code already present in the code-base, would it be possible to run the CI pipelines anyways?

PR #15686, opened 2025-07-21
providers/proxy: Fix incorrect comparison of redirect URL and CookieDomain

### LINKED ISSUE #15685: Redirect fails if non-default port is configured
**Describe the bug**
When using the Proxy provider in forward-auth mode for an application that is hosted on a non-default port (for example `https://app1.example.com:8443`), the user is not redirected to the application after successful login, but gets redirected to the "My applications page".

**To Reproduce**
1. Configure Authentik Proxy provider and Nginx in forward-auth mode according [to docs](https://docs.goauthentik.io/docs/add-secure-apps/providers/proxy/) for an application on non-defualt port.
2. Visit the application while not logged in: go to `https://app1.example.com:8443`
3. You will be redirected to Authentik login page
4. Log in
5. You get redirected to `https://auth.exmaple.com/if/user/#/library` instead of `https://app1.example.com:8443`

**Expected behavior**
After login, the user is redirected to the URL they were originally trying to visit

-   authentik version: 2025.6.3

**Additional context**
I've been digging in the code and found the source of the bug: https://github.com/goauthentik/authentik/blob/21a73fe58cf70ff1bfdb0c0a89fb5d48f4a54624/internal/outpost/proxyv2/application/oauth_state.go#L63

The code is comparing [URL.Host](https://pkg.go.dev/net/url#URL) with `a.proxyConfig.CookieDomain`, which is wrong. [URL.Hostname](https://pkg.go.dev/net/url#URL.Hostname) should be used instead.

According to docs, URL.Host contains the host and port, while Hostname returns only the host without the port. CookieDomain obviously does not contain the port. `string.HasSuffix` function is used, so this bug only appears when a non-default port is used in the redirected URL.

vitSkalicky:
This is my suggested fix for #15685

According to docs, URL.Host contains the host and port, while URL.Hostname returns only the host without the port. CookieDomain obviously does not contain the port. string.HasSuffix function is used, so if a port is set in the redirect URL, this check always fails.

There might be more occurrences of this mistake in the Authentik code.

- I'm not a go developer
- I don't know the Authentik codebase
- I have not tested these changes
- I'm not spending any more of my time on this, since I've decided to not use Authentik in the end.
- Feel free to throw away this PR


dominic-r:
you’re right that u.Host includes both the hostname and port, while u.Hostname() strips the port, per the Go documentation

That said, I’m trying to connect the dots here: the current logic checks if u.Host ends with the cookie domain, which might fail if there’s a port in u.Host (?) However, that alone should only affect domain matching cookie wise, not cause a complete redirect to the “My Applications” page.

dominic-r:
During your testing/finding the bug, did this issue occur 100% of the time? cc @BeryJu

vitSkalicky:
> That said, I’m trying to connect the dots here: the current logic checks if u.Host ends with the cookie domain, which might fail if there’s a port in u.Host (?) However, that alone should only affect domain matching cookie wise, not cause a complete redirect to the “My Applications” page.

It does cause the redirect to the “My Applications” page, because this check against the cookie domain is used in the function "checkRedirectParam".

And yes, it happened every time the user went through the Authentik interface.

BeryJu (approved):

dominic-r (approved):

vitSkalicky:
Thank you for merging it ❤️

PR #15687, opened 2025-07-21
providers/oauth2: fix clean_expired_models failing

BeryJu:
## Details

In this signal we were accessing the relation which at that point might've been deleted already, but we can use the PK

---

## Checklist

If applicable

### DIFF
--- a/authentik/providers/oauth2/signals.py
+++ b/authentik/providers/oauth2/signals.py
@@ -10,7 +10,7 @@ def user_session_deleted_oauth_tokens_removal(sender, instance: AuthenticatedSes
     """Revoke tokens upon user logout"""
     AccessToken.objects.filter(
         user=instance.user,
-        session__session__session_key=instance.session.session_key,
+        session__session__session_key=instance.pk,
     ).delete()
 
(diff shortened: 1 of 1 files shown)


rissson (approved):

rissson:
Actually, the session field on the AccessToken might be null at that point, no?

authentik/providers/oauth2/signals.py
```diff
@@ -10,7 +10,7 @@ def user_session_deleted_oauth_tokens_removal(sender, instance: AuthenticatedSes
     """Revoke tokens upon user logout"""
     AccessToken.objects.filter(
         user=instance.user,
-        session__session__session_key=instance.session.session_key,
+        session__session__session_key=instance.pk,
```
rissson, line 13: So this change should be fine, but we might need to also update the backchannel stuff that was added in the meantime.

One thing that is weird to me, is that the `session` field is marked as cascade on delete for AccessTokens, so that should happen automatically I think?

rissson (changes requested):

PR #15691, opened 2025-07-21
website/docs: change azure ad to entra id

dewi-tik:
## Details

Updates the azure ad doc to entra id. Goes with #15538 which updates the UI

ONLY MERGE AFTER 2025.8 RELEASE

---

## Checklist

-   [x] The documentation has been updated
-   [x] The documentation has been formatted (`make docs`)

PR #15749, opened 2025-07-23
sources: add Telegram source

### LINKED ISSUE #2232: Add Telegram as a social login provider
Telegram offers an authentication mechanism for websites, i've seen some websites that did this --> https///portal.bcc.no. The login page is Auth0 so integrating it is possible

https://telegram.org/blog/login
https://core.telegram.org/widgets/login

I've searched without luck for a keycloak plugin for this function but nobody seems to have investigated it but now that i've come across Authentik project it seems way more suited for it.

atereshkin:
## Details

This implements Telegram social login provider as requested in #2232

Closes #2232

---

## Checklist

If applicable


authentik/sources/telegram/admin.py
```diff
@@ -0,0 +1 @@
+# Register your models here.
```
rissson: We don't use the django admin interface, you can remove this file.

authentik/sources/telegram/models.py
```diff
@@ -0,0 +1,138 @@
...
+    GroupSourceConnection,
+    PropertyMapping,
+    Source,
+    UserSourceConnection,
+)
+from authentik.core.types import UILoginButton
+from authentik.flows.challenge import RedirectChallenge
+
+
+class TelegramSourcePropertyMapping(PropertyMapping):
```
rissson: nit: usually we put the models in this order for sources:
- source
- property mapping
- user source connection
- group source connection

authentik/sources/telegram/models.py
```diff
@@ -0,0 +1,138 @@
...
+    class Meta:
+        verbose_name = _("Telegram Source Property Mapping")
+        verbose_name_plural = _("Telegram Source Property Mappings")
+
+
+class TelegramSource(Source):
+    """Log in with Telegram."""
+
+    bot_username = models.CharField(max_length=255, help_text=_("Telegram bot username"))
+    bot_token = models.CharField(max_length=255, help_text=_("Telegram bot token"))
```
rissson:
```suggestion
    bot_username = models.TextField(help_text=_("Telegram bot username"))
    bot_token = models.TextField(help_text=_("Telegram bot token"))
```
Unless there is a specific reason to limit to 255 characters?

authentik/sources/telegram/models.py
```diff
@@ -0,0 +1,138 @@
...
+    request_access = models.BooleanField(
```
rissson:
```suggestion
    request_message_access = models.BooleanField(
```
just a nit, if it makes sense to be more descriptive?

web/src/admin/sources/SourceWizard.ts
```diff
@@ -4,6 +4,7 @@ import "#admin/sources/oauth/OAuthSourceForm";
 import "#admin/sources/plex/PlexSourceForm";
```
rissson, line 4: You also need to add telegram to `web/src/elements/user/sources/SourceSettings.ts`. This is shown in the user interface, so users can unlink the source from their account.

rissson (changes requested):

website/docs/developer-docs/index.md
```diff
@@ -67,7 +67,8 @@ authentik
 │   ├── ldap - Sync LDAP users from OpenLDAP or Active Directory into authentik
 │   ├── oauth - OAuth1 and OAuth2 Source
 │   ├── plex - Plex source
-│   └── saml - SAML2 Source
+│   ├── saml - SAML2 Source
```
tanberry:
```suggestion
│   ├── saml - SAML2 source
```

tanberry: Both "source" and "provider" should be lower-case... I see we have long had a mix. @atereshkin would you please fix the others, too? :-)

website/docs/sidebar.mjs
```diff
@@ -568,8 +568,9 @@ const items = [
                                 ],
                             },
                             "users-sources/sources/social-logins/mailcow/index",
-                            "users-sources/sources/social-logins/twitch/index",
                             "users-sources/sources/social-logins/plex/index",
+                            "users-sources/sources/social-logins/telegram/index",
+                            "users-sources/sources/social-logins/twitch/index",
```
tanberry, line 575: Thanks for fixing the wrong alphabetization for Twitch!

website/docs/users-sources/sources/social-logins/telegram/index.md
```diff
@@ -0,0 +1,52 @@
+---
+title: Telegram
+support_level: community
+---
+
+Allows users to authenticate using their Telegram account via Telegram application
```
tanberry:
```suggestion
Configuring Telegram as a source allows users to authenticate using their Telegram account credentials within authentik.
```

website/docs/users-sources/sources/social-logins/telegram/index.md
```diff
@@ -0,0 +1,52 @@
...
+
```
tanberry, line 11: Please add a Head2 (##) section for Preparation, and list any pre-reqs, things to be aware of, placeholders for the FQDN, etc.

website/docs/users-sources/sources/social-logins/telegram/index.md
```diff
@@ -0,0 +1,52 @@
...
+## Telegram
+
+To use Telegram as a source, you need to register a Telegram bot first:
```
tanberry:
```suggestion
To use Telegram as a source, you first need to register a Telegram bot:
```

website/docs/users-sources/sources/social-logins/telegram/index.md
```diff
@@ -0,0 +1,52 @@
...
+
+1. Start a chat with `@BotFather` on Telegram.
+2. Use `/newbot` command to create a new bot. Pick a username for your new bot (e.g., `my_bot`).
+   Note the username you've chosen - you'll need it when setting up the source in authentik.
+3. BotFather will provide you with a token for your bot. Make a record of the token as well.
+4. Link the bot to your authentik domain name using `/setdomain` command.
+   Please note that **the domain name must exactly match the FQDN of the authentik installation**.
+
+Now as the bot is configured you can proceed to creating a source in authentik.
```
tanberry:
```suggestion
Now that the bot is configured you can proceed to creating a source in authentik.
```

website/docs/users-sources/sources/social-logins/telegram/index.md
```diff
@@ -0,0 +1,52 @@
...
+
+## authentik
+
+In authentik, go to _Directory_ > _Federation & Social login_ and click _Create_.
```
tanberry:
```suggestion
In authentik, open the Admin interface and go to **Directory** > **Federation & Social login** and click **Create**.
```

tanberry: For UI elements we use bold. Italics are only used for emphasis.

website/docs/users-sources/sources/social-logins/telegram/index.md
```diff
@@ -0,0 +1,52 @@
...
+Select _Telegram source_ and configure your source:
```
tanberry:
```suggestion
Select **Telegram source** and configure your source:
```

website/docs/users-sources/sources/social-logins/telegram/index.md
```diff
@@ -0,0 +1,52 @@
...
+
+- _Name:_ Choose a name
```
tanberry:
```suggestion
- **Name**: Define a name.
```

website/docs/users-sources/sources/social-logins/telegram/index.md
dewi-tik: Here's an example of a modern source doc if you need something to compare against @atereshkin: https://github.com/goauthentik/authentik/pull/15691/files

authentik/sources/telegram/models.py
atereshkin: Both are short strings of the kind of data CharField is meant for. There's definitely a limit on username length in Telegram and the tokens are most likely fixed length.  I just wouldn't want to spend too much time figuring them out since they aren't documented. Plus, if a UI field is auto-generated somewhere (though probably not the case in authentik ATM), an <input> would be more appropriate for both values.
But if our convention is to use TextField for values like that, I'll change that for uniformity.

rissson: Yeah we usually don't bother and just use TextField, it's the same underlying data type in postgres anyway, so performance is not a concern.

website/docs/users-sources/sources/social-logins/telegram/index.md
```diff
@@ -0,0 +1,52 @@
...
+- _Slug:_ Set a slug
```
tanberry:
```suggestion
- **Slug**: Set a slug.
```

website/docs/users-sources/sources/social-logins/telegram/index.md
```diff
@@ -0,0 +1,52 @@
...
+- _Bot username:_ Set the username of your Telegram bot (e.g., `my_bot`)
```
tanberry:
```suggestion
- **Bot username**: Set the username of your Telegram bot (e.g., `my_bot`).
```

website/docs/users-sources/sources/social-logins/telegram/index.md
```diff
@@ -0,0 +1,52 @@
...
+- _Bot token:_ Set the token of your Telegram bot
```
tanberry:
```suggestion
- **Bot token**: Set the token of your Telegram bot.
```

website/docs/users-sources/sources/social-logins/telegram/index.md
```diff
@@ -0,0 +1,52 @@
...
+- _"Request access to send messages from your bot":_ enable this if you need your bot to be able to
```
tanberry:
```suggestion
- **Request access to send messages from your bot**: enable this if you need your bot to be able to
```

tanberry: @atereshkin are the quotation marks actually on the UI? If so please add them back...

website/docs/users-sources/sources/social-logins/telegram/index.md
```diff
@@ -0,0 +1,52 @@
...
+:::note
+For instructions on how to display the new source on the authentik login page, refer to the [Add sources to default login page documentation](../../index.md#add-sources-to-default-login-page).
+:::
+
+## Telegram source property mappings
+
+See the [overview](../../property-mappings/index.md) for information on how property mappings work.
+
+### Expression data
+
```
tanberry, line 48: Can we add a sentence or two in here, explaining why and where these data are used, etc.? Is it useful for creating property mappings, or creating an expression policy, or? And then link to the docs about any relevant tasks.

website/docs/users-sources/sources/social-logins/telegram/index.md
```diff
@@ -0,0 +1,52 @@
...
+The following variables are available to OAuth source property mappings:
+
+- `info`: A Python dictionary containing Telegram user data:
+    - `id` - Telegram user ID
+    - `username` - Username of the user. May not be present
```
tanberry:
```suggestion
    - `username` - Username of the user. Might not be present.
```

website/docs/users-sources/sources/social-logins/telegram/index.md
```diff
@@ -0,0 +1,52 @@
...
+    - `first_name` - First name of the user. May not be present
```
tanberry:
```suggestion
    - `first_name` - First name of the user. Might not be present.
```

website/docs/users-sources/sources/social-logins/telegram/index.md
```diff
@@ -0,0 +1,52 @@
...
+    - `last_name` - Last name of the user. May not be present
```
tanberry:
```suggestion
    - `last_name` - Last name of the user. Might not be present.
```

website/docs/users-sources/sources/social-logins/telegram/index.md
```diff
@@ -0,0 +1,52 @@
...
+    - `photo_url` - URL of the user's profile photo. May not be present
```
tanberry:
```suggestion
    - `photo_url` - URL of the user's profile photo. Might not be present.
```

tanberry: Let's add the same Note that is on most other Source docs at the very end:

:::note 
For instructions on how to display the new source on the authentik login page, refer to the [Add sources to default login page documentation](https://github.com/goauthentik/authentik/blob/main/website/docs/users-sources/sources/index.md#add-sources-to-default-login-page). 
:::

tanberry (changes requested):
Thanks so much @atereshkin for adding the docs for Telegram. A few nits to address please, but looks great, LGTM!!

tanberry:
Oh, also @atereshkin in order to resolve some of the build checks fails, you might need to 1. rebase your local repo on upstream `main` in order to pick up some changes in our build tools, and 2. run `make docs-install` to instlal those new tools/versions.

website/docs/users-sources/sources/social-logins/telegram/index.md
atereshkin: > @atereshkin are the quotation marks actually on the UI? If so please add them back...

@tanberry Not really, so committing this.

authentik/sources/telegram/admin.py
atereshkin: done

authentik/sources/telegram/models.py
atereshkin: done

authentik/sources/telegram/models.py
atereshkin: done

web/src/admin/sources/SourceWizard.ts
atereshkin: done

authentik/sources/telegram/models.py
atereshkin: done

website/docs/developer-docs/index.md
atereshkin: done

website/docs/users-sources/sources/social-logins/telegram/index.md
atereshkin: done

website/docs/users-sources/sources/social-logins/telegram/index.md
atereshkin: done

tanberry (approved):
Thanks @atereshkin !

authentik/sources/telegram/stage.py
```diff
@@ -0,0 +1,48 @@
...
+class TelegramLoginChallenge(LoginChallengeMixin, Challenge):
+    component = CharField(default="ak-source-telegram")
+    bot_username = CharField(help_text=_("Telegram bot username"))
+    request_message_access = BooleanField()
+
+
+class TelegramChallengeResponse(ChallengeResponse):
+    component = CharField(default="ak-source-telegram")
+
+    id = serializers.IntegerField()
```
BeryJu: What is this `id` field for?

authentik/sources/telegram/stage.py
```diff
@@ -0,0 +1,48 @@
...
+    first_name = serializers.CharField(max_length=255, required=False)
+    last_name = serializers.CharField(max_length=255, required=False)
+    username = serializers.CharField(max_length=255, required=False)
+    photo_url = serializers.URLField(required=False)
+    auth_date = serializers.IntegerField(required=True)
```
BeryJu: Minor nit but these field types should be directly imported from `rest_framework.fields` instead of serializers

authentik/sources/telegram/stage.py
```diff
@@ -0,0 +1,48 @@
...
+    hash = serializers.CharField(max_length=64, required=True)
+
+    def validate_auth_date(self, auth_date):
+        if datetime.fromtimestamp(auth_date) < datetime.now() - timedelta(minutes=5):
+            raise serializers.ValidationError(_("Authentication date is too old"))
+        return auth_date
+
+    def validate(self, attrs):
```
BeryJu: type annotations on args and return please

authentik/sources/telegram/views.py
```diff
@@ -0,0 +1,68 @@
...
+    GroupTelegramSourceConnection,
+    TelegramSource,
+    UserTelegramSourceConnection,
+)
+from authentik.sources.telegram.stage import TelegramChallengeResponse, TelegramLoginChallenge
+
+
+class TelegramStartView(View):
+    def get(self, request, source_slug):
+        source = TelegramSource.objects.get(slug=source_slug)
```
BeryJu: need to check for `.enabled` and raise Http404 if its false

authentik/sources/telegram/views.py
```diff
@@ -0,0 +1,68 @@
...
+        plan: FlowPlan = self.request.session[SESSION_KEY_PLAN]
```
BeryJu: Will need to find another solution for this as this URL might be called without any flow in the context

authentik/sources/telegram/stage.py
atereshkin: That's Telegram user id. I just named these fields the same way Telegram calls them. Do you think we should change it to something that makes sense in the context of authentik or would a simple comment suffice?

authentik/sources/telegram/stage.py
atereshkin: Fixed.

authentik/sources/telegram/views.py
atereshkin: Fixed.

authentik/sources/telegram/stage.py
atereshkin: Fixed.

authentik/sources/telegram/stage.py
BeryJu: Ah ok, makes sense, maybe `user_id` since just `id` makes me think database primary key, but no hard feelings on either

authentik/sources/telegram/views.py
atereshkin: @BeryJu could you elaborate a bit? I was assuming all of this happens within the context of a flow. IdentificationStage (which I was assuming to always run within a flow) renders the sources and when the user clicks on Telegram we use RedirectChallenge to send the user to this view which injects a detour. What are the other situations when this might be called?

One radical solution I have in mind is simply placing the Telegram widget (which is actually just a button that opens Telegram authentication popup) straight on the IdentificationStage along normal source buttons. The nice thing about it is that it will solve the need for the user to click a button twice present in the current implementation. We already discussed this with @rissson as a possible future improvement, but the reason I didn't want to do it right away is because it would require refactoring code parts used by other sources (right now there's no way for a source to redefine the behavior of the rendering of the respective button on the IdentificationStage). I thought that might be too bold for a first contribution. But now that there seems to be an issue with the redirect approach, maybe I should go ahead with this one instead?

BeryJu: One solution for this that would work is similarly to the SAML Source: https://github.com/goauthentik/authentik/blob/main/authentik/sources/saml/views.py#L67-L140

Add a pre-authentication flow to the model, and in this view create a plan for that flow and append the telegram stage to it

atereshkin: Just to make sure we are on the same page. If this view is called from within a flow, we are still injecting Telegram login into that flow, correct? Otherwise we'd deprive the users of the ability to configure a custom flow (potentially with some extra stages before and after login) that merely has Telegram as one of the login options. It's only for the cases when they want to link directly to Telegram login that we need a _pre-authentication flow_ setting. Then if there's no flow in the context we'd use that similarly to how SAML source does it. And seems like that setting should be optional since for most users it wouldn't make sense. Does that sound right?

BeryJu: The idea would be that the sources' `ui_login_button` returns a `RedirectChallenge` to this view that initiates the pre_authorization flow and does `plan.append_stage(in_memory_stage(TelegramLoginView))` (which would allow user configured stages to run before telegram. Then after the authentication is done we use the default backend flow logic with either continuing with the authentication or enrollment flow

atereshkin: OK, done that. The only thing is that it caused a bit of code duplication with SAML source because I wanted to keep this change self-contained. Is that the right call, or should I rather do a refactoring right away?

authentik/sources/telegram/stage.py
atereshkin: These field names are actually used as data when doing validation (i.e. they are a part of the string being hashed to check response authenticity). So since you've indicated it's OK I left things as is to avoid introducing special handling of this field.

BeryJu:
@atereshkin pre-authentication flow stuff looks good, resolves the last issue. Could you fix the PR to not include 300+ changed files though? I think something went wrong when merging main into this branch

atereshkin:
> @atereshkin pre-authentication flow stuff looks good, resolves the last issue. Could you fix the PR to not include 300+ changed files though? I think something went wrong when merging main into this branch

Sorry! Fixed.

dewi-tik:
Hope you don't mind, I've edited the documentation to better match our current style guide. Great PR. Looking forward to getting it merged.

atereshkin:
> Hope you don't mind, I've edited the documentation to better match our current style guide. Great PR. Looking forward to getting it merged.

Of course not. Thank you!

BeryJu (approved):

Issue #12230, opened 2024-11-29
Proxy outpost fills up TMPDIR with sessions


zbig-t:
**Describe the bug**
Every few days my self-hosted Authentik instance crashes which correlates to full `/dev/shm` filesystem inside the container.

**To Reproduce**
Deploy Authentik on Kubernetes cluster using Helm chart as per documentation.

**Logs**
Excerpt from `authentik-server` pod's log:
```
{"event": "Starting gunicorn 23.0.0", "level": "info", "logger": "gunicorn.error", "timestamp": 1732875579.247224}
Traceback (most recent call last):
File "/ak-root/venv/bin/gunicorn", line 8, in <module>
sys.exit(run())
^^^^^
File "/ak-root/venv/lib/python3.12/site-packages/gunicorn/app/wsgiapp.py", line 66, in run
WSGIApplication("%(prog)s [OPTIONS] [APP_MODULE]", prog=prog).run()
File "/ak-root/venv/lib/python3.12/site-packages/gunicorn/app/base.py", line 235, in run
super().run()
File "/ak-root/venv/lib/python3.12/site-packages/gunicorn/app/base.py", line 71, in run
Arbiter(self).run()
File "/ak-root/venv/lib/python3.12/site-packages/gunicorn/arbiter.py", line 197, in run
self.start()
File "/ak-root/venv/lib/python3.12/site-packages/gunicorn/arbiter.py", line 136, in start
self.pidfile.create(self.pid)
File "/ak-root/venv/lib/python3.12/site-packages/gunicorn/pidfile.py", line 36, in create
os.write(fd, ("%s\n" % self.pid).encode('utf-8'))
OSError: [Errno 28] No space left on device
Sentry is attempting to send 1 pending events
Waiting up to 2 seconds
Press Ctrl-C to quit
```
Output from `df`:
```
authentik@authentik-server-55d65ffb7c-8k6fj:/$ df -h
Filesystem      Size  Used Avail Use% Mounted on
overlay         118G   22G   96G  19% /
tmpfs            64M     0   64M   0% /dev
/dev/sda1       118G   22G   96G  19% /etc/hosts
shm              64M   64M     0 100% /dev/shm
tmpfs           7.8G   12K  7.8G   1% /run/secrets/kubernetes.io/serviceaccount
tmpfs           3.9G     0  3.9G   0% /proc/acpi
tmpfs           3.9G     0  3.9G   0% /proc/scsi
[...]

PR #15798, opened 2025-07-26
outpost/proxyv2: add session cleanup for filesystem session store

Tom3201:
## Details

The package used for the file system sessions `github.com/gorilla/sessions` does not seem to have a session cleanup.
This PR contains a wrapper for the session store with a session cleanup job. With the current implementation, this job will run every five minutes.
With these changes the issue #12230 should hopefully be solved.

---

## Checklist

If applicable

### DIFF
--- a/internal/outpost/proxyv2/application/session.go
+++ b/internal/outpost/proxyv2/application/session.go
@@ -20,6 +20,7 @@ import (
 	"goauthentik.io/internal/config"
 	"goauthentik.io/internal/outpost/proxyv2/codecs"
 	"goauthentik.io/internal/outpost/proxyv2/constants"
+	"goauthentik.io/internal/outpost/proxyv2/filesystemstore"
 	"goauthentik.io/internal/outpost/proxyv2/redisstore"
 	"goauthentik.io/internal/utils"
 )
@@ -90,7 +91,10 @@ func (a *Application) getStore(p api.ProxyOutpostConfig, externalHost *url.URL) 
 		return rs, nil
 	}
 	dir := os.TempDir()
-	cs := sessions.NewFilesystemStore(dir)
+	cs, err := filesystemstore.GetPersistentStore(dir)
+	if err != nil {
+		return nil, err
+	}
 	cs.Codecs = codecs.CodecsFromPairs(maxAge, []byte(*p.CookieSecret))
 	// https://github.com/markbates/goth/commit/7276be0fdf719ddff753f3574ef0f967e4a5a5f7
 	// set the maxLength of the cookies stored on the disk to a larger number to prevent issues with:
@@ -123,7 +127,7 @@ func (a *Application) getAllCodecs() []securecookie.Codec {
 }
 
 func (a *Application) Logout(ctx context.Context, filter func(c Claims) bool) error {
-	if _, ok := a.sessions.(*sessions.FilesystemStore); ok {
+	if _, ok := a.sessions.(*filesystemstore.Store); ok {
 		files, err := os.ReadDir(os.TempDir())
 		if err != nil {
 			return err
--- a/internal/outpost/proxyv2/filesystemstore/filesystemstore.go
+++ b/internal/outpost/proxyv2/filesystemstore/filesystemstore.go
@@ -1,0 +1,226 @@
+package filesystemstore
+
+import (
+	"context"
+	"errors"
+	"os"
+	"path"
+	"strings"
+	"sync"
+	"syscall"
+	"time"
+
+	"github.com/gorilla/sessions"
+	log "github.com/sirupsen/logrus"
+)
+
+const (
+	SessionCleanupInterval     = 5 * time.Minute
+	SessionCleanupLockFileName = "session-cleanup.lock"
+	SessionFilePrefix          = "session_"
+	SessionTestFile            = SessionFilePrefix + "write_test"
+)
+
+var (
+	ErrSessionCleanupAlreadyRunning = errors.New("session cleanup is already running by another instance")
+	ErrSessionStoreNoPermission     = errors.New("path is not writable")
+	ErrSessionStorePathNotExist     = errors.New("path does not exist")
+)
+
+type Store struct {
+	*sessions.FilesystemStore
+	storePath string
+	log       *log.Entry
+}
+
+// NewStore checks if the specified store path exists, is writable and creates a new filesystem session store.
+func NewStore(storePath string, keyPairs ...[]byte) (*Store, error) {
+	if storePath == "" {
+		storePath = os.TempDir()
+	}
+
+	// check if path exists
+	_, err := os.ReadDir(storePath)
+	if err != nil {
+		if errors.Is(err, os.ErrNotExist) {
+			return nil, ErrSessionStorePathNotExist
+		}
+		return nil, err
+	}
+
+	// check if path is writable
+	testPath := path.Join(storePath, SessionTestFile)
+	testFile, err := os.OpenFile(testPath, os.O_CREATE, 0600)
+	if err != nil {
+		if errors.Is(err, os.ErrPermission) {
+			return nil, ErrSessionStoreNoPermission
+		}
+		return nil, err
+	}
+	if err = testFile.Close(); err != nil {
+		return nil, err
+	}
+	if err = os.Remove(testPath); err != nil {
+		return nil, err
+	}
+
+	return &Store{
+		FilesystemStore: sessions.NewFilesystemStore(storePath, keyPairs...),
+		storePath:       storePath,
+		log:             log.WithField("logger", "authentik.outpost.proxyv2.filesystemstore"),
+	}, nil
+}
+
+// SessionCleanup acquires a file lock to ensure only one instance runs at a time,
+// then checks and deletes expired session files from the filesystem session store.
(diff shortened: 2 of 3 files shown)


internal/outpost/proxyv2/filesystemstore/filesystemstore.go
```diff
@@ -0,0 +1,185 @@
...
+	var lastErr error
+	for _, file := range files {
+		select {
+		case <-ctx.Done():
+			s.log.Warn("session cleanup interrupted during file processing")
+			return ctx.Err()
+		default:
+		}
+
+		if !strings.HasPrefix(file.Name(), "session_") {
```
dominic-r: I'd suggest not hardcoding the session prefix and instead using a package-level constant.

internal/outpost/proxyv2/filesystemstore/filesystemstore.go
```diff
@@ -0,0 +1,185 @@
...
+		modTime := stat.ModTime()
+		if time.Since(modTime) <= time.Duration(s.Options.MaxAge)*time.Second {
+			s.log.WithField("max-age", s.Options.MaxAge).WithField("modified", modTime.String()).Debug("session still valid")
+			continue
+		}
+
+		s.log.WithField("path", fullPath).WithField("modified", modTime.String()).Info("cleanup expired session")
+		if err = os.Remove(fullPath); err != nil {
+			s.log.WithError(err).WithField("path", fullPath).Warn("failed to delete session")
+			lastErr = err
```
dominic-r: Using lastErr, the last error encountered is returned. All others are silently dropped. If multiple deletions fail, you'd never know. I would figure it's best to collect all the errors with errors.Join probably doing something like this:
```
var errs []error
...
errs = append(errs, err)
...
return errors.Join(errs...)
```

internal/outpost/proxyv2/filesystemstore/filesystemstore.go
```diff
@@ -0,0 +1,185 @@
...
+			continue
+		}
+	}
+	return lastErr
+}
+
+var (
```
dominic-r, line 157: So, these vars are global and mutable but are not protected by a sync.Mutex this means that if GetPersistentStore() or StopPersistentStore() are called from multiple goroutines, a race condition *could* occur.

internal/outpost/proxyv2/filesystemstore/filesystemstore.go
```diff
@@ -0,0 +1,185 @@
...
+		log:             log.WithField("logger", "authentik.outpost.proxyv2.filesystemstore"),
+	}
+}
+
+// SessionCleanup acquires a file lock to ensure only one instance runs at a time,
+// then checks and deletes expired session files from the filesystem session store.
+// It supports context-based cancellation to allow graceful shutdowns or timeouts.
+func (s *Store) SessionCleanup(ctx context.Context) error {
+	s.log.Info("Starting session cleanup")
+	lockPath := path.Join(s.storePath, SessionCleanupLockFileName)
```
dominic-r, line 79: I mabye skipped over it, but it seems like the lock file is never removed, even after the lock is released.

dominic-r (commented):
Looks like a very solid MR overall. Thanks for opening this! Might conflict with the migration to sqlite and postgres in #15465, but I suppose we'll see as we go along.

internal/outpost/proxyv2/filesystemstore/filesystemstore.go
Tom3201: Good idea, I hadn't thought about that and therefore included the errors in the log. I have now improved it.

internal/outpost/proxyv2/filesystemstore/filesystemstore.go
Tom3201: Correct, the file has not yet been deleted, as I think the syscall lock should be sufficient. However, I've also added file deletion to the function now.

Tom3201:
> Looks like a very solid MR overall. Thanks for opening this! Might conflict with the migration to sqlite and postgres in #15465, but I suppose we'll see as we go along.

@dominic-r Thank you very much for your quick review. I have improved and implemented the points you mentioned.

internal/outpost/proxyv2/application/session.go
```diff
@@ -90,7 +91,7 @@ func (a *Application) getStore(p api.ProxyOutpostConfig, externalHost *url.URL)
 		return rs, nil
 	}
 	dir := os.TempDir()
-	cs := sessions.NewFilesystemStore(dir)
+	cs := filesystemstore.GetPersistentStore(dir)
```
dominic-r: nit: the function name GetPersistentStore is misleading - it suggests it only retrieves, but it also creates and manages cleanup jobs

internal/outpost/proxyv2/filesystemstore/filesystemstore.go
```diff
@@ -0,0 +1,203 @@
...
+	ErrSessionCleanupAlreadyRunning = errors.New("session cleanup is already running by another instance")
+)
+
+type Store struct {
+	sessions.FilesystemStore
+	storePath string
+	log       *log.Entry
+}
+
+func NewStore(path string, keyPairs ...[]byte) *Store {
```
dominic-r: Can we add some input validation for path (check if directory exists and is writable basically)

dominic-r: The embedding of sessions.FilesystemStore by value could be problematic. You could consider using a pointer

dominic-r: Can we add error return to handle cases where store creation fails

internal/outpost/proxyv2/filesystemstore/filesystemstore.go
```diff
@@ -0,0 +1,203 @@
...
+	}
+}
+
+// SessionCleanup acquires a file lock to ensure only one instance runs at a time,
+// then checks and deletes expired session files from the filesystem session store.
+// It supports context-based cancellation to allow graceful shutdowns or timeouts.
+func (s *Store) SessionCleanup(ctx context.Context) error {
+	s.log.Info("Starting session cleanup")
+	lockPath := path.Join(s.storePath, SessionCleanupLockFileName)
+	lockFile, err := os.OpenFile(lockPath, os.O_CREATE|os.O_RDWR, 0644)
```
dominic-r:
```suggestion
	lockFile, err := os.OpenFile(lockPath, os.O_CREATE|os.O_RDWR, 0600)
```

would it work with this permission?

internal/outpost/proxyv2/filesystemstore/filesystemstore.go
```diff
@@ -0,0 +1,203 @@
...
+		done <- s.sessionCleanup(ctx)
+	}()
+
+	select {
+	case <-ctx.Done():
+		return ctx.Err()
+	case err = <-done:
+		s.log.WithField("error", err).Info("Session cleanup finished")
+		return err
+	}
```
dominic-r: The sessionCleanup method already handles context cancellation,, so wouldn't the wrapper be unneeded here?

internal/outpost/proxyv2/filesystemstore/filesystemstore.go
```diff
@@ -0,0 +1,203 @@
...
+			s.log.WithField("max-age", s.Options.MaxAge).WithField("modified", modTime.String()).Debug("session still valid")
+			continue
+		}
+
+		s.log.WithField("path", fullPath).WithField("modified", modTime.String()).Info("cleanup expired session")
+		if err = os.Remove(fullPath); err != nil {
+			s.log.WithError(err).WithField("path", fullPath).Warn("failed to delete session")
+			errs = append(errs, err)
+			continue
+		}
```
dominic-r, line 152: It would be best to add a check if the file still exists before deletion

dominic-r (approved):
Looks pretty good for me, thanks again!

internal/outpost/proxyv2/application/session.go
Tom3201: I am always open to suggestions. I haven't come up with anything other than "Get" yet, as "New" doesn't fit in this context either.

internal/outpost/proxyv2/filesystemstore/filesystemstore.go
Tom3201: I have added an existence validation and writing test.

internal/outpost/proxyv2/filesystemstore/filesystemstore.go
Tom3201: Yes, also works with `0600`. Is modified 😃

internal/outpost/proxyv2/filesystemstore/filesystemstore.go
Tom3201: Is the existence validation necessary here? If the file no longer exists, the modification time check would fail beforehand.

internal/outpost/proxyv2/filesystemstore/filesystemstore.go
Tom3201: You're absolutely right. It was still an old residue from the function splitting that I overlooked.

BeryJu (approved):

PR #15825, opened 2025-07-28
internal: run server and new worker when using `ak server` in debug mode

BeryJu:
## Details

Also changes the PID file behaviour to only generate the name and let the application create the file

---

## Checklist

If applicable


rissson (approved):
apart from the `--watch` thing, LGTM.

Do we want to add an option to avoid worker start with debug=true, so that one can still start it manually separately from the server?

GirlBossRush (approved):

PR #15861, opened 2025-07-29
website/integrations: add profile scope to nextcloud oidc

rattencreep:
When configuring the OpenID Connect user backend app, the  scope profile is missing. Without that, auto-provisioned users will have the display name set to the account name - which is not easily human readable.


dewi-tik (approved):
Thanks for this @rattencreep

dominic-r (approved):
Thanks

PR #15876, opened 2025-07-30
web/a11y: Flow Search

GirlBossRush:
## Dependencies

- #16750

## Details

This PR prepares the flow search components for use in a screen reader and e2e test suite. The changes here represent a partial implementation of accessibility, curated for an approachable PR review.

- Fix issue where escape key closes modal when selecting search select item
- Clean up mismatch of instance bound methods 

---

## Checklist

-   [ ] Local tests pass (`ak test authentik/`)
-   [ ] The code has been formatted (`make lint-fix`)

If an API change has been made

-   [ ] The API schema has been updated (`make gen-build`)

If changes to the frontend have been made

-   [ ] The code has been formatted (`make web`)

If applicable

-   [ ] The documentation has been updated
-   [ ] The documentation has been formatted (`make docs`)

PR #15701, opened 2025-07-21
web/a11y: Wizards & Forms

GirlBossRush:
## Depends on

- #15877 
- #15878 

## Details

This PR continues the a11y-ification of the UI, adding additional labels for wizards, tables, inputs, navigations -- Pretty much everything the test suite needs to traverse the DOM.

Screenshot 2025-07-21 at 22 24 44

Screenshot 2025-07-21 at 22 33 55

---

## Checklist

-   [ ] Local tests pass (`ak test authentik/`)
-   [ ] The code has been formatted (`make lint-fix`)

If an API change has been made

-   [ ] The API schema has been updated (`make gen-build`)

If changes to the frontend have been made

-   [x] The code has been formatted (`make web`)

If applicable

-   [ ] The documentation has been updated
-   [ ] The documentation has been formatted (`make docs`)

## See also

- #15876

PR #15894, opened 2025-07-30
web: Form validation regressions, consistency fixes

GirlBossRush:
## Details

This PR resolves a collection of form validation issues, as well as a partial fix for our WebDriverIO tests to run while Playwright is in review.

### Fixed Issues

- `ErrorProp`s containing strings are now parsed correctly instead of defaulting to a generic network error
- Single-step modal forms perform client-side validation before submitting their data
- Wizards now perform client-side validation before progressing to the next step
- Form Groups no longer fight for focus when multiple invalid inputs exist.
- Form Groups no longer scroll into view when a previously invalid field becomes valid
- Invalid inputs within Form Groups now wait for their parent to expand before attempting to focus the element while still hidden
- Both server and client-side errors now trigger Form Group expansion.
- Inputs with server-side errors are now marked with ARIA attributes to match their status
- The inner `form` element getter within AKForm now uses `renderRoot` over `shadowRoot`, fixing selector issues when in compatibility mode. 

## See also

- #16119 (Split off from this PR)
- #15876
- #15701

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -193,7 +193,8 @@
         "@wdio/browser-runner": "^9.19.0",
         "@wdio/cli": "^9.19.0",
         "@wdio/spec-reporter": "^9.19.0",
-        "@web/test-runner": "^0.20.2"
+        "@web/test-runner": "^0.20.2",
+        "chromedriver": "^139.0.0"
     },
     "wireit": {
         "build": {
--- a/web/src/admin/applications/wizard/ApplicationWizardStep.ts
+++ b/web/src/admin/applications/wizard/ApplicationWizardStep.ts
@@ -5,6 +5,7 @@ import {
 } from "./types.js";
 
 import { serializeForm } from "#elements/forms/Form";
+import { reportValidityDeep } from "#elements/forms/FormGroup";
 
 import {
     NavigationEventInit,
@@ -18,7 +19,7 @@ import { styles } from "#admin/applications/wizard/ApplicationWizardFormStepStyl
 import { ApplicationRequest, ValidationError } from "@goauthentik/api";
 
 import { msg } from "@lit/localize";
-import { property, query } from "lit/decorators.js";
+import { property } from "lit/decorators.js";
 
 export class ApplicationWizardStep<T = Partial<ApplicationRequest>> extends WizardStep {
     static styles = [...WizardStep.styles, ...styles];
@@ -33,10 +34,39 @@ export class ApplicationWizardStep<T = Partial<ApplicationRequest>> extends Wiza
     public canCancel = true;
 
     // This should be overridden in the children for more precise targeting.
-    @query("form")
-    protected form!: HTMLFormElement;
+    public get form(): HTMLFormElement | null {
+        return this.renderRoot.querySelector("form");
+    }
+
+    /**
+     * @todo This defaults to true when the form is not yet available
+     * to ease the migration of existing wizards. This behavior should be removed.
+     */
+    public reportValidity(): boolean {
+        const { form } = this;
+
+        if (!form) return true;
+
+        return reportValidityDeep(form);
+    }
+
+    /**
+     * @todo This defaults to true when the form is not yet available
+     * to ease the migration of existing wizards. This behavior should be removed.
+     */
+    public checkValidity(): boolean {
+        const { form } = this;
+
+        if (!form) return true;
+
+        return form.checkValidity();
+    }
 
     protected get formValues(): T {
+        if (!this.form) {
+            throw new TypeError("Form reference is not set");
+        }
+
         return serializeForm<T>([
             ...this.form.querySelectorAll("ak-form-element-horizontal"),
             ...this.form.querySelectorAll("[data-ak-control]"),
--- a/web/src/admin/applications/wizard/steps/ak-application-wizard-application-step.ts
+++ b/web/src/admin/applications/wizard/steps/ak-application-wizard-application-step.ts
@@ -21,7 +21,7 @@ import { snakeCase } from "change-case";
 
 import { msg } from "@lit/localize";
 import { html } from "lit";
-import { customElement, query, state } from "lit/decorators.js";
+import { customElement, state } from "lit/decorators.js";
 import { ifDefined } from "lit/directives/if-defined.js";
 
 function trimMany<T extends object, K extends keyof T>(target: T, keys: K[]): Pick<T, K> {
@@ -43,8 +43,9 @@ export class ApplicationWizardApplicationStep extends ApplicationWizardStep {
     @state()
     errors = new Map<keyof ApplicationRequest, string>();
 
-    @query("form#applicationform")
-    form!: HTMLFormElement;
+    public get form(): HTMLFormElement | null {
+        return this.renderRoot.querySelector("form#applicationform");
+    }
 
     constructor() {
         super();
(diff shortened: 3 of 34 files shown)


web/src/components/ak-label.ts
```diff
@@ -0,0 +1,27 @@
...
+import type { LabelHTMLAttributes } from "react";
+
+import { html, nothing } from "lit";
+import { ifDefined } from "lit/directives/if-defined.js";
+
+export interface FormLabelProps extends LabelHTMLAttributes<HTMLLabelElement> {
+    required?: boolean;
+}
+
+export const AKLabel: LitFC<FormLabelProps> = (
```
kensternberg-authentik, line 13: Excellent.  Fragments rock.

web/src/elements/forms/HorizontalFormElement.ts
```diff
@@ -63,79 +42,88 @@ export class HorizontalFormElement extends AKElement {
...
+     * @attribute
+     * @deprecated Labels cannot associate with inputs across DOM roots. Use the slotted `label` element instead.
+     */
     @property({ type: String })
-    public label = "";
+    public label?: string;
 
     @property({ type: Boolean })
-    public required = false;
+    public required?: boolean;
```
kensternberg-authentik, line 63: This isn't good lit.  Remember that Boolean properties must always be false by default; if you don't want it to show up in the tag automagically, keep the `reflect: true` out of the property declaration, which you've already done here.

kensternberg-authentik (approved):
I'm approving, but I did have one thing I'd like you to look at and change. `undefined !== false`; the use case in HorizontalFormComponent works by coincidence, and coincidence makes me cross.

web/package.json
```diff
@@ -193,7 +193,8 @@
         "@wdio/browser-runner": "^9.19.0",
         "@wdio/cli": "^9.19.0",
         "@wdio/spec-reporter": "^9.19.0",
-        "@web/test-runner": "^0.20.2"
+        "@web/test-runner": "^0.20.2",
+        "chromedriver": "^139.0.0"
```
GirlBossRush, line 197: WebDriver is still plagued with download issues, but this inclusion will let the legacy e2e test suite run when Chromium is manually installed.

web/src/admin/applications/wizard/ApplicationWizardStep.ts
```diff
@@ -33,10 +34,39 @@ export class ApplicationWizardStep<T = Partial<ApplicationRequest>> extends Wiza
...
 
     // This should be overridden in the children for more precise targeting.
-    @query("form")
-    protected form!: HTMLFormElement;
+    public get form(): HTMLFormElement | null {
+        return this.renderRoot.querySelector("form");
+    }
+
+    /**
+     * @todo This defaults to true when the form is not yet available
```
GirlBossRush, line 42: TOOD: Circle back on this after #16010 gives us a reusable pattern with form internals

web/src/admin/applications/wizard/ApplicationWizardStep.ts
```diff
@@ -33,10 +34,39 @@ export class ApplicationWizardStep<T = Partial<ApplicationRequest>> extends Wiza
...
+
+        if (!form) return true;
+
+        return form.checkValidity();
+    }
 
     protected get formValues(): T {
+        if (!this.form) {
+            throw new TypeError("Form reference is not set");
+        }
```
GirlBossRush, line 68: Unlikely but this has been technically possible with all uses of `@query`

web/src/admin/applications/wizard/steps/ak-application-wizard-application-step.ts
```diff
@@ -43,8 +43,9 @@ export class ApplicationWizardApplicationStep extends ApplicationWizardStep {
     @state()
     errors = new Map<keyof ApplicationRequest, string>();
 
-    @query("form#applicationform")
-    form!: HTMLFormElement;
+    public get form(): HTMLFormElement | null {
+        return this.renderRoot.querySelector("form#applicationform");
+    }
```
GirlBossRush, line 48: Switch to getter to allow for overriding in application provider steps

web/src/admin/applications/wizard/steps/ak-application-wizard-application-step.ts
```diff
@@ -71,7 +72,7 @@ export class ApplicationWizardApplicationStep extends ApplicationWizardStep {
             this.errors.set("name", msg("An application name is required"));
         }
 
-        if (!values.metaLaunchUrl || !URL.canParse(values.metaLaunchUrl)) {
+        if (values.metaLaunchUrl && !URL.canParse(values.metaLaunchUrl)) {
```
GirlBossRush, line 75: TODO: This was subtle fix. Circle back on something like an `<ak-url-input>` to handle this sort of check.

web/src/admin/applications/wizard/steps/ak-application-wizard-provider-step.ts
```diff
@@ -33,7 +33,24 @@ export class ApplicationWizardProviderStep extends ApplicationWizardStep {
...
 
     @query("#providerform")
-    element!: ApplicationWizardProviderForm<OneOfProvider>;
+    protected element!: ApplicationWizardProviderForm<OneOfProvider>;
+
+    get form(): HTMLFormElement | null {
+        const providerForm = this.element.form;
+
+        if (!providerForm) {
+            // TODO: This needs to be removed once all steps can report their validity.
```
GirlBossRush, line 42: AFAICT this is only true of the bindings step which has no form. Leaving it in for now to reduce the number of changes.

web/src/admin/applications/wizard/steps/providers/ApplicationWizardProviderForm.ts
```diff
@@ -16,7 +16,7 @@ import { snakeCase } from "change-case";
 import { CSSResult } from "lit";
 import { property, query } from "lit/decorators.js";
 
-export class ApplicationWizardProviderForm<T extends OneOfProvider> extends AKElement {
+export abstract class ApplicationWizardProviderForm<T extends OneOfProvider> extends AKElement {
```
GirlBossRush, line 19: `ApplicationWizardStep` should also be abstract, but our typing on `CustomEmitterElement` makes this difficult to achieve without a lot of detangling.

web/src/admin/applications/wizard/steps/providers/ak-application-wizard-provider-for-oauth.ts
```diff
@@ -48,7 +48,7 @@ export class ApplicationWizardOauth2ProviderForm extends ApplicationWizardProvid
             <form id="providerform" class="pf-c-form pf-m-horizontal" slot="form">
                 ${renderForm(
                     provider ?? {},
-                    errors,
+                    errors.provider ?? {},
```
GirlBossRush, line 51: AFAICT not showing field errors in the provider step seems to be a long standing bug we managed to surface when error parsing was introduced. 

I suspect that `ak-application-wizard-submit-step` also has an expectation of field error name spacing that doesn't apply anymore.

web/src/common/styles/authentik.css
```diff
@@ -35,6 +35,31 @@
...
+}
+
+.pf-c-form__label[aria-required] .pf-c-form__label-text::after {
+    content: "*";
+    user-select: none;
+    margin-left: var(--pf-c-form__label-required--MarginLeft);
+    font-size: var(--pf-c-form__label-required--FontSize);
+    color: var(--pf-c-form__label-required--Color);
+}
+
```
GirlBossRush: Largely moved out of horizontal form for better consistency in inputs across all interfaces.

web/src/components/ak-label.ts
GirlBossRush, line 13: See #16119 for how this works with label input association.

PR #15940, opened 2025-08-03
web/a11y: Form Errors

GirlBossRush:
## Details

This PR resolves a collection of accessibility issues in form errors. The primary change here is a function component capable of rendering common form errors using our existing error parsing tools.

## See also

- #15894

### DIFF
--- a/web/src/elements/types.ts
+++ b/web/src/elements/types.ts
@@ -54,6 +54,19 @@ export type LitPropertyRecord<T extends object> = {
  */
 export type LitPropertyKey<K> = K extends string ? `.${K}` | `?${K}` | K : K;
 
+/**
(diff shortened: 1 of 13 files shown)


web/src/elements/types.ts
```diff
@@ -54,6 +54,19 @@ export type LitPropertyRecord<T extends object> = {
  */
 export type LitPropertyKey<K> = K extends string ? `.${K}` | `?${K}` | K : K;
 
+/**
+ * A React-like functional component. Used to render a component in a template.
```
kensternberg-authentik, line 58: Not really "rendering components," more like "fragments."  But it's a comment; let it pass.

kensternberg-authentik (approved):
I really like the restructuring.

PR #16007, opened 2025-08-06
website: Glossary

### LINKED ISSUE #18093: Glossary
- [ ] Review definitions
- [ ] Review look-n-feel
- [ ] Test findability/UX
- [ ] Announce in Rel Notes
- [ ] Announce in Discord ask for reviews and contributions if anyone wants

dominic-r:
Closes: AUTH-1228

Closes: #18093


website/docs/add-secure-apps/flows-stages/stages/source/index.md
```diff
@@ -18,7 +18,7 @@ Another use case is to route users to authenticate with your legacy (Okta, etc)
 
 It is very important that the configured source's authentication and enrollment flows (when set; they can be left unselected to prevent authentication or enrollment with the source) do **not** have a [User login stage](../user_login/index.md) bound to them.
 
-This is because the Source stage works by appending a [dynamic in-memory](../../../../core/terminology.md#dynamic-in-memory-stage) stage to the source's flow, so having a [User login stage](../user_login/index.md) bound will cause the source's flow to not resume the original flow it was started from, and instead directly authenticating the pending user.
+This is because the Source stage works by appending a [dynamic in-memory](../../../../core/glossary?term=dynamic_in_memory_stage) stage to the source's flow, so having a [User login stage](../user_login/index.md) bound will cause the source's flow to not resume the original flow it was started from, and instead directly authenticating the pending user.
```
dominic-r:
```suggestion
This is because the Source stage works by appending a [dynamic in-memory](../../../../core/glossary#dynamic-in-memory-stage) stage to the source's flow, so having a [User login stage](../user_login/index.md) bound will cause the source's flow to not resume the original flow it was started from, and instead directly authenticating the pending user.
```

website/docusaurus-theme/theme/DocCardList/GlossaryDocCardList.tsx
```diff
@@ -0,0 +1,325 @@
...
+                                aria-expanded={isExpanded}
+                            >
+                                {isExpanded ? "▼ Hide details" : "▶ Show details"}
+                            </button>
+                            {isExpanded && (
+                                <div className={sharedStyles.glossaryLong}>
+                                    {longParagraphs.map((paragraph, index) => (
+                                        <p
+                                            key={index}
+                                            dangerouslySetInnerHTML={{
```
dominic-r: Probably needs dompurify here and everywhere else

website/docusaurus-theme/theme/DocCardList/index.tsx
```diff
@@ -0,0 +1,106 @@
...
+): value is T & Record<K, unknown> {
+    return Object.prototype.hasOwnProperty.call(value, key);
+}
+
+/**
+ * Generates stable React keys from sidebar items, with fallback to index
+ */
+function getStableKey(item: GlossaryItem | PropSidebarItem, idx: number) {
+    if (typeof item === "object" && item !== null) {
+        if (hasOwnProperty(item, "docId") && typeof item.docId === "string") return item.docId;
```
kensternberg-authentik: Last Line Effect Risk! Consider:

```suggestion
      const match = ["docId", "id", "href", "label"].find(
        (name) => hasOwnProperty(item, name) && typeof item[name] === "string"
    );
    return match ? item[match] : idx;
```

website/docusaurus-theme/theme/utils/glossaryUtils.ts
```diff
@@ -0,0 +1,85 @@
...
+export function safeStringArrayExtract(value: unknown): string[] {
+    return Array.isArray(value)
+        ? value.filter((tag): tag is string => typeof tag === "string")
+        : [];
+}
+
+/**
+ * Extracts all available tags from a collection of terms
+ */
+export function extractAvailableTags(terms: readonly GlossaryHelperTerm[]): string[] {
```
kensternberg-authentik:
```suggestion
return Array.from(new Set(terms.flatMap(t => (t.tags ?? [])))).toSorted();
```

You could replace the whole thing with that.  Also, a lot of your functions are one-liner, perfect for arrowing.

```
export const extractAvailableTags = (terms: readonly GlossaryHelperTerm[]): string[]) =>
    Array.from(new Set(terms.flatMap(t => (t.tags ?? [])))).toSorted();
```

In either case, you should prefer `.toSorted()` here to `.sort()`; the latter takes up twice as much memory and takes twice as long to run.  I know, we're talking 2 milliseconds instead of one.

kensternberg-authentik (approved):
While I have some quibbles with the glossary, the code overall looks fine.  It would be nice if terms had tags that specified if something was an _authentik_ term or just an industry term, like "Application" vs "LDAP".  Also, the `#` symbol at the end of each entry feels... _weird_.  Like, that's an anchor, but it triggers a search-like behavior. That's unexpected.

Also: some snark about JavaScript.  :-)

tanberry (approved):
This is great, @dominic-r ! Impressive number of terms for first round.

PR #16010, opened 2025-08-06
web/a11y: Codemirror

GirlBossRush:
## Details

This PR resolves a collection of accessibility issues in the Codemirror web component. In addition to fleshing out Codemirror's lifecycle with Lit, our element now makes use of [`ElementInternals`](https://developer.mozilla.org/en-US/docs/Web/API/ElementInternals), allowing for a more native form input serialization. 

## See also

- #16011 (For full form compatibility)

### DIFF
--- a/web/src/elements/CodeMirror.ts
+++ b/web/src/elements/CodeMirror.ts
@@ -1,6 +1,6 @@
 import { EVENT_THEME_CHANGE } from "#common/constants";
 
-import { AKElement } from "#elements/Base";
+import { FormAssociatedElement } from "#elements/forms/form-associated-element";
 
 import { UiThemeEnum } from "@goauthentik/api";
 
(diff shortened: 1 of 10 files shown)


web/src/elements/mixins/internals.ts
```diff
@@ -0,0 +1,113 @@
...
+import { property } from "lit/decorators.js";
+
+/**
+ * A mixin that provides form-associated internals to the element.
+ *
+ * @see {@link https://developer.mozilla.org/en-US/docs/Web/API/ElementInternals | MDN}
+ *
+ * @see {@linkcode WithInternals}
+ */
+export interface InternalsMixin
```
kensternberg-authentik: this is a lot more `formAssociated` than it is `elementInternals`.  ElementInternals has _three_ separate uses, sadly; I think the W3C is just hanging off it "everything we wanted to do that we can't otherwise."  It does form association, [custom state management](https://www.dannymoerkerke.com/blog/the-hidden-power-of-custom-states-for-web-components/), and [ARIA integration](https://developer.mozilla.org/en-US/docs/Web/API/ElementInternals#instance_properties_included_from_aria), plus it's an extra check on components because internals is where the shadowDOM reference is actually kept.

Consider (a) changing the name to `formAssociatedMixin`, and (b) checking that the component does not already have an `internals` field.  Or make "adding internals" either it's own Mixin that you automatically include with this one (with that check, or find one of the "mixin dedup" functions Justin Fagnani has mentioned), or, you know, just add internals to the base class of the component library.

web/src/elements/CodeMirror.ts
```diff
@@ -83,131 +115,191 @@ export class CodeMirrorTextarea<T> extends AKElement {
...
+
+        if (this.#editor) {
+            console.debug("ak-codemirror: destroying editor");
+            this.#editor.destroy();
+        }
+    }
+
+    #focusListener = () => {
+        this.#editor?.contentDOM?.focus();
+    };
```
GirlBossRush, line 306: Note, while this works when a form error is shown, clicking an associated label is supported in #16011

kensternberg-authentik (approved):
Thanks for fixing the non-standard binary issue.

PR #16014, opened 2025-08-06
web/e2e: Playwright end-to-end test runner

GirlBossRush:
## Details

This PR introduces Playwright as an end-to-end test runner, replacing WebDriverIO.
The changes included in this branch do not include the tests, instead focusing on a smaller subset to review:

### What is included

- Playwright test fixtures
  - Form element selectors, buttons, etc
  - Wizard elements, buttons, etc
  - Pointer actions, clicking, pressing, etc
  - Session actions, login, logout
- An entity name generator, used to create unique names for users, groups, providers, and other resources during test runs
- Pino, a runtime-agnostic logger for use in test reporters, CLI scripts, and possibly in the front-end at a later time
- A few modifications to our TSConfig to type-check new tests
- Vitest boilerplate for unit tests

### What's not included

- Removing WebDriverIO
- Provider tests
- Accessibility improvements
- Alignment with the upcoming `@goauthentik/elements` package

## See also

- #16016

PR #15874, opened 2025-07-30
web: Form error rendering

GirlBossRush:
This PR adjusts form and wizard error handling to use existing utilities for a more consistent rendering.

## See also

- #15940 

---

## Checklist

-   [ ] Local tests pass (`ak test authentik/`)
-   [ ] The code has been formatted (`make lint-fix`)

If an API change has been made

-   [ ] The API schema has been updated (`make gen-build`)

If changes to the frontend have been made

-   [ ] The code has been formatted (`make web`)

If applicable

-   [ ] The documentation has been updated
-   [ ] The documentation has been formatted (`make docs`)

PR #16015, opened 2025-08-06
web/e2e: Session Tests

GirlBossRush:
## Dependencies

- #15894
- #15874
- #15876
- #15701
- #16014

## Details

This PR adds session tests to Playwright, allowing us to verify that authentication works for existing users. This is largely ported from the WebDriverIO tests

## See also

- #16016


GirlBossRush:
Resolved in #16014

Issue #16017, opened 2025-08-06
Failed standard search for a User with two or parts in name


tanberry:
If I have a User with the name `user 9` a standard search on the Users list page in the Admin UI doesn't return that user, but instead lists all users.

-   authentik version: [e.g. 2025.6.2]
-   Deployment: docker-compose

kensternberg-authentik:
I suspect this is an excess of enthusiasm in `core/api/users.py:UserViewSet`, which defines `search_fields = ["username", "name", "is_active", "email", "uuid", "attributes", "date_joined", "last_updated"]`.  A lot of those aren't text-searchable, like "is_active" and the Date-based fields, not without extra magic.  Our basic textual search is probably getting confused about those.  Limiting the field set to just "username", "name", and "email" would probably make more sense here.  (UUIDs _might_, but...)

I've tested this locally and am happy to say that doing so causes the search to behave as expected and doesn't seem to break anything else. The only place `UserViewSet` is mentioned is in the corresponding `url.py` file (and a structural mention in `enterprise/middleware.py`, but not relevant to this issue). Changing the `search_field` list does not break other features of a search request, such as `ordering`.

tanberry:
Also just noticed that the `X` beside the Search field doesn't clear the contents of the Search field.

PR #16020, opened 2025-08-06
core: limit User search_field to pure text only

kensternberg-authentik:
core/core: limit the User search field to pure text fields

# What

This commit limits the `core/api/users.py:UserViewSet.search_field` entries to a list of fields that, logically, an admin would want to search on when looking for a user. The search is now limited to “username”, “name” (the display name), and “email”.

# Why

[The documentation for DRF states](https://www.django-rest-framework.org/api-guide/filtering/), “The search_fields attribute should be a list of names of text-type fields on the model, such as CharField or TextField.” A quick look through Stack Overflow and Reddit reveals that lots of users have struggled with this restriction, but also that DRF can return highly erroneous lists when non-text fields are included in `search_field`.

We encountered this bug while looking for a user in a very large collection. Usernames with spaces were likely to match on a text segment rather than the whole query, and searches were failing to limit their returns to anything vaguely resembling the original query string.

Searches for non-text data should be done using the `filterset` feature.

# Fixes

[Failed standard search for a User with two or parts in name #16017](https://github.com/goauthentik/authentik/issues/16017)

# Checklist


authentik/core/api/users.py
```diff
@@ -405,21 +410,15 @@ class UserViewSet(UsedByMixin, ModelViewSet):
     ordering = ["username", "date_joined", "last_updated"]
     serializer_class = UserSerializer
     filterset_class = UsersFilter
-    search_fields = [
```
kensternberg-authentik, line 408: This is the only change to the actual functionality.  Everything else is the result of my running `black`.

dominic-r:
Ken opening a backend pr? Would not have put that on my bingo card 🤣

kensternberg-authentik:
> Ken opening a backend pr? Would not have put that on my bingo card 🤣

I used to be [very, very good](https://stackoverflow.com/search?q=user:166838+[django]) at Django.

authentik/core/api/users.py
```diff
@@ -405,21 +410,15 @@ class UserViewSet(UsedByMixin, ModelViewSet):
...
-        "username",
-        "name",
-        "is_active",
-        "email",
-        "uuid",
-        "attributes",
```
rissson, line 414: I would keep uuid, but otherwise LGTM

rissson (approved):

PR #16028, opened 2025-08-07
web: Fix initial browser color scheme.

### LINKED ISSUE #13819: UI Flashing White While Navigating
As stated the UI is flashing white while im navigating through the application. Im not sure if this is a configuration issue on my end but I don't believe it is. When I hit Admin Interface or User Interface for example it flashes white while loading the screen. I'm trying to implement this at my company as a dashboard for various groups so I need a fix for this because I can't run the risk someone has a seizure trying to use this.

GirlBossRush:
## Details

This PR fixes an issue stemming from the browser's preferred _color scheme_ not aligning with the server's preferred _UI theme_. The fix is performed by providing the browser with early hints on what color scheme has precedence before any CSS or stylesheets are loaded.

closes #13819

### DIFF
--- a/authentik/core/templates/base/theme.html
+++ b/authentik/core/templates/base/theme.html
@@ -1,0 +1,11 @@
+{% if ui_theme == "dark" %}
+  <meta name="color-scheme" content="dark" />
+  <meta name="theme-color" content="#18191a">
+{% elif ui_theme == "light" %}
+  <meta name="color-scheme" content="light" />
+  <meta name="theme-color" content="#ffffff">
+{% else %}
+  <meta name="color-scheme" content="light dark" />
+  <meta name="theme-color" content="#ffffff" media="(prefers-color-scheme: light)">
+  <meta name="theme-color" content="#18191a" media="(prefers-color-scheme: dark)">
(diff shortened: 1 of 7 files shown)


authentik/core/templates/base/theme.html
```diff
@@ -0,0 +1,11 @@
...
+  <meta name="color-scheme" content="dark" />
+  <meta name="theme-color" content="#18191a">
+{% elif ui_theme == "light" %}
+  <meta name="color-scheme" content="light" />
+  <meta name="theme-color" content="#ffffff">
+{% else %}
+  <meta name="color-scheme" content="light dark" />
+  <meta name="theme-color" content="#ffffff" media="(prefers-color-scheme: light)">
+  <meta name="theme-color" content="#18191a" media="(prefers-color-scheme: dark)">
+{% endif %}
```
GirlBossRush, line 11: A bit wordy here but the separation is intentional. When the UI theme is set to automatic, we communicate to the browser that both color schemes are available. The user's preferred scheme will take precedence.

rissson (approved):
% the failing CI, LGTM

PR #16030, opened 2025-08-07
web: Automatic reload during server start up.

GirlBossRush:
Screenshot 2025-08-25 at 18 49 35

## Details

This PR updates the initial page shown while authentik starts up. The styling and colors are defined by the browser, keeping the page small and quick to load.

## Demo

https://htmlpreview.github.io/?https://github.com/goauthentik/authentik/blob/developer-loading-screen-test/web/src/standalone/loading/startup/demo-circles-hollow.html


web/initializing.html
```diff
@@ -0,0 +1,42 @@
+<!doctype html>
```
GirlBossRush: Largely inspired from our www API redirect template.

rissson:
Let's only merge after branch-off

kensternberg-authentik (approved):
Oooh, pretty!

PR #16031, opened 2025-08-07
web: Clean up render interfaces.

GirlBossRush:
Co-Author: ken@goauthentik.io

## Details

This PR is based on the feedback provided in #15940, making use of consistent render prop interfaces, and fewer redundant [nullish coalescing operators](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing).


kensternberg-authentik (approved):
Builds, runs, looks good, behaves as expected.

PR #16065, opened 2025-08-11
core: Mark impersonation reason field as required in UI and fix status codes

### LINKED ISSUE #15991: Mandatory impersonation reason, but missing star and bad HTTP status
When impersonation reason is marked as mandatory, in the ui, the red star is missing and submitting with a blank reason (`reason: ""`) results in a 401 instead of what I assume should be a 400 with a descriptive message:

Image

dominic-r:
Signed-off-by: Dominic R <dominic@sdko.org>
Closes: https://github.com/goauthentik/authentik/issues/15991

Updates the Impersonation modal to have the reason field marked as mandatory when necessary as well as updating the API to return a 400 by raising a validationerror when the reason is not set and the tenant requires an impersonation reason.

### DIFF
--- a/authentik/core/api/users.py
+++ b/authentik/core/api/users.py
@@ -678,8 +678,7 @@ class UserViewSet(UsedByMixin, ModelViewSet):
             },
         ),
         responses={
-            "204": OpenApiResponse(description="Successfully started impersonation"),
-            "401": OpenApiResponse(description="Access denied"),
+            204: OpenApiResponse(description="Successfully started impersonation"),
         },
     )
     @action(detail=True, methods=["POST"], permission_classes=[])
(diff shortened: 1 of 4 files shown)


authentik/core/api/users.py
```diff
@@ -678,8 +678,10 @@ def recovery_email(self, request: Request, pk: int) -> Response:
             },
         ),
         responses={
-            "204": OpenApiResponse(description="Successfully started impersonation"),
-            "401": OpenApiResponse(description="Access denied"),
+            204: OpenApiResponse(description="Successfully started impersonation"),
+            400: OpenApiResponse(description="Bad request"),
```
BeryJu: 400 shouldn't be declared here as that will override the generated schema for validation errors

web/src/admin/users/UserImpersonateForm.ts
```diff
@@ -5,17 +5,29 @@ import { globalAK } from "#common/global";
...
 
+    @state()
+    private requireReason = false;
+
+    async firstUpdated(): Promise<void> {
+        try {
+            const settings = await new AdminApi(DEFAULT_CONFIG).adminSettingsRetrieve();
+            this.requireReason = settings.impersonationRequireReason ?? false;
+        } catch (e) {
+            this.requireReason = false;
```
BeryJu, line 29:
```suggestion
            // fallback to reason not required as the backend will still validate it
            this.requireReason = false;
```

BeryJu (approved):

Issue #16050, opened 2025-08-09
Nextcloud SAML support


BrandonMFong:
Seems like SAML/SSO is no longer supported by the [nextcloud application](https://github.com/nextcloud/user_saml)? 

Also looks like the [authentik documentation](https://integrations.goauthentik.io/chat-communication-collaboration/nextcloud/#preparation-1) is outdated?

dominic-r:
In the link you mentioned, what points to SAML being deprecated?

BrandonMFong:
@dominic-r , I can't see any of the [settings mentioned here](https://integrations.goauthentik.io/chat-communication-collaboration/nextcloud/#nextcloud-configuration-1):
<img width="1390" height="881" alt="Image" src="https://github.com/user-attachments/assets/7b44ef26-9102-4780-bf51-7cf2b014c89e" />

According to [this guide](https://blog.cubieserver.de/2022/complete-guide-to-nextcloud-saml-authentication-with-authentik/), looks like there is supposed to be more fields I can modify? I'm running [Nextcloud Hub 10](https://nextcloud.com/) (31.0.7)

dominic-r:
i'm on 31.0.7 as well and all worked as expected. My nextcloud is down as i'm migrating between two orchestration systems, but i'll try to take a look tonight. Did you try to restart your instance or reinstall the plugin? perhaps that might help

BrandonMFong:
@dominic-r, yes I did restart and reinstalled the nextcloud instance and the plugin. Is there other settings I'm missing?

dominic-r:
Oh I think I know why. Brought my NC back online and you need to click, in https://nextcloud.company/settings/admin/saml, this button:

<img width="1211" height="165" alt="Image" src="https://github.com/user-attachments/assets/d2b65c14-a2d6-488b-8ca1-005548759b4d" />

I'll open a PR to add a note to make this clear

(There's "Reset settings" at the bottom of the page to switch modes)

BrandonMFong:
Ah, the `Reset settings` button revealed that button. Thank you for clearing this out!

PR #16066, opened 2025-08-11
website/integrations: nextcloud: Cleanup SAML service config

dominic-r:
Signed-off-by: Dominic R <dominic@sdko.org>
CC: @dewi-tik, @tanberry 

Created following creation of https://github.com/goauthentik/authentik/issues/16050

This PR includes:

* the :::danger now uses a relative link
* updated field names and logically sorted them following what's in the UI
* Removed the section "Configure group attribute in Nextcloud" because  before it tells you to configure it before you even configure the rest of the SSO which is... interesting. Also, on the old line 310 you were adding the group mapping anyways, so I just added an (optional) nexto it


dewi-tik (approved):
LGTM. Thanks Dominic

PR #16119, opened 2025-08-12
web/a11y: Associating labels with inputs

GirlBossRush:
## Details

**Split off from #15894**

This PR resolves a collection of accessibility issues in labels within flow interfaces. The primary change here is a migration away from `<ak-form-control />` to two functional components , `AKLabel` and `AKFieldError`.

Together, they bring the `<label />` and `<input />` element into the same DOM tree, allowing them to match `htmlFor` and `id` attributes as a screen-reader friendly pairing.

## See also

- #15876
- #15701


kensternberg-authentik (approved):
A consistent set of fixes.

PR #16129, opened 2025-08-12
stages/captcha: Make stage more managed with provider-specific defaults

### LINKED ISSUE #10254: Support a self-hosted CAPTCHA provider like mCaptcha
Avoiding third party services is part of the reason I use Authentik, so needing one to support the CAPTCHA functionality feels kind of bad. Also, there is the privacy risk involved with their methodology of detection.

**Describe the solution you'd like.**
[mCaptcha](https://github.com/mCaptcha/mCaptcha) is a self-hostable alternative to CAPTCHAs, using a proof-of-work algorithm to slow down abusers.

**Describe alternatives you've considered.**
No viable option exists if I want to stay independent and not use any 3rd-party services.

**Additional context**
It doesn't have to be mCaptcha itself, it's just one I know off the top of my head.

### LINKED ISSUE #17674: Update Google reCaptcha docs
Update links and instructions for how to setup google reCaptcha

Currently configured through reCaptcha console which is being merged into google cloud

dominic-r:
Reopened to reset code owner check 

Closes: https://github.com/goauthentik/authentik/issues/10254

## Details

REPLACE ME

---

## Checklist

If applicable


web/src/admin/stages/captcha/CaptchaStageForm.ts
```diff
@@ -12,17 +12,221 @@ import { CaptchaStage, CaptchaStageRequest, StagesApi } from "@goauthentik/api";
...
+    },
+};
+
 @customElement("ak-stage-captcha-form")
 export class CaptchaStageForm extends BaseStageForm<CaptchaStage> {
+    @state()
+    selectedProvider = "custom";
+
+    @state()
+    currentPreset: CaptchaProviderPreset = CAPTCHA_PROVIDERS.custom;
```
kensternberg-authentik, line 114: `currentPreset` always changes in lock-state with `selectedProvider`; it is the output of an _expression_ with `selectedProvider` as the input.  It doesn't need to be a `@state()` in its own right; changing `selectedProvider` will trigger the update just fine.

Recommendation: remove the `@state` for `currentPreset`.

web/src/admin/stages/captcha/CaptchaStageForm.ts
```diff
@@ -12,17 +12,221 @@ import { CaptchaStage, CaptchaStageRequest, StagesApi } from "@goauthentik/api";
...
+            const minScoreInput = this.shadowRoot?.querySelector(
+                'ak-number-input[name="scoreMinThreshold"]',
+            ) as HTMLInputElement;
+            const maxScoreInput = this.shadowRoot?.querySelector(
+                'ak-number-input[name="scoreMaxThreshold"]',
+            ) as HTMLInputElement;
+
+            if (minScoreInput) minScoreInput.value = this.currentPreset.score.min.toString();
+            if (maxScoreInput) maxScoreInput.value = this.currentPreset.score.max.toString();
+        }
```
kensternberg-authentik: Why isn't setting all of this in the form itself, as simple `value=` attributes in the HTML, sufficient?  It looks like that should work.  The `ifDefined()`s look a little suspicious.  See my comment at line 351.

web/src/admin/stages/captcha/CaptchaStageForm.ts
```diff
@@ -35,10 +239,178 @@ export class CaptchaStageForm extends BaseStageForm<CaptchaStage> {
...
+                        "Minimum required score to allow continuing. Lower scores indicate more suspicious behavior.",
+                    )}
+                ></ak-number-input>
+                <ak-number-input
+                    label=${msg("Score Maximum Threshold")}
+                    required
+                    name="scoreMaxThreshold"
+                    value="${ifDefined(
+                        this.instance?.scoreMaxThreshold ?? this.currentPreset.score?.max ?? 1.0,
+                    )}"
```
kensternberg-authentik: The low-level DOM manipulation in `handleProviderChange` feels like a work-around for this bit of precedence.   

Suggestion: Split this function in half: one function chooses "`instance` or `currentPreset`?" and creates exactly the object the renderer needs.  Then make the renderer a pure function that takes that object as input, and produces the HTML as output, including the `value=` fields.

This way, the business logic is clear as to _what_ you're rendering and these cascades of defaults or in-cases go away in the renderer.

While it seems unlikely, the possibility exists that the `instance` has _some_ of these fields filled, and the `currentPreset` has others, creating an inconsistent display.  Being very explicit about _which_ object you pass to the renderer can catch and eliminate those sorts of bugs.

web/src/admin/stages/captcha/CaptchaStageForm.ts
```diff
@@ -12,17 +12,221 @@ import { CaptchaStage, CaptchaStageRequest, StagesApi } from "@goauthentik/api";
...
+            }
+        }
+        return "custom";
+    }
+
+    firstUpdated(changedProperties: Map<PropertyKey, unknown>): void {
+        super.firstUpdated(changedProperties);
+        this.selectedProvider = this.detectProviderFromInstance();
+        this.currentPreset = CAPTCHA_PROVIDERS[this.selectedProvider];
+    }
```
kensternberg-authentik, line 144: It's entirely possible the instance is available before your first render.  Checking for that after the first render can be blinky; Lit whines about setting state in a post-render lifecycle op, because doing so schedules an inefficient re-render.  Also, while unlikely and not currently done in our code base, the `instance` should be allowed to change at any time, since it's a property.

Suggestion: move this logic into `willUpdate()`.  Using `PropertyValues<this>` ensures that the `.has()` expression will be checked to ensure that the name provided is, in fact, a property or state key of `CaptchaStageForm`. 

```
    willUpdate(changed: PropertyValues<this>) {
        super.willUpdate(changed);
        if (changed.has("instance")) {
            this.selectedProvider = this.detectProviderFromInstance();
            this.currentPreset = CAPTCHA_PROVIDERS[this.selectedProvider];
        }
    }
```

kensternberg-authentik (commented):
A few comments on code quality. Overall, this looks good, but there are a few lit-specific problems that could be addressed to make it better/faster/smarter.

kensternberg-authentik (approved):
Much stronger, but the help text is redundant and could make translation more difficult.  Let's revisit that later.

melizeche (approved):

kensternberg-authentik (approved):

PR #16139, opened 2025-08-12
web: Docker versioning compatibility

rissson:
## Details

This PR updates the frontend only setup to be compatible with the latest changes to our versioning. This achieved with a combination of more accurate version parsing, and fixing an edge case where the ESBuild's versioned assets can't align with the version embedded in Docker images.   

- Update frontend dev docs to always use latest dev images
- ~~Update static assets to use query params for cache busting~~


authentik/core/templatetags/authentik_core.py
```diff
@@ -10,5 +11,6 @@
 
 @register.simple_tag()
 def versioned_script(path: str) -> str:
-    """Wrapper around {% static %} tag that supports setting the version"""
-    return static_loader(path.replace("%v", authentik_full_version()))
+    """Wrapper around {% static %} tag that appends a version query parameter to the URL"""
+
+    return static_loader(path) + "?v=" + authentik_full_version()
```
GirlBossRush: This change in particular is the crux of it. The value of `authentik_full_version` within the Docker image isn't accessible from ESBuild, but all we're hoping to achieve here is ensure that static assets have cache busting between version upgrades.

This function might be a good candidate for [Django path objects](https://docs.djangoproject.com/en/5.2/topics/forms/media/#paths-as-objects), making it possible to do thing like [subresource integrity](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity) and [CSP nonces](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/nonce#using_nonce_to_allowlist_a_script_element)

GirlBossRush (approved):

PR #16140, opened 2025-08-12
core: Fix docstring for core_users_recovery_email_create and core_users_recovery_create

LucaTheHacker:
## Details

Fixing documentation issue for `core_users_recovery_email_create` and `core_users_recovery_create`

---

## Checklist

If applicable

Notice: Wan't able to run "make web|docs"

### DIFF
--- a/authentik/core/api/users.py
+++ b/authentik/core/api/users.py
@@ -620,7 +620,7 @@ class UserViewSet(UsedByMixin, ModelViewSet):
     )
     @action(detail=True, pagination_class=None, filter_backends=[], methods=["POST"])
     def recovery(self, request: Request, pk: int) -> Response:
-        """Create a temporary link that a user can use to recover their accounts"""
+        """Create a temporary link that a user can use to recover their account"""
         link, _ = self._create_recovery_link()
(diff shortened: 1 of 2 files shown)


dominic-r:
The schema is automatically generated. You should instead find where this string is in the code, then run `make gen-build`

LucaTheHacker:
> The schema is automatically generated. You should instead find where this string is in the code, then run `make gen-build`

My bad, I've used grep to find the reference and went with the first entry. I believe we should add a "this is a generated file, do not edit" comment at the top of schema.yml tbh

melizeche (changes requested):
The schema is autogenerated from the method’s docstrings; this change should be made in the docstrings
https://github.com/goauthentik/authentik/blob/main/authentik/core/api/users.py#L623
https://github.com/goauthentik/authentik/blob/main/authentik/core/api/users.py#L644

after the changes, generate the schema using `make gen-build`

dominic-r (approved):
Thanks!

melizeche (approved):
LGTM 🚀

PR #16164, opened 2025-08-13
ci: release: consolidation bump version and on tag

rissson:
## Details

REPLACE ME

---

## Checklist

If applicable


.github/workflows/release-tag.yml
```diff
@@ -1,39 +1,179 @@
...
+        run: |
+          # ID from https://api.github.com/users/authentik-automation[bot]
+          git config --global user.name '${{ steps.app-token.outputs.app-slug }}[bot]'
+          git config --global user.email '${{ steps.get-user-id.outputs.user-id }}+${{ steps.app-token.outputs.app-slug }}[bot]@users.noreply.github.com'
+          git commit -a -m "release: ${{ inputs.version }}" --allow-empty
+          git tag "version/${{ inputs.version }}" HEAD -m "version/${{ inputs.version }}"
+          git push --follow-tags
       - name: Create Release
-        id: create_release
         uses: actions/create-release@v1.1.4
```
dominic-r: is deprecated

.github/workflows/release-tag.yml
```diff
@@ -1,39 +1,179 @@
 ---
-name: Release - On tag
+name: Release - Tag new version
```
dominic-r, line 2: can we specify shell: bash and set -euo pipefail to scripts

rissson: no need, the stuff fails hard enough when a command fails

.github/workflows/release-tag.yml
```diff
@@ -1,39 +1,179 @@
...
+          jq \
+            --arg version "${{ inputs.version }}" \
+            --arg changelog "See ${changelog_url}" \
+            --arg changelog_url "${changelog_url}" \
+            '.stable.version = $version | .stable.changelog = $changelog | .stable.changelog_url = $changelog_url' version.json > version.new.json
+          mv version.new.json version.json
+      - name: Bump feature version
+        if: "${{ inputs.release_reason != 'feature' }}"
+        run: |
+          changelog_url="https://docs.goauthentik.io/docs/releases/${{ needs.check-inputs.outputs.major_version }}#fixed-in-$(echo -n ${{ inputs.version}} | sed 's/\.//g')"
```
dominic-r, line 173: wouldn't a if version != prerelease needed here as fixed in doesn't include the rc unless it's somewhere and I missed it

.github/workflows/release-tag.yml
```diff
@@ -1,39 +1,179 @@
...
+        if: "${{ inputs.release_reason == 'feature' }}"
+        run: |
+          changelog_url="https://docs.goauthentik.io/docs/releases/${{ needs.check-inputs.outputs.major_version }}"
+          jq \
+            --arg version "${{ inputs.version }}" \
+            --arg changelog "See ${changelog_url}" \
+            --arg changelog_url "${changelog_url}" \
+            '.stable.version = $version | .stable.changelog = $changelog | .stable.changelog_url = $changelog_url' version.json > version.new.json
+          mv version.new.json version.json
+      - name: Bump feature version
```
dominic-r: Can we rename this to avoid duplicate titles might be confusing unless it's intended this way?

.github/workflows/release-tag.yml
rissson: This doesn't run on RCs, so no need

.github/workflows/release-tag.yml
```diff
@@ -1,39 +1,190 @@
...
-        env:
-          GITHUB_TOKEN: ${{ steps.generate_token.outputs.token }}
+        uses: softprops/action-gh-release@v2
         with:
-          tag_name: ${{ github.ref }}
-          release_name: Release ${{ steps.ev.outputs.version }}
+          token: "${{ steps.app-token.outputs.token }}"
+          tag_name: "version/${{ inputs.version }}"
+          name: Release ${{ inputs.version }}
           draft: true
```
dominic-r, line 90: is it intended that all releases are marked as draft?

rissson: yes, so we can review and edit before actually clicking the button that might notify many people

.github/workflows/release-tag.yml
```diff
@@ -1,39 +1,190 @@
 ---
-name: Release - On tag
+name: Release - Tag new version
```
dominic-r, line 2: to prevent "who can click the release button first", as a safety measure. mabye:
```
concurrency:
  group: release-${{ github.ref || github.run_id }}
  cancel-in-progress: false
```

.github/workflows/release-tag.yml
```diff
@@ -1,39 +1,190 @@
...
-    name: Create Release from Tag
+  check-inputs:
+    name: Check inputs validity
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v5
-      - name: Pre-release test
+      - id: check
+        run: |
+          echo "${{ inputs.version }}" | grep -E "^[0-9]{4}\.[0-9]{1,2}\.[0-9]+(-rc[0-9]+)?$"
```
dominic-r:
```suggestion
          echo "${{ inputs.version }}" | grep -E '^[0-9]{4}\.(0?[1-9]|1[0-2])\.[0-9]+(-rc[0-9]+)?$'
```

found this one online which is more strict. confirmed on regex101
it works as expected

dominic-r: and mabye enforce version early too with:
```
if [[ "${{ inputs.release_reason }}" == "prerelease" && "${{ inputs.version }}" != *-rc* ]]; then
  echo "prerelease requires rc" >&2; exit 1
fi
if [[ "${{ inputs.release_reason }}" != "prerelease" && "${{ inputs.version }}" == *-rc* ]]; then
  echo "stable version doesn't allow rc" >&2; exit 1
fi
```

.github/workflows/release-tag.yml
rissson: We have a release manager for this exact reason. Worst case, one of the CI fails

dominic-r, line 2: the checkout step for tagging related, wherever it might be, would probably need fetch-depth: 0

.github/workflows/release-tag.yml
rissson: we might want to do a 2042.1.6-rc1, who knows

dominic-r (approved):
Overall LGTM happy to see this done!

.github/workflows/release-tag.yml
rissson: it doesn't, and we create a commit on top of the last one on the branch and tag that one anyway

BeryJu (approved):

PR #16168, opened 2025-08-13
web: Fix hidden textarea `required` attribute.

GirlBossRush:
## Details

A follow up on our recent fixes to `ak-secret-text-input` having a `required` attribute when editing existing entities. This PR applies the same treatment to `ak-secret-textarea`

## Included Fixes
- Fix missing flag property
- Clarify field error reporting


kensternberg-authentik (approved):
Had to double check to make sure any field that is _not_ revealed is also _not_ required. That could be hella frustrating. Looks good.

PR #16171, opened 2025-08-13
web: Improvements to ReCaptcha resizing 

GirlBossRush:
## Details

This PR focuses on a few visual issues that arise in the Captcha stage when using ReCaptcha.
As reported in #15598, Recaptcha v2 and v3 do not resize in response to their parent iframe's change in height. Additionally, ReCaptcha's (latest?) logic for handling domain verification isn't compatible with our switch to blob URLs as a source for frames.

The current fix for this limited due to the nature of this three-layer deep iframe structure.
We have a few options for fixing this:

1. Move the captcha out of the iframe, sacrificing the encapsulation of vendor scripts
2. Apply an estimated height to reCaptcha's iframe and trigger a resize event
3. Use a different captcha provider such as Turnstile or hCaptcha

This PR uses option 2, expanding our existing resize logic to traverse the DOM tree and resize all relevant elements. This is a bit of a hack, and it's limited to expanding the captcha's iframe, but it has the least impact on the other captcha vendors.

Also included is a fix captcha iframes not inheriting the color scheme of the parent page, ensuring **fewer** slivers of white around the captcha.

Screenshot 2025-08-13 at 19 57 09

Screenshot 2025-08-13 at 19 58 47

## What's not included

- The ability to configure recaptcha's `size` parameter
- Shrinking reCaptcha's iframe after the challenge is completed

---

Closes #15598

### DIFF
--- a/web/src/common/purify.ts
+++ b/web/src/common/purify.ts
@@ -105,7 +105,9 @@ export function renderStaticHTMLUnsafe(untrustedHTML: unknown): string {
 
     render(untrustedHTML, container);
 
-    const result = container.innerHTML;
-
+    const result = container.innerHTML
+        // Remove all comments as they can interfere with the styles.
+        .replaceAll("<!---->", "")
+        .replaceAll(/<!--\?lit\$\d+\$-->/g, "");
     return result;
 }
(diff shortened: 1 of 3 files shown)


kensternberg-authentik (approved):
So much fiddling.

PR #16198, opened 2025-08-15
web/a11y: QL Search Input

GirlBossRush:
## Details

This PR polishes the UX of the new query search input element and adds additional coverage to its ARIA attributes.

Screenshot 2025-08-15 at 04 54 02

## Fixes

- Query search elements now participate in client-side form submission and validation
- Elements within the search field now participate in ARIA states for expansion and selected menu item focus
- Suggestions now reposition when scrolling the page, and resizing the page
- Suggestions now close when focus moves away from element
- Suggestions no longer render outside the viewport, opting to scroll instead.
- Suggestions are now opened on input focus 
- Suggestions now provide ARIA labels that communicate the result of each logical operator
- Keyboard input no longer selects menu indexes outside of the list
- Keyboard interactions now conform to the [recommended ARIA behavior for comboboxes](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/combobox_role#keyboard_interactions)
- The textarea has a minimal and max height when resized

### DIFF
--- a/web/src/components/ak-search-ql/index.ts
+++ b/web/src/components/ak-search-ql/index.ts
@@ -1,14 +1,18 @@
 import "#elements/buttons/Dropdown";
 
-import { AKElement } from "#elements/Base";
+import { StripHTMLTrustPolicy } from "#common/purify";
+import { rootInterface } from "#common/theme";
+
+import { FormAssociated, FormAssociatedElement } from "#elements/forms/form-associated-element";
 import { PaginatedResponse } from "#elements/table/Table";
 
 import DjangoQL, { Introspections } from "@mrmarble/djangoql-completion";
 
 import { msg } from "@lit/localize";
-import { css, CSSResult, html, nothing, TemplateResult } from "lit";
-import { customElement, property, query, state } from "lit/decorators.js";
+import { css, CSSResult, html, LitElement, nothing, PropertyValues, TemplateResult } from "lit";
+import { customElement, property } from "lit/decorators.js";
 import { ifDefined } from "lit/directives/if-defined.js";
+import { createRef, ref, Ref } from "lit/directives/ref.js";
 
 import PFFormControl from "@patternfly/patternfly/components/FormControl/form-control.css";
 import PFSearchInput from "@patternfly/patternfly/components/SearchInput/search-input.css";
@@ -25,40 +29,26 @@ export class QL extends DjangoQL {
     textareaResize() {}
 }
 
-@customElement("ak-search-ql")
-export class QLSearch extends AKElement {
-    @property()
-    value?: string;
+/**
+ * Given an array or length, return logical index of the element at the given delta.
+ * This is effectively a modulo loop, allowing for positive and negative deltas.
+ */
+function torusIndex(lengthLike: number | ArrayLike<number>, delta: number): number {
+    const length = typeof lengthLike === "number" ? lengthLike : lengthLike.length;
 
-    @query("[name=search]")
-    searchElement?: HTMLTextAreaElement;
-
-    @state()
-    menuOpen = false;
-
-    @property()
-    onSearch?: (value: string) => void;
-
-    @state()
-    selected?: number;
-
-    @state()
-    cursorX: number = 0;
-
-    @state()
-    cursorY: number = 0;
-
-    ql?: QL;
-    canvas?: CanvasRenderingContext2D;
-
-    set apiResponse(value: PaginatedResponse<unknown> | undefined) {
-        if (!value || !value.autocomplete || !this.ql) {
-            return;
-        }
-        this.ql.loadIntrospections(value.autocomplete as unknown as Introspections);
+    if (delta < 0) {
+        return (length + delta) % length;
     }
 
-    static styles: CSSResult[] = [
+    return ((delta % length) + length) % length;
+}
+
+@customElement("ak-search-ql")
+export class QLSearch extends FormAssociatedElement<string> implements FormAssociated {
+    declare anchorRef: Ref<HTMLTextAreaElement>;
+    declare anchor: HTMLTextAreaElement | null;
+
+    public static styles: CSSResult[] = [
         PFBase,
         PFFormControl,
         PFSearchInput,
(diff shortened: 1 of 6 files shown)


web/src/elements/forms/form-associated-element.ts
```diff
@@ -0,0 +1,216 @@
...
+import { LitElement } from "lit";
+import { property } from "lit/decorators.js";
+import { createRef, Ref } from "lit/directives/ref.js";
+
+/**
+ * A subset of form associated {@linkcode ElementInternals} properties.
+ *
+ * @see {@linkcode FormAssociatedElement} for usage.
+ */
+export interface FormAssociated
```
GirlBossRush, line 15: This is largely ported from the first iteration in #16010 

I've gone back and forth on whether this is better off as a mixin -- my current thought is that it would be level of indirection that could make debugging the internals object a challenge. 

However, if we do want a mixin in the future, we can incorporate this base class instead of dynamically creating one like we usually do 😊

web/src/elements/forms/types.d.ts
```diff
@@ -0,0 +1,41 @@
...
+ * While these types are part of the HTML standard, they're not yet defined
+ * in the TypeScript standard library, so we define them here.
+ *
+ * @expires 2026-01-01
+ */
+
+/**
+ * Callbacks for form-associated elements.
+ */
+interface HTMLElement {
```
GirlBossRush, line 13: @kensternberg-authentik maybe we can send the TypeScript team a PR?

web/src/elements/table/TableSearch.ts
```diff
@@ -16,17 +17,23 @@ import PFToolbar from "@patternfly/patternfly/components/Toolbar/toolbar.css";
 import PFBase from "@patternfly/patternfly/patternfly-base.css";
 
 @customElement("ak-table-search")
-export class TableSearch extends WithLicenseSummary(AKElement) {
-    @property()
-    public value?: string;
+export class TableSearchForm extends WithLicenseSummary(AKElement) {
+    @property({ type: String, reflect: false })
+    public defaultValue?: string;
```
GirlBossRush, line 22: This is a really tricky property since the element here is more like a form than a single search input. We also want to avoid communicating that the value is timely after the user interacts with the search field. IMO the name is fitting since the input `defaultValue` property [expresses a similar concept](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/defaultValue)

web/src/components/ak-search-ql/index.ts
```diff
@@ -25,248 +29,436 @@ export class QL extends DjangoQL {
...
+
+    //#endregion
+
+    //#region State
+
+    #menuRef = createRef<HTMLDivElement>();
+
+    #ql: QL | null = null;
+    #ctx: OffscreenCanvasRenderingContext2D | null = null;
+    #letterWidth = -1;
```
GirlBossRush, line 143: Calculate this once, zoom zoom

web/src/components/ak-search-ql/index.ts
```diff
@@ -25,248 +29,436 @@ export class QL extends DjangoQL {
...
+        this.internals.ariaAutoComplete = "list";
+        this.internals.role = "combobox";
+        this.internals.ariaHasPopup = "listbox";
+
+        this.#scrollContainer =
+            rootInterface<LitElement>().renderRoot.querySelector("#main-content");
+
+        this.#scrollContainer?.addEventListener("scroll", this.#updateDropdownPosition, {
+            passive: true,
+        });
```
GirlBossRush, line 170: This a bit awkward but gets us our reference. We might be able to set the interface and router to an open root to make these easier to grab in the future

web/src/components/ak-search-ql/index.ts
```diff
@@ -25,248 +29,436 @@ export class QL extends DjangoQL {
...
         }
-        context.font = window.getComputedStyle(this.searchElement).font;
-        this.canvas = context;
+
+        this.#ctx.font = window.getComputedStyle(textarea).font;
+
+        // We need the width of a letter to measure x; we use a monospaced font but still
+        // check the length for `m` as its the widest ASCII char
+        const metrics = this.#ctx?.measureText("m");
+        this.#letterWidth = Math.ceil(metrics?.width || 0) + 1;
```
GirlBossRush, line 238: Previously recomputed on every paint, now cached.

web/src/components/ak-search-ql/index.ts
```diff
@@ -25,248 +29,436 @@ export class QL extends DjangoQL {
...
-                // Control keys shouldn't trigger completion popup
-                break;
+                this.open = false;
+                return;
         }
-    }
+    };
+
+    #blurListener = ({ relatedTarget }: FocusEvent) => {
+        if (relatedTarget instanceof Node && this.renderRoot.contains(relatedTarget)) {
```
GirlBossRush, line 409: I found about `relatedTarget` today. This beats having to sniff around the window with a global click handler 😎

web/src/components/ak-search-ql/index.ts
```diff
@@ -25,248 +29,436 @@ export class QL extends DjangoQL {
...
+            <div ${ref(this.#menuRef)} class="pf-c-search-input__menu">
+                <ul
+                    class="pf-c-search-input__menu-list"
+                    role="listbox"
+                    id="ql-suggestions"
+                    aria-label=${msg("Query suggestions")}
+                >
+                    ${this.#ql.suggestions.map((suggestion, idx) => {
+                        // Cast to string to sooth Lit Analyzer's primitive type rule.
+                        const label = `${StripHTMLTrustPolicy.createHTML(suggestion.suggestionText)}`;
```
GirlBossRush, line 441: We might be able to remove this later if the backend can change the suggestion text to not include wrapping `<i>` tags

web/src/components/ak-search-ql/index.ts
```diff
@@ -25,248 +29,436 @@ export class QL extends DjangoQL {
...
+
+    @property({ type: String })
+    public get value(): string {
+        return this.#value;
+    }
+
+    public set value(value: unknown) {
+        const parsed = typeof value === "string" ? value : "";
+
+        this.setFormValue(parsed.trim(), parsed);
```
GirlBossRush, line 127: Documentation on this method is sparse, but AFAICT this is the suggested approach. Store the idealized value reported to the form, and what the user provided us verbatim for future input restoration.

BeryJu (approved):

PR #16204, opened 2025-08-15
web: Group stages by type.

GirlBossRush:
## Details

Similar to our row grouping on other tables, this PR groups stages by their type, which helped greatly when swapping between debugging different captcha configurations.

## Before

Screenshot 2025-08-15 at 14 18 03

## After

Screenshot 2025-08-15 at 14 17 39


BeryJu (changes requested):
We used to do this but removed it as the grouping isn't consistent across pages as it's only grouped in the frontend and so sorting by name might still put different categories across different pages

PR #16208, opened 2025-08-15
Change to apt-get dist-upgrade and add upgrade operations

TommyTran732:
- `apt-get upgrade` does not remove unnecessary packages, while `apt-get dist-upgrade` does. We want to limit the attack surface by removing unnecessary packages.
- It is generally good practice to have an up to date environment before building packages. We don't want stuff like vulnerable openssl or tar or whatever to affect the build.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,13 +9,15 @@ ENV NODE_ENV=production
 
 WORKDIR /work/web
 
+RUN --mount=type=cache,id=apt-$TARGETARCH$TARGETVARIANT,sharing=locked,target=/var/cache/apt \
+    apt-get update && \
+    apt-get dist-upgrade -y
+
 RUN --mount=type=bind,target=/work/web/package.json,src=./web/package.json \
     --mount=type=bind,target=/work/web/package-lock.json,src=./web/package-lock.json \
     --mount=type=bind,target=/work/web/packages/sfe/package.json,src=./web/packages/sfe/package.json \
     --mount=type=bind,target=/work/web/scripts,src=./web/scripts \
     --mount=type=cache,id=npm-ak,sharing=shared,target=/root/.npm \
-    apt-get update && \
-    apt-get dist-upgrade && \
     npm ci
 
 COPY ./package.json /work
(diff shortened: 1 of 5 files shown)


TommyTran732:
Sorry about the title, GitHub messed up my PR. Should be fixed now.

rissson:
will take a look at this after release

Dockerfile
```diff
@@ -14,6 +14,8 @@ RUN --mount=type=bind,target=/work/web/package.json,src=./web/package.json \
     --mount=type=bind,target=/work/web/packages/sfe/package.json,src=./web/packages/sfe/package.json \
     --mount=type=bind,target=/work/web/scripts,src=./web/scripts \
     --mount=type=cache,id=npm-ak,sharing=shared,target=/root/.npm \
+    apt-get update && \
+    apt-get dist-upgrade && \
```
rissson: This is not needed as only build artifacts end up in the final image.

Dockerfile
```diff
@@ -40,6 +42,7 @@ WORKDIR /go/src/goauthentik.io
 RUN --mount=type=cache,id=apt-$TARGETARCH$TARGETVARIANT,sharing=locked,target=/var/cache/apt \
     dpkg --add-architecture arm64 && \
     apt-get update && \
+    apt-get dist-upgrade && \
```
rissson: This is not needed as only build artifacts end up in the final image.

Dockerfile
```diff
@@ -103,6 +106,7 @@ ENV PATH="/root/.cargo/bin:$PATH"
 
 RUN --mount=type=cache,id=apt-$TARGETARCH$TARGETVARIANT,sharing=locked,target=/var/cache/apt \
     apt-get update && \
+    apt-get dist-upgrade && \
```
rissson: This is done in https://github.com/goauthentik/fips as one of the final steps, so we can save time here.

Dockerfile
```diff
@@ -149,7 +153,7 @@ WORKDIR /
 
 # We cannot cache this layer otherwise we'll end up with a bigger image
 RUN apt-get update && \
-    apt-get upgrade -y && \
+    apt-get dist-upgrade -y && \
```
rissson: This is done in [goauthentik/fips](https://github.com/goauthentik/fips) as one of the final steps, so we can save time here.

rissson (changes requested):

rissson:
I'm gonna close this as the changes I'm actually interested in should be done in https://github.com/goauthentik/fips

Dockerfile
TommyTran732: It still matters because you do not want to build your stuff in an outdated environment

TommyTran732: @rissson I hope you would consider because say if something uses curl or openssl during the build step and they get exploited it would be pretty bad

rissson: I will agree for `node` images, as they're not updated as often as our goauthentik/fips images (every day).

rissson:
@TommyTran732 I've updated the PRs with my comments, and also updated where needed in other docker images. Please take a look and let me know what you think.

BeryJu:
Closing in favour of having our daily updated and upgraded FIPS images and not needing to run upgrade in each docker build to have more consistent builds

PR #16223, opened 2025-08-18
website/docs: Add steps for fixing xml python errors, clean up

PeshekDotDev:
## Details

There is a bug with the SAML tests i've been able to reproduce across 3 macs. I have found the solution and fixed it across 3 macs as well. I added the steps to fix this into the dev documentation, as if this isn't fixed, when someone runs `make all`, their tests will never pass.

I did some cleanup as well in the docs while I was in there.

I made a change to the order that `make all` fires for sanity, tests at the end feels a bit better because failing on `gen` or `web` afterwards can feel frustrating

---

## Checklist

If applicable


website/docs/developer-docs/setup/full-dev-environment.mdx
```diff
@@ -237,7 +237,11 @@ When `AUTHENTIK_DEBUG` is set to `true` (the default for the development environ
 
 ## End-to-End (E2E) Setup
 
-To run E2E tests, navigate to the `/tests/e2e` directory in your local copy of the authentik git repo, and start the services by running `docker compose up -d`.
+Start the e2e test services with the following command:
```
tanberry: @ConzorKingKong I'm a big fan of telling people where they need to be when they run a command, so at first I was wondering why you removed that, but I see it is part of the command itself, so they can run it from wherever, right?

PeshekDotDev: Technically the command has to be ran from the root of the folder that the app is in. But this fits better with how we handle the other docker compose script in the initial steps and all the make commands. Basically every other command is set to run in the root of the project and this change makes this fit that pattern

website/docs/developer-docs/setup/full-dev-environment.mdx
```diff
@@ -58,16 +58,39 @@ If using locally installed databases, ensure the PostgreSQL credentials provided
...
 node@24 \
-golangci-lint
+golangci-lint \
+krb5
+```
+
+Don't forget to add `krb5` to your path or else you may run into problems with Kerberos functionality:
+
+```shell
+echo 'export PATH="$(brew --prefix krb5)/bin:$PATH"' >> ~/.bashrc
```
dominic-r: default mac shell is zsh not bash

website/docs/developer-docs/setup/full-dev-environment.mdx
```diff
@@ -58,16 +58,39 @@ If using locally installed databases, ensure the PostgreSQL credentials provided
...
+```
+
+Don't forget to add `krb5` to your path or else you may run into problems with Kerberos functionality:
+
+```shell
+echo 'export PATH="$(brew --prefix krb5)/bin:$PATH"' >> ~/.zshrc
+echo 'export PATH="$(brew --prefix krb5)/sbin:$PATH"' >> ~/.zshrc
+```
+
+Since Mac comes with a default install of `libxml2`, python may not build `xmlsec` and `lxml` properly with the `libxml2` version installed by `libxmlsec1` unless you update your build flags:
```
dominic-r:
```suggestion
Since macOS includes a default installation of `libxml2`, Python may fail to build `xmlsec` and `lxml` correctly when using the `libxml2` version provided by `libxmlsec1`, unless you update your build flags.
```

website/docs/developer-docs/setup/full-dev-environment.mdx
```diff
@@ -58,16 +58,39 @@ If using locally installed databases, ensure the PostgreSQL credentials provided
...
+
+```shell
+echo 'export LDFLAGS="-L$(brew --prefix libxml2)/lib $LDFLAGS"' >> ~/.zshrc
+echo 'export CPPFLAGS="-I$(brew --prefix libxml2)/include $CPPFLAGS"' >> ~/.zshrc
+echo 'export PKG_CONFIG_PATH="$(brew --prefix libxml2)/lib/pkgconfig:$PKG_CONFIG_PATH"' >> ~/.zshrc
+```
+
+If you ran `make install` before adding the build flags for `libxml2`, you will have to run the following in the root of your authentik project to fix any SAML related errors:
```
dominic-r:
```suggestion
If you ran `make install` before adding the `libxml2` build flags, you will need to run the following command in the root of your authentik project to resolve any SAML-related errors.
```

website/docs/developer-docs/setup/website-dev-environment.mdx
```diff
@@ -0,0 +1,80 @@
...
+
+## 1. Install dependencies
+
+<Tabs defaultValue="Mac">
+<TabItem value="Mac">
+
+Install the required dependencies on macOS using Homebrew:
+
+```shell
+xcode-select --install
```
dominic-r: i dont think we need this for docs only?

website/docs/developer-docs/setup/website-dev-environment.mdx
```diff
@@ -0,0 +1,80 @@
...
+
+## Prerequisites
+
+- [Node.js](https://nodejs.org/en) (24 or later)
+- [Make](https://www.gnu.org/software/make/) (3 or later)
+
+## 1. Install dependencies
+
+<Tabs defaultValue="Mac">
+<TabItem value="Mac">
```
dominic-r:
```suggestion
<TabItem value="macOS">
```

website/docs/developer-docs/setup/website-dev-environment.mdx
```diff
@@ -0,0 +1,80 @@
...
+If you want to only make changes to the documentation, you only need Node.js.
+
+## Prerequisites
+
+- [Node.js](https://nodejs.org/en) (24 or later)
+- [Make](https://www.gnu.org/software/make/) (3 or later)
+
+## 1. Install dependencies
+
+<Tabs defaultValue="Mac">
```
dominic-r:
```suggestion
<Tabs defaultValue="macOS">
```

website/docs/developer-docs/setup/full-dev-environment.mdx
PeshekDotDev: Thank you, I remember having to install zsh on (i think linux) many moons ago, never clicked that its default on mac

website/docs/developer-docs/setup/website-dev-environment.mdx
```diff
@@ -0,0 +1,80 @@
...
+<TabItem value="Windows">
+
+We're currently seeking community input on running the docs in Windows. If you have experience with this setup, please consider contributing to this documentation.
+
+</TabItem>
+</Tabs>
+
+## 2. Instructions
+
+### Clone the repo
```
dominic-r: do we really need headers? it's just codeblocked commands with no text and takes up unneeded space imo

website/docs/developer-docs/setup/website-dev-environment.mdx
PeshekDotDev: This installs make. we could say brew install it, but ive never installed make that way

website/docs/developer-docs/setup/website-dev-environment.mdx
```diff
@@ -0,0 +1,80 @@
...
+
+### Prepare your changes for commit
+
+This will run the linter and auto-formatter.
+
+```shell
+make docs
+```
+
+Now you can push up your changes and send us a pull request.
```
dominic-r: I feel like this lacks description compared to the rest. I'd suggest looking at our other guides for reference

website/docs/developer-docs/setup/full-dev-environment.mdx
```diff
@@ -58,16 +58,39 @@ If using locally installed databases, ensure the PostgreSQL credentials provided
...
+
+```shell
+echo 'export PATH="$(brew --prefix krb5)/bin:$PATH"' >> ~/.zshrc
+echo 'export PATH="$(brew --prefix krb5)/sbin:$PATH"' >> ~/.zshrc
+```
+
+Since macOS includes a default installation of `libxml2`, Python may fail to build `xmlsec` and `lxml` correctly when using the `libxml2` version provided by `libxmlsec1`, unless you update your build flags.
+
+```shell
+echo 'export LDFLAGS="-L$(brew --prefix libxml2)/lib $LDFLAGS"' >> ~/.zshrc
```
dominic-r: also, do we really want to define LDFLAGS on our entire system just for this singular xml thing? Seems like a bad idea

dominic-r: ldflags and the 2 below as well*

website/docs/developer-docs/setup/website-dev-environment.mdx
dominic-r: Creates an unneeded dependency on xcode, brew is easier

dominic-r: nvm doesn't look like it, but it's an extra program one needs to install

website/docs/developer-docs/setup/website-dev-environment.mdx
PeshekDotDev: updated

website/docs/developer-docs/setup/website-dev-environment.mdx
PeshekDotDev: done

website/docs/developer-docs/setup/website-dev-environment.mdx
PeshekDotDev: done

website/docs/developer-docs/setup/full-dev-environment.mdx
PeshekDotDev: I can try and add them manually to the makefile. its a bad idea for the user to not have them set imo. they'll just get stuck using the old version. but we can ignore giving the user instructions and try to update the makefile directly. May need someone like jens to check this pr out now though, I'm not a make expert or anything, could be unintended side effects, but it should be fine

website/docs/developer-docs/setup/website-dev-environment.mdx
PeshekDotDev: changed to sentences

website/docs/developer-docs/setup/website-dev-environment.mdx
PeshekDotDev: I mean it's pretty benign overall. If someone runs it and they already have x-code installed, it skips it. and if they dont have it installed, they probably should. but its a long install.

currently we list javascript and make as pre-reqs. however, we add installing node to our scripts on the full dev setup. I added make just because it made the installs bulletproof, but adding pre-reqs to the install scripts kind of breaks the pattern that we usually hold which is "prereqs are pre requirements, install them on your own beforehand".

So I think if I'm removing xcode, I should remove nodejs from the install scripts. or remove them both from the pre-reqs.

To be honest I'm not very strongly held to any option. we can

1. remove nodejs and make from the prereqs and keep the install scripts
2. keep nodejs and make in the prereqs and remove them from both docs pages install scripts (meaning docs page has no install scripts)
3. keep them both in the prereqs and the install to both guarantee the user has the proper things installed and the user has context of what is being installed by it being in the prereqs (currently implemented in the PR)

PeshekDotDev: For now, I removed xcode install from install steps and left everything else as is. The prereqs wording is a minor but notable thing. I'm willing to come to an answer on it or just leave it for another PR

tanberry: Thank you. Having fought with Xcode and all those gems and jewels, yeah, I am not a fan.

tanberry: Node is a requirement for Docs too, I am pretty sure. I vote for #3 of those choices above, but I would like @GirlBossRush to take a look too please.

BeryJu (approved):
👍 also good idea with adding it to the makefile

Makefile
```diff
@@ -18,7 +18,17 @@ pg_host := $(shell uv run python -m authentik.lib.config postgresql.host 2>/dev/
...
 
-all: lint-fix lint test gen web  ## Lint, build, and test everything
+UNAME := $(shell uname)
+
+## For macOS users, add the libxml2 and libxmlsec1 installed from brew to the build path
+## to prevent SAML-related tests from failing and ensure correct compilation
+ifeq ($(UNAME), Darwin)
+BREW_LDFLAGS := -L$(shell brew --prefix libxml2)/lib $(LDFLAGS)
+BREW_CPPFLAGS := -I$(shell brew --prefix libxml2)/include $(CPPFLAGS)
+BREW_PKG_CONFIG_PATH := $(shell brew --prefix libxml2)/lib/pkgconfig:$(PKG_CONFIG_PATH)
```
dominic-r, line 28:
```suggestion
BREW_PKG_CONFIG_PATH := $(shell brew --prefix libxml2)/lib/pkgconfig:$(shell brew --prefix libxmlsec1)/lib/pkgconfig:$(PKG_CONFIG_PATH)
```

website/docs/developer-docs/setup/website-dev-environment.mdx
```diff
@@ -0,0 +1,85 @@
...
+
+</TabItem>
+<TabItem value="Linux">
+
+For Debian/Ubuntu-based distributions:
+
+```shell
+sudo apt-get install -y \
+build-essential \
+nodejs
```
dominic-r: Many distros ship with old versions of node. Debian trixie packages node 20 (https://packages.debian.org/trixie/nodejs) and ubuntu lts packages with node 18 (https://packages.ubuntu.com/noble/nodejs). Anyways, NVM is the recommended installation method... See: https://nodejs.org/en/download

dominic-r (approved):
LTGM

Makefile
PeshekDotDev: I don't think we have to add libxmlsec1 for this to work, right? Did you run into an error testing?

dominic-r: in pyproject.toml, we use [`xmlsec==1.3.16`](https://github.com/goauthentik/authentik/blob/main/pyproject.toml#L73) and on [pypi](https://pypi.org/project/xmlsec/) we see the homepage as https://github.com/xmlsec/python-xmlsec which states requirements for libxml2 and libxmlsec1 in https://github.com/xmlsec/python-xmlsec#requirements and in the macos section it states that both must be installed: https://github.com/xmlsec/python-xmlsec#mac

but no, didn't test it

website/docs/developer-docs/setup/full-dev-environment.mdx
```diff
@@ -214,7 +214,11 @@ When `AUTHENTIK_DEBUG` is set to `true` (the default for the development environ
 
 ## End-to-End (E2E) Setup
 
-To run E2E tests, navigate to the `/tests/e2e` directory in your local copy of the authentik git repo, and start the services by running `docker compose up -d`.
+Start the e2e test services with the following command:
```
dominic-r:
```suggestion
Start the E2E test services with the following command:
```

Makefile
tanberry: Where are with this? If there is any chance that the extra clarity is needed, then let's include it, i.e. leave as @ConzorKingKong has it.

website/docs/developer-docs/setup/website-dev-environment.mdx
tanberry: True, @ConzorKingKong  since we require v.24, we should make it clear that whatever tool installs Node, it needs to be v 24.

dominic-r: basically

tanberry:
Hiya @ConzorKingKong I think the only thing that needs to be done still for this PR is to modify the section that installs node without enforcing v 24. Let me know when that's done and let's get this merged. Thanks!

PeshekDotDev:
@tanberry alright, I can do that. The only good way to really do this on linux is to have people install a version manager and install through that. Let me try to make a combination that is straightforward and clean for this

Makefile
PeshekDotDev: There should be no reason to add libxmlsec1 to the path, it already exists because we asked the user to install libxmlsec1 in our install steps. We only have to do this with libxml2 because of an error with the system default version being used when lxml gets built

PeshekDotDev:
I asked them to download the latest version from the nodejs website and linked to it

website/docs/developer-docs/setup/website-dev-environment.mdx
```diff
@@ -0,0 +1,79 @@
...
+---
+
+import TabItem from "@theme/TabItem";
+import Tabs from "@theme/Tabs";
+
+If you want to only make changes to the documentation, you only need Node.js.
+
+## Prerequisites
+
+- [Node.js](https://nodejs.org/en) (24 or later)
```
dominic-r, line 18:
```suggestion
- [Node.js v24](https://nodejs.org/en)
```

First of all 24 is the latest, but also with all the javascript world changes and compatibility/reqs it's best to pin to a specific major version

PeshekDotDev: The version "24 or later" was written like that because it was written that way in the full development environment and frontend development environment pages. If we change this, then we have to update those locations and all of our other prereqs to change their version location. For now, I'd like to focus on getting this PR in for the Makefile changes and then circle back to this later

The frontend development environment page needs some refreshing to match these pages, so we could make a PR later for that page and spruce this up in that PR

dominic-r: sure. if you want

PeshekDotDev: Yeah I can come back to it after this is all merged in it's nbd

website/docs/developer-docs/setup/website-dev-environment.mdx
```diff
@@ -0,0 +1,79 @@
...
+    - development
+    - contributor
+    - docs
+    - docusaurus
+---
+
+import TabItem from "@theme/TabItem";
+import Tabs from "@theme/Tabs";
+
+If you want to only make changes to the documentation, you only need Node.js.
```
tanberry, line 14: Does one also need Make, as the current docs say? @ConzorKingKong

tanberry (approved):
Let's merge it @ConzorKingKong !

PR #16224, opened 2025-08-18
core: allow creating tokens for other user without superuser

BeryJu:
## Details

REPLACE ME

---

## Checklist

If applicable


rissson:
We might want to add a few tests for this.

authentik/core/api/tokens.py
```diff
@@ -46,6 +47,11 @@ def validate_user(self, user: User):
         if self.instance and self.instance.user_id:
             if user.pk != self.instance.user_id:
                 raise ValidationError("User cannot be changed")
+        request: HttpRequest = self.context.get("request")
+        if not request.user.has_perm("change_user", user) and not request.user.has_perm(
+            "authentik_core.change_user"
+        ):
```
gergosimonyi, line 53: No need to check separately, `has_perm` checks both.

```suggestion
        if not request.user.has_perm("authentik_core.change_user", user):
```

authentik/core/api/tokens.py
```diff
@@ -46,6 +47,11 @@ def validate_user(self, user: User):
...
+            raise ValidationError("Cannot create token for this user")
```
gergosimonyi, line 54: Errors like this should not be raised in `validate_*` functions, but earlier in the process.

gergosimonyi (commented):
This is very suspicious. It feels like `change_user` shouldn't ever be enough to create a token that lets me do anything in that user's name.

I believe a larger restructure is needed. This shouldn't just be blanket bearer token. Instead, it should carry the information that "I am a token User A created to (in effect) impersonate User B". Then the backend could have some say in whether it wants to allow that.

PR #16231, opened 2025-08-18
web: Fix reported error precedence

GirlBossRush:
## Details

- Fix issue where controlled element is not assigned, such as when ak-search-select attempts to pluck its own name attribute from a parent form field.
- Fix the preferred order of errors to display when API response includes a more descriptive validation.
- Fix issue where impersonation form can be submitted with empty fields

### DIFF
--- a/web/src/common/errors/network.ts
+++ b/web/src/common/errors/network.ts
@@ -6,6 +6,8 @@ import {
     ValidationErrorFromJSON,
 } from "@goauthentik/api";
 
+import { sentenceCase } from "change-case";
+
 //#region HTTP
 
 /**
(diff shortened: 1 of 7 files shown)


web/src/common/errors/network.ts
```diff
@@ -233,3 +235,23 @@ export async function parseAPIResponseError<T extends APIError = APIError>(
...
+
+/**
+ * Pluck a field error from a validation error.
+ *
+ * This is used to create a fallback error message when the API returns
+ * a validation error that isn't associated with field within the form.
+ *
+ * We can still show the error message, to at least give the user some feedback.
+ */
+export function pluckFallbackFieldErrors(parsedError: APIError): string[] {
```
GirlBossRush, line 249: Largely brought out from the form element. We can use this elsewhere for QL search

web/src/elements/forms/Form.ts
```diff
@@ -143,6 +149,41 @@ export function serializeForm<T = Record<string, unknown>>(elements: Iterable<AK
     return json as unknown as T;
 }
 
+//#region Validation Reporting
+
+/**
+ * Assign all input-related errors to their respective elements.
+ */
+function reportInvalidFields(
```
GirlBossRush, line 157: Largely pulled out of the send method's catch clause to clarify the stage of  reporting of field errors from the business of parsing API errors.

web/src/elements/forms/HorizontalFormElement.ts
```diff
@@ -70,7 +70,20 @@ export class HorizontalFormElement extends AKElement {
 
     //#endregion
 
-    public controlledElement: NamedElement | AkControlElement | null = null;
+    #controlledElement: AkControlElement | NamedElement | null = null;
+
+    /**
+     * The element that should be focused when the form is submitted.
+     */
+    public get focusTarget(): AkControlElement | NamedElement<HTMLElement> | null {
```
GirlBossRush, line 78: This, along with most of horizontal form element's non-visual behavior, can later be served by the native `attachInternals`

kensternberg-authentik (approved):
@GirlBossRush  and I discussed some of the loose / suspect Typing. For the moment, しょうがない.  The flow of logic looks solid.

PR #16236, opened 2025-08-18
web: Fix ak-flow-card footer alignment.

GirlBossRush:
## Details

Caused by nested flex containers. Solution -- set parent flex group to block.

### Before

Screenshot 2025-08-18 at 19 40 36

### After
Screenshot 2025-08-18 at 19 40 09

### DIFF
--- a/web/src/flow/components/ak-flow-card.ts
+++ b/web/src/flow/components/ak-flow-card.ts
@@ -33,13 +33,10 @@ export class FlowCard extends AKElement {
         PFLogin,
         PFTitle,
         css`
-            slot[name="footer"],
-            slot[name="footer-band"] {
-                display: flex;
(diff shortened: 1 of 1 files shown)


web/src/flow/components/ak-flow-card.ts
```diff
@@ -33,13 +33,10 @@ export class FlowCard extends AKElement {
         PFLogin,
         PFTitle,
         css`
-            slot[name="footer"],
-            slot[name="footer-band"] {
-                display: flex;
-                flex-wrap: wrap;
-                justify-content: center;
-                flex-basis: 100%;
```
GirlBossRush, line 41: Note to self: I suspect this also needs refining during an a11y clean up.

BeryJu (approved):

PR #16243, opened 2025-08-18
sources/ldap: fix malformed filter error with special characters in group DN

jung0han:
## Details

Fixed LDAPInvalidFilterError that occurs when group DN contains special characters (parentheses, backslashes, asterisks) during LDAP membership synchronization.

The issue was that group DN values were being used directly in LDAP filter construction without proper escaping, causing malformed filter errors when special characters were present. This fix uses ldap3's
`escape_filter_chars()` function to properly escape these characters before constructing the LDAP filter.

Changes made:
- Added import for `escape_filter_chars` from `ldap3.utils.conv`
- Updated group filter construction in `MembershipLDAPSynchronizer.sync_page()` to escape the group DN value
- Added comprehensive test cases covering various special characters and edge cases
- Tests verify both the membership synchronization process and the escape function directly

---

## Checklist

- [x] Local tests pass (`ak test authentik/`)

- [ ] The API schema has been updated (`make gen-build`)

- [ ] The code has been formatted (`make web`)

If applicable

- [ ] The documentation has been updated
- [ ] The documentation has been formatted (`make docs`)

### DIFF
--- a/authentik/sources/ldap/tests/test_sync.py
+++ b/authentik/sources/ldap/tests/test_sync.py
@@ -528,29 +528,29 @@ class LDAPSyncTests(TestCase):
         self.source.group_object_filter = "(objectClass=groupOfNames)"
         self.source.lookup_groups_from_user = True
         self.source.group_membership_field = "memberOf"
-        
+
         # Mock connection with group DN containing special characters
         mock_conn = MagicMock()
-        
+
         # Simulate group with special characters in DN: parentheses, backslashes, asterisks
         special_group_dn = "cn=test(group),ou=groups,dc=example,dc=com"
         backslash_group_dn = "cn=test\\group,ou=groups,dc=example,dc=com"
         asterisk_group_dn = "cn=test*group,ou=groups,dc=example,dc=com"
-        
+
         # Mock the paged_search method that would be called with the filter
         mock_standard = MagicMock()
         mock_conn.extend.standard = mock_standard
-        
+
         # Test case 1: Group DN with parentheses
         with patch("authentik.sources.ldap.models.LDAPSource.connection", return_value=mock_conn):
             membership_sync = MembershipLDAPSynchronizer(self.source, Task())
-            
+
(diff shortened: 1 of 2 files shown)


authentik/sources/ldap/tests/test_sync.py
```diff
@@ -519,3 +520,90 @@ def test_batch_deletion(self):
...
+                call_args = mock_standard.paged_search.call_args
+                search_filter = call_args[1]["search_filter"]
+                # The asterisk should be escaped as \2a
+                self.assertIn("\\2a", search_filter)  # Escaped *
+            except LDAPInvalidFilterError:
+                self.fail("LDAPInvalidFilterError should not be raised with escaped filter")
+
+    def test_escape_filter_chars_function(self):
+        """Test the escape_filter_chars function directly"""
+        from ldap3.utils.conv import escape_filter_chars
```
BeryJu: Import should be at the top of the file

BeryJu (approved):
LGTM, are there any other places where we should/need to use `escape_filter_chars`?

authentik/sources/ldap/tests/test_sync.py
jung0han: Good catch — I moved the inline import to the top of authentik/sources/ldap/tests/test_sync.py and updated the test accordingly. Thanks for the nudge!

jung0han:
> LGTM, are there any other places where we should/need to use `escape_filter_chars`?

Thanks! I audited all places where we build LDAP filters.

The only spot where we interpolate directory values (DNs/attrs) into a filter is in MembershipLDAPSynchronizer when lookup_groups_from_user is enabled; that path now escapes the DN with ldap3’s escape_filter_chars.

Other usages rely on admin-configured static filters (self._source.user_object_filter / group_object_filter), fixed filters like (objectClass=*), or logic that doesn’t construct LDAP filters. So there aren’t any other places that need escaping right now.

PR #16246, opened 2025-08-19
core: remove ability to add providers with the same name

PeshekDotDev:
## Details

Currently, a user can make a provider of name "provider" and "PROVIDER". This PR prevents adding a provider of the same name with different casing when submitting a provider form.

---

## Checklist

If applicable

### DIFF
--- a/authentik/core/api/providers.py
+++ b/authentik/core/api/providers.py
@@ -6,6 +6,7 @@ from django.utils.translation import gettext_lazy as _
 from django_filters.filters import BooleanFilter
 from django_filters.filterset import FilterSet
 from rest_framework import mixins
+from rest_framework.exceptions import ValidationError
 from rest_framework.fields import ReadOnlyField, SerializerMethodField
 from rest_framework.viewsets import GenericViewSet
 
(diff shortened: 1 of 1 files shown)


rissson:
How much of an issue is it that providers have the same case insensitive name? I'm pretty sure we use the application slug (which should be case insensitive) for things like URL components

PeshekDotDev:
At the end of the day you are right, if a user wanted to do so, they could make the same named provider and could attach it to an application that would have a unique slug. However, it seems like a recipe to get unnecessary bug tickets or easily confused users. "Oh, I forgot I made a "google" provider for oauth and a "Google" provider for saml", etc. It seems like something a user would do more because they forgot a provider with that name already existed.

It's not something where I go through and change database models or anything, just make it so when you post the form, we block it if the name exists. You can still manually make this happen at the database level if you have a reason for it. I just can't really think of a good reason why a user would want the ability for the same name with different casing, and I can think of problems it would prevent if they couldn't

dominic-r (approved):
ooo Thanks!

gergosimonyi:
(Not a review, just notes.)

This is a breaking change. If someone already has `Provider` and `provider`, they will be unable to update these without a name change.

If the plan is to breaking things, perhaps break them loudly (and correctly) at the database level via an [expressional index](https://www.postgresql.org/docs/current/indexes-expressional.html) on `lower(name)` or, uh, [the hard way](https://adamj.eu/tech/2023/02/23/migrate-django-postgresql-ci-fields-case-insensitive-collation/).

PR #16259, opened 2025-08-19
website: QL Search keyboard interactions docs, examples.

GirlBossRush:
## Details

This PR refines our QL search docs to be compatible with screen readers. Changes include:

- A table of keyboard interactions with ARIA attributes to clarify their use.
- Updates the example queries to allow the screen reader to better announce the contents, and to allow the user to copy the content to try in their authentik instance

## Screenshot

Screenshot 2025-08-19 at 21 54 41

## ARIA Tree

Screenshot 2025-08-19 at 21 59 10

### DIFF
--- a/website/docs/endpoint-devices/device-compliance/device-reporting.md
+++ b/website/docs/endpoint-devices/device-compliance/device-reporting.md
@@ -37,7 +37,7 @@ For an example of the facts provided for a Linux device, see [here](./facts-linu
 
 ## Endpoint devices in event logs
 
-Authentication events involving endpoint devices are included in the [event logs](../../sys-mgmt/events/logging-events.md). For example:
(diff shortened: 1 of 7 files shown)


dewi-tik (approved):
LGTM but will need to be updated

tanberry (approved):
Wow this is really great, very useful, especially the example queries! Thanks so much @GirlBossRush and @dewi-tik !

1. We need some text between the title "Keyboard interactions" (need lower-case I there) and the start of the table... explaining where exactly these keyboard shortcuts can be used... I can deduce by looking at the right-side mini-ToC that they are for searching event logs, but it would be useful to have that spelled out at top of this page.

2. Let's lowercase the "I" on the header "Keyboard Interactions".

dominic-r (approved):

PR #16276, opened 2025-08-20
web: Fix issue where form group uses unknown slot.

GirlBossRush:
## Details

Introduced in #12963, this component was revised to use the default slot at the same time the new fields were introduced.

## See also

- Cherry-pick #16277


kensternberg-authentik (approved):
This is correct.

PR #16277, opened 2025-08-20
web: Fix form group slots (cherry-pick #16276)

GirlBossRush:
## Details

See #16276


kensternberg-authentik (approved):
Not broken, but where did the lovely slots go?

PR #16280, opened 2025-08-20
web/a11y: QL Search Input (cherry-pick #16198)

GirlBossRush:
web: Fix issues surrounding form submission, keyboard focus, alignment.

## Details

See #16198

### DIFF
--- a/web/src/admin/sources/oauth/OAuthSourceForm.ts
+++ b/web/src/admin/sources/oauth/OAuthSourceForm.ts
@@ -530,9 +530,8 @@ export class OAuthSourceForm extends WithCapabilitiesConfig(BaseSourceForm<OAuth
                     </ak-form-element-horizontal>
                 </div>
             </ak-form-group>
-            <ak-form-group>
-                <span slot="header"> ${msg("Advanced settings")} </span>
-                <div slot="body" class="pf-c-form">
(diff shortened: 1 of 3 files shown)


web/src/elements/forms/types.d.ts
```diff
@@ -0,0 +1,41 @@
+/**
+ * @file Type definitions for form-associated elements.
+ *
+ * While these types are part of the HTML standard, they're not yet defined
+ * in the TypeScript standard library, so we define them here.
+ *
+ * @expires 2026-01-01
```
kensternberg-authentik: Comment still valid. I went and looked; these are not in `dom.generated.d.ts` for the current TypeScript. WTF, guys?

web/src/elements/table/Table.ts
```diff
@@ -124,8 +124,8 @@ export abstract class Table<T extends object>
     @property({ type: String })
     public order?: string;
 
-    @property({ type: String })
-    public search: string = "";
+    @property({ type: String, attribute: false })
```
kensternberg-authentik: Why is this not an attribute?  If it isn't, should it be a `state()`?

kensternberg-authentik (approved):
Looks the same as before.  Still wonder why the search string isn't a property or state.

PR #16304, opened 2025-08-21
fix(rac): restore proper AuthenticatedSession lookup in ConnectionToken creation

### LINKED ISSUE #16303: [Bug] RAC SSH fails with UUID error - regression in ConnectionToken.objects.create since v2025.4.x
## Description
RAC SSH connections fail with a database error after upgrading to authentik 2025.4.x or later.

## Error
```
django.db.utils.DataError: invalid input syntax for type uuid: "6zpc13ds1dkwzzd5hqnnxtqmw57orkfu"
LINE 1: ...RBTFOMzXDm4zgl1DhS7uGdvA2byIxUjAzI', '{}'::jsonb, '6zpc13ds1...
```

## Root Cause
In `/authentik/providers/rac/views.py`, the code passes a session key string instead of an AuthenticatedSession object:

**Current broken code (v2025.4.1 onwards):**
```python
token = ConnectionToken.objects.create(
    provider=self.provider,
    endpoint=self.endpoint,
    settings=settings or {},
    session=self.request.session["authenticatedsession"],  # This is a STRING!
    expires=now() + timedelta_from_string(self.provider.connection_expiry),
    expiring=True,
)
```

**Previous working code (v2025.2.4):**
```python
from authentik.core.models import AuthenticatedSession

token = ConnectionToken.objects.create(
    provider=self.provider,
    endpoint=self.endpoint,
    settings=settings or {},
    session=AuthenticatedSession.objects.filter(
        session_key=self.request.session.session_key
    ).first(),
    expires=now() + timedelta_from_string(self.provider.connection_expiry),
    expiring=True,
)
```

## Affected Versions
- ❌ 2025.4.1 and later (including 2025.6.0, 2025.8.0, main)
- ✅ 2025.2.4 and earlier work correctly

## Steps to Reproduce
1. Install authentik 2025.2.4 (or earlier)
2. Configure and use RAC SSH endpoints (working at this point)
3. Upgrade to authentik 2025.4.1 or later
4. Migration `0007_migrate_session` runs during upgrade, changing the database schema
5. Try to connect via RAC SSH
6. Connection fails with UUID error because:
   - The database now expects a UUID (AuthenticatedSession reference)
   - But the v2025.4.x+ code passes a string (session key)

## Alternative (for fresh installs):
On fresh installs of 2025.4.x+, RAC SSH might work IF the migration properly handles new entries, but will fail for any sessions created before the migration.

## Workaround
For those affected, you can temporarily fix this by reverting the session_id column type:
```sql
ALTER TABLE authentik_providers_rac_connectiontoken 
ALTER COLUMN session_id TYPE varchar(40) USING session_id::text;
```

## Proposed Fix
Revert the regression by restoring the correct code from v2025.2.4 that properly fetches the AuthenticatedSession object.

## Environment
- Kubernetes deployment
- PostgreSQL database
- Authentik 2025.8.0 (upgraded from earlier version)

terratrax:
## Description
Fixes RAC SSH connection failures caused by passing a session key string instead of an AuthenticatedSession object.

## Related Issue
Fixes #16303

## The Problem
In `/authentik/providers/rac/views.py`, the code was incorrectly passing `self.request.session["authenticatedsession"]` (a session key string) to `ConnectionToken.session` which expects an AuthenticatedSession object after migration 0007_migrate_session.

## The Solution
This PR restores the correct behavior from v2025.2.4 where the AuthenticatedSession object is properly fetched using:
```python
session=AuthenticatedSession.objects.filter(
    session_key=self.request.session.session_key
).first()
```

## Changes
- Added `AuthenticatedSession` to the imports in `views.py`
- Fixed the `ConnectionToken.objects.create()` call to properly fetch the AuthenticatedSession object

## Testing
- [ ] RAC SSH connections work after this fix
- [ ] No UUID type errors in logs
- [ ] Tested upgrade path from affected versions

## Notes
This regression was introduced in v2025.4.x and affects all versions from 2025.4.1 onwards including the current main branch.

### DIFF
--- a/authentik/providers/rac/views.py
+++ b/authentik/providers/rac/views.py
@@ -8,7 +8,7 @@ from django.urls import reverse
 from django.utils.timezone import now
 from django.utils.translation import gettext as _
 
-from authentik.core.models import Application
+from authentik.core.models import Application, AuthenticatedSession
 from authentik.core.views.interface import InterfaceView
 from authentik.events.models import Event, EventAction
 from authentik.flows.challenge import RedirectChallenge
@@ -124,7 +124,9 @@ class RACFinalStage(RedirectStage):
             provider=self.provider,
             endpoint=self.endpoint,
             settings=settings or {},
-            session=self.request.session["authenticatedsession"],
+            session=AuthenticatedSession.objects.filter(
+                session_key=self.request.session.session_key
+            ).first(),
             expires=now() + timedelta_from_string(self.provider.connection_expiry),
             expiring=True,
         )


rissson:
`self.request.session["authenticatedsession"]` should be an AuthenticatedSession object

rissson:
it's set here: https://github.com/goauthentik/authentik/blob/main/authentik/core/sessions.py#L93

terratrax:
I think what's happening in that line, the authenticatedsession property is returning the UUID rather than the object itself since its a foreign key. [ConnectionToken](https://github.com/goauthentik/authentik/pull/16304/files#diff-0982e527c1d02757e8493d7bc7f09ffae07864099d3d2831d8ac3c2ee5c15b92L123) tries to use this as if it were the AuthenticatedSession object, and gets upset because its actually a string.

dewi-tik:
This shouldn't be required, can you please re-test on a more recent release and open an issue if required

PR #15774, opened 2025-07-24
website: Fix version origin detection, build-time URLs 

GirlBossRush:
## Details

This PR fixes several issues surrounding Docusaurus's ability to consistently use version URLs, such as when linking to and from the API reference, as well as linking to prerelease docs.

## What this PR does not include

- A cherry-pick for 2025.8. A PR will be opened after this is merged

## How to test linking to the API reference

1. Open a new tab to this [Netlify preview deployment](https://68a73bf839c2abc62c9911fa--authentik-docs.netlify.app)
2. From the docs sidebar, navigate to Developer Documentation -> API Overview
3. Confirm you're on the API Reference
4. Click "Back to developer docs"
5. Confirm you're on the developer docs

---

## Checklist

-   [ ] Local tests pass (`ak test authentik/`)
-   [ ] The code has been formatted (`make lint-fix`)

If an API change has been made

-   [ ] The API schema has been updated (`make gen-build`)

If changes to the frontend have been made

-   [ ] The code has been formatted (`make web`)

If applicable

-   [ ] The documentation has been updated
-   [x] The documentation has been formatted (`make docs`)

PR #16320, opened 2025-08-21
website: Move docs netlify.toml

dominic-r:
## Details

Follow up to #15774 

> Netlify doesn’t allow config files to override one another. The config closest to the root of the repo will take precedence over any other, even if the package configuration is set to a more specific file. When the Netlify config is at website, configs for integrations and api will not be picked up


GirlBossRush (approved):
Teamwork 👏

dominic-r:
FTW 🚀

PR #16324, opened 2025-08-22
 core: bump django to 5.2.7

melizeche:
## Details

REPLACE ME

---

## Checklist

If applicable


pyproject.toml
```diff
@@ -79,6 +79,8 @@ dev = [
   "aws-cdk-lib==2.188.0",
   "bandit==1.8.3",
   "black==25.1.0",
+  "bpython==0.25",
+  "bump2version==1.0.1",
```
BeryJu: what do we need that for?

rissson: We don't. I removed it

rissson (approved):

authentik/core/management/commands/shell.py
```diff
@@ -79,41 +58,7 @@ def pre_delete_handler(sender, instance: Model, **_):
...
-            import rlcompleter
-        except ModuleNotFoundError:
-            pass
-        else:
-            readline.set_completer(rlcompleter.Completer(namespace).complete)
-            readline.parse_and_bind("tab: complete")
-
-        # Run interactive shell
-        code.interact(banner=get_banner_text(), local=namespace)
+        super().handle(**options)
```
BeryJu, line 66: would be nice if we could keep `get_banner_text()` somehow

rissson: re-added as a simple print

BeryJu (approved):
do we want to add the override for createsuperuser/etc in this PR too? @rissson

rissson:
Let's do it separately

PR #16347, opened 2025-08-23
web/admin: allow blank value for User path template in User Write Stage

heavygale:
## Details

The helper text for the User path template input states:

> Path new users will be created under. If left blank, the default path will be used.

But the field itself is marked as _required_ in html.

As there's no red asterix and it work's perfecly without a value this PR removes the _required_ attribute, so that the form can be submitted while leaving this field empty.

---

## Checklist

If applicable


BeryJu (approved):

Nexus2k:
FYI cherry-pick failed: https://github.com/goauthentik/authentik/pull/17677

PR #16369, opened 2025-08-26
website: Fix stale links that depend on initial route not triggering redirects.

GirlBossRush:
## Details

This PR fixes a current and future issue that arises when...

- The user is already viewing the docs
- They search for a page which depends on a redirect created during a Netlify build
- And the to-be-redirected page is picked up by Docusaurus's client-side router first.

## See also

- #16372


tanberry (approved):
Thanks so much @GirlBossRush  fo rthis PR. What you explained earlier makes sense, so though I can't fully fully parse this code, I trust you and thank you!

Issue #10865, opened 2024-08-12
Python client slow when importing


javikalsan:
**Describe your question**
Is there a way to make imports using the Python client faster?

**Relevant info**
Here is an example I did using **Python 3.12.4** and Authentik client version **2024.6.3-1723032688**
authentik_client installations did using a virtual environment and pip.

```
time python -c "from authentik_client.api_client import ApiClient"
real    0m11,639s
user    0m11,525s
sys     0m0,115s
```

and other downgrading to version **2024.2.2.post1710464650**

```
time python -c "from authentik_client.api_client import ApiClient"

real    0m3,400s
user    0m3,344s
sys     0m0,057s
```

I read this related issue on openapi generator: https://github.com/OpenAPITools/openapi-generator/issues/18144

Thanks!

rlincke:
This issue affects me as well :(. I use the OpenAPI generator tool version 7.8.0 (latest). The generated Python API takes over 10 minutes (!!!) to load the imports, i.e., the application does not start due to this (using Python 3.11). Generated API client size is ca 379 kloc.

I am following the alternative described by fingon as a work around.

A Python client generated with an older OpenAPI generator tool version, e.g., 3.3.4 did not cause this delay, despite having the imports in the __init__.py files. (generated API is also just 159 kloc)

katrinm86:
has there been an update on this? i'm currently trying to use it in a cli tool and always waiting for the import to finish is a bit cumbersome >.>

BeryJu:
Should be able to improve this with https://github.com/OpenAPITools/openapi-generator/pull/21486 since thats released in 7.15 which makes the import much faster

PR #16440, opened 2025-08-28
root: bump openapi-generator-cli to v7.15.0

katrinm86:
## Details

This is just a small starting point to update the generator-cli version to use lazy imports in the python client (see #10865 ). Since I don't know that much about the build process, I just wanted to get the change started ...

Also see https://github.com/OpenAPITools/openapi-generator/pull/21486 for why version 7.15

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -177,7 +177,7 @@ gen-client-ts: gen-clean-ts  ## Build and install the authentik API for Typescri
 	docker run \
 		--rm -v ${PWD}:/local \
 		--user ${UID}:${GID} \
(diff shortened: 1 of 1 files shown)


rissson:
I'm probably not merging this since this whole thing will get reworked and moved to its own repo

rissson:
But thank you for your contribution, I'll make sure to do the update when moving this stuff around

rissson:
Actually, I'm gonna go ahead and approve this. FYI, the Python client generation now happens at https://github.com/goauthentik/client-python. We're currently releasing versions for several past releases.

rissson (approved):

PR #16465, opened 2025-08-29
providers/scim: fix string formatting for SCIM user filter

### LINKED ISSUE #16455: SCIM users sync failing - filter query is not valid
**Describe the bug**
Authentik is sending a request with method and path `GET /Users?filter=userName eq akadmin`, which is resulting in a 400 response. According to the SCIM protocol spec [RFC7644](https://www.rfc-editor.org/rfc/rfc7644#section-3.4.2.2), a filter comparison value should be valid JSON, so a string value should be wrapped in quotation marks.

> compValue = false / null / true / number / string
    ; rules from JSON ([RFC 7159](https://www.rfc-editor.org/rfc/rfc7159))

And from [section 7](https://www.rfc-editor.org/rfc/rfc7159#section-7) of the referenced spec describing a JSON string

> string = quotation-mark *char quotation-mark

As such, a valid request would have method and path of `GET /Users?filter=userName eq "akadmin"`

**To Reproduce**

1. Go to Applications -> Providers
2. Set up SCIM provider
3. Sync Users

**Expected behavior**
String values inside SCIM filters should be enclosed in double quotes.

**Logs**
```json
{"domain_url": null, "event": "Failed to send SCIM request", "level": "warning", "logger": "authentik.lib.sync.outgoing.base", "method": "GET", "path": "/Users?filter=userName+eq+akadmin", "pid": 204, "provider": "SCIM", "response": "{\"schemas\":[\"urn:ietf:params:scim:api:messages:2.0:Error\"],\"status\":\"400\",\"scimType\":\"invalidFilter\",\"detail\":\"The specified filter cannot be parsed.\"}", "schema_name": "public", "timestamp": "2025-08-28T20:17:42.458932"}
```

-   authentik version: 2025.8.1
-   Deployment: docker

drmathias:
## Details

Fixes formatting of SCIM filter to adhere to SCIM protocol spec

closes #16455

---

## Checklist

If applicable

### DIFF
--- a/authentik/providers/scim/clients/users.py
+++ b/authentik/providers/scim/clients/users.py
@@ -72,7 +72,7 @@ class SCIMUserClient(SCIMClient[User, SCIMProviderUser, SCIMUserSchema]):
                 if not self._config.filter.supported:
                     raise exc
                 users = self._request(
-                    "GET", f"/Users?{urlencode({'filter': f'userName eq {scim_user.userName}'})}"
+                    "GET", f"/Users?{urlencode({'filter': f'userName eq \"{scim_user.userName}\"'})}"
                 )
                 users_res = users.get("Resources", [])
                 if len(users_res) < 1:
--- a/authentik/providers/scim/clients/users.py
+++ b/authentik/providers/scim/clients/users.py
@@ -72,7 +72,8 @@ class SCIMUserClient(SCIMClient[User, SCIMProviderUser, SCIMUserSchema]):
                 if not self._config.filter.supported:
                     raise exc
                 users = self._request(
-                    "GET", f"/Users?{urlencode({'filter': f'userName eq \"{scim_user.userName}\"'})}"
+                    "GET",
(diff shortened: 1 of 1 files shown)


authentik/providers/scim/clients/users.py
```diff
@@ -72,7 +72,7 @@ def create(self, user: User):
                 if not self._config.filter.supported:
                     raise exc
                 users = self._request(
-                    "GET", f"/Users?{urlencode({'filter': f'userName eq {scim_user.userName}'})}"
+                    "GET", f"/Users?{urlencode({'filter': f'userName eq \"{scim_user.userName}\"'})}"
```
josegomezr: This is the same fix we applied to our instance. However the linter doesn't like that line because it's too long.

This seemed to keep everybody happy:
```py
filter_expr = 'userName eq "{}"'.format(scim_user.userName)
users = self._request("GET", f"/Users?{urlencode({"filter": filter_expr})}")
```

josegomezr (commented):
This fixed the error on our instance! :rocket:

BeryJu (approved):

PR #16476, opened 2025-08-29
rbac: fix typo

gergosimonyi:


rissson:
Can we edit the original migration instead of creating a new one for virtually no changes?

gergosimonyi:
@rissson that would mean anyone who installed authentik before `2025.10` would have the wrong name. Which is, obviously, a _huge_ deal.

BeryJu (approved):

rissson:
Ah yeah the name is stored in the database. Well, that is just stupid.

rissson (approved):

PR #16481, opened 2025-08-29
web: Use embedded layout.

GirlBossRush:
### Before

Screenshot 2025-08-29 at 21 24 30

### After

Screenshot 2025-08-29 at 21 22 33


kensternberg-authentik (approved):
Could be better.  Should just be `embedded`.

web/src/admin/rbac/ObjectPermissionModal.ts
```diff
@@ -35,6 +35,7 @@ export class ObjectPermissionsPageForm extends ModelForm<unknown, string> {
 
     renderForm(): TemplateResult {
         return html`<ak-rbac-object-permission-page
+            .embedded=${true}
```
kensternberg-authentik: Aroo?  Not just `embedded`?

PR #16617, opened 2025-09-04
website/docs: update create oauth provider page

dewi-tik:
## Details

Updates the page to be more consistent with upcoming changes to the saml page

---

## Checklist

If applicable

### DIFF
--- a/website/docs/add-secure-apps/providers/oauth2/create-oauth2-provider.md
+++ b/website/docs/add-secure-apps/providers/oauth2/create-oauth2-provider.md
@@ -9,3 +9,7 @@ To create a provider along with the corresponding application that uses it for a
 3. On the **New application** page, define the application settings, and then click **Next**.
 4. Select **OAuth2/OIDC** as the **Provider Type**, and then click **Next**.
 5. On the **Configure OAuth2/OpenId Provider** page, provide the configuration settings and then click **Submit** to create both the application and the provider.
+
+:::info
+Optionally, configure the provider with the `offline_access` scope mapping. By default, applications only receive an access token. To receive a refresh token, applications and authentik must be configured to request the `offline_access` scope. Do this in the Scope mapping area on the **Configure OAuth2/OpenId Provider** page.
+:::
--- a/website/docs/add-secure-apps/providers/oauth2/create-oauth2-provider.md
+++ b/website/docs/add-secure-apps/providers/oauth2/create-oauth2-provider.md
@@ -2,18 +2,10 @@
 title: Create an OAuth2 provider
 ---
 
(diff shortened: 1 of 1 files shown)


website/docs/add-secure-apps/providers/oauth2/create-oauth2-provider.md
```diff
@@ -2,18 +2,10 @@
...
-4. Select the **Provider Type** of **OAuth2/OIDC**, and then click **Next**.
-
-5. On the **Configure OAuth2/OpenId Provider** page, provide the configuration settings and then click **Submit** to create and save both the application and the provider.
-
-:::info
-Optionally, configure the provider to have the `offline_access` scope mapping. Starting with authentik 2024.2, by default applications only receive an access token. To receive a refresh token, both applications and authentik must be configured to request the `offline_access` scope. Do this in the Scope mapping area on the **Configure OAuth2/OpenId Provider** page.
-:::
+1. Log in to authentik as an administrator, and open the authentik Admin interface.
+2. Navigate to **Applications > Applications** and click **Create with provider** to create an application and provider pair.
+3. On the **New application** page, define the application settings, and then click **Next**.
```
tanberry, line 9: The use of the word "box" was to indicate that it is a modal... I am fine with calling it a page, just like all the other pages, but that is a conscious decision to be less specific, imo.

tanberry: Also, why do we want to remove line 17 with the info about `offline_access`?

dewi-tik: "Box" seems a bit weird to me in this case. I know that box is used as a CSS term but to a non-frontend type person, box isn't as clear as page/window IMO.

I removed the offline access information because it's already mentioned on the OAuth provider page and seems to fit better there: https://version-2024-8.goauthentik.io/docs/providers/oauth2/#authorization_code

tanberry: I agree about not using the term "box". I wish I had won the debate to use "modal". But sure, "page" is fine.

On the Note about offline access though I think we need to leave it here. This is the page about creating/configuring an Oauth provider, and this Note is about configuring the provider to use offline access scope, so this is where it most needs to be, so they see it as they are creating it. Thoughts?

dewi-tik: Re-added the note with slight wording changes

tanberry (approved):
Thanks @dewi-tik !!

PR #16624, opened 2025-09-04
root: move docker files to lifecycle/containers and change docker-compose to compose

PeshekDotDev:
## Details

Move docker files to `lifecycle/container` and rename all `docker-compose.yml` files to `compose.yml` for podman support

---

## Checklist

If applicable


rissson:
Can we also rename `docker-compose.yml` to just `compose.yml`? It is supported by docker and podman IIRC

PeshekDotDev:
@rissson just want to double check - Do you want me to rename the docker-compose moved in this pr, or all across the repo?

@BeryJu Have we thought about having a standard compose.yml in the root for when people work on the front-end? or should they still use the one in the lifecycle folder? my thinking was that if we had a root compose.yml like we asked people to launch an instance of authentik with, it would make front-end development easier and let people try out the project quicker if they clone us down

rissson:
> @rissson just want to double check - Do you want me to rename the docker-compose moved in this pr, or all across the repo?

probably all of them while we're at it

PeshekDotDev:
@rissson I updated all the `compose.yml` files

scripts/test_docker.sh
```diff
@@ -5,29 +5,29 @@ hash="$(git rev-parse HEAD || openssl rand -base64 36 | sha256sum)"
 AUTHENTIK_IMAGE="xghcr.io/goauthentik/server"
 AUTHENTIK_TAG="$(echo "$hash" | cut -c1-15)"
 
-if [ -f .env ]; then
```
rissson, line 8: is docker-compose smart enough to pick this one up or does it only look in the CWD?

website/docs/install-config/upgrade.mdx
```diff
@@ -33,9 +33,9 @@ import Tabs from "@theme/Tabs";
   <TabItem value="docker-compose" label="Docker Compose" default>
     In your terminal, navigate to your installation directory and follow these steps:
 
-    #### 1. Retrieve latest `docker-compose.yml` file
+    #### 1. Retrieve latest `compose.yml` file
 
-    Download the `docker-compose.yml` file using either `wget -O docker-compose.yml https://docs.goauthentik.io/docker-compose.yml` or `curl -O https://docs.goauthentik.io/docker-compose.yml` or a similar process.
```
rissson, line 38: We probably need a PR to github.com/goauthentik/website to serve that file, as well as the old one.

PeshekDotDev: I think that the file I'm linking to handles this but I don't fully understand how it works

https://github.com/goauthentik/authentik/pull/16624/files#diff-af888ed9617d810a8c361e2b121dc620fb55f0700f84bb9bfb2c00b406aab68c

scripts/test_docker.sh
PeshekDotDev: It wouldn't pick up the `.env` file if i made it in the root and called `compose -f lifecycle/container/compose.yml`. We might still be able to make the `.env` in root if i change the configuration inside the `compose.yml` if you want me to try it out

rissson: Probably best to ask @beryju about this

BeryJu: https://docs.docker.com/compose/how-tos/environment-variables/set-environment-variables/#:~:text=The%20paths%20to%20your%20.env%20file%2C%20specified%20in%20the%20env_file%20attribute%2C%20are%20relative%20to%20the%20location%20of%20your%20compose.yaml%20file.

BeryJu (commented):
LGTM, @rissson i dont remember if you had any other comments/feel free to approve

rissson (approved):

website/docs/install-config/upgrade.mdx
dominic-r: Needs update here: https://github.com/goauthentik/website/blob/HEAD/static/_redirects#L16

PeshekDotDev: Ah okay, thanks @dominic-r

dominic-r (approved):

PR #16628, opened 2025-09-05
outpost: proxyv2: Use Postgres for the Embedded Outpost

dominic-r:
Updates the proxyv2 embedded outpost to store it's sessions in Postgres instead of Redis due to the deprecation in authentik 2025.10

todo:
* [x] code review myself it
* [x] potential escaping of postgres dsn
* [x] fix the postgres tests that I need to find a way to run on mac
* [x] that todo in postgresstore


authentik/outposts/models.py
```diff
@@ -47,6 +48,52 @@
 USER_PATH_OUTPOSTS = USER_PATH_SYSTEM_PREFIX + "/outposts"
 
 
+class OutpostProxySession(models.Model):
```
rissson: This should probably go into the proxy provider app.

internal/outpost/proxyv2/postgresstore/postgresstore.go
```diff
@@ -0,0 +1,458 @@
...
+func (ProxySession) TableName() string {
+	return "authentik_outposts_proxy_session"
+}
+
+// NewPostgresStore returns a new PostgresStore with default configuration
+func NewPostgresStore() (*PostgresStore, error) {
+	cfg := config.Get().PostgreSQL
+
+	// Build connection string
+	dsn := fmt.Sprintf("host=%s port=%d user=%s password=%s dbname=%s",
```
rissson: how likely is it that we're going to get issues with quoting?

internal/outpost/proxyv2/postgresstore/postgresstore.go
```diff
@@ -0,0 +1,458 @@
...
+	if cfg.SSLCert != "" {
+		dsn += fmt.Sprintf(" sslcert=%s", cfg.SSLCert)
+	}
+	if cfg.SSLKey != "" {
+		dsn += fmt.Sprintf(" sslkey=%s", cfg.SSLKey)
+	}
+
+	// Add schema if specified
+	if cfg.DefaultSchema != "" {
+		dsn += fmt.Sprintf(" search_path=%s", cfg.DefaultSchema)
```
rissson: How will this play out with tenants?

authentik/outposts/models.py
dominic-r: Moved.

internal/outpost/proxyv2/postgresstore/postgresstore.go
dominic-r: Oh my, didn't think of that at all. I'd have to look more into how tenants work to anise this question

internal/outpost/proxyv2/postgresstore/postgresstore.go
dominic-r: Ok, went to eat dinner, and... Is this acceptable? Should be better https://github.com/goauthentik/authentik/pull/16628/commits/2b3cfb5ae4c7abbfba1de415d063c8b6fad5e435

internal/outpost/proxyv2/postgresstore/postgresstore.go
dominic-r: todo: I don't even think so, i'll add more tests for cases with quotes and other chars and modify as needed

authentik/providers/proxy/models.py
```diff
@@ -25,6 +26,52 @@
 OUTPOST_CALLBACK_SIGNATURE = "X-authentik-auth-callback"
 
 
+class OutpostProxySession(models.Model):
+    """Session storage for proxyv2 outposts using PostgreSQL"""
+
+    session_key = models.CharField(max_length=255, primary_key=True)
```
BeryJu: should probably use a UUID field for primary key and then have the session key as unique and indexed so we can show these sessions in the UI without showing session keys

internal/outpost/proxyv2/application/claims.go
```diff
@@ -1,23 +1,6 @@
...
-	Name              string       `json:"name"`
-	PreferredUsername string       `json:"preferred_username"`
-	Groups            []string     `json:"groups"`
-	Entitlements      []string     `json:"entitlements"`
-	Sid               string       `json:"sid"`
-	Proxy             *ProxyClaims `json:"ak_proxy"`
-
-	RawToken string
-}
+type ProxyClaims = types.ProxyClaims
```
BeryJu: Shouldn't need to re-export the types, can update the direct references to them

docker-compose.outposttest.yml
```diff
@@ -0,0 +1,24 @@
+services:
```
BeryJu: We have these https://github.com/goauthentik/testing-setups?tab=readme-ov-file

but also https://github.com/goauthentik/authentik/pull/11374

if we keep this, it should be in `scripts/`

dominic-r: I'll get rid of it, just wanted a easy and local way to get it working without depending on automated tests

internal/outpost/proxyv2/postgresstore/postgresstore.go
BeryJu: fwiw embedded outposts are not supported with tenants anyways so this is fine

authentik/providers/proxy/models.py
```diff
@@ -25,6 +25,27 @@
...
```
BeryJu:
```suggestion
    session_key = models.TextField(primary_key=True)
```

BeryJu: text and char result in the same thing in postgres, so we use textfield for clarity

internal/outpost/proxyv2/application/auth.go
```diff
@@ -59,14 +61,38 @@ func (a *Application) getClaimsFromSession(r *http.Request) *Claims {
...
+
+	// Try direct cast first (when using filesystem store which stores Claims struct directly)
+	if c, ok := claims.(types.Claims); ok {
+		return &c
+	}
+
+	// Try pointer cast (when using filesystem store with pointer)
+	if c, ok := claims.(*types.Claims); ok {
+		return c
+	}
```
BeryJu: this shouldn't be needed, only one of those paths should be needed

internal/outpost/proxyv2/application/auth.go
```diff
@@ -59,14 +61,38 @@ func (a *Application) getClaimsFromSession(r *http.Request) *Claims {
...
+	// Convert map back to Claims struct using JSON marshaling
+	jsonData, err := json.Marshal(claimsMap)
+	if err != nil {
+		return nil
+	}
+
+	var c types.Claims
+	if err := json.Unmarshal(jsonData, &c); err != nil {
+		return nil
+	}
```
BeryJu: Should use https://github.com/mitchellh/mapstructure instead of re-json encoding it

internal/outpost/proxyv2/application/oauth.go
```diff
@@ -16,16 +16,11 @@ const (
...
+	state, err := a.createState(r, rw, fwd)
 	if err != nil {
 		a.log.WithError(err).Warning("failed to create state")
 		return
 	}
-	s, _ := a.sessions.Get(r, a.SessionName())
-	err = s.Save(r, rw)
-	if err != nil {
-		a.log.WithError(err).Warning("failed to save session")
-	}
```
BeryJu, line 28: Why is this removed?

internal/outpost/proxyv2/postgresstore/postgresstore.go
```diff
@@ -0,0 +1,480 @@
...
+						if expTime, err := time.Parse(time.RFC3339, expiresAtStr); err == nil {
+							if now.After(expTime) {
+								expiredKeys = append(expiredKeys, session.SessionKey)
+							}
+						}
+					}
+				}
+			}
+		}
+	}
```
BeryJu: Since this is in a loop, instead of nesting ifs that deep, invert the diff and use `continue` to skip the element if its not applicable

internal/outpost/proxyv2/postgresstore/postgresstore.go
```diff
@@ -0,0 +1,480 @@
...
+		if exp, ok := claimsMap["exp"].(float64); ok {
+			claims.Exp = int(exp)
+		}
+		if verified, ok := claimsMap["email_verified"].(bool); ok {
+			claims.Verified = verified
+		}
+		if name, ok := claimsMap["name"].(string); ok {
+			claims.Name = name
+		}
+		if groups, ok := claimsMap["groups"].([]interface{}); ok {
```
BeryJu: also https://github.com/mitchellh/mapstructure

authentik/providers/proxy/models.py
BeryJu: Also this still needs a separate primary key so we can show all sessions without leaving the actually session key

BeryJu (changes requested):

rissson (approved):

internal/outpost/proxyv2/application/auth.go
dominic-r: good catch. forgot to get rid of it from when I was trying out different things

internal/outpost/proxyv2/postgresstore/postgresstore.go
```diff
@@ -0,0 +1,480 @@
...
+		sqlDB.SetConnMaxLifetime(time.Duration(cfg.ConnMaxAge) * time.Second)
+	} else {
+		sqlDB.SetConnMaxLifetime(time.Hour) // Default 1 hour
+	}
+
+	ps := &PostgresStore{
+		db: db,
+		options: sessions.Options{
+			Path:   "/",
+			MaxAge: 86400 * 30, // 30 days default todo(dominic): validate if it was that before
```
dominic-r: .

internal/outpost/proxyv2/postgresstore/postgresstore.go
dominic-r: do we still want to use it even if the repo was archived in 2024? I assume it's fine?

internal/outpost/proxyv2/application/session_postgres_test.go
```diff
@@ -0,0 +1,273 @@
...
+		require.NoError(t, err)
+	}
+
+	// Verify all sessions were created
+	var totalCount int64
+	db.Model(&postgresstore.ProxySession{}).Count(&totalCount)
+	assert.Equal(t, int64(3), totalCount)
+
+	// Simulate logout of user1 sessions by deleting them directly
+	// todo(dominic): should probably use LogoutSessions method instead to really test this
```
dominic-r: .

internal/outpost/proxyv2/postgresstore/postgresstore.go
BeryJu: https://github.com/goauthentik/authentik/blob/main/internal/outpost/proxyv2/redisstore/redisstore.go#L41 yep, doesn't matter though since we override it in the postgresstore creation based on token validity

dominic-r: yea that's what i figured as well, Added the todo since autocomplete completed that duration

dominic-r:
```suggestion
			MaxAge: 86400 * 30, // 30 days default (but overwritten in postgresstore creation based on token validation)
```

internal/outpost/proxyv2/postgresstore/postgresstore.go
BeryJu: yeah its stable and we already use it

BeryJu: also the only thing we compare on is `sid` and `sub` so we don't really need the other fields

internal/outpost/proxyv2/application/oauth.go
dominic-r: so, in the old code, `createState(r, fwd)` only took the request and forward URL, so we had to explicitly save the session afterward w `s.Save(r, rw)`. this meant that handleAuthStart was responsible for manually saving the session since createState didn't have access to the ResponseWriter. In the new version, `createState(r, rw, fwd)` now takes the ResponseWriter as a parameter, so that allows it to save the param internally.

authentik/providers/proxy/models.py
```diff
@@ -25,6 +25,28 @@
 OUTPOST_CALLBACK_SIGNATURE = "X-authentik-auth-callback"
 
 
+class OutpostProxySession(models.Model):
+    """Session storage for proxyv2 outposts using PostgreSQL"""
+
+    id = models.AutoField(primary_key=True)
```
BeryJu:
```suggestion
    uuid = models.UUIDField(default=uuid4, primary_key=True)
```

BeryJu (approved):

PR #16642, opened 2025-09-05
core: remove superuser condition on token list

gergosimonyi:
This is a remnant from before we had object-level permissions.


gergosimonyi:
I accidentally created a duplicate (and better version) over at #18684

PR #16653, opened 2025-09-06
packages/django-postgres-cache: Initial implementation of postgres cache

BeryJu:
## Details

Initial sub-project for postgres cache

Some differences to the django builtin postgres backend:
- Uses `text` field to key instead of `varchar(255)` => In postgres, they have basically the same speed and we might have longer keys
- Uses a django model for migrations compared to just creating the table in a management command (idk why django does it that way)
- Adds support for `.keys()` which django's backend doesn't have

~~Now currently this doesn't quite work as for healthchecks we rely on django-redis to be the cache and to establish a redis connection for us, and `channels-redis` doesn't expose this API easily accessible.~~

Other notes:
- We might want to add a flag to make the table created `UNLOGGED` which could have some performance improvements (although not many as far as I've seen) but isn't WAL replicated to Postgres read replicas
- No migration steps are required, like 0, actually none
- Methods that still need to be implemented:
 (`grep -ER 'cache\.\w+\(' authentik/ packages/ | grep -Ev '.(get|set|delete|keys)\('`
  - [x] `cache.delete_many`
  - [x] `cache.has_key`
  - [x] `cache.touch`
  - [x] `cache.ttl`

---

## Checklist


packages/django-postgres-cache/django_postgres_cache/tasks.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+def clear_expired_cache():
+    CacheEntry.objects.filter(expires__lt=now()).delete()
+
+
+try:
+    from dramatiq import actor
+
+    @actor
```
rissson: This will fail our tests since we require a `description` to be provided.

packages/django-postgres-cache/django_postgres_cache/models.py
```diff
@@ -0,0 +1,15 @@
+from django.db import models
+
+
+class CacheEntry(models.Model):
+
+    cache_key = models.TextField(unique=True, primary_key=True)
+    value = models.TextField()
```
rissson, line 7: IIRC this is pickled when stored, so we could use a BinaryField. But I'm quite confused about this no migration thing. How does it work?

BeryJu: Yeah probably, but the b64 encoding is quite deep in the backend

The default backend doesn't use the ORM at all really, it issues raw SQL commands to create table and CRUD entries. Don't know why, might be because its older than the ORM or because performance

authentik/core/migrations/0018_auto_20210330_1345_squashed_0028_alter_token_intent.py
```diff
@@ -53,6 +53,20 @@ class Migration(migrations.Migration):
         ("authentik_core", "0017_managed"),
     ]
 
+    def __init__(self, name, app_label):
```
BeryJu: This is questionable and AI generated, however we can't just add this as a hard dependency to this migration. Adding it as a hard dependency will fail the upgrade as we're applying the cache migration afterwards.

We only need it as a dependency here for fresh installs so that the cache is available during migrations

packages/django-postgres-cache/django_postgres_cache/models.py
BeryJu: Also after more testing, I'm pretty sure the reason the original backend doesn't use the ORM is because it makes migrations easier for this, especially with dependency management and migrating backends

packages/django-postgres-cache/django_postgres_cache/backend.py
```diff
@@ -0,0 +1,79 @@
...
+            return []
+
+    def _keys(self, keys_pattern: str, version: int | None = None) -> list[str]:
+        keys_pattern = self.make_key(keys_pattern.replace("*", ".*"), version=version)
+        db = router.db_for_read(self.cache_model_class)
+        connection = connections[db]
+        quote_name = connection.ops.quote_name
+        table = quote_name(self._table)
+
+        with connection.cursor() as cursor:
```
BeryJu: use ORM

packages/django-postgres-cache/django_postgres_cache/models.py
```diff
@@ -0,0 +1,15 @@
...
+    expires = models.DateTimeField(db_index=True)
+
+    class Meta:
+        default_permissions = []
+        db_table = ""
```
BeryJu: this probably shouldn't be here

packages/django-postgres-cache/django_postgres_cache/tasks.py
```diff
@@ -0,0 +1,19 @@
...
+from django_postgres_cache.models import CacheEntry
+
+
+def clear_expired_cache() -> None:
+    # TODO: optimize deletion, possibly chunking it like we do elsewhere
+    CacheEntry.objects.filter(expires__lt=now()).delete()
+
+
+try:
+    from dramatiq import actor
```
BeryJu: remove task here, move it into authentik/ due to `description=` dependencies

BeryJu:
@rissson addressed the points, should be all good now

authentik/root/settings.py
```diff
@@ -226,20 +227,11 @@
...
-        "BACKEND": "django_redis.cache.RedisCache",
-        "LOCATION": CONFIG.get("cache.url") or redis_url(CONFIG.get("redis.db")),
-        "TIMEOUT": CONFIG.get_int("cache.timeout", 300),
-        "OPTIONS": {
-            "CLIENT_CLASS": "django_redis.client.DefaultClient",
-        },
-        "KEY_PREFIX": "authentik_cache",
+        "BACKEND": "django_postgres_cache.backend.DatabaseCache",
         "KEY_FUNCTION": "django_tenants.cache.make_key",
         "REVERSE_KEY_FUNCTION": "django_tenants.cache.reverse_key",
```
rissson, line 233: Is this still needed since django-postgres-cache is set as a shared app?

packages/django-postgres-cache/django_postgres_cache/models.py
```diff
@@ -0,0 +1,14 @@
+from django.db import models
+
+
+class CacheEntry(models.Model):
+
+    cache_key = models.TextField(unique=True, primary_key=True)
```
rissson:
```suggestion
    cache_key = models.TextField(primary_key=True)
```
this is implicit

rissson (changes requested):

packages/django-postgres-cache/django_postgres_cache/migrations/0001_initial.py
```diff
@@ -0,0 +1,24 @@
...
+
+    initial = True
+
+    dependencies = []
+
+    operations = [
+        migrations.CreateModel(
+            name="CacheEntry",
+            fields=[
+                ("cache_key", models.TextField(primary_key=True, serialize=False, unique=True)),
```
rissson:
```suggestion
                ("cache_key", models.TextField(primary_key=True, serialize=False)),
```

authentik/root/settings.py
BeryJu: uuuh probably for the parts we don't re-implement that don't use the ORM

rissson: yeah you're probably right. I guess I'll git revert then x)

rissson (approved):

GirlBossRush (approved):

PR #16701, opened 2025-09-10
web/esbuild-plugin-live-reload: Logger options.

GirlBossRush:
## Details

This PR adds a `logger` property to the plugin options for environment-specific logging. Providing a logger such as Pino to the plugin reduces the amount of terminal noise while debugging e2e tests.


kensternberg-authentik (approved):
The semantics look good, but the commit message (or PR message) doesn't include any comments on removing the timing.  Is there a a reason for that?

PR #16725, opened 2025-09-11
core: Normalize NPM script arguments.

GirlBossRush:
## Details

This PR fixes NPM's argument parsing in some unique situations such as:

- Passing CLI arguments to a Node.js script when executed via Make
- When a NPM script uses a combination of either the `prefix` and `workspace`, while also defining script arguments in the package.json

I ran into this while attempting to run a specific Playwright e2e test with a "watch" argument. The fix is performed by reordering existing arguments so that NPM can return a subset to a script for more accurate parsing.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -238,34 +238,30 @@ node-install:  ## Install the necessary libraries to build Node.js packages
 #########################
 
 web-build: node-install  ## Build the Authentik UI
-	cd web && npm run build
+	npm run --prefix web build
 
 web: web-lint-fix web-lint web-check-compile  ## Automatically fix formatting issues in the Authentik UI source code, lint the code, and compile it
 
 web-test: ## Run tests for the Authentik UI
-	cd web && npm run test
+	npm run --prefix web test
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -238,34 +238,30 @@ node-install:  ## Install the necessary libraries to build Node.js packages
...
 web: web-lint-fix web-lint web-check-compile  ## Automatically fix formatting issues in the Authentik UI source code, lint the code, and compile it
 
 web-test: ## Run tests for the Authentik UI
-	cd web && npm run test
+	npm run --prefix web test
 
 web-watch:  ## Build and watch the Authentik UI for changes, updating automatically
-	rm -rf web/dist/
-	mkdir web/dist/
-	touch web/dist/.gitkeep
```
GirlBossRush, line 251: Vestigial code. This is handled before ESBuild runs:

https://github.com/goauthentik/authentik/blob/aa56b2fdb138db53fb81b0d6b31ed0f1d9e6c241/web/scripts/build-web.mjs#L108-L121

BeryJu (approved):

PR #16735, opened 2025-09-11
lifecycle: fix permission error when running worker as root

BeryJu:
## Details

related to https://github.com/goauthentik/authentik/pull/16298

cc @rissson 

---

## Checklist

If applicable

### DIFF
--- a/lifecycle/ak
+++ b/lifecycle/ak
@@ -2,6 +2,10 @@
 set -e -o pipefail
 MODE_FILE="${TMPDIR}/authentik-mode"
 
+if [[ -z "${PROMETHEUS_MULTIPROC_DIR}" ]]; then
+    export PROMETHEUS_MULTIPROC_DIR="${TMPDIR:-/tmp}/authentik_prometheus_tmp"
+fi
+
 function log {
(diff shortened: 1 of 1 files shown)


lifecycle/ak
```diff
@@ -68,9 +72,6 @@ function prepare_debug {
     chown authentik:authentik /unittest.xml
 }
 
-if [[ -z "${PROMETHEUS_MULTIPROC_DIR}" ]]; then
-    export PROMETHEUS_MULTIPROC_DIR="${TMPDIR:-/tmp}/authentik_prometheus_tmp"
-fi
 mkdir -p "${PROMETHEUS_MULTIPROC_DIR}"
```
cheggerdev, line 75: mkdir -p must come before chown and not after.

dominic-r: ifif i'm not mistaken, this doesn't matter much. bash executes functions when they are defined, not caller. the chown is in a function check_if_root which is, in the code before the mkdir, but called later on. so, the mkdir, which runs unconditionally, will always run before. cc @BeryJu if im missing something

PR #16771, opened 2025-09-15
website/integrations: update omada controller doc

dewi-tik:
## Details

Adds tabs, updates group name property mapping, fixes some language.

Closes #16745

Only merge after @cheggerdev confirms that it works as expected in #16745.

---

## Checklist

If applicable


tanberry (approved):
Looks great, thanks @dewi-tik! Approving now but yes, as you point out we'll wait to merge after @cheggerdev confirms that it works as expected in https://github.com/goauthentik/authentik/issues/16745.

cheggerdev (approved):
LGTM. Confirm that omada_usergroup_name works when the documented requisites (case sensitivity, user belongs to a single Omada-Group) are respected.
I also confirm that the user may also be a member of multiple non-Omada-Groups with this update.

PR #16799, opened 2025-09-16
website/docs: fix docker tabs not rendering properly

PeshekDotDev:
## Details

There was a bug in this page on the docs that caused switching between docker and kubernetes to not always render properly. This has been fixed

https://docs.goauthentik.io/troubleshooting/logs/

---

## Checklist


dewi-tik (approved):
Good catch. Thanks

PR #16854, opened 2025-09-17
website/docs: add strackgres as recommended postgresql operator

kunthar:
Stackgres added as an option for Postgresql.

## Details

REPLACE ME

---

## Checklist

If applicable


dewi-tik:
Thank you for the suggestion @kunthar.

We're not comfortable merging this at the moment because we don't have experience with Stackgres ourselves. This might change in future, in which case we'll add the URL. Thanks again.

PR #16882, opened 2025-09-18
web/flows: improvements for hCaptcha

Tealk:
## Details

Since hCaptcha recently updated its verification process to require the "host" parameter, requests without it are now rejected with a 403 Forbidden error see #16755. Although PR #16171 introduced this check for reCAPTCHA, hCaptcha also needs the same update. Without this change, hCaptcha blocks captcha verification and prevents user registrations.

This effectively blocks new user registrations through the registration flow that includes hCaptcha.

---

## Checklist

If applicable

### DIFF
--- a/web/src/flow/stages/captcha/CaptchaStage.ts
+++ b/web/src/flow/stages/captcha/CaptchaStage.ts
@@ -574,8 +574,8 @@ export class CaptchaStage extends BaseStage<CaptchaChallenge, CaptchaChallengeRe
                 theme: this.activeTheme,
             });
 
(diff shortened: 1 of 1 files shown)


GirlBossRush (approved):

Tealk:
Too bad it didn't make it into 2025.10.1 :'(

BeryJu (approved):

Tealk:
i think there is still a problem:
image

image

then it stops here and there's no button or anything similar to advance to the next stage
image

PR #16920, opened 2025-09-22
web: Remove CSS constructor polyfill.

GirlBossRush:
## Details

https://caniuse.com/mdn-api_cssstylesheet_cssstylesheet


kensternberg-authentik (approved):
`new CSSStylesheet()` is _Baseline: Newly Available_, but looking at the statistics I can't really see who will be affected by removing the polyfill.  The biggest risk is a handful of 64-bit Safari instances.

PR #16921, opened 2025-09-22
web: Replace Github Slugger package with change-case.

GirlBossRush:
## Details

Swaps out Github Slugger for `change-case` -- Our use case is small enough to not need the ID tracking features of Github Slugger


kensternberg-authentik (approved):
```
x.kebabCase("000-let's head to the  f#$!@#!@$-女子更衣室")
'000-let-s-head-to-the-f-女子更衣室'
```

Looks like kebabCase handles non-alphas correctly, leaves prefixing digits and handles Unicode without surprises.

PR #16926, opened 2025-09-22
website: fix docs links

fty4:
## Details

There are some broken links and also redirects.
Those do not work in curl without following the redirect - this will mean those curl commands currently are broken as flag `-L` would be required.
I've updated the links for the [docker-compose](https://docs.goauthentik.io/docker-compose.yml) and also other broken docs links.

_Also consider if you want to supply a [dev containtainer](https://containers.dev/) for easier contributions.
I was not able to run the checklist tests as you have too many dependencies.
Or maybe provide a image with a pre installed environment to run those.
See https://github.com/goauthentik/authentik/issues/16927_

---

## Checklist

If applicable

### DIFF
--- a/web/src/admin/admin-overview/cards/VersionStatusCard.ts
+++ b/web/src/admin/admin-overview/cards/VersionStatusCard.ts
@@ -56,7 +56,7 @@ export class VersionStatusCard extends AdminStatusCard<Version> {
         let text = this.value?.versionCurrent;
         const versionFamily = this.value?.versionCurrent.split(".");
         versionFamily?.pop();
-        let link = `https://goauthentik.io/docs/releases/${versionFamily?.join(".")}`;
+        let link = `https://docs.goauthentik.io/releases/${versionFamily?.join(".")}`;
         if (this.value?.buildHash) {
             text = this.value.buildHash?.substring(0, 7);
             link = `https://github.com/goauthentik/authentik/commit/${this.value.buildHash}`;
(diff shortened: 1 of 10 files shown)


dominic-r (commented):
I'm not sure the best idea is to update these links as they work. For example, in the security policy: https://goauthentik.io/docs/security/policy is a valid redirect to the docs.

fty4:
> I'm not sure the best idea is to update these links as they work. For example, in the security policy: https://goauthentik.io/docs/security/policy is a valid redirect to the docs.

IMHO I think a redirect is only used to ensure links from external do not break. But the links from the product itself should use the latest and correct links.

dominic-r:
Fair point, I'll ask the others of the doc team for their opinion

dominic-r (approved):
Thanks for your contribution

dewi-tik (approved):

web/src/admin/admin-overview/cards/VersionStatusCard.ts
```diff
@@ -56,7 +56,7 @@ export class VersionStatusCard extends AdminStatusCard<Version> {
         let text = this.value?.versionCurrent;
         const versionFamily = this.value?.versionCurrent.split(".");
         versionFamily?.pop();
-        let link = `https://goauthentik.io/docs/releases/${versionFamily?.join(".")}`;
+        let link = `https://docs.goauthentik.io/releases/${versionFamily?.join(".")}`;
```
GirlBossRush, line 59: This can be swapped with `import.meta.env.AK_DOCS_RELEASE_NOTES_URL`

GirlBossRush (approved):

melizeche (approved):

PR #16942, opened 2025-09-23
blueprints: add `!FindMany` tag

solidDoWant:
## Details

There's a couple of places where I'd like to be able to use the `!Find` tag to get multiple primary keys, rather than a single one. For example, when deploying an outpost via a blueprint, I'd like to be able to specify all providers, or all that match some condition.

This adds a `!FindMany` tag. It's implemented based on the existing `!Find` code, except it returns a list of all matches rather than only the first matched result. The `!Find` tag logic has not really changed, but it now inherits the `FindMany` class which now owns most of this logic.

This probably needs a test or two, but I'm not sure where to add them.

The `make lint-fix` and `make gen-build` commands changed several things that I didn't touch, but the PR body template requested that I run them.

---

## Checklist

If applicable

-   [ ] The documentation has been formatted (`make docs`) - No, `docusaurus build` is failing on master


solidDoWant:
This is now tested about as much as I reasonably can. See the tags.yaml file for some ways that this can be used.

website/docs/customize/blueprints/v1/tags.mdx
```diff
@@ -79,6 +80,24 @@ configure_flow:
...
 
+#### `!FindMany`
+
+Examples:
+
+{/* prettier-ignore-start */}
+
+```yaml
+configure_flow: !Find [authentik_flows.flow, [slug, default-password-change]]
+```
```
rissson, line 87: I'm guessing this is a leftover that was supposed to be removed?

rissson (changes requested):
Thank you for your contribution!

I'm sure the next tag request we'll get is going to be a `!Filter` to use with this :upside_down_face:

website/docs/customize/blueprints/v1/tags.mdx
solidDoWant: Yep, good catch.

solidDoWant:
> I'm sure the next tag request we'll get is going to be a `!Filter`

I would be tentatively interested in adding this if a specific UX is proposed first. Note quite sure how this would differ from the filter support built directly into `!Find` or `!FindMany`.

solidDoWant:
Closing this as there doesn't seem to be any follow-through in reviewing it after attempts to reach maintainers in several places

PR #16952, opened 2025-09-23
providers/plex: Add provider_type to PlexSource

tradiuz:
This will allow the Plex source to work with policies written against OAuth.  Closes #16951

## Details

This is a minor enhancement to add a property to the PlexSource that the OAuth source already has, and is used in example policies.

---

## Checklist

If applicable

### DIFF
--- a/authentik/sources/plex/models.py
+++ b/authentik/sources/plex/models.py
@@ -42,6 +42,7 @@ class PlexAuthenticationChallengeResponse(ChallengeResponse):
 class PlexSource(ScheduledModel, Source):
     """Authenticate against plex.tv"""
 
+    provider_type = "plex"
     client_id = models.TextField(
         default=generate_id,
         help_text=_("Client identifier used to talk to Plex."),


dominic-r:
Can you edit this pr's description to include "Closes <id>" to link the issue and PR so it gets automatically closed on merge? You can find the list of keywords here: https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

rissson:
I'm not inclined to merge this as there is really no necessity to do so. See the example of how to achieve the same thing provided in the associated issue.

tradiuz:
> I'm not inclined to merge this as there is really no necessity to do so. See the example of how to achieve the same thing provided in the associated issue.

I disagree, as this makes the plex source more in line with OAuth sources, which make up the vast majority of providers.  It also gives a very clear and easy attribute to query to see source type in expression flows.  If the alternative is to add the extra logic before doing a lookup on that attribute, then it would mean all the documentation needs to be updated that has sample code just calling `.source_type` raw without it. 

tl;dr - this:

1) Treats plex like most other sources
2) Prevents rewriting documentation

BeryJu:
`.provider_type` is only a thing on OAuth sources because of how they are implemented (one model, different UI and logic on top, decided by that field), other sources don't have this because they just have a 1:1 mapping, and the idea is that eventually we can get to a point where there's only the one OAuth Source and everything can be configured

Alas, we're not adding `provider_type` to other sources, and instead updating the docs: #17008

tradiuz:
That makes sense, it means you'll also need to update the docs for all OAuth sources that have examples (looks like only github/discord at this time).

PR #16964, opened 2025-09-24
core: Add input validation for service account creation

### LINKED ISSUE #15947: SAs can be created without username (mandatory field issue)
when creating a service account, username is marked as a mandatory field:

Image

however, if you put nothing and click submit it actually creates it:

Image

tested on 2025.6.4 EE, have not tested on authentik@next

dominic-r:
Closes: https://github.com/goauthentik/authentik/issues/15947

Not sure this is the best way of doing it, but let's try


authentik/core/tests/test_users_api.py
```diff
@@ -665,9 +676,12 @@ def test_service_account_validation_multiple_errors(self):
         )
         self.assertEqual(response.status_code, 400)
         body = loads(response.content)
-        self.assertIn("non_field_errors", body)
-        # Should have multiple error messages
-        self.assertGreater(len(body["non_field_errors"]), 1)
+        # Should have multiple field errors
+        self.assertIn("name", body)
```
BeryJu: fyi for these I usually recommend self.assertJSONEqual to compare the entire response

dominic-r: TIL (I'll mabye forget, but now I know). Fixed in latest commit.

BeryJu (approved):

PR #16977, opened 2025-09-24
ci: test with postgres read-replica

BeryJu:
## Details

Read-replicas were previously only manually tested

---

## Checklist

If applicable

### DIFF
--- a/.github/workflows/ci-main.yml
+++ b/.github/workflows/ci-main.yml
@@ -124,8 +124,8 @@ jobs:
           - 16-alpine
           - 17-alpine
         profiles:
-          - redis,postgres
-          - redis,postgres,postgres_replica
+          - ""
+          - postgres_replica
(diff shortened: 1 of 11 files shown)


.github/workflows/ci-main.yml
```diff
@@ -67,7 +67,6 @@ jobs:
       fail-fast: false
       matrix:
         psql:
-          - 15-alpine
```
rissson, line 70: We need to define a policy for those tests. I'm thinking one of:
- keep the earliest postgres version support and the latest one, so 14 and 17
- the two latest ones, so 16 and 17

BeryJu:
closing as we never got this to work with how django's test db is created, might be something we can add for the new e2e tests that run against a prod-like authentik cc @rissson, but for the time being we're real-world testing this with our internal authentik instance

PR #17031, opened 2025-09-25
web/a11y: Notifications drawer

GirlBossRush:
## Details

A first pass at the notifications drawer. This is mostly focused on presenting the drawer as its own ARIA region. 

## Sample ARIA Tree

Screenshot 2025-09-25 at 20 39 25


kensternberg-authentik (approved):
I like the changes.  But maybe `ak-notification-item` should be a thing.

PR #17050, opened 2025-09-26
core: add digraph group hierarchy

### LINKED ISSUE #1278: Allow Groups to be members of other Groups
I would like to be able to create a group to control access to each of the applications, then create a group of users, and put that user group as a member of the various application groups.

gergosimonyi:


cheggerdev:
This raises the question: What is an indirect group? So far it includes all groups with an parent.
With this PR will it be still so or do indirect groups include all children ? The answer is important because it decides whether expression policies and property mappings using user.ak_groups() will continue to work or require an adaption.

gergosimonyi:
@cheggerdev

The definition for "indirect groups of a user" doesn't change in this PR. That definition is: "any ancestor of the user's direct groups" (and by "direct groups" here, I mean `user.ak_groups.all()`).

This PR only changes the group hierarchy in one way: groups can now have multiple parents. Notably, your example of `user.ak_groups` isn't affected, it stays a standard django-managed `ManyToManyField`.

authentik/core/api/users.py
```diff
@@ -163,24 +163,26 @@ def create(self, validated_data: dict) -> User:
...
-        permissions = Permission.objects.filter(
+        perms_qs = Permission.objects.filter(
             codename__in=[x.split(".")[1] for x in validated_data.pop("permissions", [])]
-        )
-        validated_data["user_permissions"] = permissions
+        ).values_list("content_type__app_label", "codename")
+        perms_list = [f"{ct}.{name}" for ct, name in list(perms_qs)]
         instance = super().update(instance, validated_data)
         self._set_password(instance, password)
+        instance.assign_perms_to_managed_role(perms_list)
```
BeryJu, line 208: Should only be called if there are permissions to assign

authentik/rbac/models.py
```diff
@@ -45,19 +45,11 @@ class Role(SerializerModel):
     # The main advantage of that is that all the permission checking just works out of the box,
     # as these permissions are checked by default by django and most other libraries that build
     # on top of django
-    group = models.OneToOneField("auth.Group", on_delete=models.CASCADE)
+    group = models.OneToOneField("auth.Group", on_delete=models.CASCADE, null=True)
```
BeryJu, line 49: Purposefully keeping this in with `null=True` and removing it in a future release to make the migration 1% easier

authentik/root/settings.py
```diff
@@ -142,6 +142,7 @@
 PUBLIC_SCHEMA_NAME = CONFIG.get("postgresql.default_schema")
 
 GUARDIAN_MONKEY_PATCH_USER = False
```
BeryJu:
```suggestion
```

authentik/blueprints/v1/importer.py
```diff
@@ -98,6 +97,7 @@ def excluded_models() -> list[type[Model]]:
         DjangoGroup,
         ContentType,
         Permission,
+        RoleObjectPermission,
         UserObjectPermission,
         # Base classes
```
BeryJu, line 114: need to add grouparentagenode

kiaraly:
> This raises the question: What is an indirect group? So far it includes all groups with an parent. With this PR will it be still so or do indirect groups include all children ? The answer is important because it decides whether expression policies and property mappings using user.ak_groups() will continue to work or require an adaption.

I think the way you intuitively think about nested groups is that a parent can do everything the child can do, similar to https://github.com/goauthentik/authentik/discussions/16282.

However, I can also see how you can think of it as inheritance and then inheriting everything the parent can do makes perfect sense. When you can add multiple parents the outcome is the same and in this version the master admin group would simply get every app specific admin group as a parent.

authentik/core/migrations/0052_groupancestor_groupparentagenode_group_parents.py
```diff
@@ -0,0 +1,143 @@
...
+
+    for group in Group.objects.using(db_alias).all():
+        if not group.parent:
+            continue
+        group.parents.add(group.parent)
+        group.save()
+
+
+create_transitive_closure = """
+CREATE MATERIALIZED VIEW authentik_core_groupancestry AS
```
rissson: Might want to check https://django-postgres-extra.readthedocs.io/en/master/api_reference.html?highlight=view#psqlextra.models.PostgresMaterializedViewModel

authentik/core/migrations/0052_groupancestor_groupparentagenode_group_parents.py
```diff
@@ -0,0 +1,143 @@
...
+CREATE OR REPLACE FUNCTION refresh_group_ancestry()
+RETURNS trigger LANGUAGE plpgsql AS $$
+BEGIN
+    REFRESH MATERIALIZED VIEW CONCURRENTLY authentik_core_groupancestry;
+    RETURN NULL;
+END;
+$$;
+
+-- Refresh trigger
+CREATE OR REPLACE TRIGGER trigger_refresh_group_ancestry
```
rissson: we already depend on pgtrigger, might as well use it

packages/ak-guardian/guardian/models/__init__.py
```diff
@@ -0,0 +1,33 @@
+from .models import (
```
BeryJu: Move models/models.py to models.py so we can get rid of this

web/src/admin/groups/GroupForm.ts
```diff
@@ -87,34 +91,30 @@ export class GroupForm extends ModelForm<Group, string> {
...
+                                page: page,
+                                search: search,
+                            })
+                            .then((results) => {
+                                return {
+                                    pagination: results.pagination,
+                                    options: results.results.map(coreGroupPair),
+                                };
+                            });
                     }}
```
GirlBossRush: Move into private method

```ts
#fetchDataProvision = (page: number, search?: string): Promise<DataProvision> => {
                        return new CoreApi(DEFAULT_CONFIG)
                            .coreGroupsList({
                                page: page,
                                search: search,
                            })
                            .then((results) => {
                                return {
                                    pagination: results.pagination,
                                    options: results.results.map(coreGroupPair),
                                };
                            });
                   }
```

```suggestion
                    .provider=${this.#fetchDataProvision}
```

web/src/admin/groups/GroupForm.ts
```diff
@@ -87,34 +91,30 @@ export class GroupForm extends ModelForm<Group, string> {
...
-                    blankable
-                >
-                </ak-search-select>
+                    .selected=${(this.instance?.parentsObj ?? []).map(coreGroupPair)}
+                    available-label="${msg("Available Groups")}"
+                    selected-label="${msg("Selected Groups")}"
```
GirlBossRush: Can omit quotes on interpolated attributes (in most scenarios)

```suggestion
                    available-label=${msg("Available Groups")}
                    selected-label=${msg("Selected Groups")}
```

web/src/admin/groups/GroupForm.ts
```diff
@@ -87,34 +91,30 @@ export class GroupForm extends ModelForm<Group, string> {
...
+                ></ak-dual-select-provider>
+                <p class="pf-c-form__helper-text">
+                    ${msg(
+                        "Select parent groups. A group inherits every role from its ancestors (recursively).",
```
GirlBossRush:
```suggestion
                        "A group recursively inherits every role from its ancestors.",
```

web/src/admin/groups/RelatedUserList.ts
```diff
@@ -55,12 +66,23 @@ export class RelatedUserAdd extends Form<{ users: number[] }> {
...
+                        groupUuid: this.targetGroup.pk,
+                        userAccountRequest: {
+                            pk: user,
+                        },
+                    });
+                }
+                if (this.targetRole) {
+                    return new RbacApi(DEFAULT_CONFIG).rbacRolesAddUserCreate({
+                        uuid: this.targetRole.pk,
+                        // todo: rename this
```
GirlBossRush:
```suggestion
                        // TODO: Rename this.
```

web/src/admin/groups/RelatedUserList.ts
```diff
@@ -55,12 +66,23 @@ export class RelatedUserAdd extends Form<{ users: number[] }> {
     async send(data: { users: number[] }): Promise<{ users: number[] }> {
         await Promise.all(
             data.users.map((user) => {
```
GirlBossRush:
```suggestion
            data.users.map((userPK) => {
```

web/src/admin/groups/RelatedUserList.ts
```diff
@@ -136,6 +158,9 @@ export class RelatedUserList extends WithBrandConfig(WithCapabilitiesConfig(Tabl
     @property({ attribute: false })
     targetGroup?: Group;
 
+    @property({ attribute: false })
+    targetRole?: Role;
```
GirlBossRush, line 153: A bit of a legacy thing, all `@property` decorations are public. Avoid a union with `undefined`, opting for `null` to match the typical type signature of native elements -- `null` when not present, `undefined` when not supported.

```suggestion
    @property({ attribute: false })
    public targetRole: Role | null = null;
```

web/src/admin/groups/RelatedUserList.ts
```diff
@@ -175,10 +202,10 @@ export class RelatedUserList extends WithBrandConfig(WithCapabilitiesConfig(Tabl
         const disabled = this.selectedElements.length < 1;
         return html`<ak-forms-delete-bulk
             objectLabel=${msg("User(s)")}
-            actionLabel=${msg("Remove Users(s)")}
+            actionLabel=${msg("Remove User(s)")}
             action=${msg("removed")}
             actionSubtext=${msg(
-                str`Are you sure you want to remove the selected users from the group ${this.targetGroup?.name}?`,
+                str`Are you sure you want to remove the selected users from ${this.targetGroup?.name || this.targetRole?.name}?`,
             )}
```
GirlBossRush: Add fallback for no target e.g.
```suggestion
  const targetLabel = this.targetGroup?.name || this.targetRole?.name
  ...
           actionSubtext=${targetLabel ? msg(
                str`Are you sure you want to remove the selected users from ${targetLabel}?`) : msg("Are you sure you want to remove the selected users?"),
}
```

web/src/admin/rbac/ObjectPermissionModal.ts
```diff
@@ -21,7 +21,7 @@ import PFBase from "@patternfly/patternfly/patternfly-base.css";
 @customElement("ak-rbac-object-permission-modal-form")
 export class ObjectPermissionsPageForm extends ModelForm<unknown, string> {
     @property()
-    model?: RbacPermissionsAssignedByUsersListModelEnum;
+    model?: RbacPermissionsAssignedByRolesListModelEnum;
```
GirlBossRush:
```suggestion
    public model: RbacPermissionsAssignedByRolesListModelEnum | null = null
```

web/src/admin/rbac/ObjectPermissionModal.ts
```diff
@@ -57,7 +57,7 @@ export class ObjectPermissionModal extends AKElement {
     ];
 
     @property()
-    public model?: RbacPermissionsAssignedByUsersListModelEnum;
+    public model?: RbacPermissionsAssignedByRolesListModelEnum;
```
GirlBossRush:
```suggestion
    public model: RbacPermissionsAssignedByRolesListModelEnum | null = null;
```

web/src/admin/rbac/ObjectPermissionsPage.ts
```diff
@@ -22,7 +19,7 @@ import PFBase from "@patternfly/patternfly/patternfly-base.css";
 @customElement("ak-rbac-object-permission-page")
 export class ObjectPermissionPage extends AKElement {
     @property()
-    model?: RbacPermissionsAssignedByUsersListModelEnum;
+    model?: RbacPermissionsAssignedByRolesListModelEnum;
```
GirlBossRush:
```suggestion
    public model: RbacPermissionsAssignedByRolesListModelEnum | null = null;
```

web/src/admin/rbac/RoleObjectPermissionTable.ts
```diff
@@ -20,6 +20,19 @@ import { html, TemplateResult } from "lit";
...
+        next: 0,
+        previous: 0,
+        count: 0,
+        current: 1,
+        totalPages: 1,
+        startIndex: 0,
+        endIndex: 0,
+    },
+    results: [],
+};
```
GirlBossRush:
```suggestion
import type { Pagination } from "@goauthentik/api";

const FALLBACK_PAGINATED_RESPONSE: Pagination = {
    pagination: {
        next: 0,
        previous: 0,
        count: 0,
        current: 1,
        totalPages: 1,
        startIndex: 0,
        endIndex: 0,
    },
    results: [],
};
```

web/src/admin/rbac/RoleObjectPermissionTable.ts
```diff
@@ -98,27 +113,38 @@ export class RoleAssignedObjectPermissionTable extends Table<RoleAssignedObjectP
...
         const baseRow = [html` <a href="#/identity/roles/${item.rolePk}">${item.name}</a>`];
         this.modelPermissions?.results.forEach((perm) => {
-            const granted =
-                item.permissions.filter((uperm) => uperm.codename === perm.codename).length > 0;
+            const assignedToModel =
+                item.modelPermissions.filter((uperm) => uperm.codename === perm.codename).length >
+                0;
+            const assignedToObject =
+                item.objectPermissions.filter((uperm) => uperm.codename === perm.codename).length >
+                0;
```
GirlBossRush:
```suggestion
            const assignedToModel =
                item.modelPermissions.some((uperm) => uperm.codename === perm.codename);
            const assignedToObject =
                item.objectPermissions.some((uperm) => uperm.codename === perm.codename);
```

web/src/admin/rbac/RoleObjectPermissionTable.ts
```diff
@@ -98,27 +113,38 @@ export class RoleAssignedObjectPermissionTable extends Table<RoleAssignedObjectP
...
+            const message = [];
+            if (assignedToModel) {
+                message.push(msg("Global permission"));
+            }
+            if (assignedToObject) {
+                message.push(msg("Object permission"));
+            }
```
GirlBossRush: The joiner must be localized in place to make translation consistently possible:

```suggestion
            let tooltip: string | null = null
            
            if (assignedToModel && assignedToObject) {
                tooltip = msg("Global and object permission"));
            } else if (assignedToModel) {
               tooltip = msg("Global permission"));
            }
            if (assignedToObject) {
               tooltip = msg("Object permission"));
            }
```

web/src/admin/roles/RelatedRoleList.ts
```diff
@@ -0,0 +1,185 @@
...
+import { customElement, property, state } from "lit/decorators.js";
+import { ifDefined } from "lit/directives/if-defined.js";
+
+@customElement("ak-role-related-add")
+export class RelatedRoleAdd extends Form<{ roles: string[] }> {
+    @property({ attribute: false })
+    user?: User;
+
+    @state()
+    rolesToAdd: Role[] = [];
```
GirlBossRush:
```suggestion
    protected rolesToAdd: Role[] = [];
```

web/src/admin/roles/RelatedRoleList.ts
```diff
@@ -0,0 +1,185 @@
...
+
+import { msg, str } from "@lit/localize";
+import { html, nothing, TemplateResult } from "lit";
+import { customElement, property, state } from "lit/decorators.js";
+import { ifDefined } from "lit/directives/if-defined.js";
+
+@customElement("ak-role-related-add")
+export class RelatedRoleAdd extends Form<{ roles: string[] }> {
+    @property({ attribute: false })
+    user?: User;
```
GirlBossRush:
```suggestion
    public user: User | null = null;
```

web/src/admin/roles/RelatedRoleList.ts
```diff
@@ -0,0 +1,185 @@
...
+    @state()
+    rolesToAdd: Role[] = [];
+
+    getSuccessMessage(): string {
+        return msg("Successfully added user to role(s).");
+    }
+
+    async send(data: { roles: string[] }): Promise<unknown> {
+        await Promise.all(
+            data.roles.map((role) => {
```
GirlBossRush, line 37: Nit: this has potential to hang the tab if the number of requests are numerous. Depending on the possible size, you may want to batch with something like this:

```ts
// Place in "src/common/collections.ts"

/**
 * Given an iterable, returns an array of arrays of the specified size.
 *
 * ```ts
 * for (const batch of take([1, 2, 3, 4, 5], 2)) {
 *   console.log(batch)
 * }
 * // Output:
 * // [1, 2]
 * // [3, 4]
 * // [5]
 * ```
 * @param collection The collection to batch.
 * @param batchSize The size of each batch.
 *
 * @returns An iterable of arrays of the specified size.
 */
export function* take<T>(collection: Iterable<T>, batchSize: number): Iterable<T[]> {
	const batch: T[] = []

	for (const item of collection) {
		batch.push(item)

		if (batch.length === batchSize) {
			yield batch

			batch.length = 0
		}
	}

	if (batch.length > 0) {
		yield batch
	}
}
```

```
for (const batch of take(data.roles, 25)) {
 ...
}

File: web/src/admin/roles/RelatedRoleList.ts [hunk 24]
```diff
@@ -0,0 +1,185 @@
...
+        return data;
+    }
+
+    renderForm(): TemplateResult {
+        return html`<ak-form-element-horizontal label=${msg("Roles to add")} name="roles">
+            <div class="pf-c-input-group">
+                <ak-user-role-select-table
+                    .confirm=${(items: Role[]) => {
+                        this.rolesToAdd = items;
+                        this.requestUpdate();
```
GirlBossRush: `rolesToAdd` is tracked by `@state()`, is this needed?

web/src/admin/roles/RelatedRoleList.ts
```diff
@@ -0,0 +1,185 @@
...
+                            <i class="fas fa-plus" aria-hidden="true"></i>
+                        </pf-tooltip>
+                    </button>
+                </ak-user-role-select-table>
+                <div class="pf-c-form-control">
+                    <ak-chip-group>
+                        ${this.rolesToAdd.map((role) => {
+                            return html`<ak-chip
+                                removable
+                                value=${ifDefined(role.pk)}
```
GirlBossRush:
```suggestion
                                value=${ifPresent(role.pk)}
```

web/src/admin/roles/RelatedRoleList.ts
```diff
@@ -0,0 +1,185 @@
...
+                                @remove=${() => {
+                                    const idx = this.rolesToAdd.indexOf(role);
+                                    this.rolesToAdd.splice(idx, 1);
+                                    this.requestUpdate();
```
GirlBossRush: Immutable approach, something along the lines of...
```suggestion
                                    this.rolesToAdd = this.rolesToAdd.filter(($role) => $role === role);
```

web/src/admin/roles/RelatedRoleList.ts
```diff
@@ -0,0 +1,185 @@
...
+export class RelatedRoleList extends Table<Role> {
+    checkbox = true;
+    clearOnRefresh = true;
+    protected override searchEnabled = true;
+
+    @property()
+    order = "name";
+
+    @property({ attribute: false })
+    targetUser?: User;
```
GirlBossRush:
```suggestion
    @property({ type: String })
    public order = "name";

    @property({ attribute: false })
    public targetUser: User | null = null;
```

web/src/admin/roles/RelatedRoleList.ts
```diff
@@ -0,0 +1,185 @@
...
+        });
+    }
+
+    protected columns: TableColumn[] = [
+        [msg("Name"), "name"],
+        [msg("Actions"), null, msg("Row Actions")],
+    ];
+
+    renderToolbarSelected(): TemplateResult {
+        const disabled = this.selectedElements.length < 1;
```
GirlBossRush:
```suggestion
        const disabled = !this.selectedElements.length;
```

web/src/admin/roles/RelatedRoleList.ts
```diff
@@ -0,0 +1,185 @@
...
+                    </pf-tooltip>
+                </button>
+            </ak-forms-modal>`,
+        ];
+    }
+
+    renderToolbar(): TemplateResult {
+        return html`
+            ${this.targetUser
+                ? html`<ak-forms-modal>
```
GirlBossRush, line 156: Short circuit pattern:
```suggestion
        if (!this.targetUser) {
          return nothing
```

web/src/admin/roles/RolePermissionForm.ts
```diff
@@ -26,7 +26,7 @@ export class RolePermissionForm extends ModelForm<RolePermissionAssign, number>
     permissionsToAdd: Permission[] = [];
 
     @property()
-    roleUuid?: string;
+    roleUuid!: string;
```
GirlBossRush:
```suggestion
    @property({ type: string, attribute: "role-uuid" })
    public roleUuid: string | null = null
```

web/src/admin/roles/RoleViewPage.ts
```diff
@@ -126,13 +127,27 @@ export class RoleViewPage extends AKElement {
...
+                    role="tabpanel"
+                    tabindex="0"
+                    slot="page-users"
+                    id="page-users"
+                    aria-label="${msg("Users")}"
+                    class="pf-c-page__main-section pf-m-no-padding-mobile"
+                >
+                    <div class="pf-c-card">
+                        <div class="pf-c-card__body">
+                            <ak-user-related-list .targetRole=${this._role}> </ak-user-related-list>
```
GirlBossRush: Avoid underscores to indicate access. Consider rename to `protected role`

```suggestion
                            <ak-user-related-list .targetRole=${this.role}> </ak-user-related-list>
```

web/src/admin/users/RoleSelectModal.ts
```diff
@@ -0,0 +1,87 @@
...
+@customElement("ak-user-role-select-table")
+export class RoleSelectModal extends TableModal<Role> {
+    checkbox = true;
+    checkboxChip = true;
+
+    protected override searchEnabled = true;
+    public supportsQL = true;
+
+    @property()
+    confirm!: (selectedItems: Role[]) => Promise<unknown>;
```
GirlBossRush:
```suggestion
    @property({ attribute: false })
    public confirm: ((selectedItems: Role[]) => Promise<unknown>) | null = null;
```

web/src/admin/users/RoleSelectModal.ts
```diff
@@ -0,0 +1,87 @@
...
+
+    order = "name";
+
+    static styles: CSSResult[] = [...super.styles, PFBanner];
```
GirlBossRush: Move to top of class

web/src/admin/users/RoleSelectModal.ts
```diff
@@ -0,0 +1,87 @@
...
+        return new RbacApi(DEFAULT_CONFIG).rbacRolesList({
+            ...(await this.defaultEndpointConfig()),
+        });
+    }
+
+    protected columns: TableColumn[] = [[msg("Name"), "name"]];
+
+    row(item: Role): SlottedTemplateResult[] {
+        return [
+            html`<div>
```
GirlBossRush: Are two div layers necessary here?

web/src/admin/users/RoleSelectModal.ts
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    renderSelectedChip(item: Role): TemplateResult {
+        return html`${item.name}`;
+    }
+
+    renderModalInner(): SlottedTemplateResult {
+        return html`<section class="pf-c-modal-box__header pf-c-page__main-section pf-m-light">
+                <div class="pf-c-content">
+                    <h1 class="pf-c-title pf-m-2xl">${msg("Select groups to add user to")}</h1>
```
GirlBossRush: Avoid ending sentence with preposition. Something along the lines of:

- Select groups for this user
- Assign User to Groups

web/src/admin/users/ServiceAccountForm.ts
```diff
@@ -40,13 +42,16 @@ export class ServiceAccountForm extends Form<UserServiceAccountRequest> {
     result: UserServiceAccountResponse | null = null;
 
     @property({ attribute: false })
-    group?: Group;
+    targetGroup?: Group;
+
+    @property({ attribute: false })
+    targetRole?: Role;
```
GirlBossRush:
```suggestion
    public targetGroup: Group | null = null;

    @property({ attribute: false })
    public targetRole: Role | null = null;
```

web/src/admin/users/UserForm.ts
```diff
@@ -46,7 +46,10 @@ const UserTypeOptions: readonly RadioOption<UserTypeEnum>[] = [
 @customElement("ak-user-form")
 export class UserForm extends ModelForm<User, number> {
     @property({ attribute: false })
-    group?: Group;
+    targetGroup?: Group;
+
+    @property({ attribute: false })
+    targetRole?: Role;
```
GirlBossRush: Usual public/nullable stuff

web/src/admin/users/UserForm.ts
```diff
@@ -77,8 +80,11 @@ export class UserForm extends ModelForm<User, number> {
         if (this.instance) {
             return msg("Successfully updated user.");
         }
-        if (this.group) {
-            return msg(str`Successfully created user and added to group ${this.group.name}`);
+        if (this.targetGroup) {
+            return msg(str`Successfully created user and added to group ${this.targetGroup.name}`);
+        }
+        if (this.targetRole) {
+            return msg(str`Successfully created user and added to role ${this.targetRole.name}`);
```
GirlBossRush:
```suggestion
if (this.group) {
    return msg(str`User created and added to group ${this.group.name}`);
}
if (this.targetGroup) {
    return msg(str`User created and added to group ${this.targetGroup.name}`);
}
if (this.targetRole) {
    return msg(str`User created and added to role ${this.targetRole.name}`);
}
return msg("User created.");
```

web/src/admin/users/UserViewPage.ts
```diff
@@ -462,6 +463,20 @@ export class UserViewPage extends WithCapabilitiesConfig(AKElement) {
                         </div>
                     </div>
                 </div>
+                <div
+                    role="tabpanel"
+                    tabindex="0"
+                    slot="page-roles"
+                    id="page-roles"
+                    aria-label="${msg("Roles")}"
```
GirlBossRush:
```suggestion
                    aria-label=${msg("Roles")}
```

GirlBossRush (approved):

authentik/blueprints/v1/importer.py
gergosimonyi: Instead, I removed serialization from `GroupParentageNode`.

web/src/admin/rbac/RoleObjectPermissionTable.ts
BeryJu:
```suggestion
            [msg("Role"), "role"],
```

web/src/admin/rbac/RoleObjectPermissionTable.ts
```diff
@@ -35,20 +48,22 @@ export class RoleAssignedObjectPermissionTable extends Table<RoleAssignedObjectP
     clearOnRefresh = true;
```
BeryJu, line 50: should probably enable search here too

web/src/admin/roles/RolePermissionForm.ts
gergosimonyi: Initially I added `!` here because `roleUuid` should always be present.

Changed to your style, added `if (!this.roleUuid) return` below where it's used.

web/src/admin/roles/RoleViewPage.ts
gergosimonyi: I don't think this was chosen as an indication of `protected`, it's an alternate name because `role` is [taken](https://developer.mozilla.org/en-US/docs/Web/API/Element/role). I renamed it to `targetRole`, like in other similar locations.

web/src/admin/users/RoleSelectModal.ts
gergosimonyi: Changed to your style, added `if (!this.confirm) return` below where it's used.

web/src/admin/roles/RelatedRoleList.ts
gergosimonyi: I'm not sure I get this. There are other toolbar items below that should be rendered every time. Isn't the

```
html`
  ${condition ? conditionalItem : nothing}
  unconditionalItem1
  unconditionalItem2
`
```

pattern the easiest on the eye for this?

web/src/admin/roles/RelatedRoleList.ts
gergosimonyi: Removed, works!

web/src/admin/roles/RelatedRoleList.ts
gergosimonyi: The largest item size we expect here at the moment is 10 (most permissions for a single model, which is `authentik_core.user` as of now).

web/src/admin/rbac/RoleObjectPermissionTable.ts
gergosimonyi: Just `Role` for now, missed this, fixed! `User` should probably be added to a different page (in a subsequent PR).

web/src/admin/rbac/RoleObjectPermissionTable.ts
gergosimonyi: Done!

authentik/core/models.py
```diff
@@ -242,12 +227,103 @@ def is_member(self, user: "User") -> bool:
...
+
+        db_table = "authentik_core_groupparentage"
+
+        triggers = [
+            pgtrigger.Trigger(
+                name="refresh_groupancestry",
+                operation=pgtrigger.Insert | pgtrigger.Update | pgtrigger.Delete,
+                when=pgtrigger.After,
+                func="""
+                    REFRESH MATERIALIZED VIEW CONCURRENTLY authentik_core_groupancestry;
```
rissson, line 277: Can we retrieve the view name programmatically with `GroupAncestryNode._meta.view_name` or something?

authentik/core/sessions.py
```diff
@@ -41,7 +41,6 @@ def _get_session_from_db(self):
                 )
                 .prefetch_related(
                     "authenticatedsession__user__groups",
-                    "authenticatedsession__user__user_permissions",
```
rissson, line 44: Is there something else for permissions we should prefetch here instead?

website/docs/users-sources/access-control/permissions.md
```diff
@@ -13,38 +13,39 @@ For instructions to add, remove, and manage permissions, refer to [Manage Permis
 
 ## Fundamentals of authentik permissions
 
+A [role](../roles/index.md) is a collection of permissions. A user or a group may have any number of roles. A user has a certain permission if they have a role that has that permission, or if they are part of a group (either directly or indirectly) which has a role that has that permission.
```
tanberry:
```
A user or a group can be assigned to any number of roles.
```

tanberry:
```suggestion
A [role](../roles/index.md) is a collection of permissions. A user or a group can be assigned to any number of roles. A user has a certain permission if they have a role that also has that permission, or if they are part of a group (either directly or indirectly) that has a role which has that permission.
```

website/docs/users-sources/access-control/permissions.md
```diff
@@ -13,38 +13,39 @@ For instructions to add, remove, and manage permissions, refer to [Manage Permis
...
+
+- Example 1 (no group): Judith has the role "RADIUS", which has every permission for RADIUS providers and property mappings. Then, Judith has permission to add/view/change/delete RADIUS providers or RADIUS property mappings.
```
tanberry:
```suggestion
- Example 1 (no group): Judith has the role "RADIUS", which has every permission for RADIUS providers and property mappings. With this role, Judith has permission to add/view/change/delete RADIUS providers or RADIUS property mappings.
```

website/docs/users-sources/access-control/permissions.md
```diff
@@ -13,38 +13,39 @@ For instructions to add, remove, and manage permissions, refer to [Manage Permis
...
+- Example 2 (direct group): Marie is part of the group "Auditors". That group has the role "Event Log manager", which in turn has the permissions "Can view Event", "Can change Event", and "Can delete Event". Then, Marie has permissions to view, change, or delete Events.
```
tanberry:
```suggestion
- Example 2 (direct group): Marie is part of the group "Auditors". That group has the role "Event Log manager", which in turn has the permissions "Can view Event", "Can change Event", and "Can delete Event". Thus, Marie has permissions to view, change, or delete Events.
```

tanberry: "then" is typically used in a time-oriented way, as in next", or "afterwards"...

website/docs/users-sources/access-control/permissions.md
```diff
@@ -13,38 +13,39 @@ For instructions to add, remove, and manage permissions, refer to [Manage Permis
...
+- Example 3 (indirect group): Elaine is part of the group "Accounting", which has a parent group "Back office", which has a parent group "Employees". The group "Employees" has the role "Read-only, which has view permissions on all object types. Then, Elaine has the permission to view any object in authentik through indirect membership of the "Employees" group.
```
tanberry:
```suggestion
- Example 3 (indirect group): Elaine is part of the group "Accounting", which has a parent group "Back office", which has a parent group "Employees". The group "Employees" has the role "Read-only", which has view permissions on all object types. So Elaine has the permission to view any object in authentik through indirect membership of the "Employees" group.
```

website/docs/users-sources/access-control/permissions.md
```diff
@@ -13,38 +13,39 @@ For instructions to add, remove, and manage permissions, refer to [Manage Permis
...
+From 2025.12, authentik's access control is fully role-based. Before 2025.12, Admins could assign permissions to individual [users](../user/index.mdx). To mimic this behavior of "User permissions", an Admin can
+:::
+
 There are two main types of permissions in authentik:
 
 - [**Global permissions**](#global-permissions)
 - [**Object permissions**](#object-permissions)
 
-Additionally, authentik employs _initial permissions_ to streamline the process of granting object-level permissions when an object (user or role) is created. This feature enables an Admin to proactively assign specific rights to a user for object creation, as well as for viewing and managing those objects and other objects created by individuals in the same role. For more details, refer to [Initial permissions](./initial_permissions.mdx).
+Additionally, authentik employs _initial permissions_ to streamline the process of granting object-level permissions when an object (user or role) is created. This feature automatically adds permissions for newly created objects to the role that created them. For more details, refer to [Initial permissions](./initial_permissions.mdx).
```
tanberry: If an admin still has to create the IP, as show in our How To steps, then it is misleading to not say that this has to be manually implemented, and to only say that the permissions are automatically assigned. I think we can say both, something like:

"When an Admin implements initial permissions, this feature automatically adds permissions for newly created objects to the role (and thus users with that role) that created them."

website/docs/users-sources/access-control/permissions.md
```diff
@@ -13,38 +13,39 @@ For instructions to add, remove, and manage permissions, refer to [Manage Permis
...
 
 ### Global permissions
 
-Global permissions define who can do what on a global level across the entire system. Some examples in authentik are the ability to add new [flows](../../add-secure-apps/flows-stages/flow/index.md) or to create a URL for users to recover their login credentials.
-
-You can assign _global permissions_ to individual [users](../user/index.mdx) or to [roles](../roles/index.md). The most common and best practice is to assign permissions to roles.
-
-An example of a global permission is [`Can view admin interface`](./manage_permissions.md#assign-can-view-admin-interface-permissions), which an administrator can assign to a non-administrator user or role to provide view access to the Admin interface.
+Global permissions define coarse-grained access control. For example, a role with a global permission of "Can change Flow" can change any [flow](../../add-secure-apps/flows-stages/flow/index.md). Some permissions only make sense as global permissions, e.g. the permission to add a specific object type or whether a user [`Can view admin interface`](./manage_permissions.md#assign-can-view-admin-interface-permissions).
```
tanberry, line 35: This cuts out a lot of info that's I'd like readers to have.. I'll come back with a suggested rewrite that is a compromise.

website/docs/users-sources/access-control/permissions.md
```diff
@@ -13,38 +13,39 @@ For instructions to add, remove, and manage permissions, refer to [Manage Permis
...
 
 ### Object permissions
 
-Object permissions have two categories:
+An object permission grants permission on a single object (e.g. a [user](../user/index.mdx), a [group](../groups/index.mdx), a [role](../roles/index.md), a [flow](../../add-secure-apps/flows-stages/flow/index.md), etc.) instead of all objects of a specific type. For example, a role with only the object permission to change the Default Authentication flow will not be able to change any other flow.
 
-- **_User_ object permissions**: defines WHO (which user) can change the **_object_**
```
tanberry, line 35: Well I'll agree that this sentence is wrong. Jeez....

website/docs/users-sources/access-control/permissions.md
```diff
@@ -13,38 +13,39 @@ For instructions to add, remove, and manage permissions, refer to [Manage Permis
...
+## Viewing permissions
 
-Object permissions are assigned, as the name indicates, to an object ([users](../user/index.mdx), [groups](../groups/index.mdx), [roles](../roles/index.md), [flows](../../add-secure-apps/flows-stages/flow/index.md), and stages), and the assigned permissions state exactly what a user or role can do TO the object (i.e. what permissions does the user or role have on that object).
+Many objects in authentik's Admin interface have a Permissions view to double-check which roles have access to that particular object. Those permissions describe what those roles can do TO the object detailed on the page.
 
-When working with object permissions it is important to understand that when you are viewing the page for an object, the permissions table shows which users or roles have permissions ON that specific object. Those permissions describe what those users or roles can do TO the object detailed on the page.
-
-For example, the Admin interface UI shown below shows a user page for the user named Peter.
+For example, the Admin interface UI shown below shows the Role Permissions table for the user named Peter.
```
tanberry: What do we mean by the "Role Permissions table"? Does this mean the tab called "Permissions" on the user page for Peter?

website/docs/users-sources/access-control/permissions.md
tanberry: Actually, it's not wrong, criminy, words are hard, because that table *does indeed define WHO (the users listed in the table) can do what to the user object (the user on the details page)

website/docs/users-sources/access-control/permissions.md
tanberry:
```suggestion
For example, the Admin interface UI shown below shows the Permissions tab for the user named Peter, with a list of what users can do what to the user object for Peter.
```

authentik/core/sessions.py
gergosimonyi: Nothing that is obvious to me.

authentik/core/models.py
gergosimonyi: Technically no, since `GroupAncestryNode` is defined after this class. Could possibly extract table names into constants. I do feel the current approach is cleaner than that though.

website/docs/users-sources/access-control/permissions.md
gergosimonyi: This table:

user-page

It is likely very likely badly named. I just took out the word "Object" from the previous implementation, where it was called "Role Object Permissions". This was a lie, as the table contains both object and global permissions.

authentik/core/models.py
rissson: Fair enough!

authentik/core/sessions.py
rissson: So grabbing permissions will always be an extra query in the request?

website/docs/users-sources/access-control/initial_permissions.mdx
```diff
@@ -29,14 +29,14 @@ The fundamental steps to implement initial permissions are as follows:
 Because the new initial permissions object is coupled with the role (and that role is assigned to a group), the initial permissions object is applied automatically to any new objects (users or flows or any object) that the member user creates.
 
 :::info
-Typically, initial permissions are assigned to a user or role that is not a super-user nor administrator. In this scenario, the administrator needs to verify that the user has the `Can view Admin interface` permission (which allows the user to access the Admin interface). For details, see Step 5 below.
+Typically, initial permissions are assigned to a role that is not a super-user nor administrator. In this scenario, the administrator needs to verify that the user has the `Can view Admin interface` permission (which allows the user to access the Admin interface). For details, see Step 5 below.
```
dewi-tik:
```suggestion
Typically, initial permissions are assigned to non-super-user, non-administrator roles. In this scenario, the administrator needs to verify that the user has the `Can view Admin interface` permission (which allows the user to access the Admin interface). For details, see Step 5 below.
```

website/docs/users-sources/access-control/permissions.md
```diff
@@ -13,38 +13,39 @@ For instructions to add, remove, and manage permissions, refer to [Manage Permis
 
 ## Fundamentals of authentik permissions
 
+A [role](../roles/index.md) is a collection of permissions. A user or a group can be assigned to any number of roles. A user has a certain permission if they have a role that also has that permission, or if they are part of a group (either directly or indirectly) that has a role which has that permission.
```
dewi-tik, line 16:
```suggestion
A [role](../roles/index.md) is a collection of permissions and can be assigned to users or groups. A user or a group can be assigned to any number of roles. 

Users inherit permissions from their direct roles or from roles assigned to groups that they belong to either directly or indirectly.
```

website/docs/users-sources/access-control/permissions.md
```diff
@@ -13,38 +13,39 @@ For instructions to add, remove, and manage permissions, refer to [Manage Permis
...
+
+- Example 1 (no group): Judith has the role "RADIUS", which has every permission for RADIUS providers and property mappings. With this role, Judith has permission to add/view/change/delete RADIUS providers or RADIUS property mappings.
```
dewi-tik, line 18:
```suggestion
- Example 1 (no group): Judith has the role "RADIUS", which has every permission for RADIUS providers and property mappings. With this role assigned, Judith has permission to add/view/change/delete RADIUS providers or RADIUS property mappings.
```

website/docs/users-sources/access-control/permissions.md
```diff
@@ -13,38 +13,39 @@ For instructions to add, remove, and manage permissions, refer to [Manage Permis
...
+- Example 2 (direct group): Marie is part of the group "Auditors". That group has the role "Event Log manager", which in turn has the permissions "Can view Event", "Can change Event", and "Can delete Event". Thus, Marie has permissions to view, change, or delete Events.
+- Example 3 (indirect group): Elaine is part of the group "Accounting", which has a parent group "Back office", which has a parent group "Employees". The group "Employees" has the role "Read-only", which has view permissions on all object types. So Elaine has the permission to view any object in authentik through indirect membership of the "Employees" group.
+
+:::info
+From 2025.12, authentik's access control is fully role-based. Before 2025.12, Admins could assign permissions to individual [users](../user/index.mdx). To mimic this behavior of "User permissions", an Admin can
```
dewi-tik:
```suggestion
From 2025.12, authentik's access control is fully role-based. Before 2025.12, Admins could assign permissions to individual [users](../user/index.mdx). To mimic this behavior of "User permissions", an Admin can create a role with the required permissions and assign that role to the user.
```

website/docs/users-sources/access-control/permissions.md
```diff
@@ -13,38 +13,39 @@ For instructions to add, remove, and manage permissions, refer to [Manage Permis
...
-
-An example of a global permission is [`Can view admin interface`](./manage_permissions.md#assign-can-view-admin-interface-permissions), which an administrator can assign to a non-administrator user or role to provide view access to the Admin interface.
+Global permissions define coarse-grained access control. For example, a role with a global permission of "Can change Flow" can change any [flow](../../add-secure-apps/flows-stages/flow/index.md). Some permissions only make sense as global permissions, e.g. the permission to add a specific object type or whether a user [`Can view admin interface`](./manage_permissions.md#assign-can-view-admin-interface-permissions).
 
 ### Object permissions
 
-Object permissions have two categories:
+An object permission grants permission on a single object (e.g. a [user](../user/index.mdx), a [group](../groups/index.mdx), a [role](../roles/index.md), a [flow](../../add-secure-apps/flows-stages/flow/index.md), etc.) instead of all objects of a specific type. For example, a role with only the object permission to change the Default Authentication flow will not be able to change any other flow.
```
dewi-tik:
```suggestion
An object permission grants permission on a single object (e.g. a [user](../user/index.mdx), a [group](../groups/index.mdx), a [role](../roles/index.md), a [flow](../../add-secure-apps/flows-stages/flow/index.md), etc.) instead of all objects of a specific type. 

For example, a role with only the object permission to change the Default Authentication flow will not be able to change any other flow.
```

website/docs/users-sources/access-control/permissions.md
```diff
@@ -13,38 +13,39 @@ For instructions to add, remove, and manage permissions, refer to [Manage Permis
...
 
-- **_User_ object permissions**: defines WHO (which user) can change the **_object_**
-- **_Role_ object permissions**: defines which ROLE can change the **_object_**
+## Viewing permissions
 
-Object permissions are assigned, as the name indicates, to an object ([users](../user/index.mdx), [groups](../groups/index.mdx), [roles](../roles/index.md), [flows](../../add-secure-apps/flows-stages/flow/index.md), and stages), and the assigned permissions state exactly what a user or role can do TO the object (i.e. what permissions does the user or role have on that object).
+Many objects in authentik's Admin interface have a Permissions view to double-check which roles have access to that particular object. Those permissions describe what those roles can do TO the object detailed on the page.
```
dewi-tik:
```suggestion
Many objects in authentik's Admin interface have a **Permissions** view to check which roles have access to that particular object. Those permissions describe what those roles can do _to_ the object detailed on the page.
```

website/docs/users-sources/access-control/permissions.md
```diff
@@ -13,38 +13,39 @@ For instructions to add, remove, and manage permissions, refer to [Manage Permis
...
 
-When working with object permissions it is important to understand that when you are viewing the page for an object, the permissions table shows which users or roles have permissions ON that specific object. Those permissions describe what those users or roles can do TO the object detailed on the page.
-
-For example, the Admin interface UI shown below shows a user page for the user named Peter.
+For example, the Admin interface UI shown below shows the Permissions tab for the user named Peter, with a list of what users can do what to the user object for Peter.
```
dewi-tik:
```suggestion
For example, the screenshot below shows the Permissions tab for the user named Peter, listing what actions users can perform on Peter's user object.
```

website/docs/users-sources/access-control/permissions.md
```diff
@@ -13,38 +13,39 @@ For instructions to add, remove, and manage permissions, refer to [Manage Permis
...
 
 ![](./user-page.png)
 
-You can see in the **User Object Permissions** table that the Admin user (`akadmin`) and one other user (roberto) has permissions on Peter (that is, on the user object named Peter).
-
-Looking at another example, with a flow object called `default-recovery-flow`, you can see that the Admin user (akadmin) has all object permissions on the flow, but roberto only has a few permissions on that flow.
+You can see in the **Role Permissions** table that the Admin role and one other role (Read-only) have permissions on Peter (that is, on the user object named Peter). The Admin role has all object permissions on this object, while the Read-only role view permission.
```
dewi-tik:
```suggestion
You can see in the **Role Permissions** table that the Admin role and one other role (Read-only) have permissions on Peter (that is, on the user object named Peter). The Admin role has all object permissions on this object, while the Read-only role only has the view permission.
```

authentik/core/sessions.py
rissson: The answer is yes, specifically:
```sql
SELECT DISTINCT 1 AS "a" FROM "authentik_core_group" LEFT OUTER JOIN "authentik_core_groupancestry" ON ("authentik_core_group"."group_uuid" = "authentik_core_gro
upancestry"."ancestor_id") WHERE (("authentik_core_group"."group_uuid" IN (SELECT U0."group_uuid" AS "pk" FROM "authentik_core_group" U0 INNER JOIN "authentik_core_user_ak_groups" U1 ON (U0."group_uuid" = U1."grou
p_id") WHERE U1."user_id" = 6) OR "authentik_core_groupancestry"."descendant_id" IN (SELECT U0."group_uuid" AS "pk" FROM "authentik_core_group" U0 INNER JOIN "authentik_core_user_ak_groups" U1 ON (U0."group_uuid"
= U1."group_id") WHERE U1."user_id" = 6)) AND "authentik_core_group"."is_superuser") LIMIT 1
```

dewi-tik (commented):
Some suggestions

website/docs/users-sources/access-control/permissions.md
tanberry: added

website/docs/users-sources/access-control/permissions.md
tanberry: yep already caught that one, thanks!

website/docs/users-sources/access-control/permissions.md
tanberry: seems superfluous...

website/docs/users-sources/access-control/permissions.md
tanberry: But i think now it is still a lie...? That button _assigns selected roles the permission to alter the user object_. The table lists which roles have permissions to alter that User. 

So what the button is really doing (I think?) is adding new roles, specifically those roles than can alter the user object Peter. Or are you saying the button now _also_ adds role that have global permissions on the user object Peter?

Either way I think a more clear button label would be "Assign role permissions". The word "to" is confusing, not correct, imo.

Honestly the title of the page is confusing... "Role Permissions" very much sounds like this is where you assign permissions to roles, which it absolutely is not. That's done over on the Roles page. I would call this page simply "Permissions" (as the tab is also called).

tanberry (approved):
Thank you @gergosimonyi for making our RBAC truly RBAC-y and role-based, and updating the docs!

website/docs/users-sources/access-control/permissions.md
gergosimonyi: So we have two issues here, yes?

The button, currently labelled "Assign to new role". What the button does: presents a popup, where the admin can select a and toggle permissions. The selected role will receive the selected permissions on this object.

So the button is labelled "Assign to new role" because it assigns object permissions to a new role. I'm happy with this, but also happy to entertain an alternative.

---

The table, currently titled "Role Permissions". This title carries absolutely no meaning and should be changed. The previous screenshot doesn't address the whole picture, it gets more complex when you're looking at a role's Permissions tab, where in addition to the problematic "Role permissions", we also have "Assigned global permissions" and "Assigned object permissions" tables. The label has to fit in that left sidebar:

kép

So just calling it "Permissions" doesn't really work in my opinion. If I had infinite characters, I'd title this table "Roles that have the ability do something to this object, regardless of whether the source of that ability is an object permission or a global permission".

For now I'll change this label to "Permissions on this object", that's still terrible, but at least has meaning. Happy to entertain an alternative.

tanberry: Different UI page @gergosimonyi .. I was looking at the User detail page.. as in my screenshot above.

gergosimonyi: Offline resolution: changed button label to "Assign role permissions", keeping table title (for now at least).

BeryJu (approved):

BeryJu:
oh yeah before we merge is it possible to merge at least some of the migrations in authentik_core? @gergosimonyi

gergosimonyi:
@BeryJu (I don't like it styled this way but) I merged the last 3.

Issue #13463, opened 2025-03-11
s3 custom_url makes signature verification fail


benaryorg:
**Describe the bug**
When I upload a file to S3 and I have `custom_domain` set, I get an error.
I do not get the error with the `custom_domain` setting removed, even retroactively for existing files.

**To Reproduce**

1. use s3 backend with `custom_url`
1. upload flow background image
1. open flow
1. background image not loaded

**Expected behavior**
URL both with and without `custom_url` works.

**Screenshots**
not applicable

**Logs**
not applicable (signature calculates properly, it's just somehow wrong)

**Version and Deployment (please complete the following information):**

-   authentik version: 2024.12.1 (the code around the signature doesn't seem like it has changed recently, and I don't have a good way to quickly upgrade for testing)
-   Deployment: manual (issue is not deployment-related)

**Additional context**

This occurs with a radosgw running with Ceph squid (v19.2.0).

When the `custom_url` setting is off I get this URL:

https://s3.ovh.xn--idk5byd.net/benaryorg-authentik/media/public/application-icons/Grafana_logo.svg?X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Credential=ISXPUYU60051HVZ8LXX0%2F20250314%2Fdefault%2Fs3%2Faws4_request&X-Amz-Date=20250314T181538Z&X-Amz-Expires=3600&X-Amz-SignedHeaders=host&X-Amz-Signature=253b6e3e445dff3b553b228a01f5ed685d30f5f6d6ef93db3c350c813e53c531

When it is on I get this one:

https://benaryorg-authentik.s3.ovh.ニャー.net/media/public/application-icons/Grafana_logo.svg?X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Credential=ISXPUYU60051HVZ8LXX0%2F20250314%2Fdefault%2Fs3%2Faws4_request&X-Amz-Date=20250314T181657Z&X-Amz-Expires=3600&X-Amz-SignedHeaders=host&X-Amz-Signature=56a737db3aa170b5909fec7a6639cb9b96ccd48513db8666d30d56203001c5e1

Since it works with one but not the other (and the signature differs) we can *probably* assume that the following line is okay and works, despite the punycode domain:

https://github.com/goauthentik/authentik/blob/e87a17fd8169d3fa92bcc47eb2743928df83bc95/authentik/root/storages.py#L109

The error I get is this:

```xml
[...]

PR #17087, opened 2025-09-29
storages: s3 custom domain signature fix

benaryorg:
## Details

- #13463

Effectively this fixes s3 signatures for custom domains.
The solution for this is somewhat hacky, however it does apply cleanly on *version/2025.2.4* and I have tested it and can confirm that it works.
I'll set this to draft state because I do not currently have the capacity to finish the testing and linting, so if anyone wants to pick this up, please go ahead.
To be perfectly honest, I just copied the stuff I use to generate such links personally straight out of my shell history into the source code.
It is however the only solution that I have managed to find which actually works.
**NB**: this has *only* been tested when cherry-picked onto *2025.2.3* and *2025.8.3*!
**NB**: this is more or less a bandaid, a proper rewrite of the method to generate the URl would be desirable IMHO, especially since it will compute the signature even when it is not needed. Using the proper class for the URl would also be good, to avoid any weirdnesses for edge-cases. Right now the onus of providing a working string is put on the admin.

The commit message is included verbatim below.

---

S3 signatures are fickle since there are several hashes of components which must ultimately be correct for the signature to match with no room for error.
Since both the *host* header as well as the path, the former by default, the latter necessarily.
The issue with custom domains is therefore that the *host* header mismatches unless the signature was created for that specific domain.
This is complicated further by *boto3* making it incredibly difficult to actually sign a link with the path *not* containing the bucket prefix.
Therefore setting the endpoint alone does not work with a truly custom domain (i.e. one where the data resides on the root of the domain).

Changing this to a rather hacky workaround accessing fields of the internal data structures and grabbing the signer directly allows for signing the request.
This comes with several caveats such as the parameters going largely untouched, however in preliminary testing it works fine.
There are no superfluous slashes (i.e. no `https://example.com//file.svg`), no superfluous prefixes (i.e. `https://example.com/example.com/example.com/file.svg`), and there are no signature errors in general.
Uploads still work.
Reminder that the application cache may need clearing before the changes reflect in the non-admin UI.

Relates:

- [issue 13463](https://redirect.github.com/goauthentik/authentik/issues/13463)
- [pr 14706 ](https://redirect.github.com/goauthentik/authentik/pull/14706)

---

## Checklist

If applicable

### DIFF
--- a/authentik/root/storages.py
+++ b/authentik/root/storages.py
@@ -3,13 +3,13 @@
 import os
 from urllib.parse import parse_qsl, urlsplit
 
+from botocore.awsrequest import AWSRequest
 from django.conf import settings
 from django.core.exceptions import SuspiciousOperation
 from django.core.files.storage import FileSystemStorage
 from django.db import connection
 from storages.backends.s3 import S3Storage as BaseS3Storage
 from storages.utils import clean_name, safe_join
-import botocore
 
 from authentik.lib.config import CONFIG
 
@@ -99,8 +99,10 @@ class S3Storage(BaseS3Storage):
 
         if self.custom_domain:
             custom_url = f"{self.url_protocol}//{self.custom_domain}/{name}"
-            request = botocore.awsrequest.AWSRequest(http_method or "GET", custom_url)
-            self.bucket.meta.client._request_signer.sign("GetObject", request, signing_type="presign-url", expires_in=expire)
+            request = AWSRequest(http_method or "GET", custom_url)
+            self.bucket.meta.client._request_signer.sign(
+                "GetObject", request, signing_type="presign-url", expires_in=expire
+            )
             prepared_request = request.prepare()
             url = prepared_request.url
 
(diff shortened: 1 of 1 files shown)


benaryorg:
> sh: 1: cross-env: not found

and

> npm error No workspaces found

I'm not 100% sure, but the CI failures sound unrelated to this PR.

benaryorg:
I have confirmed this patch to work on both *2025.2.3* and *2025.8.3* on my end.

Some of the files I uploaded intermediately (I think with a different patch) needed a reupload since their URLs had duplicate prefixes (`https://id-static.benary.org/media/public/media/public/application-icons/50352242-5439-41e8-bc38-94a7a28535d2.svg`).
Given that the files did not work at all prior to that I don't think this is much of a problem.

Removing the `custom_domain` option will yield the usual endpoint-based URls with valid signatures all the same.

BeryJu:
We're still planning to rework file management but for the time being this should be a good fix

BeryJu:
this shouldn't be needed anymore with #17535 merged @dominic-r @rissson

dominic-r:
What Jens said ^^ Thanks for this PR

PR #17114, opened 2025-09-29
web: Fix table child alignment

GirlBossRush:
## Details

This PR fixes two issues stemming from table styles:

- Modals are subject to CSS inheritance of a parent table row
- Table action columns wrap awkwardly on smaller viewports

The fix here is a stop gap while modals get migrated to `<dialog>`

### DIFF
--- a/web/src/admin/applications/ApplicationListPage.ts
+++ b/web/src/admin/applications/ApplicationListPage.ts
@@ -130,12 +130,17 @@ export class ApplicationListPage extends WithBrandConfig(TablePage<Application>)
                   </a>`
                 : html`-`,
             html`${item.providerObj?.verboseName || msg("-")}`,
-            html`<ak-forms-modal>
+            html`<div>
+                <ak-forms-modal>
                     <span slot="submit">${msg("Update")}</span>
                     <span slot="header">${msg("Update Application")}</span>
                     <ak-application-form slot="form" .instancePk=${item.slug}>
                     </ak-application-form>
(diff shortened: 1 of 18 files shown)


web/src/admin/applications/ApplicationListPage.ts
```diff
@@ -130,12 +130,17 @@ export class ApplicationListPage extends WithBrandConfig(TablePage<Application>)
                   </a>`
                 : html`-`,
             html`${item.providerObj?.verboseName || msg("-")}`,
-            html`<ak-forms-modal>
+            html`<div>
```
GirlBossRush, line 133: The fix here is having one degree of separation from the parent `<td>`. We might benefit from moving this into all table cells in the future to ensure that the cell contents always have a blocking parent without CSS table semantics.

web/src/elements/table/Table.ts
```diff
@@ -74,18 +74,11 @@ export abstract class Table<T extends object>
                 .presentational {
                     --pf-c-table--cell--MinWidth: 0;
                 }
-                @container (width > 600px) {
+                @container (width > 1200px) {
```
GirlBossRush, line 77: The previous number was just as arbitrary but lacked consideration for some of our busier pages like the user's list.

kensternberg-authentik: Got it.

kensternberg-authentik (approved):

PR #17115, opened 2025-09-29
web: Additional text field properties, ARIA fixes

GirlBossRush:
## Details

This PR fixes several input issues discovered while preparing for e2e user creation tests:

- Fix ARIA grouping, label association.
- Add placeholder style.
- Wrap helper text on larger viewports.
- Fix label ARIA.

### DIFF
--- a/web/src/common/styles/authentik.css
+++ b/web/src/common/styles/authentik.css
@@ -225,12 +225,21 @@ html > form > input {
 .pf-c-form__helper-text {
     text-wrap: pretty;
 }
+
+::placeholder {
+    font-style: italic;
+}
+
 /* #endregion */
 
 /* #region Fonts */
 
 .pf-m-monospace {
     font-family: var(--pf-global--FontFamily--monospace);
+
+    &::placeholder {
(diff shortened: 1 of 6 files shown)


web/src/common/styles/authentik.css
```diff
@@ -219,12 +219,24 @@ html > form > input {
...
+}
+
 /* #endregion */
 
 /* #region Fonts */
 
 .pf-m-monospace {
     font-family: var(--pf-global--FontFamily--monospace);
+
+    &::placeholder {
```
GirlBossRush, line 240: I've found a great deal of writing on why placeholders shouldn't be used to replace labels, but little on the preferred content of placeholders when used. VoiceOver will announce placeholders, followed by their parent region (dialog in this case), then the label, followed by the `aria-describedby` element. The label is repeated once more to indicate the field announcement has concluded.

**In situations which mirror the label or offer an example, the content can be quite strange and repetitive.**

Image

---

**The _Type a [blank]..._ formatting here attempts to give a clearer usage without too much repetition.**

Image

kensternberg-authentik (approved):

PR #17116, opened 2025-09-29
web/a11y: Table header -- Fix pagination jitter, prepare alignment

GirlBossRush:
## Details

This PR is in preparation for ARIA-friendly table headers. While the layout is improved, the changes included represent a smaller scope of the final result to aid in review.


kensternberg-authentik (approved):
Looks good.

PR #17117, opened 2025-09-29
web/a11y: Table header -- Search input

GirlBossRush:
## Depends on

- #17116 

## Details

Both the basic table search and QL search inputs have improved keyboard access, as well a more consistent layout on smaller screens. QL search inputs now announce suggestions according to the ARIA combobox role.

Screenshot 2025-09-29 at 21 56 16

### DIFF
--- a/web/src/components/ak-search-ql/index.ts
+++ b/web/src/components/ak-search-ql/index.ts
@@ -5,6 +5,7 @@ import { rootInterface } from "#common/theme";
 
 import { FormAssociated, FormAssociatedElement } from "#elements/forms/form-associated-element";
 import { PaginatedResponse } from "#elements/table/Table";
+import { ifPresent } from "#elements/utils/attributes";
 
 import DjangoQL, { Introspections } from "@mrmarble/djangoql-completion";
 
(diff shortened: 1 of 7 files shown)


web/src/components/ak-search-ql/index.ts
```diff
@@ -123,8 +149,16 @@ export class QLSearch extends FormAssociatedElement<string> implements FormAssoc
 
     public set value(value: unknown) {
         const parsed = typeof value === "string" ? value : "";
+        const trimmed = parsed.trim();
+
+        this.setFormValue(trimmed, parsed);
+
+        if (trimmed) {
```
kensternberg-authentik, line 156: Suggestion: If we find ourselves doing this a lot, we should add, monkeypatch, or otherwise fix `this.internals.states` to have a `toggle(name, value?)` where we can say things like `this.internals.states.toggle("present")` or `this.internals.states.set("present", !!trimmed)`

kensternberg-authentik: What I _really_ want to be able to say is `this.internals.states.present = !this.internals.states.present`, but I don't live in the Good Timeline.  :-)

kensternberg-authentik (approved):
Not too shabby.

PR #17144, opened 2025-09-30
web/a11y: Fix dark theme color contrast

GirlBossRush:
## Details

This PR covers a few situations where the dark theme doesn't apply colors that can be read with low-contrast backgrounds. Some CSS rules may appear twice to take precedence when combined with certain elements.

### DIFF
--- a/web/src/common/styles/theme-dark.css
+++ b/web/src/common/styles/theme-dark.css
@@ -98,21 +98,11 @@ body {
     --pf-c-table--tr--m-hoverable--active--BackgroundColor: var(--ak-dark-background-lighter);
     --pf-global--link--Color: var(--pf-global--link--Color--light);
     --pf-global--link--Color--hover: var(--pf-global--link--Color--light--hover);
-    /* --pf-c-table__sort__button__text--Color: var(--pf-global--Color--light-200);
-    --pf-c-table__button--hover--Color: var(--pf-global--link--Color--light--hover); */
 }
 
 .pf-c-table__text {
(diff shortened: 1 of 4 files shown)


web/src/common/styles/theme-dark.css
```diff
@@ -86,21 +89,31 @@ body {
...
+.pf-c-table__sort.pf-m-selected {
+    .pf-c-table__button .pf-c-table__text {
+        --pf-c-table__sort__button__text--Color: var(--pf-global--Color--light-200);
 
-.pf-c-table__sort:not(.pf-m-selected) .pf-c-table__button .pf-c-table__text {
-    color: var(--ak-dark-foreground) !important;
-}
+        --pf-c-table__button--hover--Color: var(--pf-global--link--Color--light--hover);
+    }
+} */
```
BeryJu:
```suggestion
```

Remove commented out code

web/src/common/styles/theme-dark.css
```diff
@@ -14,6 +14,9 @@
     --ak-global--Color--100: var(--ak-dark-foreground) !important;
     --pf-c-page__main-section--m-light--BackgroundColor: var(--ak-dark-background-darker);
     --pf-global--BorderColor--100: var(--ak-dark-background-lighter) !important;
+    --pf-global--danger-color--100: #e6190a;
```
BeryJu: I'd prefer if we keep the original patternfly colours, although their #C9190B is quite dark, so maybe this is fine

BeryJu (approved):

PR #17223, opened 2025-10-02
Makefile: Fix kerberos tests for brew users

PeshekDotDev:
## Details

We tell people to install krb5 with brew on dev env setup. The output from running `brew install krb5` tells people they have to update their path to add krb5 manually. This is very easy to miss, and without it in your path, the tests in `authentik/sources/kerberos` will hang infinitely.

This PR automatically updates your path to add the brew install of krb5 when the test suite is run via `make test` or `make all`

---

## Checklist

If applicable

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,21 +17,20 @@ pg_user := $(shell uv run python -m authentik.lib.config postgresql.user 2>/dev/
 pg_host := $(shell uv run python -m authentik.lib.config postgresql.host 2>/dev/null)
 pg_name := $(shell uv run python -m authentik.lib.config postgresql.name 2>/dev/null)
 
-UNAME := $(shell uname)
-
 # For macOS users, add the libxml2 installed from brew libxmlsec1 to the build path
 # to prevent SAML-related tests from failing and ensure correct pip dependency compilation
-ifeq ($(UNAME), Darwin)
-# Only add for brew users who installed libxmlsec1
-	BREW_EXISTS := $(shell command -v brew 2> /dev/null)
-	ifdef BREW_EXISTS
-		LIBXML2_EXISTS := $(shell brew list libxml2 2> /dev/null)
-		ifdef LIBXML2_EXISTS
-			BREW_LDFLAGS := -L$(shell brew --prefix libxml2)/lib $(LDFLAGS)
-			BREW_CPPFLAGS := -I$(shell brew --prefix libxml2)/include $(CPPFLAGS)
-			BREW_PKG_CONFIG_PATH := $(shell brew --prefix libxml2)/lib/pkgconfig:$(PKG_CONFIG_PATH)
-		endif
-	endif
+# These functions are only evaluated when called in specific targets
+LIBXML2_EXISTS = $(shell brew list libxml2 2> /dev/null)
+KRB5_EXISTS = $(shell brew list krb5 2> /dev/null)
+
+LIBXML2_LDFLAGS = -L$(shell brew --prefix libxml2)/lib $(LDFLAGS)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -51,7 +52,11 @@ go-test:
 	go test -timeout 0 -v -race -cover ./...
 
 test: ## Run the server tests and produce a coverage report (locally)
-	uv run coverage run manage.py test --keepdb authentik
+ifdef KRB5_EXISTS
+	PATH="$(shell brew --prefix krb5)/sbin:$(shell brew --prefix krb5)/bin:$$PATH" uv run coverage run manage.py test --keepdb authentik
```
BeryJu: I think you could probably just `export` the changed path and not need to duplicate command

PeshekDotDev: each physical line in a Makefile recipe is executed in a separate shell instance, so we can't export, but I did find a way to inject the path in that is much cleaner

PeshekDotDev:
@BeryJu I made an additional minor update with this — you can now run `make test path/to/tests` to run specific folders tests. Let me know what you think of that. `make test` will still default to running all tests

PeshekDotDev:
Using `:=` causes the variable to be given a value ever time a make command is ran. Using `=` means the variable only gets a value when it is referenced, meaning that the brew commands no longer run every time a command such as `make help` is ran

rissson:
Some of this won't be necessary anymore after we fix krb5 for Heimdal, as the lib discovery has been improved to find the correct paths on Linux and macos

PeshekDotDev:
Ah glad to hear. What are your thoughts @rissson ? Should we merge this for now and remove it after? or just remove the `krb5` parts?

rissson:
The kerberos PR should be ready soon, I'll edit this PR once that is merged

PeshekDotDev:
Awesome, much appreciated

rissson:
actually no, it won't fix them, we can go ahead with this.

PeshekDotDev:
Do you have time for a review on this @rissson ?

Makefile
```diff
@@ -17,21 +17,20 @@ pg_user := $(shell uv run python -m authentik.lib.config postgresql.user 2>/dev/
 pg_host := $(shell uv run python -m authentik.lib.config postgresql.host 2>/dev/null)
 pg_name := $(shell uv run python -m authentik.lib.config postgresql.name 2>/dev/null)
 
-UNAME := $(shell uname)
-
 # For macOS users, add the libxml2 installed from brew libxmlsec1 to the build path
 # to prevent SAML-related tests from failing and ensure correct pip dependency compilation
-ifeq ($(UNAME), Darwin)
```
rissson, line 24: Why are we loosing that check? `brew` doesn't exist on Linux, we shouldn't require users to have it installed.

rissson (changes requested):

Makefile
PeshekDotDev: 1. Brew exists as linuxbrew on linux (what im using on my linux machine)

2. There isn't a requirement to have brew installed. We check if the user has installed libxml2 or krb5 via brew, and if they haven't, we don't do anything

`LIBXML2_EXISTS = $(shell brew list libxml2 2> /dev/null)` will be falsy both if brew is not installed or if libxml2 is not installed by brew

The original implementation of this checked for mac users, but the feedback was that broke peoples setups, so we did a check to see if brew installed the packages in question, which was reported to fix this setup for machines not using brew. Now that it does, I don't see the need for an OS check, as the libxml2 and krb5 check are the actual safeguards

rissson: Fair enough

rissson (approved):

PR #17233, opened 2025-10-03
Fixed CORS requests failing with forward auth via proxy provider

### LINKED ISSUE #10057: application/o/authorize endpoint missing CORS headers
**Describe the bug**
Not sure it is a bug or a misconfiguration somewhere...
Web application is calling the /application/o/authorize/ endpoint from the browser.

It is failing with the message in the Chrome Debug Console

`Access to fetch at 'https://auth.mydomain.com/application/o/authorize/?client_id=xxxxxxxxxx&redirect_uri=https%3A%2F%2Forigin.domain.com%2Foutpost.goauthentik.io%2Fcallback%3FX-authentik-auth-callback%3Dtrue&response_type=code&scope=email+profile+ak_proxy+openid&state=m2qJvqqq3tNbVgvzIgz7h6l3w4-hErhUX1fd5WmLJ_g' (redirected from 'https://origin.domain.com/info') from origin 'https://origin.domain.com' has been blocked by CORS policy: Response to preflight request doesn't pass access control check: Redirect is not allowed for a preflight request.`

I do not see the headers : access-control-allow-origin: https://origin.domain.com

**To Reproduce**

Create a Forward auth (Single Applicaiton) provider and set `External host` to `https://origin.domain.com`
Make sure the user was already logged in --> no new login requested.
Try to get authorisation from https://auth.mydomain.com/application/o/authorize/

**Expected behavior**
Preflight contains CORS headers 
`access-control-allow-origin: https://origin.domain.com`

-   authentik version: 2024.4.2

### LINKED ISSUE #25263: OAuth authorize and revoke responses omit CORS headers
## What

OAuth endpoints used from a browser SPA omit CORS headers (/authorize and /revoke). Also: missing Access-Control-Allow-Origin on API/OAuth calls from a browser. Confirm it is the same allow-origin config before closing.

## Want

- OAuth endpoints used from a browser SPA omit CORS headers (/authorize
- Related missing Access-Control-Allow-Origin on API/OAuth calls from a browser
- Confirm it is the same allow-origin config before closing

## Ideas this came from

- [#10057 application/o/authorize endpoint missing CORS headers](https://github.com/goauthentik/authentik/issues/10057) — Not sure it is a bug or a misconfiguration somewhere... Web application is calling the /application/o/authorize/ endpoint from the browser. It is failing with t
- [#11057 `/revoke` end point missing CORS header](https://github.com/goauthentik/authentik/issues/11057) — After reading the source code, I've discovered the implementation for token revocation was followed by the RFC 7009: https://datatracker.ietf.org/doc/html/rfc70
- [#12159 Missing CORS header: Cannot perform an API request - can I configure allowed origins?](https://github.com/goauthentik/authentik/issues/12159) — When trying to fetch data from the authentik-API i run into an error: Cross-source (cross-origin) request blocked: The same-source rule prohibits reading the ex

solidDoWant:
## Details

This allows CORS requests to go through the proxy provider forward auth "plan" (not sure if this is the right word), fixing XHR and fetch requests, as well as WebSockets when the forward auth plan needs to be ran again.

I'm not really familiar with this codebase and I don't do much front end web dev, so there may be issues (including security vulnerabilities) introduced by this. Please review this carefully and assume that I don't know what I'm doing.

I've tested this patch against a branch based off of the latest release, and I have it working with Radarr, an app that uses credentialed XMLHTTPRequests and SignalR over WebSockets. It's working great, and renewing the proxy cookie is 100% transparent to me as a user. Note that I have _only_ tested with Envoy/Istio. I don't have (or plan on having) traefik, caddy, or nginx deployed as a reverse proxy, so I'll be unable to test the changes that are specific to these reverse proxies.

One new requirement of this is that front-end application requests must set `XMLHTTPRequest.withCredentials = true`. Without this, the browser won't send the authentik session cookie to the authentik callback endpoint, making authentik think that every request needs to go through the application's configured auth flow. This needs to be set even if the application otherwise does not use credentials for these requests. I don't think that there is a way around this with how forward auth is handled with the proxy cookie. I'm not sure where to document this.

Closes https://github.com/goauthentik/authentik/issues/10057

---

## Checklist

-   [ ] Local tests pass (`ak test authentik/`) - These take like an hour to run locally. If they're not ran automatically in CI then I'll run them overnight.

If applicable

-   [ ] The documentation has been updated - Please see above note about application needing `XMLHTTPRequest.withCredentials = true`

Closes #25263

### DIFF
--- a/authentik/policies/views.py
+++ b/authentik/policies/views.py
@@ -68,13 +68,20 @@ class PolicyAccessView(AccessMixin, View):
         is not caught, and will return directly"""
         raise NotImplementedError
 
+    def supports_cors_preflight_requests(self) -> bool:
+        """If true, OPTIONS requests will be answered without authentication or permission checks.
+        This is useful for CORS preflight requests."""
+        return hasattr(self, "options") and callable(self.options)
+
     def dispatch(self, request: HttpRequest, *args: Any, **kwargs: Any) -> HttpResponse:
+        # Validate request before doing anything else
         try:
             self.pre_permission_check()
         except RequestValidationError as exc:
             if exc.response:
                 return exc.response
             return self.handle_no_permission()
+
         try:
             self.resolve_provider_application()
         except (Application.DoesNotExist, Provider.DoesNotExist) as exc:
@@ -82,14 +89,19 @@ class PolicyAccessView(AccessMixin, View):
             return self.handle_no_permission_authenticated(
                 PolicyResult(False, _("Failed to resolve application"))
             )
-        # Check if user is unauthenticated, so we pass the application
-        # for the identification stage
-        if not request.user.is_authenticated:
-            return self.handle_no_permission()
-        # Check permissions
-        result = self.user_has_access()
-        if not result.passing:
-            return self.handle_no_permission_authenticated(result)
+
+        # CORS preflight requests should not require authentication
+        if request.method != "OPTIONS" or not self.supports_cors_preflight_requests():
+            # Check if user is unauthenticated, so we pass the application
+            # for the identification stage
+            if not request.user.is_authenticated:
+                return self.handle_no_permission()
+
+            # Check permissions
+            result = self.user_has_access()
+            if not result.passing:
+                return self.handle_no_permission_authenticated(result)
+
         return super().dispatch(request, *args, **kwargs)
 
     def handle_no_permission(self) -> HttpResponse:
--- a/authentik/providers/oauth2/views/authorize.py
+++ b/authentik/providers/oauth2/views/authorize.py
@@ -427,9 +427,10 @@ class AuthorizationFlowInitView(BufferedPolicyAccessView):
                     parsed_url._replace(query=urlencode(args, quote_via=quote, doseq=True))
                 )
 
-        # Add CORS headers based on the provider's redirect URIs
+        # Add CORS headers based on the provider's redirect URIs, if a provider exists and has
+        # redirect URIs configured
         allowed_origins = []
-        if self.provider and hasattr(self.provider, "redirect_uris"):
+        if hasattr(self, "provider") and self.provider and hasattr(self.provider, "redirect_uris"):
             allowed_origins = [x.url for x in self.provider.redirect_uris]
         cors_allow(self.request, response, *allowed_origins)
 
--- a/authentik/providers/oauth2/views/authorize.py
+++ b/authentik/providers/oauth2/views/authorize.py
@@ -62,7 +62,7 @@ from authentik.providers.oauth2.models import (
     ResponseTypes,
     ScopeMapping,
 )
-from authentik.providers.oauth2.utils import HttpResponseRedirectScheme
+from authentik.providers.oauth2.utils import HttpResponseRedirectScheme, TokenResponse, cors_allow
 from authentik.providers.oauth2.views.userinfo import UserInfoView
 from authentik.stages.consent.models import ConsentMode, ConsentStage
 from authentik.stages.consent.stage import (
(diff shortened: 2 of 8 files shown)


solidDoWant:
I'm aware code coverage is lacking. I'll add tests when/if this gets a first pass review. Running tests in this project locally is pretty difficult and time consuming, so I don't really want to put in this work if the actual code change needs a major rework.

solidDoWant:
Friendly ping for reviewers (@BeryJu maybe). This is starting to get stale and I really don't want this to get lost/abandoned. This is affecting at least 25 other people, probably significantly more (see linked issue).

solidDoWant:
CI failures are all completely unrelated (something related to NPM packages which I didn't touch)

dewi-tik:
Thanks for the contribution @solidDoWant!

A member of the backend team will review this ASAP. Apologies for the delay, lots of tasks being juggled at the moment.

solidDoWant:
Thanks @dewi-tik. Completely understand, just wanted to make sure this didn't get lost :)

I see the code coverage check failing. I'll get this fixed if I can get a first-pass review on this PR first to make sure that it doesn't need large changes/a rewrite.

solidDoWant:
Been testing this for a few weeks and I think that there may be an issue with how proxy provider auth is implemented that makes it technically impossible for a secure CORS implementation. Looking for some feedback to see if I'm write, or misunderstanding something.

Here is what the XHR flow looks like for the authentication middleware (Istio/envoy external auth + Authentik) that I'd like to use this with:

1. Front-end application sends HTTP request to backend, which has a reverse proxy in the middle (`frontend -> reverse proxy -> backend`)
2. If the request has a specific valid credential cookie attached, it is considered "authenticated" by the reverse proxy and passed through to the backend (`frontend -> reverse proxy -> backend`, with a side-channel between `reverse proxy <-> authentik` prior to forwarding the request to the backend)
3. If the cookie is missing or invalid, the reverse proxy forwards the request to Authentik, which returns a 302 redirect. This redirect changes the domain from `application.domain.com` to `authentik.domain.com`, as well as the path.
4. The front-end receives the redirect and executes it.
5. Authentik receives this request and does some processing, returning another 302 redirect. This redirects back to the original domain name (`application.domain.com`), but with an Authentik-specific `/outpost.goauthentik.io` path.
6. The front-end receives the redirect and executes it. PRIOR TO THIS POINT all requests had the `Origin` header set. AT AND AFTER THIS POINT, the `Origin` header is set to a literal string value of `null` for XHR requests. This is a browser security feature.
7. The reverse proxy intercepts this request and forwards it to Authentik. Authentik returns another 302 response, and includes the `Set-Cookie` header to set the cookie mentioned in (2). This 302 response redirects back `//application.domain.com/<originally requested path from (1)>`.
8. The front-end executes this redirect, going back to (1) but now with the authentication cookie.

Because the browser passes `Origin: null` in part of this flow for XHR requests, the backend _must_ set `Access-Control-Allow-Origin: null` for this to work. The problem I see is that this basically bypasses all CORS protections. If an attacker can get the user's browser to execute a request to `application.domain.com` from a separate `attacker.web.site` while setting `Origin: null` (which is trivial), then request will succeed.

As I understand it, the browser will always set `Origin: null` with this redirect chain for cross-origin requests. The only way to avoid the backend needing to set `Access-Control-Allow-Origin: null` would be for the Authentik-imposed flow to take place all under the same domain. The obvious issue with this is that domain-specific cookies for Authentik (`authentik_session`) could not be sent.

Is there any path forward here for the proxy-provider to work with CORS?

solidDoWant:
Closing this as there doesn't seem to be any follow-through in reviewing it after attempts to reach maintainers in several places

PR #17255, opened 2025-10-06
web: Fix flow autofocus element targeting.

GirlBossRush:
## Details

This PR refines #16739 with additional autofocus fixes such as when...

- Autofocus is lost after switching tabs
- A stage element intercepts a focus target before rendering completes
- Compatibility mode delays the visibility of nested custom elements between Lit lifecycle callbacks

### DIFF
--- a/web/src/elements/decorators/intersection-observer.ts
+++ b/web/src/elements/decorators/intersection-observer.ts
@@ -1,0 +1,95 @@
+/**
+ * @file Intersection Observer Decorator for LitElement
+ */
+
+import { LitElement } from "lit";
+import { property } from "lit/decorators.js";
+
+/**
+ * Type for the decorator
+ */
+export type IntersectionDecorator = <T extends LitElement>(target: T, propertyKey: keyof T) => void;
+
+/**
+ * A decorator that applies an IntersectionObserver to the element.
+ * This is useful for lazy-loading elements that are not visible on the screen.
+ *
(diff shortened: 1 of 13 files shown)


web/src/elements/decorators/intersection-observer.ts
```diff
@@ -0,0 +1,95 @@
+/**
+ * @file Intersection Observer Decorator for LitElement
+ */
```
GirlBossRush, line 3: This decorator was originally designed to help with stale form issues in the upcoming dialog/modals fixes. The implementation can also handle our form viewport checks.

web/src/flow/stages/authenticator_totp/AuthenticatorTOTPStage.ts
```diff
@@ -114,15 +115,22 @@ export class AuthenticatorTOTPStage extends BaseStage<
...
+                        msg("TOTP Code"),
+                    )}
                     <input
                         id="totp-code-input"
                         type="text"
                         name="code"
                         inputmode="numeric"
                         pattern="[0-9]*"
-                        placeholder="${msg("Please enter your TOTP Code")}"
-                        autofocus=""
```
GirlBossRush, line 125: Omitting autofocus here to avoid scrolling past the QR Code

BeryJu (approved):

PR #17279, opened 2025-10-07
web/a11y: Prefers more field contrast

GirlBossRush:
## Details

This PR adds additional support for a user's preferred level of contrast on inputs and fieldsets. The choice of field colors is deferred to the browser, allowing for default colors (with respect to our theme) to take effect.

## Screenshots

Screenshot 2025-10-07 at 05 44 05

Screenshot 2025-10-07 at 05 43 31

### DIFF
--- a/web/src/elements/forms/SearchSelect/ak-search-select-view.ts
+++ b/web/src/elements/forms/SearchSelect/ak-search-select-view.ts
@@ -9,7 +9,7 @@ import type { GroupedOptions, SelectOption, SelectOptions } from "#elements/type
 import { randomId } from "#elements/utils/randomId";
 
 import { msg } from "@lit/localize";
-import { CSSResult, html, nothing, PropertyValues } from "lit";
+import { css, CSSResult, html, nothing, PropertyValues } from "lit";
 import { customElement, property, state } from "lit/decorators.js";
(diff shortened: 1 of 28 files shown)


web/src/elements/forms/SearchSelect/ak-search-select-view.ts
```diff
@@ -69,7 +69,17 @@ export interface ISearchSelectView {
...
 export class SearchSelectView extends AKElement implements ISearchSelectView {
-    static styles: CSSResult[] = [PFBase, PFForm, PFFormControl, PFSelect];
+    static styles: CSSResult[] = [
+        PFBase,
+        PFForm,
+        PFFormControl,
+        PFSelect,
+        css`
+            .pf-c-select {
+                --pf-c-select__toggle-wrapper--MaxWidth: unset;
```
kensternberg-authentik: MaxWidth is a non-heritable property; this should be `initial`, not `unset`.

> For a property that is inherited (e.g. color) it means inherit, and for a property that isn’t inherited (e.g. float) it means initial. ([inherit, initial, unset, revert](https://css-tricks.com/inherit-initial-unset-revert/)).

So this will do what you mean, but it's... unclear why it means that.  :-)

kensternberg-authentik (approved):

PR #17300, opened 2025-10-07
website: Enable legacy integrations redirects.

GirlBossRush:
## Details

This PR enables our legacy redirects for integrations, allowing URLs such as `https://integrations.goauthentik.io/services/[some integration name]/` to redirect to their current path.

This functionality has existed for sometime but was disabled during the Netlify redirect clean up.


dominic-r (approved):

tanberry (approved):
Thanks so much @GirlBossRush !

PR #17359, opened 2025-10-09
tasks: store messages in separate table

rissson:
## Details

allows for instantaneous message reporting

in a future release we can add a migration to remove the previous field and migrate the data over. Not doing it in this release to avoid breaking changes and missed rows during migrations

---

## Checklist

If applicable


authentik/tasks/models.py
```diff
@@ -115,37 +115,48 @@ def _make_message(
...
-    def warning(self, message: str | Exception, save: bool = False, **attributes):
-        self.log(self.uid, TaskStatus.WARNING, message, save=save, **attributes)
 
-    def error(self, message: str | Exception, save: bool = False, **attributes):
-        self.log(self.uid, TaskStatus.ERROR, message, save=save, **attributes)
+class TaskLog(models.Model):
+    id = models.UUIDField(default=uuid4, primary_key=True, editable=False)
+
+    task = models.ForeignKey(Task, on_delete=models.CASCADE)
+    log = models.JSONField()
```
BeryJu: I'm leaning on the side of having maybe level and message be concrete fields and use a JSON field for kv-attrs

rissson: Something like this? (API not yet updated)

BeryJu (approved):

PR #17365, opened 2025-10-09
blueprints: regenerate schema

dominic-r:
The result of running `make gen-build` following today's PR


BeryJu (changes requested):
Actually this object should be listed here https://github.com/goauthentik/authentik/blob/main/authentik/blueprints/v1/importer.py#L142 as it mustn't be created in blueprints

BeryJu (approved):

PR #17414, opened 2025-10-13
outpost/proxyv2: postgresstore: credential refresh

dominic-r:
## Details

REPLACE ME

---

## Checklist

If applicable

### DIFF
--- a/internal/outpost/proxyv2/postgresstore/connpool.go
+++ b/internal/outpost/proxyv2/postgresstore/connpool.go
@@ -1,0 +1,292 @@
+package postgresstore
+
+import (
+	"context"
+	"database/sql"
+	"database/sql/driver"
+	"strings"
+	"sync"
+	"time"
+
+	log "github.com/sirupsen/logrus"
+	"gorm.io/driver/postgres"
+	"gorm.io/gorm"
+
+	"goauthentik.io/internal/config"
+)
+
+// RefreshableConnPool wraps sql.DB and refreshes PostgreSQL credentials on authentication errors
+// This implements gorm.ConnPool interface to allow credential rotation
+type RefreshableConnPool struct {
+	mu         sync.RWMutex
+	db         *sql.DB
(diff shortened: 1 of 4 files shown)


internal/outpost/proxyv2/postgresstore/connpool.go
```diff
@@ -0,0 +1,296 @@
...
+	// 28P02: invalid_password (deprecated but still used)
+	errMsg := strings.ToLower(err.Error())
+	return strings.Contains(errMsg, "password authentication failed") ||
+		strings.Contains(errMsg, "no password supplied") ||
+		strings.Contains(errMsg, "invalid authorization specification") ||
+		strings.Contains(errMsg, "pq: password authentication failed") ||
+		strings.Contains(errMsg, "authentication failed") ||
+		strings.Contains(errMsg, "28p01") || // PostgreSQL error code
+		strings.Contains(errMsg, "28000") || // PostgreSQL error code
+		strings.Contains(errMsg, "28p02") // PostgreSQL error code (deprecated)
```
BeryJu: Yeah this is pretty terrible, and makes me wonder if postgres does localised error messages (I'd hope not)..ideally we just catch the actual error codes and nothing else

dominic-r: yeaaaaa.... I think we could just use the error codes like "28P01"

dominic-r: `ws.com): failed SASL auth: FATAL: password authentication failed for user \"myuser\" (SQLSTATE 28P01)","event":"` 

we could mabye just check for FATAL: but that does not garantee a password refresh or stick to the error codes like in this log ^^

dominic-r: https://github.com/goauthentik/authentik/pull/17414/commits/e348d2807e21db4b2007c7f2d8030bccb0143b11

BeryJu (approved):

PR #17470, opened 2025-10-14
providers/oauth2: allow setting logout method always

BeryJu:
cc @PeshekDotDev 
required for terraform as we specify this value as required in the schema and have a default, so terraform will set the default value, which this doesn't allow

see https://github.com/goauthentik/terraform-provider-authentik/pull/764


PeshekDotDev:
Ah, fair play. Okay thank you for updating and letting me know

PR #17280, opened 2025-10-07
web: Minimal mobile flow

GirlBossRush:
## Details

This PR is a first-pass at reducing visual noise when authenticating on smaller screens.

## Screenshots

Screenshot 2025-10-07 at 05 55 01

Screenshot 2025-10-07 at 05 55 10

Screenshot 2025-10-07 at 15 01 10
Screenshot 2025-10-07 at 15 00 21
Screenshot 2025-10-07 at 14 57 25

Screenshot 2025-10-07 at 16 45 25

Screenshot 2025-10-07 at 16 47 05

PR #17472, opened 2025-10-14
web: Flow Layout Refinements

GirlBossRush:
## Dependencies

- #17444

## Details

This PR is a follow up on #17280, fixing layout and palette issues.

- Sidebar and content left/right positioning now reflows on mobile viewports
- Card padding now gracefully adjusts to mobile viewports, no longer snapping from each breakpoint
- Over-scrolling on the page no longer revealing background image
- Background now remains obscured when resizing
- Background image no longer repaints twice after first render
- Footer contents are now consistently visible when moving from sidebar layout to mobile viewport

## Caveats

This iteration includes a few strategies to retain visual stability as flow cards load, with some caveats:

- Keeping all the content within the viewport is prioritized over the card’s baseline centering. For mobile and desktop screens, there is no shift, however on tablet sized screens there can be a slight shift as the upper padding shrinks. IMO this is a good balance since the two most common device sizes have a stable layout without any scroll bars

- When using a sidebar layout, the flow card will default to stacked if the screen’s aspect ratio would result in a just a sliver of background visibility. This also keeps the contents stable without scroll bars

### DIFF
--- a/web/src/common/theme.ts
+++ b/web/src/common/theme.ts
@@ -287,6 +287,44 @@ export function applyDocumentTheme(hint: CSSColorSchemeValue | UIThemeHint = "au
 }
 
 /**
+ * A CSS variable representing the global background image.
+ */
+export const AKBackgroundImageProperty = "--ak-global--background-image";
+
+/**
+ * Applies the given background image URL to the document body.
+ *
+ * This method is very defensive to avoid unnecessary DOM repaints.
+ */
+export function applyBackgroundImageProperty(value?: string | null): void {
+    const fallbackOrigin = window.location.origin;
+
+    if (!value || !URL.canParse(value, fallbackOrigin)) {
+        return;
+    }
+
+    const nextBackgroundURL = new URL(value, fallbackOrigin);
+
+    const currentBackgroundImage = getComputedStyle(document.body, "::before").backgroundImage;
+    let currentBackgroundImageURL: URL | null = null;
+
+    if (currentBackgroundImage && currentBackgroundImage !== "none") {
+        // Extract URL from background-image property
+        const [, urlMatch] = currentBackgroundImage.match(/url\(["']?([^"']*)["']?\)/) || [];
+
+        if (URL.canParse(urlMatch)) {
+            currentBackgroundImageURL = new URL(urlMatch, fallbackOrigin);
+        }
+    }
+
+    if (currentBackgroundImageURL && currentBackgroundImageURL.href === nextBackgroundURL.href) {
+        return;
+    }
+
+    document.body.style.setProperty(AKBackgroundImageProperty, `url("${nextBackgroundURL.href}")`);
+}
+
+/**
  * Returns the root interface element of the page.
  *
  * @todo Can this be handled with a Lit Mixin?
--- a/web/src/elements/ak-locale-context/ak-locale-context.ts
+++ b/web/src/elements/ak-locale-context/ak-locale-context.ts
@@ -8,7 +8,6 @@ import { EVENT_LOCALE_CHANGE, EVENT_LOCALE_REQUEST } from "#common/constants";
 import { AKElement } from "#elements/Base";
 import { customEvent } from "#elements/utils/customEvents";
 
-import { html } from "lit";
 import { customElement, property } from "lit/decorators.js";
 
 /**
@@ -26,6 +25,10 @@ import { customElement, property } from "lit/decorators.js";
  */
 @customElement("ak-locale-context")
 export class LocaleContext extends WithBrandConfig(AKElement) {
+    protected createRenderRoot(): HTMLElement | DocumentFragment {
+        return this;
+    }
+
     /// @attribute The text representation of the current locale */
     @property({ attribute: true, type: String })
     locale = DEFAULT_LOCALE;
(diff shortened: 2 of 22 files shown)


web/src/elements/ak-locale-context/ak-locale-context.ts
```diff
@@ -26,6 +25,10 @@ import { customElement, property } from "lit/decorators.js";
  */
 @customElement("ak-locale-context")
 export class LocaleContext extends WithBrandConfig(AKElement) {
+    protected createRenderRoot(): HTMLElement | DocumentFragment {
```
GirlBossRush, line 28: Ensures one less level of injected global styles. We’ve slowly moved towards this in all “context-like” components which have no styles of their own.

web/src/common/theme.ts
```diff
@@ -286,6 +286,44 @@ export function applyDocumentTheme(hint: CSSColorSchemeValue | UIThemeHint = "au
...
+ * A CSS variable representing the global background image.
+ */
+export const AKBackgroundImageProperty = "--ak-global--background-image";
+
+/**
+ * Applies the given background image URL to the document body.
+ *
+ * This method is very defensive to avoid unnecessary DOM repaints.
+ */
+export function applyBackgroundImageProperty(value?: string | null): void {
```
GirlBossRush, line 299: Extracted from the FlowExecutor. We can use this elsewhere in the future

web/src/flow/components/ak-flow-card.ts
```diff
@@ -30,58 +31,30 @@ export class FlowCard extends AKElement {
...
         // No title if the challenge doesn't provide a title and no custom title is set
-        let title = undefined;
+        let title: null | SlottedTemplateResult = null;
         if (this.hasSlotted("title")) {
             title = html`<h1 class="pf-c-title pf-m-3xl"><slot name="title"></slot></h1>`;
         } else if (this.challenge?.flowInfo?.title) {
             title = html`<h1 class="pf-c-title pf-m-3xl">${this.challenge.flowInfo.title}</h1>`;
         }
-        return html`${title ? html`<div class="pf-c-login__main-header">${title}</div>` : nothing}
+        const footer = this.hasSlotted("footer") ? html`<slot name="footer"></slot>` : null;
```
GirlBossRush, line 48: For ternaries that potentially represent a template result, preferring `null` over `nothing` allows us to use the value’s truthiness in predicates — less boilerplate to check if we should render a container if its children are existent

web/src/flow/stages/authenticator_validate/AuthenticatorValidateStage.css
```diff
@@ -0,0 +1,22 @@
+.authenticator-button {
+    /* compatibility-mode-fix */
+    & {
+        align-items: center;
+        width: 100%;
+        display: grid;
+        grid-template-columns: minmax(auto, 2rem) minmax(33%, max-content);
```
GirlBossRush, line 7: Ported from our current design. This needs some attention to make use of PatternFly’s existing full width card buttons

web/src/flow/FlowExecutor.ts
```diff
@@ -48,20 +51,14 @@ import PFLogin from "@patternfly/patternfly/components/Login/login.css";
...
-    [FlowLayoutEnum.Stacked]: "ak-login-container",
-    [FlowLayoutEnum.UnknownDefaultOpenApi]: "ak-login-container",
-} as const satisfies Record<FlowLayoutEnum, string>;
-
 @customElement("ak-flow-executor")
 export class FlowExecutor
     extends WithCapabilitiesConfig(WithBrandConfig(Interface))
     implements StageHost
 {
+    static readonly DefaultLayout: FlowLayoutEnum =
```
GirlBossRush, line 59: @BeryJu 

Not sure if this is a bug per se, but if the authentication flow’s layout does not match the invalidation flow’s layout, the contents will shift when clicking the “not you?” link.

web/src/standalone/loading/index.entrypoint.ts
```diff
@@ -38,24 +47,17 @@ export class Loading extends AKElement {
...
-                    >
-                        <span class="pf-c-spinner__clipper"></span>
-                        <span class="pf-c-spinner__lead-ball"></span>
-                        <span class="pf-c-spinner__tail-ball"></span>
-                    </span>
-                    <h1 class="pf-c-title pf-m-lg">${msg("Loading...")}</h1>
-                </div>
-            </div>
-        </section>`;
+        return html`<span class="pf-c-spinner pf-m-xl" aria-hidden="true">
```
GirlBossRush, line 50: BeryJu @kensternberg-authentik 

The loading spinner needs some additional support to fully work with the screen reader and the “ARIA live” concept, but this iteration ensures that the progress bar is read as a momentary and indefinite value.

In some cases, the timing of this render path can also result in the screen reader discovering the progress bar after the parent component has already rendered. IMO a future revision might be more practical as a Django template partial rather than a full web component which depends on JavaScript to initialize.

web/src/user/LibraryPage/ak-library-application-list.css
```diff
@@ -32,7 +32,6 @@
 
 .pf-c-card {
     --pf-c-card--BoxShadow: var(--pf-global--BoxShadow--md-bottom);
-    --pf-c-card--BackgroundColor: var(--pf-global--BackgroundColor--150);
```
GirlBossRush, line 35: Removed since cards now inherit the color from the global styles

GirlBossRush:
Resolved via #17444

PR #17486, opened 2025-10-15
sources/oauth: Fallback user id to sub field

### LINKED ISSUE #16115: Cannot find id field on OIDC sources
**Describe the bug**
When logging in via Gitlab OIDC source, the error "Could not determine id." is thrown. Because there is no field named "id" in the oauth callback from Gitlab.

**To Reproduce**
On a fresh install, create a GitLab social login, add it to the default identification stage and try to log in via gitlab.

**Expected behavior**
It should be able to create a user based on the Gitlab callback.
I patched the [get_user_id function](https://github.com/goauthentik/authentik/blob/08ddd3e8d76a1d49aaec687910471dc2fd1373f7/authentik/sources/oauth/views/callback.py#L90) locally to fallback to the "sub" field if "id" does not exist and it works flawlessly.

**Version and Deployment:**

-   authentik version: 2025.6.3

luechtdev:
## Details

Adding a fallback for OAuth Sources IDs via the sub field. 
I encountered some problems where external OAuth Providers did not have any id field on the callback request. 

closes #16115

### DIFF
--- a/authentik/sources/oauth/views/callback.py
+++ b/authentik/sources/oauth/views/callback.py
@@ -94,6 +94,8 @@ class OAuthCallback(OAuthClientMixin, View):
         """Return unique identifier from the profile info."""
         if "id" in info:
             return info["id"]
+        if "sub" in info:
+            return info["sub"]
         return None
 
     def handle_login_failure(self, reason: str) -> HttpResponse:


dewi-tik:
This was fixed in #22672. But thank you for the PR.

PR #17505, opened 2025-10-16
web/a11y: More Consistent UI Labelling

GirlBossRush:
## Details

This PR updates our UI to use a more consistent labelling for placeholders, buttons, empty states, and other references to API provided objects.

This iteration is not all-encompassing, focusing at least one level of navigation depth in the UI -- i.e. the UI at a first glance.


web/src/admin/groups/RelatedUserList.ts
dominic-r, line 504:
```suggestion
                        <span class="pf-c-switch__label">${msg("Hide service accounts")}</span>
```

dominic-r: dunno why its showing weirdly. anyways, just service-accounts -> service accounts

PR #17511, opened 2025-10-16
outpost/proxyv2: postgresstore: db/pool/misc cleanup and enhancement

dominic-r:


internal/outpost/proxyv2/application/session_postgres_test.go
```diff
@@ -9,7 +9,6 @@ import (
 	"github.com/google/uuid"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
-	"gorm.io/driver/postgres"
```
rissson, line 12: Is this intended?

BeryJu, line 12:
```suggestion
	_ "gorm.io/driver/postgres"
```

internal/outpost/proxyv2/application/session_postgres_test.go
```diff
@@ -9,7 +9,6 @@ import (
 	"github.com/google/uuid"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
```
dominic-r, line 11:
```suggestion
	"github.com/stretchr/testify/require"
	_ "gorm.io/driver/postgres"
```

internal/outpost/proxyv2/application/session_postgres_test.go
dominic-r: Applied in https://github.com/goauthentik/authentik/pull/17511#discussion_r2444825919 cause accepting suggestions to deleted lines is not supported in GH

internal/outpost/proxyv2/refresh.go
```diff
@@ -49,7 +49,7 @@ func (ps *ProxyServer) Refresh() error {
 			existing.Stop()
 		}
 		if err != nil {
-			ps.log.WithError(err).Warning("failed to setup application")
+			ps.log.WithError(err).Fatal("failed to setup application")
```
BeryJu:
```suggestion
			ps.log.WithError(err).Warning("failed to setup application")
```

I dont think we'd want the outpost to crash in this case, since other applications might be running fine

BeryJu (approved):

authentik/lib/default.yml
```diff
@@ -21,6 +21,7 @@ postgresql:
   user: authentik
   port: 5432
   password: "env://POSTGRES_PASSWORD"
+  sslmode: disable
```
severin, line 24: @BeryJu did this "break"/invalidate the documentation? https://version-2025-10.goauthentik.io/install-config/configuration/#ssltls-settings says that the default ssl mode is `verify-ca`, but with this change it seems to be `disable` now.

Disclaimer: I don't fully understand the configuration loading sequence and what overwrites what (defaults in the code, config files, env vars). But in our case we had to set `AUTHENTIK_POSTGRESQL__SSLMODE` explicitly to `verify-ca` after the upgrade from 2025.8 to 2025.10 because otherwise postgres refused connections with the message "no encryption".

PR #17522, opened 2025-10-17
web/a11y: User library -- fix issues surrounding element focus, ARIA labeling.

GirlBossRush:
## Details

This PR fixes several issues which prevented consistent element focusing in the user library such as...

- When the search input renders for the first time with a prefilled value
- When accepting an autocomplete suggestion while only one card is available
- When a focus target is a button, such as a RAC modal
- When using a keyboard to tab through elements outside of the search input's focus

Additionally, a few ARIA issues have been fixed which prevented search results from consistently communicating the number of cards, which order they render when in column format, and which card is in focus in each group. This is achieved by porting some of the library elements into function components, removing the shadow root boundaries which prevent a direct association of search input to card group. 

The visual layout is the same, however card groups now use our fieldset elements which support high-contrast mode:

### Normal contrast
 
No extra padding or margin:

Screenshot 2025-10-17 at 06 48 21

### More contrast

Group labels moved inward, matching our design in other fieldsets with visible legends, as well as `<ak-form-group />` in high-contrast mode.

Screenshot 2025-10-17 at 06 47 44


web/src/user/LibraryPage/ApplicationList.ts
```diff
@@ -0,0 +1,96 @@
+import type { AppGroupEntry } from "./types.js";
```
GirlBossRush, line 1: This is largely ported from `ak-application-list` as function component, making a lot of the focus logic possible when connecting search inputs to their respective cards.

web/src/user/index.entrypoint.ts
```diff
@@ -261,6 +261,10 @@ class UserInterfacePresentation extends WithBrandConfig(AKElement) {
 //
 @customElement("ak-interface-user")
 export class UserInterface extends WithBrandConfig(AuthenticatedInterface) {
+    public static shadowRootOptions = { ...AKElement.shadowRootOptions, delegatesFocus: true };
+
+    public override tabIndex = -1;
+
```
GirlBossRush, line 267: Applied so that `document.activeElement` targets deeper into the `<ak-library-impl>`

kensternberg-authentik (approved):
This one looks good.

PR #17636, opened 2025-10-21
crypto: improve discovery for mounted k8s TLS Secrets

ember-ana:
## Details

When discovering certificates, authentik currently has special support for Certbot-style file naming. This adds the same for [Kubernetes TLS Secrets that have been mounted as a volume](https://kubernetes.io/docs/concepts/configuration/secret/#using-secrets-as-files-from-a-pod).

This is already somewhat possible, but has the drawback of needing to use [subPath](https://kubernetes.io/docs/concepts/storage/volumes/#using-subpath) to set the file names (else the secrets will always be named `tls`).

The Kubernetes documentation on Secrets explains why this is sub-optimal (emphasis added):
> When a [...] Secret is updated, Kubernetes tracks this and updates the data in the volume
> [...]
> A container using a Secret as a subPath volume mount does <ins>not</ins> receive automated Secret updates.

Handling the files in this way allows the Secret to instead be mounted under a subdirectory without the need for subPath.

---

When using a Secret generated with cert-manager's [additionalOutputFormats](https://cert-manager.io/docs/usage/certificate/#additional-certificate-output-formats) option, some of the extra files produced might lead to an unnecessary warning as they fail to parse.

This also includes support for handling those gracefully by ignoring files named `ca.crt`, `tls-combined.pem`, or `key.der`. I'm not sure whether this constitutes a breaking change.

---

## Checklist


ember-ana:
rebased onto main & added test coverage

rissson (approved):

PR #17759, opened 2025-10-28
providers/proxy: fix missing JWT/claims header

### LINKED ISSUE #17750: x-authentik-jwt header empty after upgrade to  2025.10.0
**Describe the bug**
JWT header ist empty, after upgrade to 2025.10.0

**To Reproduce**

Use a "Proxy Provider".
Check the headers, the x-authentik-jwt is still available but contains an empty string

**Expected behavior**
x-authentik-jwt should contain its value as in the releases before

-   authentik version: [e.g. 2025.10.0]

Thank you for any help

### LINKED ISSUE #17753: Scope mapping for proxy / additional header does not work any more with 2025.10.0
**Describe the bug**
We defined the following scope mapping and added it to an proxy-provider:

```
token = request.user.attributes.get("influxdbtoken")

return {
    "ak_proxy": {
        "user_attributes": {
            "additionalHeaders": {
                "Authorization": f"Bearer {token}"
            }
        }
    }
}
```

**To Reproduce**

1. Create the scope mapping within "Customization - Property Mappings"
2. Add the scope to your proxy provider
3. Check the headers, no "Authorization" is added to the request

**Expected behavior**
Scope mapping should still work and add extra headers to the request

-   authentik version:2025.10.0

BeryJu:
raw JWT was not correctly de-serialized when loading session

also fix other fields not being correctly serialized

fixes #17750
fixes #17753

### DIFF
--- a/internal/outpost/proxyv2/application/auth_test.go
+++ b/internal/outpost/proxyv2/application/auth_test.go
@@ -7,6 +7,7 @@ import (
 	"testing"
 
 	"github.com/gorilla/sessions"
+	"github.com/mitchellh/mapstructure"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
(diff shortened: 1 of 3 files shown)


dominic-r (approved):

internal/outpost/proxyv2/application/auth_test.go
```diff
@@ -70,35 +71,33 @@ func TestClaimsJSONSerialization(t *testing.T) {
...
 	}
 
-	// Convert map to Claims using JSON marshaling (like getClaimsFromSession does)
-	jsonData, err := json.Marshal(claimsMap)
-	require.NoError(t, err)
-
+	// Convert map to Claims using mapstructure marshaling (like getClaimsFromSession does)
 	var claims types.Claims
-	err = json.Unmarshal(jsonData, &claims)
+	err := mapstructure.Decode(claimsMap, &claims)
```
dominic-r, line 100: more mapstructure, good catch

PR #17771, opened 2025-10-28
packages/django-postgres-cache: rework to use ORM

rissson:
## Details

REPLACE ME

---

## Checklist

If applicable


BeryJu (approved):

packages/django-postgres-cache/django_postgres_cache/backend.py
```diff
@@ -66,59 +41,34 @@ def _base_set_expiry(self, timeout: float | None) -> datetime:
         exp.replace(microsecond=0)
```
melizeche, line 41: afaik replace creates a new datetime, this should be
`exp = exp.replace(microsecond=0)`

packages/django-postgres-cache/django_postgres_cache/backend.py
```diff
@@ -1,62 +1,37 @@
...
-class DatabaseCache(BaseDatabaseCache):
-    def __init__(self, table: str, params: dict[str, Any]) -> None:
-        super().__init__(table, params)
-        self.reverse_key_func = import_string(params["REVERSE_KEY_FUNCTION"])
-        self._table = CacheEntry._meta.db_table
-        self.cache_model_class = CacheEntry
+class DatabaseCache(BaseCache):
+    pickle_protocol = pickle.HIGHEST_PROTOCOL
 
-    def _cull(self, *args: Any, **kwargs: Any) -> None:
```
melizeche, line 25: nit: This backend doesn't handle expired entry cleanup but since this is a standalone package, we should document that consumers must implement their own periodic cleanup imho

packages/django-postgres-cache/django_postgres_cache/backend.py
melizeche, line 41: afaik this returns a new datetime. this should be something like
`exp = exp.replace(microsecond=0)`

packages/django-postgres-cache/django_postgres_cache/backend.py
```diff
@@ -137,5 +89,99 @@ def set(
...
+        timeout: float | None = DEFAULT_TIMEOUT,
+        version: int | None = None,
+    ) -> bool:
+        key = self.make_and_validate_key(key, version=version)
+        expiry = self._make_expiry(timeout)
+        return bool(CacheEntry.objects.filter(cache_key=key).update(expires=expiry))
+
+    def delete(self, key: Any, version: int | None = None) -> bool:
+        key = self.make_and_validate_key(key, version=version)
+        return bool(CacheEntry.objects.filter(cache_key=key).delete())
```
melizeche, line 104: `QuerySet.delete()` returns a tuple `(count, {label: count})`, so _bool(...)_ is always True here. 
Needs something like `count, _ = ...delete()` then return `count !=0`

PR #17783, opened 2025-10-29
outposts: update permissions more eagerly

### LINKED ISSUE #17740: Outposts not working with 2025.10.0
**Describe the bug**
When accessing an outpost after all components were upgraded to 2025.10.0 I get this error (real domain is obfuscated):
**{
	"Message": "no app for hostname",
	"Host": "whoami-xxxx.fqdn",
	"Detail": "Check the outpost settings and make sure 'whoami-xxxx.fqdn' is included."
}**

**To Reproduce**

Setup an outpost with releases before 2025.10.0 and upgrade it to 2025.10.0. That's all.

**Expected behavior**
Outpost should behave like with older releasees.

**Screenshots**
Error log from above is enough.

**Logs**
See above.

2025.10.0

-   authentik version: 2025.10.0

**Additional context**
Outpost is running on seperated docker host with container image ghcr.io/goauthentik/proxy:2025.10.0.
Authentik host is using ghcr.io/goauthentik/server:2025.10.0

dominic-r:
Closes https://github.com/goauthentik/authentik/issues/17740

Some writings on the subject: https://www.notion.so/authentiksecurity/29aaee05b24e80f0ac77ffc777f7a892?source=copy_link


internal/outpost/ak/api.go
```diff
@@ -28,6 +28,23 @@ import (
 
 const ConfigLogLevel = "log_level"
 
+// loggingRoundTripper wraps an http.RoundTripper and logs all requests
+type loggingRoundTripper struct {
+	inner  http.RoundTripper
+	logger *log.Entry
+}
+
+func (l *loggingRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
```
BeryJu: should probably be in a file in https://github.com/goauthentik/authentik/tree/main/internal/utils/web

internal/outpost/proxyv2/proxyv2.go
```diff
@@ -28,11 +28,12 @@ type ProxyServer struct {
...
-	apps        map[string]*application.Application
-	log         *log.Entry
-	mux         *mux.Router
-	akAPI       *ak.APIController
+	cryptoStore    *ak.CryptoStore
+	apps           map[string]*application.Application
+	log            *log.Entry
+	mux            *mux.Router
+	akAPI          *ak.APIController
+	sessionBackend string // "postgres" or "filesystem"
```
BeryJu: I'd rather just have a reference to the session store and call the function to get the type than store the type in a string

internal/outpost/ak/api.go
dominic-r: This was temporary so I could debug why the outpost was not receiving any providers from authentik. I've removed it as we don't need logging for outbound request urls made by the outpost.

dominic-r:
Failing CI depends on: https://github.com/goauthentik/authentik/pull/17806

BeryJu (approved):

PR #17806, opened 2025-10-30
internal: fix go deprecation for +build

dominic-r:
Fixes golangci-lint too.

See: https://stackoverflow.com/questions/68360688/whats-the-difference-between-gobuild-and-build-directives


rissson (approved):

BeryJu (approved):
This was only there cause the Go extension for vscode added it automatically, but it doesn't seem to be doing that anymore so /shrug

PR #17812, opened 2025-10-30
web/a11y: Modals, Command Palette (Merge branch)

GirlBossRush:
This PR is a merge branch for `<ak-modal>`, a custom element built on the native `<dialog>` element, with support for lazy-loading and screen-reader compatibility.

Given our heavy use of modals in the UI, the new component is fully backwards compatible with existing implementations. To illustrate the migration path, this PR applies `<ak-modal>` to a few examples: **About authentik**, **Create Application**, and **Edit Application**.

- **About authentik** extends the base modal class with an updated template. The modal is added to the DOM only when the button is clicked and removed when closed — ideal for infrequently used or content-heavy modals.
- **Create Application** uses a `renderModal` utility function to trigger a modal from an event listener, with the same lazy-loading behaviour.
- **Edit Application** is similar, but also demonstrates passing data to the modal without a bound event listener per table row.

## Details

Introduces `<ak-modal>`, a new custom element built on the native `<dialog>` element, with support for lazy-loading, screen-reader accessibility (`delegatesFocus`, proper ARIA labeling), and a `renderModal` / `modalInvoker` utility API for programmatic and declarative usage.
- Migrates modals across the admin UI to use `<ak-modal>`, replacing the legacy `TableModal` and `ModalButton`-based patterns. Affected areas include: application create/edit, user management (password reset, impersonation, bulk revoke, activate/deactivate), group/role member selection, provider selection, permission selection, RAC launch endpoint, policy bindings, About authentik, and more.
- Removes deprecated components: `TableModal`, `ProviderSelectModal`, `PermissionSelectModal`, `GroupSelectModal`, and the old `Modal/modal.css` styles. Replaces them with leaner form-based equivalents (`ProviderSelectForm`, `PermissionSelectForm`, `UserGroupSelectForm`, `MemberSelectForm`).
- Extracts form serialization into a dedicated `serialization.ts` module and validation error reporting into `errors.ts`, simplifying `Form.ts`.
- Adds a new command palette (`<ak-command-palette>`) with admin and user variants, including keyboard navigation and search.
- Refactors the wizard component (`WizardStep.ts`, `ak-wizard-steps.ts`) and the application wizard to align with the new modal system, cleaning up step types and provider form composition.
- Moves several utility files to more appropriate locations (`policies/utils.ts`, `sources/utils.ts`, `events/ObjectChangelog.ts`, `events/UserEvents.ts` into `admin/` or `common/`).

### Migration highlights

| Pattern | Before | After |
|---|---|---|
| **Infrequent modals** (e.g. About) | Always in DOM | Lazy — added on open, removed on close |
| **Table row actions** (e.g. Edit Application) | `TableModal` subclass with embedded table | `renderModal()` / `modalInvoker` directive with a plain form |
| **Selection modals** (e.g. Group/Provider/Permission select) | `*SelectModal` extending `TableModal` | `*SelectForm` rendered inside `<ak-modal>` |
| **Programmatic open** | Manual orchestration controller | `renderModal(MyForm, init)` returns a `Promise<void>` |

### Misc changes

- `TableModal` is removed — any downstream subclasses must migrate to `<ak-modal>` + form pattern.
- `AkControlElement` renamed to `ControlElement`.
- Several internal module paths changed (see file renames above).


web/src/admin/applications/wizard/steps/ProviderChoices.ts
```diff
@@ -1,19 +1,28 @@
...
+export interface ProviderRendererInit {
+    render: ProviderRenderer;
+    order: number;
+}
 
-export const providerTypeRenderers: Record<
-    string,
-    { render: () => TemplateResult; order: number }
-> = {
+export const WizardProviderRenderRecord = {
```
kensternberg-authentik: This is a variable that can be referenced.  I'd really prefer to keep to the convention that they start with a lower case letter.  I'd like to reserve capital letters for Types and Interfaces (and Classes which, yes, _sigh_, can be referenced at runtime, but they do dual duty as types).

kensternberg-authentik (approved):
Overall, this looks good, and I've run it to see that it doesn't _seem_ broken.  :-)

PR #17858, opened 2025-10-31
internal: add default go http server timeouts

BeryJu:
add default timeouts

### DIFF
--- a/internal/utils/web/server.go
+++ b/internal/utils/web/server.go
@@ -1,0 +1,17 @@
+package web
+
+import (
+	"net/http"
+	"time"
+)
+
+func Server(h http.Handler) *http.Server {
+	return &http.Server{
+		Handler:           h,
+		ReadHeaderTimeout: 5 * time.Second,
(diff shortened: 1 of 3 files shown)


internal/utils/web/server.go
```diff
@@ -0,0 +1,17 @@
...
+)
+
+func Server(h http.Handler) *http.Server {
+	return &http.Server{
+		Handler:           h,
+		ReadHeaderTimeout: 5 * time.Second,
+		ReadTimeout:       30 * time.Second,
+		WriteTimeout:      60 * time.Second,
+		IdleTimeout:       120 * time.Second,
+		MaxHeaderBytes:    http.DefaultMaxHeaderBytes,
```
rissson, line 15: Should these be configurable somehow?

BeryJu: Seeing as we don't have these timeouts configurable for LDAP/Radius/etc servers, I think this is fine for now

Issue #8708, opened 2024-02-27
Unable to use Date prompt type in enrollment flow.


CookieGMVN:
**Describe the bug**
I wanted to add a Date prompt in the enrollment flow, asking the user to provide their birthdate (`attributes.birthday`), just for verification purpose. But, after I added it to the flow and bind the prompt to a stage, when user filled their birthdate info in and click "Continue", authentik returned with the error "Request has been denied. Failed to update user. Please try again later.".

**To Reproduce**

1. Create a new prompt with Date format.
2. Bind that prompt to a prompt stage in an enrollment flow.
3. Go to the enrollment flow and fill all the information.
4. See error.

**Expected behavior**
The user will be written to authentik, and the verification email should be sent to the user.

**Screenshots**
![image](https://github.com/goauthentik/authentik/assets/73780382/4b7ccd75-a053-475b-bbf6-8001eb7e1f40)
![image](https://github.com/goauthentik/authentik/assets/73780382/d6828a2c-a10f-4a10-981a-f5a6a3f78880)
![image](https://github.com/goauthentik/authentik/assets/73780382/d2056849-498e-49bd-9c22-40042811a1ef)
![image](https://github.com/goauthentik/authentik/assets/73780382/eff8a039-936d-4150-be00-6d9d1287e6b2)

**Logs**
From Server:
```
{"auth_via": "unauthenticated", "domain_url": "domain", "event": "/api/v3/flows/executor/nzid-register/?query=next%3D%252F", "host": "domain", "level": "info", "logger": "authentik.asgi", "method": "GET", "pid": 44, "remote": "x.x.x.x", "request_id": "94ca682b026941abaa8b853b09380512", "runtime": 610, "schema_name": "public", "scheme": "https", "status": 200, "timestamp": "2024-02-27T18:41:10.517363", "user": "", "user_agent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/122.0.0.0 Safari/537.36"}
```
I can't find the logs from worker (LOG_LEVEL already TRACE). Will send the logs asap.

**Version and Deployment (please complete the following information):**

-   authentik version: [2024.2.1](https://goauthentik.io/docs/releases/2024.2)
-   Deployment: docker-compose

**Additional context**
I tested the flow using Text type. It worked, but I still want client to render a date picker to user for easier input.

balderekjk:
[...]

PR #17890, opened 2025-11-01
stages/user_write: Fix user attributes are not sanitized under certains conditions

se7kn8:
## Details

This PR fixes a case where certain prompt types (like the date prompt) lead to a failure in the user write stage, because of unsanitized attributes. The original issue was reported in #8708 and later fixed in #8926. However, the fix in #8926 only worked if the prompt would write to `attributes_foo` or `attributes.foo.bar`. If the normal syntax is used `attributes.foo` the stage would try to directly update the user attributes, which lead to unsanitized values. This PR fixes that case.

---

## Checklist

If applicable


BeryJu:
needs tests @BeryJu

BeryJu (approved):

PR #17898, opened 2025-11-02
website/integrations: macmon NAC

nicedevil007:
## Details

Keeping the doc momentum going with a guide for macmon NAC.

---

## Checklist

If applicable

-   [X] The documentation has been formatted (`make docs`)


dewi-tik:
Thanks again @nicedevil007. Are you happy with the changes that I made?

nicedevil007 (commented):
mhhh I know that this seems to be the case for most applications but not for macmon.

`This documentation lists only the settings that you need to change from their default values. Be aware that any changes other than those explicitly mentioned in this guide could cause issues accessing your application.`

You cannot disable the "local login" right now in macmon. So you are only able to missconfigure the SSO part and still be able to login with the local admin.

Is there a "standardized sentence" that we can add here?

nicedevil007:
> mhhh I know that this seems to be the case for most applications but not for macmon.
> 
> `This documentation lists only the settings that you need to change from their default values. Be aware that any changes other than those explicitly mentioned in this guide could cause issues accessing your application.`
> 
> You cannot disable the "local login" right now in macmon. So you are only able to missconfigure the SSO part and still be able to login with the local admin.
> 
> Is there a "standardized sentence" that we can add here?

screenshot here, upper part will always be visible.

image

this is still the case for the most recent version 6.5

dewi-tik:
> mhhh I know that this seems to be the case for most applications but not for macmon.
> 
> `This documentation lists only the settings that you need to change from their default values. Be aware that any changes other than those explicitly mentioned in this guide could cause issues accessing your application.`
> 
> You cannot disable the "local login" right now in macmon. So you are only able to missconfigure the SSO part and still be able to login with the local admin.
> 
> Is there a "standardized sentence" that we can add here?

This is our standardized sentence. The "cause issues accessing your application" means via authentik. We use the same info box on all guides.

For the time being we can go with this version, but we can look to potentially change it on all guides with a different PR.

dewi-tik (approved):
Thanks again!

tanberry (approved):
Thanks @nicedevil007 !

PR #17903, opened 2025-11-02
web: Lit Session Context

GirlBossRush:
## Details

This PR cleans up much of the client-side session fetching, allowing for fewer re-renders and a smoother recovery from stale sessions, i.e. less UI jank when navigating to a page that your user is no longer authorized to view.

Previously, this behavior was handled with a combination of `me()` calls and sentinel promises to ensure the page is ready. The core fix in this PR is a Lit context.

`SessionContext` handles most scenarios where you'd reach for `me()`, managing the loading, error, and what could be considered a _model view_ of the session. 

For page-level components the session state is available to determine whether to continue showing a loading spinner, or redirecting the user to a more appropriate route. "Smaller but still considered singleton-ish" elements can consume the session information at a higher-level via `this.currentUser`, as well as common variation properties like impersonation, whether the user is no longer authenticated, and the user's UI config.

### Fixes

- Authenticated API endpoints such as version, license, and notification information now wait until a session has been established, avoiding partial page rendering on slow connections and when credentials are stale.
- Several components no longer depend on `rootInterface` to reach up and through the DOM for session and UI config information. This function has slowly been replaced with more idiomatic Lit patterns and will likely be vanquished soon.
- Sentry's initialization is no longer dependent on calls to `me()`, detangling some odd code paths when session expires while debugging. 
- When impersonating a user without admin privileges, the admin UI no longer renders in a broken state, instead redirecting the user from a loading spinner to their library page.
- When fetching applications, the application library now parses and displays API errors, similar to our tabs and table components.

### DIFF
--- a/web/src/admin/AdminInterface/index.entrypoint.ts
+++ b/web/src/admin/AdminInterface/index.entrypoint.ts
@@ -15,13 +15,14 @@ import {
     renderSidebarItems,
 } from "./AdminSidebar.js";
 
+import { isAPIResultReady } from "#common/api/responses";
 import { EVENT_API_DRAWER_TOGGLE, EVENT_NOTIFICATION_DRAWER_TOGGLE } from "#common/constants";
 import { configureSentry } from "#common/sentry/index";
-import { me } from "#common/users";
 import { WebsocketClient } from "#common/ws";
 
 import { AuthenticatedInterface } from "#elements/AuthenticatedInterface";
 import { WithCapabilitiesConfig } from "#elements/mixins/capabilities";
+import { canAccessAdmin, WithSession } from "#elements/mixins/session";
 import { getURLParam, updateURLParams } from "#elements/router/RouteMatch";
 
 import { PageNavMenuToggle } from "#components/ak-page-navbar";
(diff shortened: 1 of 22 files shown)


web/src/admin/AdminInterface/index.entrypoint.ts
```diff
@@ -135,22 +133,21 @@ export class AdminInterface extends WithCapabilitiesConfig(AuthenticatedInterfac
...
+            if (isAPIResultReady(this.session) && !canAccessAdmin(this.session.user)) {
                 window.location.assign("/if/user/");
             }
-        });
+        }
     }
 
     render(): TemplateResult {
+        if (!isAPIResultReady(this.session) || !canAccessAdmin(this.session.user)) {
+            return html`<slot></slot>`;
```
GirlBossRush, line 148: e.g. continue showing the loading spinner

web/src/elements/controllers/SessionContextController.ts
```diff
@@ -0,0 +1,89 @@
...
+
+    constructor(
+        host: ReactiveElementHost<SessionMixin & AKConfigMixin>,
+        initialValue?: APIResult<SessionUser>,
+    ) {
+        this.#host = host;
+
+        this.#context = new ContextProvider(this.#host, {
+            context: SessionContext,
+            initialValue: initialValue ?? { loading: true, error: null },
```
GirlBossRush, line 37: @BeryJu 

In a future revision we can inject the `/core/users/me/` payload into the Django template like the authentik config and skip a render cycle 😎

kensternberg-authentik: I have a philosophy that our product should be able to standalone and served independently of the Django application, rather than be married to it.  So let's think instead that the entire "context of this session" (not just `me()`, but all of the other stuff) has a data shape (a spec in JSON) and that there are multiple ways of getting it: look in the header and if it's not there, ask the server for it, but as _separate and independent_ ways of retrieving the data and handing it to to the respective contexts.

Then we could make decisions about the priority of that information, so that we could (for example) store light/dark mode and locale in `localStorage`, so we'd have those preferences even before the whole page was parsed.  Using a little PWA magic (I know, I know, Jens shudders when he hears the phrase "service worker"), we could make the App much smoother.

kensternberg-authentik (approved):
Well, that was a lot.  But is very good.  You commit, ya.

PR #17913, opened 2025-11-03
web/i18n: Locale message fixes

GirlBossRush:
## Details

This PR includes a few small fixes to make translations work consistently:

- Pre-defined constants such as sidebar entries are lazy rendered, deferring their value to the current locale, rather than the locale at initial page loading.
- Captchas now respect the page locale, falling back to auto detection if no supported language is provided


kensternberg-authentik (approved):
Also straightforward.  Gettin' a lot of mileage out of `ifPresent`, which shows how useful it is. Good call.

PR #16185, opened 2025-08-14
brands: add more matching tests

rissson:

PR #17920, opened 2025-11-03
 brands: sort matched brand by match length

rissson:
Second take on #15413, this time with more tests

Needs #16185


authentik/brands/tests.py
```diff
@@ -27,9 +35,9 @@ def setUp(self):
             if _flag.visibility == "public":
                 self.default_flags[_flag.key] = _flag.get()
 
+    @clear_brands
```
BeryJu: wouldn't it be enough to delete all brands in setUp()?

rissson: It might be? Let's try it out

rissson: works locally, let's see in CI

BeryJu (approved):
postgres go brr

PR #17921, opened 2025-11-03
crypto: update certificate api and component

PeshekDotDev:
## Details

This PR improves how we use the crypto api for certificates, including adding for filtering on the front-end and better typing throughout

---

## Checklist

If applicable

### DIFF
--- a/web/src/admin/common/ak-crypto-certificate-search.ts
+++ b/web/src/admin/common/ak-crypto-certificate-search.ts
@@ -68,8 +68,13 @@ export class AkCryptoCertificateSearch extends CustomListenerElement(AKElement) 
     public singleton = false;
 
     /**
-     * Set to `true` to include certificate details (fingerprints, expiry, private key type, etc.)
-     * in the API response. Defaults to `false` for performance.
+     * Set to `true` to include certificate details (fingerprints, expiry, certificate subject, key type)
+     * in the API response.
+     * Each returned certificate's PEM data must be parsed using cryptography library,
+     * public keys extracted, and hashes computed. With large result sets, this can add a lot of time
+     * to responses.
+     * Only enable when you actually need the detailed fields displayed in the UI.
+     * For simple certificate selection dropdowns, leave this as `false` (default).
      * @attr
      */
     @property({ type: Boolean, attribute: "include-details" })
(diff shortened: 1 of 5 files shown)


web/src/admin/common/ak-crypto-certificate-search.ts
```diff
@@ -66,6 +67,24 @@ export class AkCryptoCertificateSearch extends CustomListenerElement(AKElement)
...
+    /**
+     * Set to `true` to include certificate details (fingerprints, expiry, private key type, etc.)
+     * in the API response. Defaults to `false` for performance.
+     * @attr
+     */
+    @property({ type: Boolean, attribute: "include-details" })
+    public includeDetails = false;
+
+    /**
+     * Optional array of allowed key algorithm types to filter certificates.
```
BeryJu: Add comment about performance implications

BeryJu (approved):
LGTM with one small nit

web/src/admin/common/ak-crypto-certificate-search.ts
PeshekDotDev: I've updated them, let me know what you think

BeryJu (approved):

web/src/admin/common/ak-crypto-certificate-search.ts
kensternberg-authentik: I know my grammar queen shouldn't get too sharp on comments, but that's... awkward.  Maybe something like?

> When `allowedKeyTypes` is set, only certificates or keypairs with matching key algorithms will be shown. Since certificates must be parsed to extract algorithm details, an instance with many certificates may experience long delays and server performance slowdowns.  Avoid setting this field whenever possible.

... which raises the question, @BeryJu: Why isn't that information stored separately from the certificate?

kensternberg-authentik (approved):
I made a suggestion for rewording a comment; that shouldn't delay merging, but if @PeshekDotDev wants to update the comment before the merge, I'd welcome it.

PeshekDotDev:
Storing the keytype in the database is something I'm willing to add in a different PR

web/src/admin/common/ak-crypto-certificate-search.ts
BeryJu: that's a good idea, we could cache the key/cert data alongside the cert instead of re-parsing it every time

maybe something to look into rather than just having this warning

PR #17992, opened 2025-11-06
website/docs: Update Homarr instructions to v1.43.1.

holocronweaver:
## Details

A few corrections to Homarr integration guide to comply with recent Homarr releases:
- adding quotes around AUTH_PROVIDERS values leads to parsing errors in Homarr.
- `AUTH_OIDC_ENABLE_DANGEROUS_ACCOUNT_LINKING` is the new name for `OAUTH_ALLOW_DANGEROUS_EMAIL_ACCOUNT_LINKING`, and probably does not need to be enabled for most installs (hence the DANGEROUS bit) as it is only used for migrating Homarr accounts between OIDC providers.


dominic-r (approved):
thanks!

PR #17994, opened 2025-11-06
blueprints: new meta clean model

anpavlov:
## Details

This PR adds a new meta blueprint model — MetaCleanBlueprint.
This new meta-blueprint removes all objects of a given model name.
For example, run it with `model_name: authentik_providers_oauth2.oauth2provider` - all oauth2 providers will be deleted.
Or with `model_name: authentik_core.application` - all applications will be removed from the database.

**Motivation for this addition:**
I am developing a solution to dynamically set up applications and providers in a docker environment based on container discovery.
In particular, I want to automatically create a new provider and application when a container with specific labels spins up, and remove the corresponding provider and application when that container is destroyed.
To achieve this, I generate blueprint file with applications and providers based on the currently running containers.
To avoid figuring out which providers need to be deleted, I looked for a way to clean all existing providers and applications so they can be recreated from scratch according to the actual running containers.

---

## Checklist

If applicable

### DIFF
--- a/authentik/blueprints/apps.py
+++ b/authentik/blueprints/apps.py
@@ -132,6 +132,7 @@ class AuthentikBlueprintsConfig(ManagedAppConfig):
     def import_models(self):
         super().import_models()
         self.import_module("authentik.blueprints.v1.meta.apply_blueprint")
+        self.import_module("authentik.blueprints.v1.meta.clean_blueprint")
 
     @ManagedAppConfig.reconcile_global
     def tasks_middlewares(self):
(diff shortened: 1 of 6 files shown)


BeryJu (commented):
Haven't reviewed the code yet, but I think there should be the require option to give a filter for what objects to delete and a very explicit flag to delete everything to prevent people from doing this unintentionally

dewi-tik:
Hi @anpavlov are you interested in continuing with this PR?

PR #18013, opened 2025-11-06
crypto: Store details parsed from includeDetails in database instead

PeshekDotDev:
## Details

Stores the key_type of certificates in the database to prevent expensive parsing of certificates

Also, including details was set to true by default it seems, so that has been changed to be false by default

---

## Checklist

If applicable


PeshekDotDev:
For the certificates page, it looks like we have to parse every certificate and send down all the details like fingerprint, expiry, subject, etc

Do we still want to parse all this information for this page? or should I update this PR to pull all the details that would be a part of includeDetails and store them in the database as well?

@BeryJu @rissson

authentik/crypto/models.py
```diff
@@ -121,19 +128,38 @@ def kid(self):
...
+                return KeyType.DSA
+            if isinstance(public_key, Ed25519PublicKey):
+                return KeyType.ED25519
+            if isinstance(public_key, Ed448PublicKey):
+                return KeyType.ED448
+        except (ValueError, TypeError, AttributeError) as exc:
+            LOGGER.warning("Failed to detect key type", exc=exc)
         return None
 
+    def save(self, *args, **kwargs):
```
BeryJu: should be done in a signal

authentik/crypto/models.py
```diff
@@ -55,6 +55,13 @@ class CertificateKeyPair(SerializerModel, ManagedModel, CreatedUpdatedModel):
         blank=True,
         default="",
     )
+    key_type_stored = models.CharField(
```
BeryJu:
```suggestion
    key_type = models.CharField(
```

I'd rather change the name here to remove the property

BeryJu (commented):
I think we might as well save other details about the cert in the database, like subject, expiry, etc, then we can just get rid of the `include_details`

authentik/crypto/models.py
PeshekDotDev: moved to a signal

authentik/crypto/models.py
PeshekDotDev: this was fixed

authentik/crypto/apps.py
```diff
@@ -18,6 +18,11 @@ class AuthentikCryptoConfig(ManagedAppConfig):
     verbose_name = "authentik Crypto"
     default = True
 
+    def ready(self) -> None:
+        """Import signals"""
+        super().ready()
+        import authentik.crypto.signals  # noqa: F401
+
```
BeryJu:
```suggestion
```

ManagedAppConfig already does this if .signals exists

authentik/crypto/signals.py
```diff
@@ -0,0 +1,83 @@
...
+    except (ValueError, TypeError, AttributeError):
+        metadata["fingerprint_sha256"] = None
+
+    # Get SHA1 fingerprint
+    try:
+        metadata["fingerprint_sha1"] = hexlify(
+            certificate.fingerprint(hashes.SHA1()), ":"  # nosec
+        ).decode("utf-8")
+    except (ValueError, TypeError, AttributeError):
+        metadata["fingerprint_sha1"] = None
```
BeryJu: I think I'd prefer to keep this in a single try block

authentik/crypto/signals.py
```diff
@@ -0,0 +1,83 @@
...
+
+    return metadata
+
+
+@receiver(pre_save, sender="authentik_crypto.CertificateKeyPair")
+def certificate_key_pair_pre_save(sender: type[Model], instance, **_):  # noqa: ARG001
+    """Automatically populate certificate metadata fields before saving"""
+    from authentik.crypto.models import CertificateKeyPair
```
BeryJu: should be top-level import

authentik/crypto/signals.py
```diff
@@ -0,0 +1,83 @@
...
+
+    if not isinstance(instance, CertificateKeyPair):
```
BeryJu: Not needed when sender is specified

authentik/crypto/tests.py
```diff
@@ -167,21 +167,24 @@ def test_list_has_key_false(self):
...
         self.assertEqual(200, response.status_code)
         body = loads(response.content.decode())
         api_cert = [x for x in body["results"] if x["name"] == cert.name][0]
-        self.assertEqual(api_cert["fingerprint_sha1"], None)
-        self.assertEqual(api_cert["fingerprint_sha256"], None)
+        # All details should now always be included
+        self.assertEqual(api_cert["fingerprint_sha1"], cert.fingerprint_sha1)
+        self.assertEqual(api_cert["fingerprint_sha256"], cert.fingerprint_sha256)
+        self.assertIsNotNone(api_cert["cert_expiry"])
+        self.assertIsNotNone(api_cert["cert_subject"])
```
BeryJu, line 194: Add test for new keypair with cert and without cert to test extraction

authentik/crypto/signals.py
```diff
@@ -0,0 +1,83 @@
...
+    # Only extract metadata if certificate_data is present
+    if not instance.certificate_data:
+        return
+
+    # Try to extract metadata from the certificate
+    # If the certificate data is invalid, skip metadata extraction
+    try:
+        metadata = extract_certificate_metadata(instance.certificate)
+    except (ValueError, TypeError, AttributeError):
+        # Invalid certificate data, skip metadata extraction
```
BeryJu: Log warning

authentik/crypto/signals.py
```diff
@@ -0,0 +1,83 @@
...
+            certificate.fingerprint(hashes.SHA1()), ":"  # nosec
+        ).decode("utf-8")
+    except (ValueError, TypeError, AttributeError):
+        metadata["fingerprint_sha1"] = None
+
+    return metadata
+
+
+@receiver(pre_save, sender="authentik_crypto.CertificateKeyPair")
+def certificate_key_pair_pre_save(sender: type[Model], instance, **_):  # noqa: ARG001
```
BeryJu:
```suggestion
def certificate_key_pair_pre_save(sender: type[CertificateKeyPair], instance: CertificateKeyPair, **_):  # noqa: ARG001
```

authentik/crypto/signals.py
```diff
@@ -0,0 +1,83 @@
...
+from cryptography.hazmat.primitives import hashes
+from cryptography.x509 import Certificate
+from django.db.models import Model
+from django.db.models.signals import pre_save
+from django.dispatch import receiver
+
+from authentik.crypto.models import detect_key_type, fingerprint_sha256
+
+
+def extract_certificate_metadata(certificate: Certificate) -> dict:
```
BeryJu:
```suggestion
def extract_certificate_metadata(certificate: Certificate) -> dict[str, Any | str]:
```

authentik/crypto/models.py
```diff
@@ -99,16 +147,6 @@ def private_key(
...
-    def fingerprint_sha256(self) -> str:
-        """Get SHA256 Fingerprint of certificate_data"""
-        return fingerprint_sha256(self.certificate)
-
-    @property
-    def fingerprint_sha1(self) -> str:
-        """Get SHA1 Fingerprint of certificate_data"""
-        return hexlify(self.certificate.fingerprint(hashes.SHA1()), ":").decode("utf-8")  # nosec
-
     @property
```
BeryJu: move to db too, backfill values with MD5, create new values with SHA256 and urlsafe_b64encode encode

BeryJu (commented):
Couple of nits

authentik/crypto/signals.py
PeshekDotDev: updated

authentik/crypto/signals.py
PeshekDotDev: fixed

authentik/crypto/signals.py
PeshekDotDev: removed

authentik/crypto/signals.py
PeshekDotDev: logged

authentik/crypto/tests.py
PeshekDotDev: done

authentik/crypto/models.py
PeshekDotDev: Handled, let me know what you think

authentik/crypto/signals.py
```diff
@@ -0,0 +1,65 @@
...
+        return
+
+    instance.key_type = metadata["key_type"]
+    instance.cert_expiry = metadata["cert_expiry"]
+    instance.cert_subject = metadata["cert_subject"]
+    instance.fingerprint_sha256 = metadata["fingerprint_sha256"]
+    instance.fingerprint_sha1 = metadata["fingerprint_sha1"]
+
+    # Generate kid only if not already set (preserves backfilled MD5 values)
+    if not instance.kid and instance.key_data:
```
BeryJu: we'll probably need a bit more logic here to account for someone updating the key of a keypair, since that does require updating the kid

PeshekDotDev: Ah alright, let me handle that. Give me about 30 mins

authentik/crypto/models.py
```diff
@@ -47,6 +48,32 @@ def fingerprint_sha256(cert: Certificate) -> str:
...
+    except (ValueError, TypeError, AttributeError) as exc:
+        LOGGER.warning("Failed to detect key type", exc=exc)
+    return None
+
+
+def generate_key_id(key_data: str) -> str:
+    """Generate Key ID using SHA512 + urlsafe_b64encode."""
+    if not key_data:
+        return ""
+    return urlsafe_b64encode(sha512(key_data.encode("utf-8")).digest()).decode("utf-8").rstrip("=")
```
BeryJu, line 74: I dont remember what we discussed but sha256 might be enough for this and not have to be quite as long

authentik/crypto/signals.py
PeshekDotDev: Done with tests added

BeryJu (approved):

PR #18022, opened 2025-11-08
providers/saml: generate issuer url when provider is set on app

PeshekDotDev:
…created

## Details

Currently, the default saml provider Entity ID/issuer is `authentik`. This PR updates the default Entity ID to `https://domain.com/application/saml/<app_slug>/`

---

## Checklist

If applicable


authentik/core/api/applications.py
```diff
@@ -81,6 +82,36 @@ def __init__(self, *args, **kwargs) -> None:
...
+    def _set_saml_issuer(self, instance: Application) -> None:
+        """Set default SAML provider issuer"""
+
+        if instance.provider.issuer:
+            return
+
+        request = self.context.get("request")
+
+        if request:
+            instance.provider.issuer = request.build_absolute_uri(
```
BeryJu: Instead of setting the field to the URL I think we should just keep the field empty and if its empty set the issuer to the URL in the SAML processor

PeshekDotDev: Fixed. If you can give me a yay or nay on how I handle some off the fallback stuff that would be cool.

Logout request parser line 71. Some tests don't pass if we don't fallback check the provider and it kind of makes sense as a fallback if someone upgrades mid session.

However, it probably makes more sense to rewrite the tests to have a samlsession created and read from that and remove the line 71 fallback, but I think some people will have to log out and back into their saml provider after upgrade i believe

Option 2 is better here? makes more sense long term I think but let me know which one I should do

authentik/providers/saml/api/providers.py
```diff
@@ -84,6 +85,18 @@ def get_url_download_metadata(self, instance: SAMLProvider) -> str:
...
 
+    def get_url_issuer(self, instance: SAMLProvider) -> str:
+        """Get Issuer/EntityID URL"""
+        if instance.issuer:
+            return instance.issuer
+        if "request" not in self._context:
+            return "authentik"
+        request: HttpRequest = self._context["request"]._request
+        try:
+            return request.build_absolute_uri(f"/application/saml/{instance.application.slug}/")
```
BeryJu: add route and use `reverse()`

authentik/providers/saml/processors/assertion.py
```diff
@@ -136,10 +137,25 @@ def get_attributes(self) -> Element:
...
+    def _get_issuer_value(self) -> str:
+        """Get issuer value, with fallback to generated URL if empty"""
+        # If user has set an override issuer, use it
+        if self.provider.issuer:
+            return self.provider.issuer
+
+        # Otherwise, build off of request if application is linked
+        application = getattr(self.provider, "application", None)
+        if self.http_request and application:
+            return self.http_request.build_absolute_uri(f"/application/saml/{application.slug}/")
```
BeryJu: use `reverse()`

authentik/providers/saml/processors/assertion.py
```diff
@@ -136,10 +137,25 @@ def get_attributes(self) -> Element:
...
         return attribute_statement
 
+    def _get_issuer_value(self) -> str:
+        """Get issuer value, with fallback to generated URL if empty"""
+        # If user has set an override issuer, use it
+        if self.provider.issuer:
+            return self.provider.issuer
+
+        # Otherwise, build off of request if application is linked
+        application = getattr(self.provider, "application", None)
```
BeryJu: this check shouldn't be required as we get here from a PolicyAccessView which checks that this is set

authentik/providers/saml/processors/metadata.py
```diff
@@ -39,6 +39,20 @@ def __init__(self, provider: SAMLProvider, request: HttpRequest):
...
+    def _get_issuer_value(self) -> str:
+        """Get issuer value, with fallback to generated URL if empty"""
+        # If user has set an override issuer, use it
+        if self.provider.issuer:
+            return self.provider.issuer
+
+        # Otherwise, build off of request if application is linked
+        application = getattr(self.provider, "application", None)
+        if self.http_request and application:
+            return self.http_request.build_absolute_uri(f"/application/saml/{application.slug}/")
```
BeryJu: `reverse()`

authentik/providers/saml/models.py
```diff
@@ -77,7 +77,12 @@ class SAMLProvider(Provider):
             "no audience restriction will be added."
         ),
     )
-    issuer = models.TextField(help_text=_("Also known as EntityID"), default="authentik")
+    issuer = models.TextField(
+        blank=True,
+        null=True,
+        default=None,
```
BeryJu:
```suggestion
        blank=True,
        default="",
```

For user-editable objects

authentik/providers/saml/models.py
```diff
@@ -317,6 +322,9 @@ class SAMLSession(SerializerModel, ExpiringModel):
     session_index = models.TextField(help_text=_("SAML SessionIndex for this session"))
     name_id = models.TextField(help_text=_("SAML NameID value for this session"))
     name_id_format = models.TextField(default="", blank=True, help_text=_("SAML NameID format"))
+    issuer = models.TextField(
+        default="", blank=True, help_text=_("SAML Issuer used for this session")
```
BeryJu:
```suggestion
        default=None, null=True, help_text=_("SAML Issuer used for this session")
```

since this is not a user-facing model

web/src/admin/providers/saml/SAMLProviderViewPage.ts
```diff
@@ -389,7 +389,9 @@ export class SAMLProviderViewPage extends AKElement {
                                               class="pf-c-form-control"
                                               readonly
                                               type="text"
-                                              value="${ifDefined(this.provider?.issuer)}"
+                                              value="${ifDefined(
+                                                  this.provider?.urlIssuer ?? undefined,
```
BeryJu:
```suggestion
                                                  this.provider?.urlIssuer
```

authentik/providers/saml/processors/metadata.py
PeshekDotDev: updated

authentik/providers/saml/processors/assertion.py
PeshekDotDev: I removed this check and the http  _request check, I assume that if we are generating an assertion, a request has been sent, so it seems redundant now that i think about it

authentik/providers/saml/processors/assertion.py
PeshekDotDev: done

authentik/providers/saml/api/providers.py
PeshekDotDev: updated

authentik/providers/saml/models.py
```diff
@@ -77,7 +77,11 @@ class SAMLProvider(Provider):
             "no audience restriction will be added."
         ),
     )
-    issuer = models.TextField(help_text=_("Also known as EntityID"), default="authentik")
+    issuer = models.TextField(
+        blank=True,
+        default="",
+        help_text=_("Also known as EntityID"),
```
BeryJu: Add note about why this is blank by default and that it can be used for overrides

PeshekDotDev: updated

BeryJu (approved):

authentik/providers/saml/migrations/0021_remove_samlprovider_issuer_and_more.py
```diff
@@ -10,20 +10,24 @@ class Migration(migrations.Migration):
     ]
 
     operations = [
-        migrations.AddField(
-            model_name="samlsession",
+        migrations.RemoveField(
```
BeryJu: should be migrations.RenameField

BeryJu (approved):
aside from the migrations LGTM

authentik/providers/saml/processors/logout_response_processor.py
```diff
@@ -0,0 +1,207 @@
...
+        self._issue_instant = get_time_string()
+        self._response_id = get_random_id()
+
+    def _get_issuer_value(self) -> str:
+        """Get issuer value from session, with fallback to provider"""
+        if self.issuer:
+            return self.issuer
+        if self.provider.issuer_override:
+            return self.provider.issuer_override
+        return "authentik"
```
BeryJu: `common/saml/constants.py` `DEFAULT_ISSUER`

BeryJu (approved):
1 nit

PR #18029, opened 2025-11-09
website/integrations: Frappe: update instructions

Daniel15:
- Make it clearer that `provider` has to be replaced with the actual provider name.
- Add `ERPNext` into the title for SEO and clarity.


dewi-tik:
Thanks for this @Daniel15. I've made a small change to the PR so that the same note about `provider` is mentioned in the authentik section. Can you confirm that you're happy with this change? Then we'll get this merged.

dominic-r (approved):

Daniel15:
Looks good to me. Thanks!

dewi-tik (approved):
Thanks @Daniel15!

dominic-r (approved):

PR #18031, opened 2025-11-10
web: Fix tab activation, blank provider URLs

GirlBossRush:
## Details

This PR fixes a regression introduced in #17804. The removal of extra re-renders in the tab component still requires that a activation event is dispatched to the parent, such as when viewing a provider page.


kensternberg-authentik (approved):
Much cleaner than before. Thanks.

PR #18041, opened 2025-11-10
website/release notes: fix broken urls

kensternberg-authentik:
website: fix bad escaping of URLs in release notes

## What

Fixes bad escaping of URLs in the release notes that resulted in mangled output.

v2024.6.4 had entries that looked like this:

    ##### `GET` /providers/google_workspace/{#123;id}#125;/

v2025.4.md had entries that looked like this:

    ##### `GET` /policies/unique_password/{#125;#123;policy_uuid}/

A correct version looks like this:

    ##### `GET` /policies/geoip/&#123;policy_uuid&#125;/used_by/

A couple of straightforward search-and-replaces has fixed the issue.

## Notes

Two of the release notes had bad escaping of URLs. I’m not sure how the error was made or got past, but it was obvious when visiting the page.

@Beryju suggested that the bug is due to our using `{...}` to symbolize parameters in a URL while Docusaurus wants to interpret `{...}` as an internal template instruction, resulting in odd behavior. In either case, docusarus interpreted the hashtagged entries as links to unrelated issues in Github (the same two issues, which were “bump version of pylint” and “bump version of sentry”), which could be very confusing.

The inconsistencies between the two releases, and the working releases, suggests that the error was introduced manually.


tanberry (approved):
Thanks you so much @kensternberg-authentik !

dominic-r (approved):

PR #18063, opened 2025-11-11
outpost/proxyv2: revalidate auth if session fails to load

chetan:
## Bug

The forward auth handler in the outpost does not properly handle a session load failure. When the session fails to load, it simply bails out, returning a `200` to the reverse proxy which then allows the request to proceed. Depending on the configuration of the upstream service, this can result in a 401, basic auth prompt, or simply allowing access. With one of my applications, the app set it's own cookies which still granted access despite the invalid authentik session.

Similar issue is mentioned in https://github.com/goauthentik/authentik/issues/2023#issuecomment-2962624284

Setup:
- https://auth.example.com
- https://app.example.com (using fwd auth to outpost [I'm using traefik but should apply to others])

Flow:
1. Visit https://app.example.com -> redir to auth, sign in -> redir to app
2. Visit https://auth.example.com -> sign out
3. Session is properly invalidated at the outpost, e.g.:

```json
{"event":"Logging out","level":"debug","logger":"authentik.outpost.proxyv2","provider":"app.example.com","timestamp":"2025-11-10T17:22:54Z"}
{"event":"deleting session","level":"trace","logger":"authentik.outpost.proxyv2.application","name":"provider-app","path":"/dev/shm/session_6JCRTWRRAVWIPWGLBIWZZPLTYUI27PYAVIEQKH5YSDQOMBNMMIWA","timestamp":"2025-11-10T17:22:54Z"}
```

4. Visit https://app.example.com - warning logged but fwd auth returns `200` - app returns 401

Step 2 can also be a restart of the outpost due to crash, reboot, or redeployment.

## Fix

After session validation falls through, fix the session/state setup so we can redirect to the auth server and complete the flow. User will either be prompted to log in or, if they had already logged back in again, they will be sent back and the new session will be established properly.

### DIFF
--- a/internal/outpost/proxyv2/application/oauth.go
+++ b/internal/outpost/proxyv2/application/oauth.go
@@ -1,6 +1,7 @@
 package application
 
 import (
+	"context"
 	"net/http"
 	"net/url"
 	"strings"
@@ -19,8 +20,41 @@ func (a *Application) handleAuthStart(rw http.ResponseWriter, r *http.Request, f
 	state, err := a.createState(r, rw, fwd)
 	if err != nil {
 		a.log.WithError(err).Warning("failed to create state")
-		rw.WriteHeader(400)
-		return
+		if !strings.HasPrefix(err.Error(), "failed to get session") {
+			rw.WriteHeader(400)
+			return
+		}
+
+		// Client has a cookie but we're unable to load the session from
+		// storage (TMPDIR=/dev/shm). This can happen if the session file
+		// was deleted due to container restart or session invalidation
+		// (e.g., logout on auth server).
+		//
+		// Re-save an empty session and try again.
+
+		session, err := a.sessions.Get(r, a.SessionName())
+		if err != nil && !strings.HasSuffix(err.Error(), "no such file or directory") {
+			a.log.WithError(err).Warning("failed to get session")
+			rw.WriteHeader(400)
+			return
+		}
+		err = a.sessions.Save(r, rw, session)
+		if err != nil {
+			a.log.WithError(err).Warning("failed to save session")
+			rw.WriteHeader(400)
+			return
+		}
+
+		// The registry caches the previous attempt to open the session so it
+		// needs to be cleared in order to get the session in createState().
+		*r = *r.WithContext(context.Background())
+
+		state, err = a.createState(r, rw, fwd)
+		if err != nil {
+			a.log.WithError(err).Warning("failed to create state on retry")
+			rw.WriteHeader(400)
+			return
+		}
 	}
 	http.Redirect(rw, r, a.oauthConfig.AuthCodeURL(state), http.StatusFound)
 }


dominic-r (approved):
LGTM. Thanks!

dominic-r (changes requested):
.

internal/outpost/proxyv2/application/oauth.go
```diff
@@ -19,7 +20,37 @@ func (a *Application) handleAuthStart(rw http.ResponseWriter, r *http.Request, f
...
+		// storage (TMPDIR=/dev/shm). This can happen if the session file
+		// was deleted due to container restart or session invalidation
+		// (e.g., logout on auth server).
+		//
+		// Create a new session/state and continue with redirect flow.
+
+		// The registry caches the previous attempt to open the session so
+		// it needs to be cleared in order to create a new session/state
+		// below.
+		*r = *r.WithContext(context.Background())
```
dominic-r: Creating a new session as you mentioned above involves generating and using a new cookie, but this logic would not work:

Summarizing what's getting modified:

  1. `createState()` fails because session file is missing (container restart cleared /dev/shm or whatnot)
  2. Clear the gorilla/sessions registry cache by replacing the request context
  3. Create a new session with `sessions.New()`
  4. Save the new session with `sessions.Save()`
  5. Retry `createState()`, expecting it to use the newly created session

 This doesn't work because he HTTP request object is read-only for incoming data. When you save a new session: `err = a.sessions.Save(r, rw, newSession)`

The new session cookie is written to `rw (the http.ResponseWriter) `which will be sent to the client in the HTTP response. But the request object `r` still contains the original cookie in its headers.

This means when the code wants to retry `createState()`: `state, err = a.createState(r, rw, fwd)  // Line 49 in this diff`

Inside `createState()`at [oauth_state.go:72](https://github.com/goauthentik/authentik/blob/c37ef6d72840e89406ab86dcbd9b24b45137e2b9/internal/outpost/proxyv2/application/oauth_state.go#L72): `s, err := a.sessions.Get(r, a.SessionName())`

The `Get() ` method reads the session cookie from the request headers, not from the `ResponseWriter`. It finds the old cookie referencing the deleted session file.

chetan: I think you're right. It probably still works fine without steps 3&4, I can give it a try. We don't really need a totally new session ID, we just need to rehydrate the claims from the auth server. At least that's my take from what I understand of the flow. What do you think?

chetan: Retested and 3&4 are both needed; it won't work without resaving the session before calling `createState()` unless that method is rewritten to accept a session object (then we can skip the Save() and just pass a new session).

> Inside createState()at [oauth_state.go:72](https://github.com/goauthentik/authentik/blob/c37ef6d72840e89406ab86dcbd9b24b45137e2b9/internal/outpost/proxyv2/application/oauth_state.go#L72): s, err := a.sessions.Get(r, a.SessionName())
> The Get()  method reads the session cookie from the request headers, not from the ResponseWriter. It finds the old cookie referencing the deleted session file.

It's a bit confusing, but `sessions.New()` is the method which actually reads the cookie from the request and we end up reusing the same `session.ID` from the cookie anyway. This can actually be replaced by a `Get()` and then reset the registry afterwards (so that createState() can ready from it properly). Maybe that's a bit cleaner and more obvious in terms of what's happening?

dewi-tik:
Closes https://github.com/goauthentik/authentik/issues/19740

dominic-r (approved):

chetan:
Rebased/resolved conflict

BeryJu (approved):

dewi-tik:
Thanks for this @chetan !

PR #18068, opened 2025-11-11
events: add reason to the MODEL_UPDATED event for password hash updates

### LINKED ISSUE #17840: Add reason for model_updated of a user after authentik upgrade
After authentik has been upgraded you might be running a different django version which might have more rounds in password hashes, which causes the password hash in the model to be updated.

This is correct behaviour, however in our events system we don't capture the reason for this, so after upgrading you'll just see changed users with no obvious difference.

danidee10:
## Checklist

Initially, I started with a more complex Implementation which detected the password hash change in all cases but it felt a bit hacky as I had to use `Queryset.update` to avoid emitting a signal.

I switched to a simpler approach that uses `model_utils.FieldTracker`. It only detects changes when the audit middleware runs and register the signal handlers i.e It will not detect changes that happen outside the request/response cycle.

_The rationale behind this is that anyone who changes the `iterations/rounds` of the hashers is probably an advanced user that knows that the hashes will be updated the next time django runs `check_password` and they shouldn't need an explicit event log_

closes #17840


scripts/test_docker.sh
```diff
@@ -14,7 +14,7 @@ echo PG_PASS="$(openssl rand -base64 36 | tr -d '\n')" >.env
 echo AUTHENTIK_SECRET_KEY="$(openssl rand -base64 60 | tr -d '\n')" >>.env
 export COMPOSE_PROJECT_NAME="authentik-test-${AUTHENTIK_TAG}"
 
-if [[ -v BUILD ]]; then
+if [ -n "${BUILD:-}" ]; then
```
danidee10, line 17: When trying to test with `docker`. I discovered that this syntax wasn't POSIX compliant and didn't work on macos (with bash 3.2)

melizeche: I can confirm the same with zsh 5.9 (arm64-apple-darwin25.0)

authentik/events/middleware.py
```diff
@@ -202,8 +203,14 @@ def post_save_handler(
             return
         user = self.get_user(request)
 
+        context = {}
+        if isinstance(instance, User) and instance.tracker.has_changed("password"):
+            context["reason"] = PASSWORD_HASH_UPGRADE_REASON
```
rissson: How do we know this is not just the user changing their password?

rissson (commented):
@danidee10 fancy seeing you here! Thank you for the contribution!

authentik/events/middleware.py
danidee10: We don't know as there's no way to distinguish why the hash changed here
But I thought it might be acceptable since there's already an existing `PASSWORD_SET` event that tracks password changes. Thus, anyone looking at the event log can figure out why the hash changed.

I had a different approach [here](https://github.com/goauthentik/authentik/pull/18068/commits/147cd3461bd83d83df5e4ccf17b01481a631beda#diff-b205ba7fbd31915270d068e7eac8d856affde1fb2bac250a0c06ea78e400f50eR395-R407) which shouldn't log the event (when the user changes their password) as it doesn't depend on the User model's `post_save` signal.

I guess it depends on how you answer these questions:
Should we track all password hash changes regardless of how they happened ? or
Should we only track password hash changes because Django changed the `iterations` of the hasher ?

BeryJu:
ideas from chatting with @rissson 
- check if we can override django to know when the password is upgraded
- context manager for authenticate() call to add this to event since that's the only time this should be triggered

danidee10:
> check if we can override django to know when the password is upgraded

That was my Initial approach by overriding `User.check_password` but I didn't like the fact that I had to disable the `post_save` signal. I've been thinking about it again and I think there's a better way to Implement it without disabling the signal entirely.


> context manager for authenticate() call to add this to event since that's the only time this should be triggered
I'm less inclined towards this approach

It has more moving parts (We have to apply the context manager to all existing auth backends and future auth backends)
I also have some open questions about the Implementation:

1. is the plan to get the context manager to attach some extra context so we can reuse the existing code in the audit middleware ? If yes then it's even more moving parts because the middleware runs before `authenticate` is called.
2. Are we sending a new Event from the context manager ? If yes then we still have to find a way to disable the signal that's triggered in `check_password`.

I will go with the `check if we can override django to know when the password is upgraded` approach

danidee10:
@BeryJu I went with Option 1 as stated earlier. I replaced `model_utils.FieldTracker` with a class attribute to track the password hash change. I also added a new test.

BeryJu:
We wanted to look into overwriting the check_password function from django to detect when it's doing a hash upgrade and there using `audit_ignore()` to prevent events from being created, but we haven't gotten around to that yet.

danidee10:
> We wanted to look into overwriting the check_password function from django to detect when it's doing a hash upgrade and there using `audit_ignore()` to prevent events from being created, but we haven't gotten around to that yet.

I wasn't aware of the `audit_ignore` context manager. Now, I've reverted the code to an earlier version (that worked like you described above) and integrated it.

rissson (approved):

rissson:
Thanks for being so responsive, even if we have been anything but. The changes look alright to me, but I'll let the team take another pass to determine if any change is still needed.

Issue #17755, opened 2025-10-28
Firefox for Android hangs/freezes on user library view


Xyaren:
**Describe the bug**
Source: https://discord.com/channels/809154715984199690/1432652378250940516

**Since updating** to `2025.10` I notice reproducable crashing in Firefox on Android.

Firefox Desktop and Chrome Mobile/Desktop work just fine.
I already cleared the app data and even reinstalled firefox - no change.

Sometimes it's just the tab not working, sometimes the whole app needing a restart.

In the profiling you can see high CPU usage in the `set CharacterData.data` function.

**To Reproduce**

1. Login to Authentik (get redirected to User Homepage/Library) or naviage to authentik while already logged in
2. Tab/App becomes unresponsive on this screen: 

<img width="172" height="350" alt="Image" src="https://github.com/user-attachments/assets/7f2633e7-3c3d-46fb-94e3-46780e94afc6" />

Sometimes after a restart it shows the apps and works fine but a reload will trigger the bug again 100% of all tries.

**Screenshots**
I was able to capture a firefox profiling via USB Debugging:
Profiling link: https://share.firefox.dev/4ocomFP

<img width="1908" height="786" alt="Image" src="https://github.com/user-attachments/assets/5c1c453b-5e73-4cc8-a994-837e99a3a6e6" />

**Version and Deployment**
-   authentik version: 2025.10
-   Deployment: docker-compose

dominic-r:
Thanks for opening this issue. We'll investigate and try to open a PR for the next patch release

rissson:
I can't reproduce this on firefox android:
Firefox 144.0.1
Android 15 (oneplus 11 15.0)

melizeche:
I'm not able to reproduce this bug on firefox mobile 144.0.1, Android 16 (Pixel 10 Pro)

Xyaren:
Another Capture: https://share.firefox.dev/43zPCWh (I don't see the difference).

Worth noting is that the acutal login works.
SSO via any of the confiugred apps works without an issue. 
Also (with an admin user) directly going to the admin console also works wihtout issues.
It must be something in the display of the library that breaks firefox.

Anything I can do to provide more information ? 

I have a testuser I can provide, just not happy to post the url publicly.
Feel free to contact me via discord and I can provide url and credentials.

BeryJu:
Also can't reproduce this on Firefox 144.0.2, Android 16, Pixel 8 Pro

BeryJu:
[...]

PR #18103, opened 2025-11-13
web: Disable library `<datalist>` on Firefox.

GirlBossRush:
## Details

This PR is a follow up on #17522, disabling the `<datalist>` element on Firefox. Unfortunately, datalists exhibit inconsistent performance characteristics when combined with web components in Firefox's implementation.

The Datalist was first introduced to help navigate app libraries with many results and serves as a supplemental feature. While we'd love to maintain feature parity across all browsers, Firefox users will have the best experience with this feature disabled until the following implementation issues are resolved:

- [Keystrokes on <datalist> fire keyboard events on associated <input>](https://bugzilla.mozilla.org/show_bug.cgi?id=1360755)
- [HTML Datalist field not working after last update](https://bugzilla.mozilla.org/show_bug.cgi?id=1880457)
- [Inconsistent behaviour of datalist + autocomplete="off" between Desktop and Mobile](https://bugzilla.mozilla.org/show_bug.cgi?id=1293664)
- https://bugzilla.mozilla.org/show_bug.cgi?id=1638433
- [Failing WPT html/semantics/forms/the-datalist-element/datalistoptions.html](https://bugzilla.mozilla.org/show_bug.cgi?id=1886914)

### DIFF
--- a/web/src/elements/utils/useragent.ts
+++ b/web/src/elements/utils/useragent.ts
@@ -1,0 +1,19 @@
+/**
+ * @file User agent utilities.
+ */
+
+/**
+ * Predicate to determine if the current browser is Safari.
+ */
+export function isSafari(): boolean {
+    const ua = navigator.userAgent.toLowerCase();
+    return ua.includes("safari") && !ua.includes("chrome") && !ua.includes("chromium");
+}
+
+/**
+ * Predicate to determine if the current browser is Firefox.
+ */
+export function isFirefox(): boolean {
+    const ua = navigator.userAgent.toLowerCase();
+    return ua.includes("firefox");
+}
--- a/web/src/elements/utils/isSafari.ts
+++ b/web/src/elements/utils/isSafari.ts
@@ -1,4 +1,0 @@
-export const isSafari = () =>
-    navigator.userAgent.includes("Safari") && !navigator.userAgent.includes("Chrome");
-
-export default isSafari;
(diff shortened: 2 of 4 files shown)


GirlBossRush:
Possible fix for #17755

web/src/elements/utils/useragent.ts
```diff
@@ -0,0 +1,19 @@
...
+    const ua = navigator.userAgent.toLowerCase();
+    return ua.includes("safari") && !ua.includes("chrome") && !ua.includes("chromium");
+}
+
+/**
+ * Predicate to determine if the current browser is Firefox.
+ */
+export function isFirefox(): boolean {
+    const ua = navigator.userAgent.toLowerCase();
+    return ua.includes("firefox");
```
dominic-r, line 18: Do we also need this on mobile/iOS. If so, the UA string is this:

> Mozilla/5.0 (iPhone; CPU iPhone OS 8_3 like Mac OS X) AppleWebKit/600.1.4 (KHTML, like Gecko) FxiOS/1.0 Mobile/12F69 Safari/600.1.4

as per https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/User-Agent/Firefox#firefox_for_ios

so matching should be done for `like Gecko` instead

dominic-r: Also, is this issue specific to firefox or all Gecko based browsers like Zen? If so, the function should probably be renamed to isGecko()

BeryJu (approved):

PR #18122, opened 2025-11-13
The Coup™

rissson:
Ready for review:
- core server functionality
- static files (mising etag, push welcomed for that)
- proxy protocol
- subpath
- tls
- tls with client certs, for mtls stage
- source IP and other data (scheme, host, etc.)
- config loading
- logging & observability setup (sentry, etc.)
- metrics
- worker functionality
- allinone functionality
- needs testing and reporting for error handling:
  - to test:
    - what if a long running task (managed by arbiter::Tasks) panics
    - what if a request returns `Result::Err`
    - what if a request panics
    - what if a "short" running task (started with tokio::task::spawn, like websocket connections) panics?
  - how to test:
    - modify some part of the code to make do what you want to test
    - run it
    - document what happens (probably in a comment here)
    - is it what you expect to happen?

Notes to reviewer:
- some stuff in the go code is currently broken, specifically because of config changes, it will get fixed at a later time, ignore it for now
- the python code changes should be reviewed as well, and should be working

Not ready for review:
- the actual forwarding code for core. The code is currently very oriented for our use case, but should be made more generic for use in the proxy outpost, so this will change quite a bit again, but the functionality should be the same
- anything proxy outpost related

How to test:
- install rust (see rust-toolchain.toml for version needed)
- `cargo build` and grab a coffee for the first build
- install watchexec for reloading
- `make run` or `make run-watch`
- `cargo t` to run tests
- `cargo clippy` for linting
- `cargo +nightly fmt` for auto formatting
- `cargo install cargo-deny` && `cargo deny check` for licensing and denied features checks (for instance we ban ring and force aws-lc-rs with FIPS for compliance) (currently fails on not allowed git dependencies, needs the config to be updated)
- `cargo install cargo-machete` && `cargo machete` to check for unused dependencies

TODO:
- etag for static files
- ci (lint, test, build)
- proxy outpost (ez)
- http compression
- fix outposts healthchecks


rissson:
Memory usage findings for the worker:
1 process, 2 threads:
Python: 461MB, Rust: 242MB
2 process, 2 threads:
Python: 685MB, Rust: 466MB
Which is logical, since we have one less process in there

Current annoyances: logs from rust don't work (cause it isn't setup). Double ctrl+c doesn't fast exit the worker (we should probably just reduce the timeout in dev). Also, healthchecks aren't doing anything, nor the worker_status, but metrics work! Also, config is hardcoded. Also, no sentry and things.

rissson:
see https://github.com/goauthentik/authentik/pull/24320

PR #18162, opened 2025-11-14
web: Fix ESBuild hanging process

GirlBossRush:
## Details

REPLACE ME

---

## Checklist

-   [ ] Local tests pass (`ak test authentik/`)
-   [ ] The code has been formatted (`make lint-fix`)

If an API change has been made

-   [ ] The API schema has been updated (`make gen-build`)

If changes to the frontend have been made

-   [ ] The code has been formatted (`make web`)

If applicable

-   [ ] The documentation has been updated
-   [ ] The documentation has been formatted (`make docs`)

PR #18164, opened 2025-11-14
web/i18n: Remove English Locale

GirlBossRush:
## Depends on

- #18162
- #18163

## Details

This PR removes the English and pseudo locale XLIFF files from the repo.

### Why no English?

Since English is our source language and the front-end automatically falls back to English strings, keeping a separate English XLIFF file means maintaining duplicate content. The pseudo locale is also now fully auto-generated, so we don't need to track it in version control.

### What about future support for regional varieties?  

The changes in this PR remove files which do not have an effect on authentik's language support. However, while American English is currently the only supported variety, automatic browser support for localization on dates, times, and other units of measure will continue to work.


kensternberg-authentik (approved):
Small commits make me happy!  Small commits that run the first time without breakage make me very happy!

Issue #16627, opened 2025-09-05
SAML Source POST binding error


bobbyjenrow:
**Describe the bug**
I've been attempting to use a third party IdP to act as a source for my Authentik server via SAML.
Using SAML tracer I've been able to verify successful flow from authentik, to the third party sign-on, but on the final SAML post to authentik, a 405 Error is returned.

**To Reproduce**

1. Set up SAML Source with correct SSO URL, Verification Cert, and Redirect Binding
2. Add the source to a stage in default auth flow
3. Attempt to log into the service via third party IdP
4. After successful authentication, the final SAML POST submission with base64 encoded data fails with a 405 Error

**Expected behavior**
Authentik shouldn't return a 405 Error, since the SAML protocol should expect a POST request.

<img width="862" height="482" alt="Image" src="https://github.com/user-attachments/assets/2f05ba12-4b0a-4abf-b0cb-b612755aa50d" />

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/495cbbeb-4a34-44b7-ac09-d446e23381ed" />

Authentik Logs on the 405 Response:
<img width="522" height="474" alt="Image" src="https://github.com/user-attachments/assets/25ccbcd3-3cd1-40b0-b2fd-4df6aee48613" />

**Logs**
Output of docker-compose logs or kubectl logs respectively

**Version and Deployment (please complete the following information):**

-   authentik version: . 2025.6.3
-   Deployment: docker-compose (behind traefik reverse proxy)

**Additional context**

loffa:
Hello,

I have tried to set this up today as well using Google Workspace as the SAML source. Got to the same step and fail with a 405 error on the last POST request. I have tested this on the `2025.8.1` version.

Trying a few different things led me to discover that Authntik responds with a 405 on all URL:s that should be 404. I tried removing trailing `/` and switching to a non-existing slug. All URL:s return 405 if they are invalid, even for GET.

Edit: Also confirmed that the same issue is left after updating to `2025.8.3`

hagaibarel:
Hi,

We're seeing the same issue with version `2025.8.1`.

Following this guide https://docs.goauthentik.io/users-sources/sources/social-logins/google/workspace/, the final step while trying to test the login leads to a 405 error on the `/source/saml/google/acs/` endpoint.

Inspecting the headers, it seems that the endpoint only allows `GET, HEAD, OPTIONS`, but the request method is `POST`

shsjshentao:
[...]

PR #18176, opened 2025-11-16
sources/saml: move SAML Response signature verification before decryption

ikob:
## Details

When a SAML Response is encrypted, signature verification may fail because it is performed after decryption, when the original signed structure has changed.
This PR moves the verification step for signed responses to before decryption to handle encrypted signed responses correctly.
For backward compatibility, the post-decryption check remains as a fallback.

Hopefully close 405 errors at the step 4 of https://github.com/goauthentik/authentik/issues/16627

---

## Checklist

If applicable


authentik/sources/saml/processors/response.py
```diff
@@ -77,11 +77,36 @@ def parse(self):
...
+                post_error = self._verify_signed("/samlp:Response")
+                if post_error == "":
+                    self.response_signature_verified = True
+                else:
+                    self.response_signature_verified = False
+                    sig_errors.append(resp_error)
+
+        if self._source.verification_kp and self._source.signed_response:
+            if self.response_signature_verified is False:
+                raise InvalidSignature(f"SAML Response signature invalid: {'; '.join(sig_errors)}")
```
PeshekDotDev:
```suggestion
                else:
                    sig_errors.append(post_error)
                    raise InvalidSignature(f"SAML Response signature invalid: {'; '.join(sig_errors)}")           
```

authentik/sources/saml/tests/test_response.py
```diff
@@ -258,3 +258,56 @@ def test_verification_no_signature(self):
...
+            parser.parse()
+
+    def test_signed_encrypted_response(self):
+        """Test signed & encrypted response"""
+        key = load_fixture("fixtures/signature_cert2.pem")
+        kp = CertificateKeyPair.objects.create(
+            name=generate_id(),
+            certificate_data=key,
+        )
+        self.source.verification_kp = kp
```
PeshekDotDev:
```suggestion
        self.source.verification_kp = kp
        self.source.encryption_kp = kp
```

It looks like we need to add the encryption kp's here too

PeshekDotDev (changes requested):
Can you also add a couple comments just saying "This is the fallback verification", etc?

Really appreciate you getting to this @ikob

authentik/sources/saml/processors/response.py
PeshekDotDev: Would this work the same or am I misreading the code?

authentik/sources/saml/tests/test_response.py
ikob: Thank for your review. We need some fix on test_response.py, which require some test fixtures. So, I update my branch.

authentik/sources/saml/processors/response.py
ikob: Thank for your review. I agree with your update. But, we need some fix on test_response.py bellow.
So, I update my branch.

tfc:
Hey @ikob is this here fixing https://github.com/goauthentik/authentik/issues/16627 ? If yes - what can I do to help push this further?

ikob:
> Hey @ikob is this here fixing https://github.com/goauthentik/authentik/issues/16627 ? 

Partially yes as as mentioned...

> Hopefully close 405 errors at the step 4 of https://github.com/goauthentik/authentik/issues/16627

> If yes - what can I do to help push this further?

I am not sure the review/merging priority, but your interests may help including this conversation.

tfc:
@rissson @PeshekDotDev anything we can do to advance this further?

authentik/sources/saml/processors/response.py
```diff
@@ -114,45 +138,34 @@ def _decrypt_response(self):
             decrypted_assertion,
         )
 
-    def _verify_signed(self):
+    def _verify_signed(self, xpath: str) -> str:
```
PeshekDotDev: Returning a string is a change that feels a little strong

Instead of working around the current error raising, can we instead make the logic work like 2 try catch blocks back to back with a boolean?

something like

make clone
decrypt
try verify decrypted
if it didn't work, try verify untouched clone
if continue is false return
rest of logic

ikob: Yes, returning a string is better than the current error raising.

> make clone
> decrypt
> try verify decrypted
> if it didn't work, try verify untouched clone
> if continue is false return
> rest of logic

The above flow will work fine. 
However, I cannot understand why we should keep a clone.

PeshekDotDev: The thought process is:

Decrypt first, because if decryption fails, we know we should skip everything else
By having a clone of the request before and after decryption, we can verify both back to back and keep the logic organized

ikob: I agree with you, because the verification blocks aggregate one part is simple.
I misunderstood the decrypt behavior that removes encrypted node.

ikob: I updated PR to follow as:

> make clone
> decrypt
> try verify decrypted
> if it didn't work, try verify untouched clone
> if continue is false return
> rest of logic

hassidan:
any update about that?

authentik/sources/saml/processors/response.py
PeshekDotDev: Apologies, I might of worded this wrong. Can you, instead of altering the verify_signed and decrypt_response to return a string, have them raise exceptions and catch them instead?

tfc:
Is fixing google login with signed callbacks no priority, or is there anything controversial about this PR?

PeshekDotDev:
Hey everyone, I have made a different version of this PR that is more in line with our engineering style. I appreciate everyone's patience, and if there are any questions — or especially if there are any other saml source/provider related issues you all have found — please let me know. I am committed to getting our SAML experience as nice as possible

Thank you

Alternate PR — #19593

ikob:
@PeshekDotDev,
I made a PR to add test cases based on this PR. Take look https://github.com/goauthentik/authentik/pull/19647

PR #18210, opened 2025-11-18
website: Unlisted & Draft Release Notes

GirlBossRush:
## Details

This PR adds support for Docusaurus's `unlisted` and `draft` front-matter properties to the version selector. 

### How does it work?

Markdown files within Docusaurus can make use the `unlisted` front-matter property to hide an entry from the sidebar:
 
For example to unlist a future version of the release notes:

```md
---
title: Release 2025.11
slug: "/releases/2025.11"
unlisted: true
---

## Highlights

...etc
```

The result is a file which renders typically in development:

Screenshot 2025-11-19 at 16 48 24

---

However in production the page will not be listed unless the URL is visited directly. In the case of a release page, the entry will not be visible in the sidebar. The warning banner is rendered automatically by Docusaurus.

- `draft`
  - ✅ [Visible in sidebar during development](https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-content-pages#draft) 
  - 🚫 Not visible in sidebar when deployed to production
  - 🚫 Not published when deployed to production
- `unlisted`
  - ✅ [Visible in sidebar during development](https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-content-pages#unlisted) 
  - ✅ Published in production, viewable with a direct link
  - 🚫 Not visible in sidebar when deployed to production

### When should I use `draft` instead of `unlisted`?

We don't yet have use-case for `draft`, but the effect on release files is similar to `unlisted`. Our support for `draft` is included to keep our version selector aligned with Docusaurus's Markdown parser.

### DIFF
--- a/website/docusaurus-theme/components/VersionPicker/VersionDropdown.tsx
+++ b/website/docusaurus-theme/components/VersionPicker/VersionDropdown.tsx
@@ -2,9 +2,13 @@ import "./styles.css";
 
 import { createVersionURL, parseBranchSemVer } from "#components/VersionPicker/utils.ts";
 
+import type {
+    AKReleaseFrontMatter,
+    AKReleasesPluginEnvironment,
+} from "@goauthentik/docusaurus-theme/releases/common";
+
 import clsx from "clsx";
 import React, { memo } from "react";
-import { AKReleasesPluginEnvironment } from "releases/node.mjs";
 
 export interface VersionDropdownProps {
     /**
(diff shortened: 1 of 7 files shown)


dominic-r (approved):

BeryJu (approved):
According to https://github.com/facebook/docusaurus/issues/8285 this banner should only show up in dev mode, however there seems to be something that this also shows in prod. I don't think we should have this in prod, if we need a banner like this it makes more sense to have one specific to the page and just add it as an info box

BeryJu (dismissed):
Meant to request changes oops

tanberry (approved):
Thank you @GirlBossRush !

PR #18211, opened 2025-11-18
outpost/proxyv2: more tests, fix pg password with spaces, and existing session on restart

### LINKED ISSUE #18130: Proxy Outpost Basic Auth 2025.10.x Breaks after Outpost Restart
**Describe the bug**
As described in [this other](https://github.com/goauthentik/authentik/issues/17779) (now resolved) bug, basic auth through the proxy provider wasn't working on 2025.10.0. The solution (included in 2025.10.1) does work on the embedded outpost, but I can't seem to get *non*-embedded outposts running 2025.10.1 to send basic auth credentials *after they restart* and the sessions are reset.

I made an issue [here](https://github.com/goauthentik/authentik/issues/18098) about session storage with the outpost proxy that's worth noting. There I learned that currently, because the session storage is in memory & can't be stored elsewhere, when my single outpost pod restarts, I lose all my sessions stored by the outpost.

When this happens, every application that relied on basic auth credential sending from the provider, now prompts me for basic auth.

I can get the basic auth to work again as intended when I fully clear my browser's cache and log back into Authentik. Simply logging out & back into Authentik does not fix it - I need to clear the cookies.

**To Reproduce**

1. Add a proxy outpost (I made mine manually on k8s)
2. Add a proxy provider w/ basic auth credentials
3. Assign the application to your external proxy outpost
4. Log into the application using the outpost
5. Kill the outpost container, clearing out its stored sessions. Wait for new outpost pod to start.
6. Refresh the page of the application

**Expected behavior**
The application shouldn't prompt me for basic auth after an outpost restart

**Logs**
Startup logs are normal after restart

When refreshing the application page after an outpost restart, logs like this one occur.
```bash
{"error":"failed to get session: open /dev/shm/session_TH4LDU4USSDPDED2TEYUG6OV5BHLG3TFBRPB45M2WNFTIQAVI3XQ: no such file or directory","event":"failed to create state","level":"warning","logger":"authentik.outpost.proxyv2.application","name":"Provider for Radarr","timestamp":"2025-11-13T20:10:24Z"}
```
I'd expect that for the first request, but it doesn't seem to be recovering after, it just logs the same thing for every request I send.

-   authentik version: 2025.10.1
-   Deployment: helm

dominic-r:
And modify / add some more tests and a bit of refactoring

Closes: https://github.com/goauthentik/authentik/issues/18175

## Details

REPLACE ME

---

## Checklist

If applicable

### DIFF
--- a/internal/outpost/proxyv2/postgresstore/postgresstore.go
+++ b/internal/outpost/proxyv2/postgresstore/postgresstore.go
@@ -2,16 +2,20 @@ package postgresstore
 
 import (
 	"context"
+	"crypto/tls"
+	"crypto/x509"
 	"encoding/json"
 	"errors"
 	"fmt"
 	"net/http"
+	"os"
 	"strings"
 	"time"
 
 	"github.com/google/uuid"
 	"github.com/gorilla/sessions"
-	_ "github.com/jackc/pgx/v5/stdlib"
+	"github.com/jackc/pgx/v5"
+	"github.com/jackc/pgx/v5/stdlib"
 	"github.com/mitchellh/mapstructure"
 	log "github.com/sirupsen/logrus"
 	_ "gorm.io/driver/postgres"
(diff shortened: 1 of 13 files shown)


internal/outpost/proxyv2/postgresstore/postgresstore.go
```diff
@@ -49,60 +53,121 @@ func (ProxySession) TableName() string {
...
+					return nil, fmt.Errorf("failed to load SSL client certificate: %w", err)
+				}
+				tlsConfig.Certificates = []tls.Certificate{cert}
+			}
+
+			// Set verification mode
+			switch cfg.SSLMode {
+			case "require":
+				// Verify the server certificate (secure by default)
+				tlsConfig.InsecureSkipVerify = false
```
JohnTheNerd: I think this change, as-is, is a breaking change that does not conform to the Postgres or Authentik documentation. If SSL mode is set to require, the documented behavior is to _not_ check certificate validity:

- https://www.postgresql.org/docs/current/libpq-ssl.html

- https://docs.goauthentik.io/install-config/configuration/ (currently not viewable due to the ongoing CloudFlare outage)

This change instead makes "require" equivalent to "verify-ca". If this is intended to be the case, it should be documented as a breaking change.

dominic-r: Yea, I need to change that, Good catch.

dominic-r: I've reverted the CodeQL alert

BeryJu (approved):

PR #18230, opened 2025-11-18
providers/scim: fix PATCH for AWS

### LINKED ISSUE #12321: AWS SCIM Provisioning Failure
**Describe the bug**
AWS is returning 400s during the SCIM sync. It looks like authentik is trying to do a PATCH request that AWS doesn't like.

{"domain_url": null, "event": "Failed to send SCIM request", "level": "warning", "logger": "authentik.lib.sync.outgoing.base", "method": "PATCH", "path": "/Groups/__uuid__", "provider": "AWS SCIM", "response": "{\"schema\":[\"urn:ietf:params:scim:api:messages:2.0:Error\"],\"schemas\":[\"urn:ietf:params:scim:api:messages:2.0:Error\"],\"detail\":\"Current PATCH support allows omitting \\\"path\\\" attribute only for Add and Replace operations modifying a single attribute.\",\"status\":\"400\",\"exceptionRequestId\":\"__uuid__\",\"timeStamp\":\"2024-12-10 19:50:00.531\"}", "schema_name": "public", "task_id": "task-ID", "timestamp": "2024-12-10T19:50:00.539728"}

-   authentik version: 2024.10.4

bitpavel-l25:
## Details
Resolves https://github.com/goauthentik/authentik/issues/12321
According to the [AWS SCIM reference](https://docs.aws.amazon.com/pdfs/singlesignon/latest/developerguide/sso-scim.pdf), `PatchGroup` operation must contain `path` field with supported attributes: displayName, members, and externalId.

The current PR enforces this behaviour based on SCIMCompatibilityMode.

---

## Checklist

If applicable

### DIFF
--- a/authentik/providers/scim/clients/groups.py
+++ b/authentik/providers/scim/clients/groups.py
@@ -129,27 +129,36 @@ class SCIMGroupClient(SCIMClient[Group, SCIMProviderGroup, SCIMGroupSchema]):
     def _update_patch(
         self, group: Group, scim_group: SCIMGroupSchema, connection: SCIMProviderGroup
     ):
-        """Apply provider-specific parameters for PATCH request"""
-        if connection.provider.compatibility_mode == SCIMCompatibilityMode.AWS:
-            for attr in ("displayName", "externalId"):
-                self._update_patch_call(group, scim_group, connection, attr)
-        else:
-            self._update_patch_call(group, scim_group, connection)
+        """Apply provider-specific PATCH requests"""
+        match connection.provider.compatibility_mode:
+            case SCIMCompatibilityMode.AWS:
+                self._update_patch_aws(group, scim_group, connection)
+            case _:
+                self._update_patch_general(group, scim_group, connection)
         return self.patch_compare_users(group)
 
-    def _update_patch_call(
-        self,
-        group: Group,
-        scim_group: SCIMGroupSchema,
-        connection: SCIMProviderGroup,
-        attr: str = None,
+    def _update_patch_aws(
+        self, group: Group, scim_group: SCIMGroupSchema, connection: SCIMProviderGroup
+    ):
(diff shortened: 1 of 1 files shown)


authentik/providers/scim/clients/groups.py
```diff
@@ -127,18 +128,37 @@ def update(self, group: Group, connection: SCIMProviderGroup):
...
+        else:
+            self._update_patch_call(group, scim_group, connection)
+        return self.patch_compare_users(group)
+
+    def _update_patch_call(
+        self,
+        group: Group,
+        scim_group: SCIMGroupSchema,
+        connection: SCIMProviderGroup,
+        attr: str = None,
```
BeryJu: insteaed of `attr` I think it'd be cleaner to just pass *ops and use that as operations and then the call above can patch its specific attributes in one call

bitpavel-l25: @BeryJu I've refactored the PR to aggregate patch updates into one batch. Should be more efficient (and the code is cleaner now).

authentik/providers/scim/clients/groups.py
```diff
@@ -127,6 +128,34 @@ def update(self, group: Group, connection: SCIMProviderGroup):
 
     def _update_patch(
         self, group: Group, scim_group: SCIMGroupSchema, connection: SCIMProviderGroup
+    ):
+        """Apply provider-specific PATCH requests"""
+        match connection.provider.compatibility_mode:
+            case SCIMCompatibilityMode.AWS:
+                self._update_patch_aws(group, scim_group, connection)
+            case _:
+                self._update_patch_general(group, scim_group, connection)
```
BeryJu, line 137: small nit, I think an if/else would be more readable for this but good either way

BeryJu (approved):

PR #18273, opened 2025-11-19
website: Auto-categories.

GirlBossRush:
## Details

Replaces the hand-maintained `sidebar.mjs` (~1000 lines) with an autogenerated one, following the same pattern the integrations site already uses. Top-level order and labels live in `categories.mjs`; everything under each section is generated from the folder tree.

Ordering and labels the static sidebar encoded are kept as `sidebar_position`/`sidebar_label` frontmatter, plus `_category_.json` for folders with no index doc. Workflow-ordered sections (Providers, Add and Secure Apps, Users and Sources) keep their sequence; flat topic lists stay alphabetical.

Docs the old sidebar had orphaned (`sys-mgmt/tenancy`, `providers/radius`) now appear automatically.


dominic-r (approved):
!!!

PR #18274, opened 2025-11-19
core: remove filter_not_expired for QS

BeryJu:
in wake of the recent CVE due to a missing `filter_not_expired`, I think it's a better DX to filter out expired objects by default and make getting expired objects an opt-in. This PR does that.

- Adds `.objects.including_expired()`, this will return a queryset with all objects
- Default `.objects` will exclude expired objects by default

Models to check:
- [x] `<class 'authentik.core.models.Token'>`
- [x] `<class 'authentik.core.models.Session'>`
- [x] `<class 'authentik.enterprise.models.LicenseUsage'>`
- [x] `<class 'authentik.events.models.Event'>`
- [x] `<class 'authentik.stages.invitation.models.Invitation'>`
- [x] `<class 'authentik.providers.oauth2.models.AuthorizationCode'>`
- [x] `<class 'authentik.providers.oauth2.models.AccessToken'>`
- [x] `<class 'authentik.providers.oauth2.models.RefreshToken'>`
- [x] `<class 'authentik.providers.oauth2.models.DeviceToken'>`
- [x] `<class 'authentik.enterprise.providers.ssf.models.StreamEvent'>`
- [x] `<class 'authentik.policies.reputation.models.Reputation'>`
- [x] `<class 'authentik.providers.proxy.models.ProxySession'>`
- [x] `<class 'authentik.providers.rac.models.ConnectionToken'>`
- [x] `<class 'authentik.providers.saml.models.SAMLSession'>`
- [x] `<class 'authentik.stages.consent.models.UserConsent'>`


authentik/core/models.py
```diff
@@ -1125,9 +1138,32 @@ def expire_action(self, *args, **kwargs):
...
+            event.save()
+
+        for obj in (
+            cls.objects.including_expired()
+            .filter(**kwargs)
+            .filter(Q(expires__lt=now(), expiring=True))
+        ):
             obj.delete()
-        return cls.objects.filter(**kwargs)
+        return cls.objects.including_expired().filter(**kwargs)
```
rissson: Should this call `including_expired`?

BeryJu: d'oh it should not


```suggestion
        return cls.objects.filter(**kwargs)
```

rissson (approved):

PR #18291, opened 2025-11-20
web/admin: add entitlement search

### LINKED ISSUE #18314: Application entitlement search field

.

A search field for application entitlements. This would make it easier to locate entitlements for applications that use a lot of them.

Alternative is using CTRL + F to find the correct entitlements but since not that many are displayed per page this is not the best solution.

### Additional context

_No response_

BeryJu:

closes #18314

### DIFF
--- a/web/src/admin/applications/entitlements/ApplicationEntitlementPage.ts
+++ b/web/src/admin/applications/entitlements/ApplicationEntitlementPage.ts
@@ -37,7 +37,7 @@ export class ApplicationEntitlementsPage extends Table<ApplicationEntitlement> {
 
     order = "order";
 
-    protected searchEnabled: boolean = true;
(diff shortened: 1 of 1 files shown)


web/src/admin/applications/entitlements/ApplicationEntitlementPage.ts
```diff
@@ -37,6 +37,8 @@ export class ApplicationEntitlementsPage extends Table<ApplicationEntitlement> {
 
     order = "order";
 
+    protected searchEnabled: boolean = true;
+
```
GirlBossRush:
```suggestion
    protected override searchEnabled = true;

```

GirlBossRush (approved):

PR #18293, opened 2025-11-20
flows: return correct status code on error

### LINKED ISSUE #18294: User login fail return 4xx HTTP code
### Discussed in https://github.com/goauthentik/authentik/discussions/17996

<div type='discussions-op-text'>

<sup>Originally posted by **aktk946** November  6, 2025</sup>
hey all, is there a way for authentik to send back 4xx HTTP code back to reverse proxy when user fails authentication?
Currently it sends back 200 which flies under the radar of my nginx proxy manager + fail2ban and repeated failures do not trigger a ban.
I would like to be able to ban the IPs failing authentication repeatedly.</div>

BeryJu:
currently we return status 200 for everything; errors are structured the same as in any other API endpoint

closes #18294


rissson:
Does this require special handling in the frontend/outposts?

BeryJu:
probably, haven't tested yet

PR #18129, opened 2025-11-13
crypto: update certificates on fs event

PeshekDotDev:
## Details

Closes #18009

---

## Checklist

-   [ ] Local tests pass (`ak test authentik/`)
-   [ ] The code has been formatted (`make lint-fix`)

If an API change has been made

-   [ ] The API schema has been updated (`make gen-build`)

If changes to the frontend have been made

-   [ ] The code has been formatted (`make web`)

If applicable

-   [ ] The documentation has been updated
-   [ ] The documentation has been formatted (`make docs`)

Issue #7242, opened 2023-10-20
Renewal of expired certificate


depuits:
**Describe your question/**
The default authentik Self-signed Certificate has expired and the one I've created to use with ldap will expire soon. The [documentation](https://goauthentik.io/docs/core/certificates) doesn't mention what the correct way to renew this is. Should these be deleted and then be replaced by newly generated ones or is there a way to renew these?

**Relevant infos**
i.e. Version of other software you're using, specifics of your setup

**Screenshots**
![image](https://github.com/goauthentik/authentik/assets/3755674/27cb08fc-235a-4d0a-a46a-a9ecdbfd3d97)

**Logs**
Output of docker-compose logs or kubectl logs respectively

**Version and Deployment (please complete the following information):**

-   authentik version: 2023.8.3
-   Deployment: docker-compose

**Additional context**

fansari:
We have the same issue with our letsenrypt certificate. How can a renewal done automatically? Delete, recreate and the reassociate to all providers (we have quite a lot) cannot be the idea.

bdomars:
Just hit the same snag with our SAML providers, quite a lot of work to go trhoug everything and update a certificate. Seems like there should be a better way to do this automatically.

angelnu:
This also affects [external certificates ](https://docs.goauthentik.io/docs/core/certificates).

The certificate in my Kubernetes cluster was correctly updated by cert-manager and the new certificate files were mounted to authentik in the /certs folder. Authentik was still using the old (expired) certificate.

After I deleted the certificate in the authentik UI and restarted authentik, the certificate was re-imported again. So it seems that authentik does not check if the certificate has changed in the filesystem - it keeps using the old values until manually deleted.

Toakan:
So the biggest problem here is that whilst you can overwrite the local /certs/domainname/ with the newer chains, the Import isn't setup to overwrite the current data in DB.

IE Below is a failure reported by the certificate_discovery task. It looks like the job should be changed to a IF EXISTS check before INSERT, otherwise UPDATE.

`
[...]

PR #18295, opened 2025-11-20
website/docs: update certificate doc

dewi-tik:
## Details

Updates the certificate doc based on the changes from Connor's PR #18129

Addresses some comments from https://github.com/goauthentik/authentik/issues/7242#issuecomment-3508677182

---

## Checklist

If applicable


website/docs/sys-mgmt/certificates.md
```diff
@@ -20,6 +20,22 @@ While this certificate can be used for SAML providers/sources, remember that it'
 
 For SAML use-cases, you can generate a certificate with a longer validity period (at your own risk).
 
+## Certificate considerations
+
+### OAuth and SAML
+
+For OAuth and SAML providers, in the vast majority of cases, certificate expiry does not matter. Most service providers don't check whether certificate are expired. What usually matters is that the signature is correct.
+
+However, there are some notable exceptions to this. For example, the Slack SAML integration checks for certificate expiry.
```
dominic-r: […] We recommend checking your service provider's documentation for bla bla bla

website/docs/sys-mgmt/certificates.md
```diff
@@ -20,6 +20,22 @@ While this certificate can be used for SAML providers/sources, remember that it'
...
+
+### Proxy provider and brands
+
+We recommend using a certificate generated outside of authentik so that it can correlate with your FQDN, preferably, generated by a publicly trusted certificate authority such as Let's Encrypt.
```
dominic-r: ~~such as Let's encrypt~~

dominic-r: > so that it

not sure about this wording

website/docs/sys-mgmt/certificates.md
```diff
@@ -20,6 +20,22 @@ While this certificate can be used for SAML providers/sources, remember that it'
...
+
+### Radius EAP-TLS
+
+We recommend using a certificate generated outside of authentik, however an internally generated cert which isn't publicly trusted is sufficient.
```
dominic-r: do we have another section in our docs or this page which covers internally generated certs? we could link it

website/docs/sys-mgmt/certificates.md
```diff
@@ -49,7 +65,9 @@ Certificate discovery can be manually initiated by restarting the `certificate_d
 - **Docker Compose**: A `certs` directory is mapped to `/certs` within the worker container.
 - **Kubernetes**: You can mount custom Secrets or Volumes under `/certs` and configure them in the worker Pod specification.
 
-authentik checks for new or changed files every hour and automatically triggers an outpost refresh when changes are detected.
+When a new keypair is added or changed, authentik automatically triggers an outpost refresh.
+
+When a new keypair is added that has a private key that already exists in the database, authentik will automatically update the keypair’s certificate in the database rather than creating a new keypair.
```
dominic-r:
```suggestion
When a new keypair is added that has a private key that already exists in the database, authentik will automatically update the keypair's certificate in the database rather than creating a new keypair.
```

dominic-r (dismissed):
.

website/docs/sys-mgmt/certificates.md
dewi-tik: Not AFAIK. The only certs doc we have is this one. Not sure why we'd have a document covering how to create certs with internal CAs

dominic-r: 🤷 i prob misread or "internally generated cert" seems a bit confusing. if you just read that bit it looks like it's an authentik internal cert

dewi-tik: Fair point. I've improved the language, what do you think?

website/docs/sys-mgmt/certificates.md
```diff
@@ -20,6 +20,24 @@ While this certificate can be used for SAML providers/sources, remember that it'
 
 For SAML use-cases, you can generate a certificate with a longer validity period (at your own risk).
 
+## Certificate considerations
+
+### OAuth and SAML
+
+For OAuth and SAML providers, in the vast majority of cases, certificate expiry does not matter. Most service providers don't check whether certificate are expired. What usually matters is that the signature is correct.
```
dominic-r:
```suggestion
For OAuth and SAML providers, in the vast majority of cases, certificate expiry does not matter. Most service providers don't check whether certificates are expired. What usually matters is that the signature is correct.
```

dominic-r: that the signature is correct -> What usually matters is that the signature is valid.

website/docs/sys-mgmt/certificates.md
```diff
@@ -20,6 +20,24 @@ While this certificate can be used for SAML providers/sources, remember that it'
...
+
+However, there are some notable exceptions to this. For example, the Slack SAML integration checks for certificate expiry.
```
dominic-r: ~~ to this~~

dominic-r: checks for --> does check for

website/docs/sys-mgmt/certificates.md
```diff
@@ -20,6 +20,24 @@ While this certificate can be used for SAML providers/sources, remember that it'
...
+
+We recommend checking your service provider's documentation for further details.
```
dominic-r: further details  --> specific requirements 

or anything you find or think of which is a bit more specific

website/docs/sys-mgmt/certificates.md
```diff
@@ -20,6 +20,24 @@ While this certificate can be used for SAML providers/sources, remember that it'
...
+
+### Proxy provider and brands
+
+We recommend using a certificate generated outside of authentik that correlates with your FQDN, preferably, generated by a publicly trusted certificate authority.
```
dominic-r: "correlates with your FQDN" is the wrong verb; certificates match or cover FQDNs.
comma after "preferably" is incorrect.

probably something along the lines of:

> We recommend using a certificate generated outside of authentik that matches your FQDN, preferably one issued by a publicly trusted certificate authority.

btw do we every spell out the acronym FQDN since the style guide says we should do that at first occurrence

website/docs/sys-mgmt/certificates.md
```diff
@@ -20,6 +20,24 @@ While this certificate can be used for SAML providers/sources, remember that it'
...
+
+### Radius EAP-TLS
+
+We recommend using a certificate generated outside of authentik. A non-publicly trusted certificate is sufficient.
```
dominic-r: non-publicly trusted certificate --> private CA certificate

or "a privately issued certificate"

website/docs/sys-mgmt/certificates.md
```diff
@@ -49,7 +67,9 @@ Certificate discovery can be manually initiated by restarting the `certificate_d
 - **Docker Compose**: A `certs` directory is mapped to `/certs` within the worker container.
 - **Kubernetes**: You can mount custom Secrets or Volumes under `/certs` and configure them in the worker Pod specification.
 
-authentik checks for new or changed files every hour and automatically triggers an outpost refresh when changes are detected.
+When a new keypair is added or changed, authentik automatically triggers an outpost refresh.
```
dominic-r: shouldn't keypairs be 2 words: key pairs 

from what I can see, that's the case
 and the aws overlords agree : https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-key-pairs.html

website/docs/sys-mgmt/certificates.md
```diff
@@ -49,7 +67,9 @@ Certificate discovery can be manually initiated by restarting the `certificate_d
...
+
+When a new keypair is added that has a private key which already exists in the database, authentik will automatically update the keypair's certificate in the database rather than creating a new keypair.
```
dominic-r: same here

dominic-r: 1) a bit confusing i find
2) private key which already exists --> that already exists

if i understand it correctly, this is what's happening: so mabye something along those lines

When a new key pair is added with a private key that already exists in the database, authentik updates the existing key pair'scertificate instead of creating a duplicate one.

dominic-r (dismissed):
just a few more things pls

website/docs/sys-mgmt/certificates.md
tanberry: Ooooff let's not spell out FDQN, tho yeah technically we should. Let'e rebel against that Style Guide.

website/docs/sys-mgmt/certificates.md
tanberry: If this is what we call an "external certificate" elsewhere, let's stick with calling it an external cert.

dominic-r (approved):

PR #18296, opened 2025-11-20
web/a11y: Accessible preference classes.

GirlBossRush:
## Details

Similar to our theme effect listeners, these listeners apply classes to the DOM for a preferred contrast, transparency, and motion.


GirlBossRush:
Closing after meditating on the developer experience. This approach feels more like "a solution in search of a problem"

Issue #12766, opened 2025-01-22
Federation SAML Source Property Mapping


katrinm86:
**Describe your question/**
So I want to hook up my Authentik instance to a SAML IdP source for federated login.

I checked the following documentation pages
- https://docs.goauthentik.io/docs/users-sources/sources/protocols/saml/
- https://docs.goauthentik.io/docs/users-sources/sources/property-mappings/

and tried following the "see documentation for a list of all variables" link in the "SAML Source Property Mapping" which points to a non-existing page "https://version-2024-12.goauthentik.io/docs/user-sources/sources/property-mappings/expressions?utm_source=authentik"

but I'm not really seeing how one is supposed to do the attribute mappings.

Is it really required to write your own lookup code for each property like it is done for "name_id" (https://github.com/goauthentik/authentik/blob/main/authentik/sources/saml/processors/response.py#L209)?

Am I something missing?

Thanks

rissson:
The latest link is https://docs.goauthentik.io/docs/users-sources/sources/property-mappings/expressions. @tanberry let's check we're linking probably from the UI

rissson:
As for the source property mappings code themselves, yes, you need to write whatever you want in there, allowing you to get as much customization as neede

katrinm86:
ok thats cool if it would say so there (less ambiguity for users).

and i would also propose a starting example on how to do that ... i myself know etree a bit but from my point of view its more complicated than the other methods... would be really helpful to have a saml xml response with attributes and then example code to extract the fields 

adding to that: there is a test button with the mappings. how would i go about suppyling a saml response there to test the code? can i just use the context there and say e.g. { root = b'mysamlresponse' }?

katrinm86:
i have an additional question regarding ids. 

the idp(s) we would like to integrate are from a federation and they suggest to find the ids based on the following rules

extract attributes
- urn:oasis:names:tc:SAML:attribute:subject-id -> subject-id
- urn:oasis:names:tc:SAML:attribute:pairwise-id -> pairwise-id
- urn:oid:1.3.6.1.4.1.5923.1.1.1.6 AND urn:mace:dir:attribute-def:eduPersonPrincipalName -> eppn
- urn:oid:1.3.6.1.4.1.5923.1.1.1.10 -> targeted-id
- urn:oasis:names:tc:SAML:2.0:nameid-format:persistent -> persistent-id

[...]

PR #18331, opened 2025-11-23
sources/saml: Allow user matching via attributes when NameID is transient

ikob:
## Details

This PR extends the behavior when handling SAML Assertions with a transient NameID, as discussed in https://github.com/goauthentik/authentik/issues/12766

In the current implementation, if the NameID is transient, the system immediately creates a temporary user without evaluating any SAML attributes. However, some Identity Providers (IdPs) expect that attribute-based user identification will still occur in such cases.

This patch introduces a proof-of-concept behavior in which, when NameID is transient, certain attributes are used to construct a provisional username, which is then passed to the user matching process according to the configured SourceUserMatchingModes. This approach is more aligned with how platforms like Authentik handle such scenarios.

Specifically:
- If SourceUserMatchingModes is EMAIL_LINK or DENY, and an email attribute is present, it is used as the provisional username.
- If USERNAME_LINK or DENY, and an ePPN attribute is present, it is used as the provisional username.
- In all other cases (e.g., modes including IDENTITY), the system falls back to the existing behavior, using the transient NameID as the username to create a temporary user, which will be deleted upon logout or after a timeout.

In addition, this patch changes the behavior so that if the user's `username` is explicitly set via property mapping, the user is no longer treated as temporary — even if the original NameID was transient.

---

## Checklist

If applicable

### DIFF
--- a/authentik/sources/saml/processors/response.py
+++ b/authentik/sources/saml/processors/response.py
@@ -32,8 +32,7 @@ from authentik.core.models import (
     USER_ATTRIBUTE_EXPIRES,
     USER_ATTRIBUTE_GENERATED,
     USER_ATTRIBUTE_SOURCES,
-    USERNAME_MAX_LENGTH,
-    User,
+    SourceUserMatchingModes,
 )
 from authentik.core.sources.flow_manager import SourceFlowManager
 from authentik.lib.utils.time import timedelta_from_string
@@ -99,6 +98,7 @@ class ResponseProcessor:
         self._verify_request_id()
         self._verify_status()
         self._verify_conditions()
+        self._saml_attributes = self._source.extract_attributes(self._root, self.get_assertion())
 
     def _decrypt_response(self):
         """Decrypt SAMLResponse EncryptedAssertion Element"""
@@ -237,46 +237,6 @@ class ResponseProcessor:
         if message_text or detail_text:
             LOGGER.debug("SAML Status message", message=message_text, detail=detail_text)
 
-    def _handle_name_id_transient(self) -> SourceFlowManager:
-        """Handle a NameID with the Format of Transient. This is a bit more complex than other
-        formats, as we need to create a temporary User that is used in the session. This
-        user has an attribute that refers to our Source for cleanup. The user is also deleted
-        on logout and periodically."""
-        # Create a temporary User
-        name_id_el, name_id = self._get_name_id()
-        # trim username to ensure it is max 150 chars
-        username = f"ak-{name_id[: USERNAME_MAX_LENGTH - 14]}-transient"
-        expiry = mktime(
-            (now() + timedelta_from_string(self._source.temporary_user_delete_after)).timetuple()
-        )
-        user: User = User.objects.create(
-            username=username,
-            attributes={
-                USER_ATTRIBUTE_GENERATED: True,
-                USER_ATTRIBUTE_SOURCES: [
-                    self._source.name,
-                ],
-                USER_ATTRIBUTE_DELETE_ON_LOGOUT: True,
-                USER_ATTRIBUTE_EXPIRES: expiry,
-            },
-            path=self._source.get_user_path(),
-        )
-        LOGGER.debug("Created temporary user for NameID Transient", username=name_id)
-        user.set_unusable_password()
-        user.save()
-        UserSAMLSourceConnection.objects.create(source=self._source, user=user, identifier=name_id)
-        return SAMLSourceFlowManager(
-            source=self._source,
-            request=self._http_request,
-            identifier=str(name_id),
-            user_info={
-                "root": self._root,
-                "assertion": self.get_assertion(),
-                "name_id": name_id_el,
-            },
-            policy_context={},
-        )
-
     def get_assertion(self) -> Element | None:
         """Get assertion element, if we have a signed assertion"""
         if self._assertion is not None:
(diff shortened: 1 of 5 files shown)


authentik/core/tasks.py
```diff
@@ -52,11 +54,18 @@ def clean_temporary_users():
...
+            user.save()
+            continue
         delta: timedelta = _now - datetime.fromtimestamp(
             user.attributes.get(USER_ATTRIBUTE_EXPIRES)
         )
         if delta.total_seconds() > 0:
             LOGGER.debug("User is expired and will be deleted.", user=user, delta=delta)
             user.delete()
             deleted_users += 1
+        print(f"Delta {delta}")
```
PeshekDotDev: Can you remove this print statement?

authentik/sources/saml/processors/constants.py
```diff
@@ -65,3 +65,19 @@
...
+#    "urn:mace:dir:attribute-def:mail",
+#    "mail",
+
+SAML_ATTR_EPPN = "urn:oid:1.3.6.1.4.1.5923.1.1.1.6"
+#    "urn:mace:dir:attribute-def:eduPersonPrincipalName",
+#   "eduPersonPrincipalName",
+
+SAML_ATTR_EPTID = "urn:oid:1.3.6.1.4.1.5923.1.1.1.10"
+#    "urn:mace:dir:attribute-def:eduPersonTargetedID",
+#    "eduPersonTargetedID",
```
PeshekDotDev: Can we rename these to SAML_ATTRIBUTE_xxx to match the rest of the constants naming conventions?

authentik/sources/saml/processors/response.py
```diff
@@ -268,7 +221,8 @@ def prepare_flow_manager(self) -> SourceFlowManager:
                 got=name_id.attrib["Format"],
             )
         # transient NameIDs are handled separately as they don't have to go through flows.
-        if name_id.attrib["Format"] == SAML_NAME_ID_FORMAT_TRANSIENT:
+        # if name_id.attrib["Format"] == SAML_NAME_ID_FORMAT_TRANSIENT:
+        if False:
             return self._handle_name_id_transient()
```
PeshekDotDev: Can we clean these last 4 lines up?

PeshekDotDev (changes requested):
Outside of the general cleanup, we may need to convert this to use the "delete on logout" instead of the new "transient token"

BeryJu:
instead of adding the transient ID to the user attributes, it should be saved in a `UserSAMLSourceConnection` object (will also need to add a field to that, probably a nullable text field `transient_id`)

ikob:
I have a question regarding the role of the SourceConnection object.

Currently, in `sources/saml/processors/response.py`, when handing off to `SAMLFlowManager`, a transient NameID is passed as `identifier = name_id.text`. From the perspective of identifier stabilization, this doesn’t seem appropriate.

Before passing control to `SAMLFlowManager`, we should extract SAML attributes and determine a more stable identifier (for example, based on ePPN or email when available).  

In that case, would it be acceptable to store such attributes on the `UserSAMLSourceConnection` object?

ikob:
> instead of adding the transient ID to the user attributes, it should be saved in a `UserSAMLSourceConnection` object (will also need to add a field to that, probably a nullable text field `transient_id`)

A temporary token derived from the transient NameID may be stored to support lifecycle-related behavior such as logout and expiry. Although it is expected to expire, the original transient NameID is intentionally not stored in clear form.

Note:
The primary goal of this PR is to preserve and make SAML assertion attributes available even when the NameID format is transient. The temporary token exists only as a supporting mechanism for this behavior.

PeshekDotDev:
@ikob Apologies, are you saying that there is an implementation reason as to why we cannot store a representation of the transient NameID in the `UserSAMLSourceConnection` instead of the User Attributes? If so, could you point it out to me so we can discuss alternatives?

ikob:
@PeshekDotDev, I’d like to better understand the reasoning behind the suggestion to store the transient NameID (or a derived representation) in UserSAMLSourceConnection rather than in user attributes.

At this stage, I’m not trying to push a particular design decision in this PR; I mainly want to understand the motivation and assumptions behind that suggestion.

For example, is the concern primarily about visibility — that storing any representation of a transient NameID in user attributes is undesirable because attributes may be visible to users or administrators?

Or is there a more general architectural reason why user attributes are not considered suitable for holding transient, lifecycle-scoped data (used only for logout or expiry handling)?

PeshekDotDev:
It's mainly just a organization/cleanlyness concern. We want to prevent the user attributes as light as possible, and the session is already designed to be temporary, so it makes sense to keep it in that space organizationally to us

ikob:
@PeshekDotDev, I understand the preference to keep user attributes as lightweight as possible, and that transient data fits better in a session- or connection-scoped structure.

I’ll take this into account and consider alternatives (e.g. using UserSourceConnection) when looking at next steps.

ikob:
@PeshekDotDev, 

Following the discussion,  I decided to keep this PR minimal: it does not add transient data to user attributes and does not modify UserSAMLSourceConnection.

The change is limited to adjusting how attributes are used during user matching when a transient NameID is received.

ikob:
Updated this branch to the latest upstream/main (5a3383338e) and rebuilt the transient SAML changes on top of it.

PR #18332, opened 2025-11-23
website/integrations: harbor: fix slashes in URLs and group claim name

adns44:
Hello!

When I configured the Harbor integration I ran into two issues and fixed them in the documentation.

Fixes:
- URI fix: The slash required in Harbor and not required in Authentik strict URL
- Group Claim Name fix: If provided Harbor fetches groups and admin group can be used.


dewi-tik (approved):
Thanks @adns44 !

PR #18335, opened 2025-11-24
web: Update Deprecated NPM Packages

GirlBossRush:
## Details

This PR upgrades several long-standing packages which report outdated and deprecation warnings during installation.

#### SFE Rollup

- The `FormData` polyfill package has been vendored to remove several warnings about dependencies which the author needed to _build_ the package. The code executed during runtime is quite small.
- Replaced `rollup-plugin-copy` with Node's built-in `fs` API. We're only copying a single Bootstrap.css file so this was pretty painless. 

#### Lex (QL Search)

This package has been [marked as deprecated](https://www.npmjs.com/package/lex) but is required for our QL search. Lex is quite small and has been vendored into a single JS file.

#### `*-config` packages

Our shared config files for ESLint, Prettier, and TypeScript have had all their peer dependencies bumped to the latest ranges, allowing Dependabot's PRs to work again.

Additionally, each of these dependencies now supports linting and formatting before publishing.

#### E2E Test Dependencies

The diamond-shaped dependency graph of ESBuild, Vitest, Storybook, and Playwright has been detangled, for now. ESBuild's sub-zero package numbers continues to make this a recurring issue, but this gets us on the latest everything.


packages/esbuild-plugin-live-reload/plugin/index.js
```diff
@@ -123,6 +123,13 @@ export function createRequestHandler({ pathname, dispatcher, logger = createLogg
         dispatcher.addEventListener("esbuild:error", listener);
         dispatcher.addEventListener("esbuild:end", listener);
 
+        const keepAliveInterval = setInterval(() => {
```
GirlBossRush, line 126: Fixes linter error about variable hoisting

packages/eslint-config/lib/javascript-config.js
```diff
@@ -68,7 +69,7 @@ export const javaScriptConfig = tseslint.config({
         "no-fallthrough": "error",
         "no-func-assign": "error",
         "no-implied-eval": "error",
-        "no-implicit-coercion": "error",
+        "no-implicit-coercion": ["error", { allow: ["!!"] }],
```
GirlBossRush, line 72: This is more consistent with the usage we use in `web` where the rule is disabled.

packages/eslint-config/lib/javascript-config.js
```diff
@@ -8,7 +9,7 @@ const MAX_PARAMS = 5;
 /**
  * ESLint configuration for JavaScript authentik projects.
  */
-export const javaScriptConfig = tseslint.config({
+export const javaScriptConfig = defineConfig({
```
GirlBossRush, line 12: Fixes TSLint's warning about the deprecated config helper.

packages/eslint-config/index.js
```diff
@@ -39,12 +48,18 @@ export const DefaultIgnorePatterns = [
...
-    return tseslint.config(
+export function createESLintPackageConfig({
+    ignorePatterns = DefaultIgnorePatterns,
+    parserOptions = {},
+} = {}) {
+    return defineConfig(
         {
             ignores: ignorePatterns,
+            languageOptions: {
+                parserOptions,
```
GirlBossRush, line 62: Allows consumer of this package to provide a root directory to resolve types. This allows VS Code's ESLint extension to correctly report errors.

packages/eslint-config/package.json
```diff
@@ -27,29 +31,35 @@
...
-        "eslint-plugin-react-hooks": "^7.0.0",
-        "eslint-plugin-wc": "^3.0.1"
+        "eslint-plugin-react-hooks": "^7.0.1",
+        "eslint-plugin-wc": "^3.0.2"
     },
     "devDependencies": {
-        "@goauthentik/prettier-config": "^3.1.0",
-        "@goauthentik/tsconfig": "^1.0.1",
+        "@goauthentik/prettier-config": "../prettier-config",
+        "@goauthentik/tsconfig": "../tsconfig",
```
GirlBossRush, line 43: `*-config` packages are published without developer dependencies, allowing us to avoid a hard version here. This also fixes an timing issue when attempting to bump the version of multiple packages at the same time.

web/packages/node-domexception/index.cjs
```diff
@@ -0,0 +1,17 @@
+/**
```
GirlBossRush, line 1: This technically isn't used, but it is a transitive dependency of `rapidoc` via Swagger API. The original package is a linchpin for several legacy dependencies which are either no longer in use, or have a more recent patch without reported vulnerabilities.

web/packages/sfe/rollup.config.mjs
```diff
@@ -1,20 +1,46 @@
...
+ * @import { Plugin, RollupOptions } from "rollup";
  */
+
+import * as fs from "node:fs/promises";
+import { join } from "node:path";
+
 import commonjs from "@rollup/plugin-commonjs";
 import resolve from "@rollup/plugin-node-resolve";
 import swc from "@rollup/plugin-swc";
-import copy from "rollup-plugin-copy";
```
GirlBossRush, line 9: A similar story here with the copy plugin being the source of `glob` related vulnerabilities.

web/package.json
```diff
@@ -229,8 +229,7 @@
                 "./dist/styles/**"
             ],
             "dependencies": [
-                "build-locales",
-                "build:sfe"
+                "build-locales"
```
GirlBossRush, line 232: I believe this became vestigial in the past year or so. The SFE's build script always runs _after_ because `build-web.mjs` cleans the directory `dist` directory.

web/package.json
```diff
@@ -301,14 +300,20 @@
     ],
     "prettier": "@goauthentik/prettier-config",
     "overrides": {
-        "chromedriver": {
-            "axios": "^1.8.4"
```
GirlBossRush, line 305: No longer needed after WebDriver was swapped with Playwright

BeryJu (approved):

PR #18350, opened 2025-11-24
endpoints: AuthN and AuthZ

BeryJu:
based on the initial endpoint devices structure, the API for authn and authz


rissson (approved):
Approving database changes, although one migration needs to be merged.

PR #18427, opened 2025-11-28
web: Adjust colors

GirlBossRush:
## Details

- Secondary buttons use a color closer to the primary in light mode
- Sidebar border colors and width better align with the docs.


kensternberg-authentik (approved):
Still think some of the red is too dark and has poor contrast, but this isn't about that.

PR #18446, opened 2025-11-29
core: Per-request caching for serialized users

joaocfernandes:
## Details

Improve execution time for calls to groups/.

For situations where a huge number of groups exists and a huge number of users scattered around them, we can save time avoiding re-serialization, if the serialization result for each user is cached by request.

This has the tradeoff of creating a cache per request. The invalidation model is simple as it only lives for the period of the request and therefore it's expected a low-memory footprint , if page_size is within a reasonable value.

---

## Checklist

If applicable


dewi-tik:
Could this please be rebased on main?

PR #18503, opened 2025-12-01
translate: fix source locale not matching transifex

GirlBossRush:
## Details

REPLACE ME

---

## Checklist

If applicable


BeryJu (approved):

kensternberg-authentik (approved):
Isn't broken, as far as I can tell, at least not on the front-end.  Code changes look solid.  Doesn't necessitate needing to know the full locale, which is something we avoided early on and shouldn't re-introduce.

PR #18517, opened 2025-12-02
sources/telegram: implement connecting existing user to a Telegram account

atereshkin:
## Details

This PR implements missing "connect account" functionality for the Telegram source. When a Telegram source is configured, existing users should be able connect their Telegram account by going to Settings -> Connected services. Currently this functionality is missing and this PR addresses that. 

A respective view is added in sources/telegram/api/source.py  and renderConnectButton is implemented in SourceSettingsTelegram.ts.

This closes #18467

## Checklist

If applicable

-   [n/a] The documentation has been updated
-   [n/a] The documentation has been formatted (`make docs`)

### DIFF
--- a/authentik/sources/telegram/api/source.py
+++ b/authentik/sources/telegram/api/source.py
@@ -65,13 +65,6 @@ class TelegramSourceViewSet(UsedByMixin, ModelViewSet):
             201: UserTelegramSourceConnectionSerializer,
             403: OpenApiResponse(description="Access denied"),
         },
-        # parameters=[
-        #     OpenApiParameter(
-        #         name="slug",
-        #         location=OpenApiParameter.PATH,
-        #         type=OpenApiTypes.STR,
-        #     )
-        # ],
     )
     @action(
         methods=["POST"],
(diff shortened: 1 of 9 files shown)


authentik/sources/telegram/api/source.py
```diff
@@ -39,3 +58,45 @@ class TelegramSourceViewSet(UsedByMixin, ModelViewSet):
...
+        detail=True,
+        pagination_class=None,
+        filter_backends=[],
+        permission_classes=[IsAuthenticated],
+    )
+    def connect_user(self, request: Request, slug: str) -> Response:
+
+        source: TelegramSource = get_object_or_404(TelegramSource, slug=slug)
+        serializer = TelegramAuthSerializer(bot_token=source.bot_token, data=request.data)
+        serializer.is_valid(raise_exception=True)
```
BeryJu, line 81: fyi https://github.com/goauthentik/authentik/pull/17435 to make this a bit cleaner

atereshkin: @BeryJu I actually used that initially, but then had to change to invoking the serializer explicitly because it needs some extra initialization with the data from the DB ("...bot_token=source.bot_token..."). It's a bit awkward, but can't see a better way immediately.

BeryJu: ah didn't catch that, fair enough, all good

rissson (approved):

kensternberg-authentik (approved):
Runs, isn't broken, no console warnings, looks like it does what it says it does (I have no  Telegram credentials to test against).

PR #18541, opened 2025-12-02
core, web: Fix Han language codes.

GirlBossRush:
## Details

Swap underscores for dashes in Chinese translation files


rissson (approved):

GirlBossRush:
Resolved via #18426

PR #18560, opened 2025-12-03
web: Locale selector

GirlBossRush:
## Details

This PR adds a locale selector to the flow executor.

### Changes

- API middleware now synchronizes with the current locale
- Selecting a locale persists the preference for the browser session. After authentication, user attributes take priority over the session preference (if any user attributes exist)
- Removed commented-out context controller code

## Screenshots

Screenshot 2025-12-03 at 07 33 49

Screenshot 2025-12-03 at 07 34 04

Screenshot 2025-12-03 at 07 34 13

### DIFF
--- a/web/src/common/global.ts
+++ b/web/src/common/global.ts
@@ -1,3 +1,6 @@
+import { TargetLocale } from "#common/ui/locale/definitions";
+import { autoDetectLanguage } from "#common/ui/locale/utils";
+
 import {
     Config,
     ConfigFromJSON,
@@ -6,9 +9,11 @@ import {
     FlowLayoutEnum,
 } from "@goauthentik/api";
 
+const convertedSymbol = Symbol("ak-converted");
+
 export interface GlobalAuthentik {
-    _converted?: boolean;
-    locale?: string;
+    [convertedSymbol]?: boolean;
+    locale: TargetLocale;
     flow?: {
         layout: FlowLayoutEnum;
     };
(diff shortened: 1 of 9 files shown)


BeryJu (commented):
Looks good, two things:
- Should this persist the the locale change to the URL? (I dont have an opinion on this, both methods have upsides and downsides)
- It should refresh at least the challenge and probably use a middleware to set the `Accept-Language` header for API requests since that'll influence which strings we get from the server

GirlBossRush:
@BeryJu

Thank you for the review 😊

>Should this persist the the locale change to the URL? (I dont have an opinion on this, both methods have upsides and downsides)

IMO the query param is nice because it keeps the locale stable, but it makes link sharing a bit troublesome and may not survive across redirects. The latest iteration uses [`sessionStorage`](https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage). Selecting a locale persists the preference for the tab's lifecycle. After authentication, user attributes take priority over the session preference (if any user attributes exist)

>It should refresh at least the challenge and probably use a middleware to set the Accept-Language header for API requests since that'll influence which strings we get from the server

I've updated the API middleware to synchronize with the current locale, which has a nice effect on the UI for the client-side strings, but I've observed that the flow executor doesn't handle repeat requests. Invoking `FlowsApi.flowsExecutorGet` after changing the locale can result in an access denied -- maybe something to do with an expectation of subsequent API calls being tied to flow progression?

web/src/flow/FlowExecutor.ts
```diff
@@ -481,6 +485,51 @@ export class FlowExecutor
         </button>`;
     }
 
+    #localeChangeListener = (event: Event) => {
```
kensternberg-authentik, line 488: So, like, weird thought: Why isn't this a component?  It take a single scalar argument, `locale`, renders a single control, and sends a single event (`setSessionLocale`)?  You could avoid a lot of extra surgery on `FlowExecutor` by giving this some concrete, independent existence.

web/src/common/global.ts
```diff
@@ -6,9 +9,11 @@ import {
     FlowLayoutEnum,
 } from "@goauthentik/api";
 
+const convertedSymbol = Symbol("ak-converted");
+
 export interface GlobalAuthentik {
-    _converted?: boolean;
-    locale?: string;
+    [convertedSymbol]?: boolean;
```
kensternberg-authentik, line 15: 👍

kensternberg-authentik (approved):
Looks good, runs, passes lint, produces no errors on the console.  

1. Suggestion: Make the control its own component.
2. Possible improvement: Provide an option for admins to "always put browser's locale at the top" or "always put brand's preferred locale at the top".

PR #18588, opened 2025-12-04
crypto: separate permissions for certificate and private keydownload

BeryJu:


BeryJu:
@goauthentik/docs Not sure if we need to document this/how we want to handle migrations

rissson (approved):
This might need a release notes entry for users that have automation setup for certificates via the API

PR #18610, opened 2025-12-04
web: Codemirror fixes

GirlBossRush:
## Details

This PR fixes several issues with Codemirror after the latest dependencies upgrade.

### Fixes

- Load Codemirror dynamically, fixing page stutters before Codemirror is rendered
- Clean up error reporting on user provided input
- Fix ARIA attributes which prevented tabbing into the editor element


web/src/elements/codemirror/editor.ts
```diff
@@ -0,0 +1,149 @@
...
+            return htmlLang();
+        case CodeMirrorMode.Python:
+            return python();
+        case CodeMirrorMode.CSS:
+            return cssLang();
+        case CodeMirrorMode.YAML:
+            return new LanguageSupport(StreamLanguage.define(yamlMode.yaml));
+    }
+
+    throw new TypeError(`Unrecognized CodeMirror mode: ${mode}`) as never;
```
kensternberg-authentik:
```suggestion
return match(mode)
    .with(CodeMirrorMode.XML, () => xml())
    .with(CodeMirrorMode.JavaScript, () => javascript())
    .with(CodeMirrorMode.HTML, () => htmlLang())
    .with(CodeMirrorMode.Python, () => python())
    .with(CodeMirrorMode.CSS, () => cssLang())
    .with(
       CodeMirrorMode.YAML, 
       () => new LanguageSupport(StreamLanguage.define(yamlMode.yaml)))
    .exhaustive();
```

`.exhaustive()` will enforce type-checking so you _can't_ pass it a type it doesn't recognize.  As long as CodeMirrorMode is a `Pick<>` of whatever CodeMirror's full collection is, this will work just fine internally.

kensternberg-authentik (approved):
Works and nicely organized.  Small suggestion to restrict our codeset via `Pick<>` and let Typescript and ts-pattern do the magic, but it's not necessary.

PR #18615, opened 2025-12-04
enterprise: account lockdown

### LINKED ISSUE #18160: "Panic button" in admin interface
"Panic button" in admin interface so that admins can forcefully sign a user or multiple users out of their account/sessions.

- Resets account password
- Locks the account
- Option to send email to user
- Option to send email to admins
- Option to send email to security email address

dominic-r:
Closes: https://github.com/goauthentik/authentik/issues/18160


web/src/admin/users/UserBulkPanicButtonForm.ts
```diff
@@ -0,0 +1,105 @@
...
+
+    @state()
+    private defaultNotifyAdmins = true;
+
+    @state()
+    private defaultNotifySecurity = false;
+
+    async firstUpdated(): Promise<void> {
+        try {
+            const settings = await new AdminApi(DEFAULT_CONFIG).adminSettingsRetrieve();
```
kensternberg-authentik: Since this component's server-side information is not dependent upon any public properties (`this.users`), this could be run at construction. Consider moving it there. Lit is... unhappy... if a render, even the first render, automagically schedules another.

web/src/admin/users/UserListPage.ts
```diff
@@ -172,44 +173,55 @@ export class UserListPage extends WithBrandConfig(
...
+            >
+                ${shouldShowWarning
+                    ? html`<div slot="notice" class="pf-c-form__alert">
+                          <div class="pf-c-alert pf-m-inline pf-m-warning">
+                              <div class="pf-c-alert__icon">
+                                  <i class="fas fa-exclamation-circle" aria-hidden="true"></i>
+                              </div>
+                              <h4 class="pf-c-alert__title">
+                                  ${msg(
+                                      str`Warning: You're about to delete the user you're logged in as (${shouldShowWarning.username}). Proceed at your own risk.`,
```
kensternberg-authentik: I know this is from the original, but the grammar here is awful. Consider re-writing it?  "You're about to delete user Joe, but you are currently logged in as Joe. Proceed at your own risk."  Something like that. Maybe ask @tanberry for advice.

kensternberg-authentik (approved):

website/docs/security/account-lockdown.md
```diff
@@ -0,0 +1,73 @@
...
+
+## What Account Lockdown does
+
+When triggered, Account Lockdown performs the following actions:
+
+- **Deactivates the user account**: The user can no longer log in
+- **Resets the user's password**: Sets a new random password, invalidating the old one
+- **Terminates all active sessions**: Immediately logs the user out of all devices and applications
+
+An event is created that can be used to trigger notifications via Notification Rules.
```
dewi-tik:
```suggestion
An event is created that can be used to [trigger notifications via Notification Rules](#configure-notifications).
```

website/docs/security/account-lockdown.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Trigger an Account Lockdown for a single user
+
+1. Log in to authentik as an admin, and open the Admin interface.
```
dewi-tik:
```suggestion
1. Log in to authentik as an administrator and open the authentik Admin interface.
```

website/docs/security/account-lockdown.md
```diff
@@ -0,0 +1,73 @@
...
+2. Navigate to **Directory** > **Users**.
+3. Click on the user you want to lock down.
+4. Click the **Account Lockdown** button.
+5. Enter a reason for the lockdown (this is logged for audit purposes).
+6. Click **Trigger Lockdown**.
+
+## Trigger an Account Lockdown for multiple users
+
+1. Log in to authentik as an admin, and open the Admin interface.
```
dewi-tik:
```suggestion
1. Log in to authentik as an administrator and open the authentik Admin interface.
```

website/docs/security/account-lockdown.md
```diff
@@ -0,0 +1,73 @@
...
+3. Select the users you want to lock down using the checkboxes.
+4. Click the **Account Lockdown** button in the toolbar.
+5. Enter a reason for the lockdown.
+6. Click **Trigger Lockdown**.
+
+## Configure notifications
+
+Account lockdown events can trigger notifications via the Notification Rules system. To set up notifications:
+
+1. Log in to authentik as an admin, and open the Admin interface.
```
dewi-tik:
```suggestion
1. Log in to authentik as an administrator and open the authentik Admin interface.
```

website/docs/security/account-lockdown.md
```diff
@@ -0,0 +1,73 @@
...
+    - **Transports**: Select the notification transports (e.g., email, webhook)
+    - **Severity**: Select the notification severity level
+10. Click **Create** to save the notification rule.
+11. In the notification rules list, click the arrow next to your new rule to expand it.
+12. Click **Bind existing Policy/Group/User**.
+13. Select the Event Matcher Policy you created and click **Create**.
+
+## Restore access after lockdown
+
+1. Log in to authentik as an admin, and open the Admin interface.
```
dewi-tik:
```suggestion
1. Log in to authentik as an administrator and open the authentik Admin interface.
```

website/docs/security/account-lockdown.md
```diff
@@ -0,0 +1,73 @@
...
+
+Account Lockdown is a security feature that allows administrators to quickly secure a user account in emergency situations, such as suspected compromise or unauthorized access.
+
+:::info
+Account Lockdown can be enabled or disabled in **System** > **Settings** under **Enable account lockdown**.
+:::
+
+:::info
+The security email address for notifications can be configured in **System** > **Settings** under **Security email**.
+:::
```
dewi-tik:
```suggestion
:::info Security email address
The security email address for notifications can be configured in **System** > **Settings** under **Security email**.
:::
```

website/docs/security/account-lockdown.md
```diff
@@ -0,0 +1,73 @@
+---
+title: Account Lockdown
+authentik_version: "2026.2"
+---
+
+Account Lockdown is a security feature that allows administrators to quickly secure a user account in emergency situations, such as suspected compromise or unauthorized access.
+
+:::info
+Account Lockdown can be enabled or disabled in **System** > **Settings** under **Enable account lockdown**.
+:::
```
dewi-tik:
```suggestion
:::info Enable or disable account lockdown
Account Lockdown can be enabled or disabled in **System** > **Settings** under **Enable account lockdown**.
:::
```

dewi-tik (changes requested):
Few minor changes to docs.

web/src/admin/events/RuleForm.ts
```diff
@@ -112,6 +112,15 @@ export class RuleForm extends ModelForm<NotificationRule, string> {
                 )}
             >
             </ak-switch-input>
+            <ak-switch-input
+                name="destinationSecurityEmail"
+                label=${msg("Send email to security address")}
+                ?checked=${this.instance?.destinationSecurityEmail ?? false}
```
kensternberg-authentik: I _think_ this could be easier written as `!!this.instance?.destinationSecurityEmail` or `Boolean(this.instance?.destinationSecurityEmail)` to make the type checker happy.

web/src/admin/users/AccountLockdownFormBase.ts
```diff
@@ -0,0 +1,176 @@
...
+                placeholder=${this.reasonPlaceholder}
+                help=${this.reasonHelp}
+                required
+            ></ak-textarea-input>
+        `;
+    }
+
+    renderForm(): TemplateResult {
+        return html`
+            ${this.renderWarningAlert()} ${this.renderInfoAlert()} ${this.renderAffectedUsers()}
```
kensternberg-authentik: If you're not happy with the way prettier lays this out, put comments between them: `<!-- -->`.  I do that sometimes.  This is fine, it's short and not gonna get lost in a tag soup, but in other cases it's important to help future-you find stuff.

web/src/admin/users/UserBulkAccountLockdownForm.ts
```diff
@@ -0,0 +1,228 @@
...
+    private result?: AccountLockdownBulkResponse;
+
+    protected override get accountCount(): number {
+        return this.users.length;
+    }
+
+    /**
+     * Find the parent modal and update the submit button visibility
+     */
+    private setSubmitButtonVisible(visible: boolean): void {
```
kensternberg-authentik: ... this feels like a shortcoming of the ModalForm, that you can't do this with an event. Especially since you're looking _up_ the DOM tree, so an event should be the way to communicate "Yo, parent, I need your attention." **_Events_ go up, and _Attributes_ go down.**

Not your job to fix it, but something @GirlBossRush and I should think about.

kensternberg-authentik (approved):

website/docs/security/account-lockdown.md
```diff
@@ -0,0 +1,106 @@
...
+- **Resets the user's password**: Sets a new random password, invalidating the old one
+- **Terminates all active sessions**: Immediately logs the user out of all devices and applications
+- **Revokes all tokens**: Invalidates all API tokens, OAuth access tokens, and refresh tokens
+
+An event is created that can be used to [trigger notifications via Notification Rules](#configure-notifications).
+
+:::note Protected accounts
+Account Lockdown cannot be triggered on:
+
+- Your own account
```
dominic-r: rm since self service

authentik/events/models.py
```diff
@@ -102,6 +102,9 @@ class EventAction(models.TextChoices):
     IMPERSONATION_STARTED = "impersonation_started"
     IMPERSONATION_ENDED = "impersonation_ended"
 
+    ACCOUNT_LOCKDOWN_TRIGGERED = "account_lockdown_triggered"
+    ACCOUNT_LOCKDOWN_SELF_TRIGGERED = "account_lockdown_self_triggered"
```
gergosimonyi: One of these is enough, then set `event.user`.

authentik/tenants/models.py
```diff
@@ -123,6 +123,15 @@ class Tenant(InternallyManagedMixin, TenantMixin, SerializerModel):
         default=100,
     )
 
+    account_lockdown_enabled = models.BooleanField(
+        help_text=_("Enable the account lockdown feature for administrators."), default=True
+    )
+    security_email = models.EmailField(
```
gergosimonyi: Why is this its own field? As far as I can tell, this is better done by `NotificationTransport.mode = TransportMode.EMAIL`.

authentik/core/api/users.py
```diff
@@ -383,6 +386,31 @@ class UserPasswordSetSerializer(PassiveSerializer):
     password = CharField(required=True)
 
 
+class UserAccountLockdownSerializer(PassiveSerializer):
+    """Payload to trigger account lockdown for a user"""
+
+    reason = CharField(
+        required=True,
+        min_length=1,
+        max_length=500,
```
gergosimonyi: I don't think this needs a `max_length`. And for that matter, I don't think it needs `min_length` either, but I can see the argument there.

Same goes for the bulk serializer.

authentik/core/api/users.py
```diff
@@ -383,6 +386,31 @@ class UserPasswordSetSerializer(PassiveSerializer):
...
+class UserBulkAccountLockdownSerializer(PassiveSerializer):
+    """Payload to trigger account lockdown for multiple users"""
+
+    users = PrimaryKeyRelatedField(
+        many=True, queryset=User.objects.all().exclude_anonymous(), help_text="Users to lock"
+    )
+    reason = CharField(
+        required=True,
+        min_length=1,
+        max_length=500,
```
gergosimonyi: I don't think this needs a max_length. (And for that matter, I don't think it needs min_length either, but I can see the argument there.)

authentik/core/api/users.py
```diff
@@ -851,6 +879,244 @@ def impersonate_end(self, request: Request) -> Response:
...
+    def _validate_lockdown_target(self, request: Request, user: User) -> str | None:
+        """Validate if a user can be targeted for account lockdown.
+
+        Returns an error message if validation fails, None if valid.
+        """
+        from guardian.shortcuts import get_anonymous_user
+
+        # Cannot lock down yourself
+        if user.pk == request.user.pk:
+            return _("Cannot trigger account lockdown on yourself.")
```
gergosimonyi: You can remove this special case by merging `account_lockdown` and `account_lockdown_self`. I think this should be done.

authentik/core/api/users.py
```diff
@@ -851,6 +879,244 @@ def impersonate_end(self, request: Request) -> Response:
...
+    ) -> Response:
+        """Trigger account lockdown for a user"""
+        if disabled_response := self._check_lockdown_enabled(request):
+            return disabled_response
+
+        user: User = self.get_object()
+        reason = body.validated_data["reason"]
+
+        if validation_error := self._validate_lockdown_target(request, user):
+            LOGGER.debug(
```
gergosimonyi: You're logging only in this function, but not in `account_lockdown_bulk`. Is there a reason for that?

authentik/core/api/users.py
```diff
@@ -851,6 +879,244 @@ def impersonate_end(self, request: Request) -> Response:
...
+                "Account lockdown validation failed",
+                user=user.username,
+                reason=validation_error,
+            )
+            return Response(
```
gergosimonyi: Instead of manually writing the response, you can raise `ValidationError({"non_field_errors": [_("message")]})` from `_validate_lockdown_target`.

(Same goes for `_check_lockdown_enabled` above.)

authentik/core/api/users.py
```diff
@@ -851,6 +879,244 @@ def impersonate_end(self, request: Request) -> Response:
...
+        Args:
+            request: The HTTP request
+            user: The user to lock down
+            reason: The reason for the lockdown
+            self_service: If True, creates a self-service event and omits triggered_by
+        """
+        with atomic():
+            user.is_active = False
+            new_password = generate_key()
+            user.set_password(new_password)
```
gergosimonyi:
```suggestion
            user.set_unusable_password()
```

authentik/core/api/users.py
```diff
@@ -851,6 +879,244 @@ def impersonate_end(self, request: Request) -> Response:
...
+
+            # Delete all sessions
+            Session.objects.filter(authenticatedsession__user=user).delete()
+
+            # Revoke all API and app password tokens
+            Token.objects.filter(user=user).delete()
+
+            # Revoke OAuth2 tokens
+            AccessToken.objects.filter(user=user).delete()
+            RefreshToken.objects.filter(user=user).delete()
```
gergosimonyi: Already done by [signal](https://github.com/goauthentik/authentik/blob/f268858d33a36df78ad71cd2dd4ab89fe3ee28eb/authentik/providers/oauth2/signals.py#L138)

authentik/core/api/users.py
```diff
@@ -851,6 +879,244 @@ def impersonate_end(self, request: Request) -> Response:
...
+            self_service: If True, creates a self-service event and omits triggered_by
+        """
+        with atomic():
+            user.is_active = False
+            new_password = generate_key()
+            user.set_password(new_password)
+            user.save()
+
+            # Delete all sessions
+            Session.objects.filter(authenticatedsession__user=user).delete()
```
gergosimonyi: Also `AuthenticatedSession`?

authentik/core/api/users.py
```diff
@@ -851,6 +879,244 @@ def impersonate_end(self, request: Request) -> Response:
...
+            LOGGER.info(
+                "Account lockdown triggered",
+                user=user.username,
+                triggered_by=request.user.username,
+            )
+            Event.new(
+                EventAction.ACCOUNT_LOCKDOWN_TRIGGERED,
+                reason=reason,
+                affected_user=user.username,
+                triggered_by=request.user.username,
```
gergosimonyi: This is recorded by `Event.user`.

authentik/core/api/users.py
```diff
@@ -851,6 +879,244 @@ def impersonate_end(self, request: Request) -> Response:
...
+            ).from_http(request, user)
+
+    @permission_required(None, ["authentik_core.reset_user_password", "authentik_core.change_user"])
```
gergosimonyi: Isn't an object permission good enough?

authentik/tenants/models.py
dominic-r: This went under the assumption that the security email could be used for more than lockdown in the future (based on team meeting) and I figured this was the best way to do it

authentik/core/api/users.py
dominic-r: In https://github.com/goauthentik/authentik/blob/e56ab5aa9eafdd8e2e92491e77974a5cc979cf59/authentik/core/models.py#L1299, we see it has `on_delete=models.CASCADE` from `Session`, so I don't think it's needed

website/docs/security/account-lockdown.md
```diff
@@ -0,0 +1,105 @@
+---
+title: Account Lockdown
+authentik_version: "2026.2"
+authentik_enterprise: true
+---
+
+Account Lockdown is a security feature that allows administrators to quickly secure a user account in emergency situations, such as suspected compromise or unauthorized access. Users can also lock their own account if they suspect it has been compromised.
```
dewi-tik:
```suggestion
Account Lockdown is a security feature that allows administrators to quickly secure a user account during emergencies, such as suspected compromise or unauthorized access. Users can also lock down their own account if they believe it has been compromised.
```

website/docs/security/account-lockdown.md
```diff
@@ -0,0 +1,105 @@
...
+:::
+
+## What Account Lockdown does
+
+When triggered, Account Lockdown performs the following actions:
+
+- **Deactivates the user account**: The user can no longer log in
+- **Resets the user's password**: Sets a new random password, invalidating the old one
+- **Terminates all active sessions**: Immediately logs the user out of all devices and applications
+- **Revokes all tokens**: Invalidates all API tokens, OAuth access tokens, and refresh tokens
```
dewi-tik:
```suggestion
- **Revokes all tokens**: Invalidates all API tokens, OAuth access tokens, and refresh tokens associated with the user account
```

website/docs/security/account-lockdown.md
```diff
@@ -0,0 +1,105 @@
...
+- Internal service accounts
+  :::
+
+## Trigger an Account Lockdown for a single user
+
+1. Log in to authentik as an administrator and open the authentik Admin interface.
+2. Navigate to **Directory** > **Users**.
+3. Click on the user you want to lock down.
+4. Click the **Account Lockdown** button.
+5. Enter a reason for the lockdown.
```
dewi-tik: Where/how does this reason get logged. Might be good to add some info about it.

website/docs/security/account-lockdown.md
```diff
@@ -0,0 +1,105 @@
...
+6. Click **Trigger Lockdown**.
+
+## Trigger an Account Lockdown for multiple users
+
+1. Log in to authentik as an administrator and open the authentik Admin interface.
+2. Navigate to **Directory** > **Users**.
+3. Select the users you want to lock down using the checkboxes.
+4. Click the **Account Lockdown** button in the toolbar.
+5. Enter a reason for the lockdown.
+6. Click **Trigger Lockdown**.
```
dewi-tik: I think that these two sections can be merged. Just modify  line 3 to "select one or more users" or something similar.

website/docs/security/account-lockdown.md
```diff
@@ -0,0 +1,105 @@
...
+
+## Self-service Account Lockdown
+
+Users can lock their own account if they suspect it has been compromised. This is useful when a user notices suspicious activity and wants to immediately secure their account without waiting for an administrator.
```
dewi-tik:
```suggestion
Users can lock down their own account if they suspect it has been compromised. This is useful when a user notices suspicious activity and wants to immediately secure their account without waiting for an administrator.
```

website/docs/security/account-lockdown.md
```diff
@@ -0,0 +1,105 @@
...
+5. Click **Lock My Account**.
+
+After locking your account, you will be immediately logged out and will not be able to log back in until an administrator restores your access.
+
+## Configure notifications
+
+Account lockdown events can trigger notifications via the Notification Rules system. There are two event types:
+
+- **Account Lockdown Triggered**: When an administrator locks a user's account
+- **Account Lockdown Self Triggered**: When a user locks their own account
```
dewi-tik: The wording here doesn't match line 53 and 84. Self-service vs self triggered. We should make it consistent throughout.

website/docs/security/account-lockdown.md
```diff
@@ -0,0 +1,105 @@
...
+
+1. Log in to authentik and open the User interface.
+2. Navigate to **Settings** > **Security**.
+3. Click the **Lock my account** button.
+4. Enter a reason describing why you are locking your account.
+5. Click **Lock My Account**.
+
+After locking your account, you will be immediately logged out and will not be able to log back in until an administrator restores your access.
+
+## Configure notifications
```
dewi-tik, line 103: Could we offer a blueprint for this? Not essential.

dewi-tik (changes requested):

authentik/core/api/users.py
```diff
@@ -382,6 +383,33 @@ class UserPasswordSetSerializer(PassiveSerializer):
...
+        required=True,
+        help_text="Reason for triggering account lockdown",
+    )
+
+
+class UserBulkAccountLockdownSerializer(PassiveSerializer):
+    """Payload to trigger account lockdown for multiple users"""
+
+    users = PrimaryKeyRelatedField(
+        many=True, queryset=User.objects.all().exclude_anonymous(), help_text="Users to lock"
```
BeryJu:
```suggestion
        many=True, queryset=User.objects.all().exclude_anonymous().exclude(type=UserTypes.INTERNAL_SERVICE_ACCOUNT), help_text="Users to lock"
```

authentik/core/api/users.py
```diff
@@ -382,6 +383,33 @@ class UserPasswordSetSerializer(PassiveSerializer):
     password = CharField(required=True)
 
 
+class UserAccountLockdownSerializer(PassiveSerializer):
+    """Payload to trigger account lockdown for a user"""
+
+    user = PrimaryKeyRelatedField(
+        queryset=User.objects.all().exclude_anonymous(),
```
BeryJu:
```suggestion
        queryset=User.objects.all().exclude_anonymous().exclude(type=UserTypes.INTERNAL_SERVICE_ACCOUNT),
```

authentik/core/api/users.py
```diff
@@ -850,6 +878,219 @@ def impersonate_end(self, request: Request) -> Response:
...
+        # Cannot lock down internal service accounts
+        if user.type == UserTypes.INTERNAL_SERVICE_ACCOUNT:
+            raise ValidationError(
+                {
+                    "non_field_errors": [
+                        _("Cannot trigger account lockdown on internal service accounts.")
+                    ]
+                }
+            )
+
```
BeryJu:
```suggestion
```

authentik/core/api/users.py
```diff
@@ -850,6 +878,219 @@ def impersonate_end(self, request: Request) -> Response:
...
+        """Check if account lockdown feature is enabled.
+
+        Raises ValidationError if disabled.
+        """
+        if not request.tenant.account_lockdown_enabled:
+            LOGGER.debug("Account lockdown feature is disabled")
+            raise ValidationError(
+                {"non_field_errors": [_("Account lockdown feature is disabled.")]}
+            )
+
```
BeryJu: should use `@validate` with the serializer, that'll validate no anonymous user and no internal service account form above

authentik/core/api/users.py
```diff
@@ -850,6 +878,219 @@ def impersonate_end(self, request: Request) -> Response:
...
+            LOGGER.info(
+                "Self-service account lockdown triggered",
+                user=user.username,
+            )
+        else:
+            LOGGER.info(
+                "Account lockdown triggered",
+                user=user.username,
+                triggered_by=request.user.username,
+            )
```
BeryJu:
```suggestion
```

not needed since event.new will always log and should be used for log filtering

authentik/tenants/models.py
BeryJu: I dont think we need to have this as a separate field, I'd rather just update the default notification rules to notify on this event.

The default rule is set to send notifications to the authentik Admins group which in most simple cases would be the same as this

BeryJu (commented):
imo this should be a flow on the brand as a "default_lockdown_flow", which would allow the admin to specify any custom text (for example what to do afterwards, how to contact your IT team, etc), any custom validation (?) and there would also be the ability for the admin to customize any kind of reason inputs.

With that, this would turn into an enterprise stage that would do the actual lockdown, would take all the data from flow context `prompt_data` for the event. We could probably also include a default flow for this that only applies when enterprise is active.

dewi-tik (approved):
LGTM for docs

authentik/core/api/users.py
```diff
@@ -382,6 +389,43 @@ class UserPasswordSetSerializer(PassiveSerializer):
     password = CharField(required=True)
 
 
+class UserAccountLockdownSerializer(PassiveSerializer):
```
BeryJu: Instead of being defined here, these should be in the mixin class in the respective django app, and then use `ConditionalInheritance` to pull them into the viewset.

Ref https://github.com/goauthentik/authentik/pull/19485 for not including static enterprise imports

authentik/blueprints/v1/common.py
```diff
@@ -210,7 +210,12 @@ class YAMLTag:
     """Base class for all YAML Tags"""
 
     def __repr__(self) -> str:
-        return str(self.resolve(BlueprintEntry(""), Blueprint()))
+        attrs = ", ".join(
+            f"{key}={value!r}" for key, value in vars(self).items() if not key.startswith("_")
+        )
+        if attrs:
+            return f"!{self.__class__.__name__}({attrs})"
+        return f"!{self.__class__.__name__}"
```
BeryJu: ...why?

authentik/enterprise/stages/account_lockdown/tests/test_api.py
```diff
@@ -0,0 +1,253 @@
...
+# Patch for enterprise license check
+patch_license = patch(
+    "authentik.enterprise.models.LicenseUsageStatus.is_valid",
+    MagicMock(return_value=True),
+)
+
+
+@patch_license
+class TestUsersAccountLockdownAPI(APITestCase):
+    """Test Users Account Lockdown API - Flow-based approach"""
```
BeryJu: should also be in the enterprise app

authentik/enterprise/stages/account_lockdown/migrations/__init__.py
```diff
@@ -0,0 +1 @@
+"""Account lockdown stage migrations"""
```
BeryJu:
```suggestion
```

authentik/enterprise/stages/account_lockdown/tests/__init__.py
```diff
@@ -0,0 +1 @@
+"""Account lockdown stage tests"""
```
BeryJu:
```suggestion
```

BeryJu (changes requested):
partial review

authentik/blueprints/v1/common.py
dominic-r: introduced in https://github.com/goauthentik/authentik/pull/18615/changes/bea56137098b412ef62c80b072acd0b2f1c9908c to fix tests

dominic-r: i reverted and ran tests again and it works

authentik/enterprise/stages/account_lockdown/models.py
```diff
@@ -0,0 +1,91 @@
...
+from django.db import models
+from django.utils.translation import gettext_lazy as _
+from django.views import View
+from rest_framework.serializers import BaseSerializer
+
+from authentik.flows.models import Stage
+
+# Context keys for account lockdown
+PLAN_CONTEXT_LOCKDOWN_TARGET = "lockdown_target_user"
+PLAN_CONTEXT_LOCKDOWN_TARGETS = "lockdown_target_users"  # List of users for bulk lockdown
```
BeryJu: I'd rather just have one key that's always a list and might only contain one user

authentik/enterprise/stages/account_lockdown/stage.py
```diff
@@ -0,0 +1,194 @@
...
+        """Execute lockdown actions on a single user."""
+        with atomic():
+            if stage.deactivate_user:
+                user.is_active = False
+            if stage.set_unusable_password:
+                user.set_unusable_password()
+            user.save()
+
+            if stage.delete_sessions:
+                Session.objects.filter(authenticatedsession__user=user).delete()
```
BeryJu: TODO: Check if that correctly cascades to SAML Session and OAuth Access tokens

Also it's probably a good idea to revoke OAuth Refresh tokens, but that might fall under `revoke_tokens`

authentik/events/models.py
```diff
@@ -102,6 +102,8 @@ class EventAction(models.TextChoices):
     IMPERSONATION_STARTED = "impersonation_started"
     IMPERSONATION_ENDED = "impersonation_ended"
 
+    ACCOUNT_LOCKDOWN_TRIGGERED = "account_lockdown_triggered"
```
```suggestion
    USER_LOCKDOWN_TRIGGERED = "user_lockdown_triggered"
```
to match the `user_write` event in name

authentik/stages/prompt/models.py
```diff
@@ -303,6 +308,10 @@ def field(  # noqa PLR0915
                 kwargs["default"] = self.placeholder
                 kwargs["required"] = False
                 kwargs["label"] = ""
+            case FieldTypes.ALERT_INFO | FieldTypes.ALERT_WARNING | FieldTypes.ALERT_DANGER:
+                kwargs["default"] = self.placeholder
+                kwargs["required"] = False
+                kwargs["label"] = ""
```
BeryJu, line 315: Could be a separate PR maybe

authentik/stages/prompt/stage.py
```diff
@@ -245,7 +245,12 @@ def clean_choices(self, choices):
 
     def get_challenge(self, *args, **kwargs) -> Challenge:
         fields: list[Prompt] = list(self.executor.current_stage.fields.all().order_by("order"))
-        context_prompt = self.executor.plan.context.get(PLAN_CONTEXT_PROMPT, {})
+        # Merge plan context with prompt context so expressions can access flow variables
+        # Prompt context values take precedence for field-specific overrides
+        context_prompt = {
+            **self.executor.plan.context,
+            **self.executor.plan.context.get(PLAN_CONTEXT_PROMPT, {}),
+        }
```
BeryJu: I dont think this is needed, it drastically changes the behaviour for existing flows that might have overlapping fields in the context

blueprints/default/flow-default-account-lockdown.yaml
```diff
@@ -0,0 +1,316 @@
...
+
+      def esc(value):
+          text = str(value or "")
+          return (
+              text.replace("&", "&amp;")
+              .replace("<", "&lt;")
+              .replace(">", "&gt;")
+              .replace('"', "&quot;")
+              .replace("'", "&#x27;")
+          )
```
BeryJu: should use django's escape_html

authentik/enterprise/stages/account_lockdown/models.py
```diff
@@ -0,0 +1,226 @@
...
+        default=True,
+        help_text=_(
+            "Revoke all tokens for the user " "(API, app password, recovery, verification, OAuth2)"
+        ),
+    )
+    self_service_message_title = models.TextField(
+        default=DEFAULT_SELF_SERVICE_MESSAGE_TITLE,
+        help_text=_("Title shown to users after self-service lockdown"),
+    )
+    self_service_message = models.TextField(
```
gergosimonyi:
```suggestion
    self_service_message_body = models.TextField(
```

authentik/enterprise/stages/account_lockdown/stage.py
```diff
@@ -0,0 +1,229 @@
...
+        """Get the lockdown reason from the plan context.
+
+        Priority:
+        1. prompt_data['reason'] (from a prompt stage)
+        2. PLAN_CONTEXT_LOCKDOWN_REASON (explicitly set)
+        3. Empty string as fallback
+        """
+        prompt_data = self.executor.plan.context.get(PLAN_CONTEXT_PROMPT, {})
+        if "reason" in prompt_data:
+            return prompt_data["reason"]
```
gergosimonyi: This should probably be namespaced more, e.g.

```suggestion
        if "lockdown_reason" in prompt_data:
            return prompt_data["lockdown_reason"]
```

authentik/enterprise/stages/account_lockdown/stage.py
```diff
@@ -0,0 +1,229 @@
...
+
+    def _lockdown_user(
+        self,
+        request: HttpRequest,
+        stage: AccountLockdownStage,
+        user: User,
+        reason: str,
+    ) -> None:
+        """Execute lockdown actions on a single user."""
+        with atomic():
```
gergosimonyi, line 98: Possible timing attack.

Imagine that two transactions are started database at the same time:

- this one to delete current tokens
- another one to create a new token

Since this transaction only deletes the ones it sees when it starts, the token created by the other transaction will survive.

Is that correct?

authentik/flows/views/executor.py
```diff
@@ -137,6 +138,34 @@ def _check_flow_token(self, key: str) -> FlowPlan | None:
...
+            FlowAuthenticationRequirement.REQUIRE_SUPERUSER,
+        ]:
+            if not self.request.user.is_authenticated:
+                LOGGER.warning(
+                    "f(exec): flow token used without authenticated user",
+                    flow=self.flow.slug,
+                    token=token.identifier,
+                )
+                return None
+            if token.user_id != self.request.user.pk:
```
gergosimonyi: Why is this needed? + Is there a test for this?

authentik/enterprise/stages/account_lockdown/stage.py
```diff
@@ -0,0 +1,229 @@
...
+        try:
+            self._lockdown_user(request, stage, user, reason)
+            self.logger.info("Account lockdown completed", user=user.username)
+            result = {"user": user, "success": True, "error": None}
+        except Exception as exc:  # noqa: BLE001
+            self.logger.warning("Account lockdown failed", user=user.username, exc=exc)
+            result = {"user": user, "success": False, "error": str(exc)}
+
+        # Store the result in plan context for the completion prompt.
+        self.executor.plan.context[PLAN_CONTEXT_LOCKDOWN_RESULT] = result
```
gergosimonyi: Where is this used?

authentik/enterprise/stages/account_lockdown/stage.py
```diff
@@ -0,0 +1,229 @@
...
+QS_LOCKDOWN_USER = "user_uuid"
+
+TARGET_REQUIRED_MESSAGE = "No target user specified for account lockdown"
+PERMISSION_DENIED_MESSAGE = "You do not have permission to lock down this account."
+ACCOUNT_LOCKDOWN_FAILED_MESSAGE = "Account lockdown failed for this account."
+SELF_SERVICE_FAILURE_MESSAGE_TITLE = "Account lockdown failed"
+SELF_SERVICE_FAILURE_MESSAGE = (
+    "<p>We could not lock your account. Please contact your administrator or "
+    "security team for assistance.</p>"
+)
```
gergosimonyi: Localize these please.

authentik/enterprise/stages/account_lockdown/stage.py
```diff
@@ -0,0 +1,229 @@
...
+
+                Token.objects.filter(user=user).delete()
+                AuthorizationCode.objects.filter(user=user).delete()
+                AccessToken.objects.filter(user=user).delete()
+                RefreshToken.objects.filter(user=user).delete()
+                DeviceToken.objects.filter(user=user).delete()
+
+        # Create event outside atomic block - lockdown succeeded, now log it
+        # This ensures the lockdown happens even if event creation fails
+        try:
```
gergosimonyi, line 126: Is there any reason to put this in a `try/catch`? If `Event.new` somehow raises an exception, that will be logged anyway.

dominic-r: `why try catch`

authentik/enterprise/stages/account_lockdown/stage.py
dominic-r: nowhere needed. removed in `drop lockdown result`

authentik/enterprise/stages/account_lockdown/stage.py
dominic-r: done in `Localize lockdown messages`

authentik/enterprise/stages/account_lockdown/stage.py
dominic-r: as to why

authentik/enterprise/stages/account_lockdown/stage.py
dominic-r: done via `Serialize token issuance during lockdown`

authentik/enterprise/stages/account_lockdown/stage.py
dominic-r: done via `Namespace lockdown reason`

authentik/enterprise/stages/account_lockdown/models.py
dominic-r: done via `Rename lockdown message body`

blueprints/example/flow-default-account-lockdown.yaml
```diff
@@ -0,0 +1,313 @@
+version: 1
+metadata:
+  name: Default - Account lockdown flow
```
gergosimonyi, line 3:
```suggestion
  name: Example - Account lockdown flow
```

blueprints/example/flow-default-account-lockdown.yaml
```diff
@@ -0,0 +1,313 @@
...
+          if pending_user:
+              email = escape(pending_user.email or pending_user.name or "No email")
+              user_html = f"<p><code>{escape(pending_user.username)}</code> ({email})</p>"
+          else:
+              user_html = "<p>the account selected when this one-time lockdown link was created</p>"
+
+          return (
+              "<p><strong>You are about to lock down the following account:</strong></p>"
+              f"{user_html}"
+              "<p>This is an emergency action for cutting off access to that account right away. "
```
gergosimonyi, line 74:
```suggestion
              "<p>This is an emergency action for cutting off access to the account right away. "
```

blueprints/example/flow-default-account-lockdown.yaml
```diff
@@ -0,0 +1,313 @@
...
+              "It does not lock the administrator who opened this page.</p>"
+              "<p><strong>This will immediately:</strong></p>"
+              "<ul>"
+              "<li>Invalidate passwords - Passwords will be set to random values</li>"
+              "<li>Deactivate accounts - Accounts will be disabled</li>"
```
gergosimonyi, line 79:
```suggestion
              "<li>Invalidate the user's password</li>"
              "<li>Deactivate the user</li>"
```

authentik/enterprise/stages/account_lockdown/api.py
```diff
@@ -0,0 +1,146 @@
...
+class AccountLockdownStageSerializer(EnterpriseRequiredMixin, StageSerializer):
+    """AccountLockdownStage Serializer"""
+
+    def validate_self_service_completion_flow(self, flow):
+        if flow and flow.authentication != FlowAuthenticationRequirement.NONE:
+            raise ValidationError(
+                _("Completion flow must not require authentication for self-service lockdown.")
+            )
+        return flow
+
```
gergosimonyi, line 42: This should be a frontend-only suggestion, because it causes problems on the backend. Scenario:

1. Set a `completion_flow` that has `authentication != FlowAuthenticationRequirement.NONE`
2. Change `completion_flow.authentication` to something else
3. Now admins will get an unrelated `ValidationError` when trying to change something like `self_service_message_title`

authentik/enterprise/stages/account_lockdown/api.py
```diff
@@ -0,0 +1,146 @@
...
+        """
+        target_user = body.validated_data.get("user")
+        self_service = target_user is None or target_user.pk == request.user.pk
+
+        if self_service:
+            user = request.user
+        else:
+            user = target_user
+            perm = "authentik_core.change_user"
+            if not request.user.has_perm(perm) and not request.user.has_perm(perm, user):
```
gergosimonyi, line 140:
```suggestion
            if not request.user.has_perm(perm, user):
```

should be enough, it's supposed to check both

gergosimonyi: Also should there be two of these checks? There's also one in `stage.py`.

If the answer is yes, I'd extract these to a common place, since they're necessarily the same.

blueprints/default/default-brand.yaml
```diff
@@ -17,13 +17,20 @@ entries:
...
+      required: false
   - attrs:
       flow_authentication:
         !Find [authentik_flows.flow, [slug, default-authentication-flow]]
       flow_invalidation:
         !Find [authentik_flows.flow, [slug, default-invalidation-flow]]
       flow_user_settings:
         !Find [authentik_flows.flow, [slug, default-user-settings-flow]]
+      flow_lockdown:
+        !Find [authentik_flows.flow, [slug, default-account-lockdown]]
```
gergosimonyi, line 33:
```suggestion
```

blueprints/default/default-brand.yaml
```diff
@@ -17,13 +17,20 @@ entries:
       identifiers:
         name: Default - User settings flow
       required: false
+  - model: authentik_blueprints.metaapplyblueprint
+    attrs:
+      identifiers:
+        name: Default - Account lockdown flow
+      required: false
```
gergosimonyi, line 24:
```suggestion
```

authentik/core/models.py
```diff
@@ -1213,6 +1213,14 @@ def __str__(self):
...
         return description
 
+    def save(self, *args, **kwargs):
+        """Serialize user-scoped token creation with other token mutations."""
+        if self._state.adding:
+            with transaction.atomic():
+                User.objects.select_for_update().only("pk").get(pk=self.user_id)
+                return super().save(*args, **kwargs)
+        return super().save(*args, **kwargs)
+
```
gergosimonyi, line 1223: This is against the timing attack discussed above, yes? It's a bit out of place here. I would have no idea what it does if I were to stumble upon this function without context.

Optimally measures against the attack should be right next to where we're deleting the tokens. Is that possible?

authentik/enterprise/stages/account_lockdown/stage.py
```diff
@@ -0,0 +1,228 @@
...
+        get_params = request.session.get(SESSION_KEY_GET, request.GET)
+        return get_params.get(QS_LOCKDOWN_USER)
+
+    def get_target_user(self, request: HttpRequest) -> User | None:
+        """Get the target user from the plan context or the authenticated request.
+
+        Priority:
+        1. Explicit user_uuid query parameter
+        2. PLAN_CONTEXT_PENDING_USER (compatibility fallback)
+        3. request.user (direct self-service execution)
```
gergosimonyi, line 49: This should not be considered. We should always have the query parameter.

(I just accidentally disabled my admin account when testing things because of this, just to mention my pain point :))

If you agree that we always have the query parameter, perhaps also remove/rework `PLAN_CONTEXT_PENDING_USER`, since it's never gonna be reached.

gergosimonyi (commented):
I have no opinion on the changes made to `authentik/blueprint`

blueprints/example/flow-default-account-lockdown.yaml
dominic-r: `Rename Example Flow`

blueprints/example/flow-default-account-lockdown.yaml
dominic-r: `Tighten Admin Copy`

blueprints/example/flow-default-account-lockdown.yaml
dominic-r: `Adjust Alert Wording`

authentik/enterprise/stages/account_lockdown/api.py
dominic-r: `Drop Backend Validation`

authentik/enterprise/stages/account_lockdown/api.py
dominic-r: `Share Lock Permissions`

blueprints/default/default-brand.yaml
dominic-r: `Drop Brand Lockdown`

blueprints/default/default-brand.yaml
dominic-r: `Drop Lockdown Blueprint`

authentik/enterprise/stages/account_lockdown/stage.py
dominic-r: `Require Query Target`

authentik/core/models.py
dominic-r: `Name Token Lock`

authentik/core/models.py
```diff
@@ -1213,6 +1213,10 @@ def __str__(self):
             description += f" (expires={self.expires})"
         return description
 
+    def save(self, *args, **kwargs):
+        """Persist the token."""
```
dominic-r:
```suggestion
```

Image

authentik/enterprise/stages/account_lockdown/stage.py
```diff
@@ -0,0 +1,258 @@
...
+            user = User.objects.get(pk=user.pk)
+            self._apply_lockdown_actions(stage, user)
+            self._delete_lockdown_artifacts(stage, user)
+
+        # Step 1: the first transaction updates the user and deletes matching
+        # sessions and tokens.
+        # Step 2: another request can create a new token or session for the
+        # same user at the same time.
+        # Step 3: keep deleting outside the first transaction until no matching
+        # artifacts remain.
```
gergosimonyi:
```suggestion
        # These additional checks/deletes are done to prevent a timing attack that creates tokens
        # with a compromised token that is simultaneously being deleted.
```

authentik/enterprise/stages/account_lockdown/tests/test_api.py
```diff
@@ -0,0 +1,161 @@
...
+        """Test account lockdown requires authentication"""
+        response = self.client.post(
+            reverse("authentik_api:user-account-lockdown"),
+            data={"user": self.user.pk},
+            format="json",
+        )
+
+        self.assertEqual(response.status_code, 403)
+
+    def test_account_lockdown_non_admin(self):
```
gergosimonyi: Test for `change_user` (3 cases: positive and negative and negative with self-service) instead of admin account please.

authentik/enterprise/stages/account_lockdown/tests/test_api.py
```diff
@@ -0,0 +1,161 @@
...
+            reverse("authentik_api:user-account-lockdown"),
+            data={"user": self.user.pk},
+            format="json",
+        )
+
+        self.assertEqual(response.status_code, 403)
+
+
+@patch_license
+class TestUsersAccountLockdownSelfServiceAPI(AccountLockdownAPITestCase):
```
gergosimonyi: I don't see a need for this anymore.

authentik/enterprise/stages/account_lockdown/tests/test_stage.py
```diff
@@ -0,0 +1,480 @@
...
+        target = view.get_target_user(request)
+        self.assertEqual(target.pk, self.target_user.pk)
+        self.assertTrue(can_lock_user(request.user, target))
+        view._lockdown_user(request, self.stage, target, view.get_reason())
+
+        self.target_user.refresh_from_db()
+        self.assertFalse(self.target_user.is_active)
+
+    def test_lockdown_reason_from_prompt(self):
+        """Test lockdown stage reads the namespaced reason from prompt data."""
```
gergosimonyi:
```suggestion
        """Test lockdown stage reads the reason from prompt data."""
```

gergosimonyi (approved):
As far as I'm concerned, this is good to go.

dominic-r:
🎉

GirlBossRush:
@dominic-r latest iteration looks great. I've rebased to the best of my ability, with a merge to prep for #21697. Will circle back once that's merged to update this branch

authentik/events/models.py
```diff
@@ -105,6 +105,8 @@ class EventAction(models.TextChoices):
     IMPERSONATION_STARTED = "impersonation_started"
     IMPERSONATION_ENDED = "impersonation_ended"
 
+    USER_LOCKDOWN_TRIGGERED = "user_lockdown_triggered"
```
BeryJu: still on the fence whether we want a dedicated action for this

authentik/stages/prompt/models.py
```diff
@@ -87,6 +87,11 @@ class FieldTypes(models.TextChoices):
     HIDDEN = "hidden", _("Hidden: Hidden field, can be used to insert data into form.")
     STATIC = "static", _("Static: Static value, displayed as-is.")
 
+    # Alert box types for displaying styled messages
+    ALERT_INFO = "alert_info", _("Alert (Info): Static alert box with info styling")
+    ALERT_WARNING = "alert_warning", _("Alert (Warning): Static alert box with warning styling")
+    ALERT_DANGER = "alert_danger", _("Alert (Danger): Static alert box with danger styling")
+
```
BeryJu, line 94: I dont think this is enough to make it work, they probably need some logic to render it as static fields...unless that's the fallback logic?

authentik/enterprise/stages/account_lockdown/models.py
```diff
@@ -0,0 +1,70 @@
...
+        help_text=_("Set an unusable password for the user"),
+    )
+    delete_sessions = models.BooleanField(
+        default=True,
+        help_text=_("Delete all active sessions for the user"),
+    )
+    revoke_tokens = models.BooleanField(
+        default=True,
+        help_text=_(
+            "Revoke all tokens for the user " "(API, app password, recovery, verification, OAuth2)"
```
BeryJu:
```suggestion
            "Revoke all tokens for the user (API, app password, recovery, verification, OAuth)"
```

authentik/enterprise/stages/account_lockdown/stage.py
```diff
@@ -0,0 +1,253 @@
...
+SELF_SERVICE_FAILURE_MESSAGE_TITLE = _("Account lockdown failed")
+SELF_SERVICE_FAILURE_MESSAGE = _(
+    "<p>We could not lock your account. Please contact your administrator or "
+    "security team for assistance.</p>"
+)
+
+
+def render_lockdown_message_html(title: str, body: str) -> str:
+    """Render simple lockdown message markup."""
+    return f"<h1>{escape(title)}</h1>{body}"
```
BeryJu: there shouldn't be any markdown for this on the backend ideally

authentik/enterprise/stages/account_lockdown/stage.py
```diff
@@ -0,0 +1,253 @@
...
+
+
+def get_lockdown_target_users() -> QuerySet[User]:
+    """Return users that can be targeted by account lockdown."""
+    return User.objects.exclude_anonymous().exclude(type=UserTypes.INTERNAL_SERVICE_ACCOUNT)
+
+
+def get_lockdown_token_models() -> tuple[type[Model], ...]:
+    """Return token and grant models removed by account lockdown."""
+    from authentik.providers.oauth2.models import (
```
BeryJu: That's incomplete, there's SAMLSession too, and I dont think this should be static

authentik/enterprise/stages/account_lockdown/stage.py
```diff
@@ -0,0 +1,253 @@
...
+            with atomic():
+                self._delete_lockdown_artifacts(stage, user)
+
+        self._emit_lockdown_event(request, user, reason)
+
+    def dispatch(self, request: HttpRequest) -> HttpResponse:
+        """Execute account lockdown actions."""
+        stage: AccountLockdownStage = self.executor.current_stage
+
+        user = self.get_target_user(request)
```
BeryJu: this should use `self.get_pending_user()`, to specify a different user it should be set in the flow planner and its context

authentik/enterprise/stages/account_lockdown/stage.py
```diff
@@ -0,0 +1,253 @@
...
+    def get_target_user_uuid(self, request: HttpRequest) -> str | None:
+        """Read the requested lockdown target from the flow query parameters."""
+        get_params = request.session.get(SESSION_KEY_GET, request.GET)
+        return get_params.get(QS_LOCKDOWN_USER)
+
+    def get_target_user(self, request: HttpRequest) -> User | None:
+        """Get the target user from the flow query parameters."""
+        if not (target_uuid := self.get_target_user_uuid(request)):
+            return None
+        return get_lockdown_target_users().filter(pk=target_uuid).first()
```
BeryJu:
```suggestion
```

not needed with `get_pending_user`

authentik/enterprise/stages/account_lockdown/stage.py
```diff
@@ -0,0 +1,253 @@
...
+            self.logger.info("Account lockdown completed", user=user.username)
+        except Exception as exc:  # noqa: BLE001
+            self.logger.warning("Account lockdown failed", user=user.username, exc=exc)
+            if self_service:
+                return self._self_service_message_response(request, stage, success=False)
+            return self.executor.stage_invalid(ACCOUNT_LOCKDOWN_FAILED_MESSAGE)
+
+        if self_service:
+            if stage.delete_sessions:
+                return self._self_service_completion_response(request)
```
BeryJu, line 317: I was almost tempted to say that we don't need all of this and can just continue the same flow, and defer session deletion to the last stage (in_memory_stage) but yeah people are able to add a bunch of stages after this which might delay session revocation a lot

authentik/enterprise/stages/account_lockdown/api.py
```diff
@@ -0,0 +1,158 @@
...
+                    "AccountLockdownFlowResponse",
+                    {
+                        "flow_url": CharField(
+                            help_text=_(
+                                "URL to redirect to for lockdown flow, including the "
+                                "user_uuid query parameter consumed by the stage."
+                            )
+                        ),
+                    },
+                ),
```
BeryJu: This could be `RedirectChallenge` as a response serializer (maybe unnecessary optimization but I prefer not having 20 serializers that all contain a single URL field. We can put the docs for it on the action instead.

authentik/enterprise/stages/account_lockdown/api.py
```diff
@@ -0,0 +1,158 @@
...
+        returned URL carries the same target into the account lockdown stage
+        via the ``user_uuid`` query parameter.
+        """
+        flow = request._request.brand.flow_lockdown
+        if flow is None:
+            raise ValidationError({"non_field_errors": [_("No lockdown flow configured.")]})
+        querystring = f"?{urlencode({QS_LOCKDOWN_USER: str(user.pk)})}"
+        return request.build_absolute_uri(
+            reverse_lazy("authentik_core:if-flow", kwargs={"flow_slug": flow.slug}) + querystring
+        )
```
BeryJu: should use FlowPlanner to supply the user to the flow plan

BeryJu (changes requested):

authentik/enterprise/stages/account_lockdown/stage.py
```diff
@@ -0,0 +1,253 @@
...
+        if PLAN_CONTEXT_LOCKDOWN_REASON in prompt_data:
+            return prompt_data[PLAN_CONTEXT_LOCKDOWN_REASON]
+        return self.executor.plan.context.get(PLAN_CONTEXT_LOCKDOWN_REASON, "")
+
+    def _apply_lockdown_actions(self, stage: AccountLockdownStage, user: User) -> None:
+        """Apply the configured account changes to the target user."""
+        if stage.deactivate_user:
+            user.is_active = False
+        if stage.set_unusable_password:
+            user.set_unusable_password()
```
BeryJu, line 152: oh yeah we might want to do something else here, not sure if `set_unusable_password` will propagate to LDAP/etc (and that maybe should be configurable?)

authentik/enterprise/stages/account_lockdown/stage.py
```diff
@@ -0,0 +1,253 @@
...
+        """
+        prompt_data = self.executor.plan.context.get(PLAN_CONTEXT_PROMPT, {})
+        if PLAN_CONTEXT_LOCKDOWN_REASON in prompt_data:
+            return prompt_data[PLAN_CONTEXT_LOCKDOWN_REASON]
+        return self.executor.plan.context.get(PLAN_CONTEXT_LOCKDOWN_REASON, "")
+
+    def _apply_lockdown_actions(self, stage: AccountLockdownStage, user: User) -> None:
+        """Apply the configured account changes to the target user."""
+        if stage.deactivate_user:
+            user.is_active = False
```
BeryJu, line 150: Since this might de-provision/deactivate the user in remote systems via SCIM/etc, we might want to trigger those tasks directly here and wait for them

authentik/events/models.py
dominic-r: I think there should be one as lockdown as a whole is a very separate action than just deactivating the user and changing its password, for example. Also, it's a core part of the rest of the feature, which is alerting admins/groups/user when their account is locked down, which is much easily done with an event matcher policy

authentik/stages/prompt/models.py
dominic-r: Addressed

authentik/enterprise/stages/account_lockdown/models.py
dominic-r: Addressed

authentik/enterprise/stages/account_lockdown/stage.py
dominic-r: Addressed

authentik/enterprise/stages/account_lockdown/stage.py
dominic-r: Addressed

authentik/enterprise/stages/account_lockdown/stage.py
dominic-r: Addressed

authentik/enterprise/stages/account_lockdown/stage.py
dominic-r: Addressed

authentik/enterprise/stages/account_lockdown/api.py
dominic-r: Addressed

authentik/enterprise/stages/account_lockdown/api.py
dominic-r: Addressed

authentik/enterprise/stages/account_lockdown/stage.py
dominic-r: I think the behaviour is fine for now. set_unusable_password() only affects the local authentik/django/thing hash. Propagation is wired thru User.set_password(raw_password) and emits password_changed with the raw password and then handlers listen to that signal. set_unusable_password() does not provide a password, so it won't send a signal or propagate. instead it stores a hash marker (https://github.com/django/django/blob/61a62be313e395ce1265132bfc99f51476fb3c95/django/contrib/auth/base_user.py#L122).

as to making it configurable, unusable is an authentik state and is not really portable. For external directories, vendor specific implementation would probably be needed (like setting a random password instead of the django method) and that would blow up the scope of this feature. I'd be open to considering it, but in a future PR

authentik/enterprise/stages/account_lockdown/stage.py
dominic-r: Addressed

authentik/enterprise/stages/account_lockdown/api.py
```diff
@@ -0,0 +1,149 @@
...
+            plan = planner.plan(request._request, {PLAN_CONTEXT_PENDING_USER: user})
+        except EmptyFlowException, FlowNonApplicableException:
+            raise ValidationError(
+                {"non_field_errors": [_("Lockdown flow is not applicable.")]}
+            ) from None
+        request.session[SESSION_KEY_HISTORY] = []
+        request.session[SESSION_KEY_PLAN] = plan
+        return request.build_absolute_uri(
+            reverse("authentik_core:if-flow", kwargs={"flow_slug": flow.slug})
+        )
```
BeryJu: `return plan.to_redirect()`

authentik/enterprise/stages/account_lockdown/api.py
```diff
@@ -0,0 +1,149 @@
...
+        """
+        user = body.validated_data.get("user") or request.user
+
+        if not can_lock_user(request.user, user):
+            LOGGER.debug("Permission denied for account lockdown", user=request.user)
+            self.permission_denied(request)
+
+        flow_url = self._create_lockdown_flow_url(request, user)
+        LOGGER.debug("Returning lockdown flow URL", flow_url=flow_url, user=user.username)
+        return Response({"component": "xak-flow-redirect", "to": flow_url})
```
BeryJu: What's the reason for using a RedirectChallenge here? Was this based on a previous comment of mine?, I don't remember tbh

authentik/enterprise/stages/account_lockdown/stage.py
```diff
@@ -0,0 +1,331 @@
...
+                        rel_obj=provider,
+                        time_limit=time_limit,
+                        uid=f"{provider.name}:user:{user.pk}:direct",
+                    )
+                )
+                wait_timeout += time_limit
+
+        if not messages:
+            return
+        group(messages).run().wait(timeout=wait_timeout)
```
BeryJu: I wonder if we can start these tasks in a way that they'll continue to run even if the timeout is too short

authentik/enterprise/stages/account_lockdown/stage.py
```diff
@@ -0,0 +1,331 @@
...
+            deactivate_user=stage.deactivate_user,
+            set_unusable_password=stage.set_unusable_password,
+            delete_sessions=stage.delete_sessions,
+            revoke_tokens=stage.revoke_tokens,
+        )
+
+        try:
+            self._lockdown_user(request, stage, user, reason)
+            self.logger.info("Account lockdown completed", user=user.username)
+        except Exception as exc:  # noqa: BLE001
```
BeryJu, line 309: not a big fan of all these broad catches but I see why they're here

authentik/events/models.py
BeryJu: yeah my thinking was more so in the direction of having both an action group and action ID to be able to group certain things without just having 23952438234 different actions for slightly different things

authentik/policies/event_matcher/migrations/0028_alter_eventmatcherpolicy_action.py
```diff
@@ -0,0 +1,58 @@
...
+                    ("update_available", "Update Available"),
+                    ("export_ready", "Export Ready"),
+                    ("review_initiated", "Review Initiated"),
+                    ("review_overdue", "Review Overdue"),
+                    ("review_attested", "Review Attested"),
+                    ("review_completed", "Review Completed"),
+                    ("custom_", "Custom Prefix"),
+                ],
+                default=None,
+                help_text="Match created events with this action type. When left empty, all action types will be matched.",
```
BeryJu: For both this and the prompt migration I wouldn't bother adding new migrations. These choices are not reflected in the actual DB schema and adding migrations for this just further increases new installs' startup time. I usually edit the previous migration and add this operation to that

blueprints/example/flow-default-account-lockdown.yaml
```diff
@@ -0,0 +1,306 @@
...
+                  "<li><strong>Invalidate your password</strong> - Your password will be set to a random value "
+                  "and cannot be recovered</li>"
+                  "<li><strong>Deactivate your account</strong> - Your account will be disabled</li>"
+                  "<li><strong>Terminate all your sessions</strong> - You will be logged out everywhere</li>"
+                  "<li><strong>Revoke all your tokens</strong> - All your API, app password, recovery, "
+                  "verification, and OAuth2 tokens and grants will be revoked</li>"
+                  "</ul>"
+                  "<p><strong>This action cannot be easily undone.</strong></p>"
+              )
+
```
BeryJu, line 62: I think all of this is something we should rework in the future tbh

web/src/admin/brands/BrandForm.ts
```diff
@@ -285,6 +323,29 @@ export class BrandForm extends ModelForm<Brand, string> {
...
+                        ></ak-flow-search>
+                        <p class="pf-c-form__helper-text">
+                            ${msg(
+                                "Flow used when a user triggers account lockdown (e.g. in case of compromise). Should contain an Account Lockdown stage.",
+                            )}
+                        </p>
+                        ${this.lockdownWarningVisible
+                            ? html`<ak-alert inline>
+                                  ${msg(
+                                      "Recommended: set Flow Authentication to 'Require authentication' for lockdown flows. Other brand edits can become confusing later if this flow is changed away from that setting.",
```
BeryJu: I am confused by this sentence tbh

BeryJu (changes requested):
Couple of small nits and a couple of things I'd like to see fixed otherwise LGTM

blueprints/example/flow-default-account-lockdown.yaml
dominic-r: Agree

website/docs/security/account-lockdown.md
```diff
@@ -105,9 +105,12 @@ Since the user's session is deleted, the stage redirects to a separate unauthent
 Use Notification Rules to alert when lockdowns occur:
 
 1. Navigate to **Customization** > **Policies** and create an **Event Matcher Policy**
-2. Set **Action** to **User Lockdown Triggered**
-3. Navigate to **Events** > **Notification Rules** and create a rule
-4. Bind the Event Matcher Policy to the rule
+2. Set **Action** to **User Write**
+3. Set **Query** to `context.action_id = "account_lockdown"`
```
BeryJu:
```suggestion
3. Set **Query** to `action = "user_write" and context.action_id = "account_lockdown"`
```

BeryJu (approved):

PR #18618, opened 2025-12-04
web: Remove Patternfly 4 base CSS stub

GirlBossRush:
## Details

This PR removes references to Patternfly's base css in our elements. Prior to #17444, it was necessary to insert base styles into components. That PR moved much of the styles into global scope, or into our Lit element `AKBase` class.

#17444 avoided a larger(er) file count change by stubbing any references to the `base.css` import with an empty stylesheet. This PR removes the references and the stub logic from our bundler.


GirlBossRush:
Resolved in #19623

PR #18646, opened 2025-12-06
web/admin: prevent file upload attempt when backend not managed

BeryJu:
uploading with an un-manageable backend results in a 405

This will only show up if you deploy in k8s and don't configure S3 and dont have a PVC

### DIFF
--- a/web/src/admin/files/FileListPage.ts
+++ b/web/src/admin/files/FileListPage.ts
@@ -129,11 +129,11 @@ export class FileListPage extends WithCapabilitiesConfig(TablePage<FileItem>) {
                     ><span
                         >${msg("Configured file backend does not support file management.")}</span
                     >
-                    <span slot="secondary"
-                        >${msg(
+                    <div slot="body">
+                        ${msg(
                             "Please ensure the data folder is mounted or S3 storage is configured.",
-                        )}</span
-                    >
+                        )}
+                    </div>
                     <div slot="primary">
                         <a
                             target="_blank"
--- a/web/src/admin/files/FileListPage.ts
+++ b/web/src/admin/files/FileListPage.ts
@@ -128,13 +128,9 @@ export class FileListPage extends WithCapabilitiesConfig(TablePage<FileItem>) {
         }
         return super.renderEmpty(
             html`<ak-empty-state icon=${this.pageIcon}
-                ><span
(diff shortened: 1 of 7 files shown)


dominic-r (approved):
🚀

web/src/admin/files/FileListPage.ts
```diff
@@ -116,7 +122,36 @@ export class FileListPage extends TablePage<FileItem> {
...
+                            "Please ensure the data folder is mounted or S3 storage is configured.",
+                        )}
+                    </div>
+                    <div slot="primary">
+                        <a
+                            target="_blank"
+                            class="pf-c-button pf-m-primary"
+                            href=${docLink("/install-config/configuration/#storage-settings")}
+                            >${msg("Learn more")}</a
+                        >
```
GirlBossRush:
```suggestion
                        ${msg("Read more about")}&nbsp;
                        <a
                            target="_blank"
                            rel="noopener noreferrer"
                            href=${docLink("/install-config/configuration/#storage-settings")}
                            >${msg("Storage Settings")}</a
                        >.
```

web/src/admin/files/FileListPage.ts
```diff
@@ -116,7 +122,36 @@ export class FileListPage extends TablePage<FileItem> {
         ];
     }
 
-    protected renderObjectCreate(): TemplateResult {
+    protected renderEmpty(inner?: TemplateResult) {
+        if (!this.can(CapabilitiesEnum.CanSaveMedia)) {
```
GirlBossRush: Short circuit flip:

```ts
if (this.can(CapabilitiesEnum.CanSaveMedia)) {
  return super.renderEmpty(inner);
}
```

GirlBossRush (approved):

web/src/admin/files/FileListPage.ts
BeryJu: I prefer keeping it as a single string in the link since that makes translation a lot easier (with regards to where the button is placed in the sentence)

rissson (approved):

rissson:
Should we somehow do the same thing with the data exports? The task will just fail if there is no storage backend configured.

BeryJu:
> Should we somehow do the same thing with the data exports? The task will just fail if there is no storage backend configured.

Added

PR #18649, opened 2025-12-06
web/admin: fix event volume chart not updating with query

BeryJu:


GirlBossRush (approved):

web/src/admin/events/EventVolumeChart.ts
```diff
@@ -20,7 +20,7 @@ export class EventVolumeChart extends EventChart {
 
     @property({ attribute: false })
     set query(value: EventsEventsListRequest | undefined) {
-        if (JSON.stringify(value) !== JSON.stringify(this._query)) return;
+        if (JSON.stringify(value) === JSON.stringify(this._query)) return;
```
GirlBossRush, line 23: This _should_ work as long as the query object is created in a similar way, but it may need something like `deepEquals` if the object keys are applied different order

PR #18670, opened 2025-12-08
sources/oauth/twitter: update fields, scopes and profile URL to match current X API 

### LINKED ISSUE #18466: Improve Twitter/X OAuth provider to support fetching user email with OAuth in the X API v2

Currently, the Twitter/X OAuth source in Authentik does not fetch the user email (because not requested). This prevents the system from populating the `email` field for users authenticating via Twitter/X. Unlike Facebook, which specifies custom fields (e.g., `email`) via the fields query parameter and additional scopes in `get_additional_parameters`, the Twitter provider does not handle it.

[Here](https://devcommunity.x.com/t/announcing-support-for-email-address-retrieval-with-oauth-2-0-in-the-x-api-v2/240555) is the announcing support for email address retrieval with OAuth 2.0 in the X API v2.

> Important: To receive the email from X, the app must have the “Request email from users” permission enabled in the X Developer dashboard. Without this permission, the API will not return the email even if requested.

Ideally, have the possibility for each source to add custom url parameters for the `profile_url`. 
Like today, we have the possibility to add custom scopes for each source.

- Updating the Twitter provider to request the `confirmed_email` field via the profile URL:
`https://api.twitter.com/2/users/me?user.fields=confirmed_email`
- To be able to fetch email, we need to update the OAuth scopes: `tweet.read users.read users.email`.
- Mapping the `confirmed_email` field from the API response to the `email` property in Authentik.

The idea is to populate the `email` field for Twitter/X users and ideally support custom fields in a consistent way across providers.

### Additional context

_No response_

muien5080:
This updates the Twitter OAuth2 source so it aligns with the current X API requirements. The previous implementation did not request the required user.fields and was missing newer scopes, which caused the Twitter login flow to fail.

This PR updates the profile URL to include the correct fields, adds the required scopes, and expands the user property mapping to match the response returned by X.

closes #18466

### DIFF
--- a/authentik/sources/oauth/types/twitter.py
+++ b/authentik/sources/oauth/types/twitter.py
@@ -15,9 +15,16 @@ class TwitterOAuthRedirect(OAuthRedirect):
     """Twitter OAuth2 Redirect"""
 
     def get_additional_parameters(self, source):  # pragma: no cover
+        scopes = ["users.read", "tweet.read"]
+        # If admin has defined custom scopes in the provider UI, merge them
+        configured = getattr(source, "scope", []) or []
+        for s in configured:
+            if s not in scopes:
+                scopes.append(s)
         return {
-            "scope": ["users.read", "tweet.read"],
+            "scope": scopes,
         }
+
 
 
 class TwitterOAuthCallback(OAuthCallback):
@@ -40,13 +47,20 @@ class TwitterType(SourceType):
 
     authorization_url = "https://twitter.com/i/oauth2/authorize"
     access_token_url = "https://api.twitter.com/2/oauth2/token"  # nosec
-    profile_url = "https://api.twitter.com/2/users/me"
+    profile_url = "https://api.twitter.com/2/users/me?user.fields=verified,username,name,profile_image_url,confirmed_email"
+
     pkce = PKCEMethod.S256
 
     def get_base_user_properties(self, info: dict[str, Any], **kwargs) -> dict[str, Any]:
         data = info.get("data", {})
+        email = (
+            data.get("confirmed_email")
+            or data.get("email")
+            or None
+        )
         return {
             "username": data.get("username"),
-            "email": None,
+            "email": email,
             "name": data.get("name"),
         }
+
--- a/authentik/sources/oauth/types/twitter.py
+++ b/authentik/sources/oauth/types/twitter.py
@@ -15,7 +15,7 @@ class TwitterOAuthRedirect(OAuthRedirect):
     """Twitter OAuth2 Redirect"""
 
     def get_additional_parameters(self, source):  # pragma: no cover
-        scopes = ["users.read", "tweet.read"]
+        scopes = ["users.read", "tweet.read","users.email"]
         # If admin has defined custom scopes in the provider UI, merge them
         configured = getattr(source, "scope", []) or []
         for s in configured:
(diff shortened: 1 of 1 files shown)


authentik/sources/oauth/types/twitter.py
```diff
@@ -40,13 +47,20 @@ class TwitterType(SourceType):
 
     authorization_url = "https://twitter.com/i/oauth2/authorize"
     access_token_url = "https://api.twitter.com/2/oauth2/token"  # nosec
-    profile_url = "https://api.twitter.com/2/users/me"
+    profile_url = "https://api.twitter.com/2/users/me?user.fields=verified,username,name,profile_image_url,confirmed_email"
```
sayah-y: According to this [doc](https://docs.x.com/x-api/fundamentals/data-dictionary#user), fields `id`, `name` and `username` are marked as default (always included).
I think it's better to specify them not to depends on the X Api changes. But in this case, `id` is missing.
It will give:
`profile_url = "https://api.twitter.com/2/users/me?user.fields=id,name,username,confirmed_email,profile_image_url,verified"`

authentik/sources/oauth/types/twitter.py
```diff
@@ -40,13 +47,20 @@ class TwitterType(SourceType):
...
+
     pkce = PKCEMethod.S256
 
     def get_base_user_properties(self, info: dict[str, Any], **kwargs) -> dict[str, Any]:
         data = info.get("data", {})
+        email = (
+            data.get("confirmed_email")
+            or data.get("email")
```
sayah-y: I think we can delete this line because this [doc](https://docs.x.com/x-api/fundamentals/data-dictionary#user) tells thats this field is named `confirmed_email`, no need to check for an alternative name.

authentik/sources/oauth/types/twitter.py
```diff
@@ -15,11 +15,18 @@ class TwitterOAuthRedirect(OAuthRedirect):
     """Twitter OAuth2 Redirect"""
 
     def get_additional_parameters(self, source):  # pragma: no cover
+        scopes = ["users.read", "tweet.read"]
```
sayah-y: According to this [doc](https://docs.x.com/fundamentals/authentication/oauth-2-0/authorization-code#scopes), to get the email, we must add the following scope: `users.email`

sayah-y (commented):
Thanks

muien5080:
#18670
The requested changes have been pushed. Let me know if anything else should be updated.

sayah-y (approved):
Thanks for taking into account my feedbacks 👍

authentik/sources/oauth/types/twitter.py
```diff
@@ -15,11 +15,18 @@ class TwitterOAuthRedirect(OAuthRedirect):
     """Twitter OAuth2 Redirect"""
 
     def get_additional_parameters(self, source):  # pragma: no cover
+        scopes = ["users.read", "tweet.read","users.email"]
+        # If admin has defined custom scopes in the provider UI, merge them
+        configured = getattr(source, "scope", []) or []
+        for s in configured:
+            if s not in scopes:
+                scopes.append(s)
```
BeryJu, line 23:
```suggestion
        default_scopes = ["users.read", "tweet.read","users.email"]
        scopes = list(set(*default_scopes, *source.scope))
```

PR #18684, opened 2025-12-08
core: remove superuser check from `Token` list

gergosimonyi:
With this change, users with global `view_token` permissions will be able to see other users' tokens (but not their keys).


rissson (approved):

rissson:
@gergosimonyi do we need to backport this?

gergosimonyi:
@rissson

No, and deliberately so. `2025.12` is the big "groups/roles" changer, while this is for next release which is the big "permissions" changer :)

PR #18686, opened 2025-12-09
core: support hashed password in users API + automated install

### LINKED ISSUE #5471: Allow passing a password hash instead of a plaintext password when bootstrapping akadmin account
When bootstrapping Authentik, the `AUTHENTIK_BOOTSTRAP_PASSWORD` field can be used to set the default password for the `akadmin` account. This requires a password to be stored in plaintext when not using a secrets manager.

When creating an Authentik setup, I should be able to set the environment variable, `AUTHENTIK_BOOTSTRAP_PSASWORD_HASH`, which accepts a password hash that could, for example, be created from the command line.

I have noticed the `AUTHENTIK_BOOTSTRAP_TOKEN` field but this was mentioned to only provide API access and has the same issue of leaving an authentication secret in plaintext.

dominic-r:
Overview: 

Adds support for bootstrapping authentik using pre-hashed passwords. Instead of passing plaintext passwords through environment variables, admins can generate a hash locally and configure authentik with the hash directly. This also extends to runtime as admins can set password hashes on existing users via API or the admin UI as they would for setting. password.

Motivation:

The current AUTHENTIK_BOOTSTRAP_PASSWORD requires plaintext passwords in environment variables which is just bad practice and some environments may require users to be able to set their password without admins knowing it (?).

Closes: https://github.com/goauthentik/authentik/issues/5471


authentik/core/management/commands/hash_password.py
```diff
@@ -0,0 +1,34 @@
...
+        )
+
+    def handle(self, *args, **options):
+        password = options["password"]
+
+        if not password:
+            raise CommandError("Password cannot be empty")
+
+        if len(password) < 1:
+            raise CommandError("Password must be at least 1 character long")
```
dominic-r:
```suggestion
```

authentik/core/management/commands/hash_password.py
```diff
@@ -0,0 +1,32 @@
...
+            help="Password to hash",
+        )
+
+    def handle(self, *args, **options):
+        password = options["password"]
+
+        if not password:
+            raise CommandError("Password cannot be empty")
+
+
```
dominic-r:
```suggestion

```

authentik/core/api/users.py
```diff
@@ -187,35 +188,54 @@ def create(self, validated_data: dict) -> User:
...
-        string then use an unusable password"""
-        if SERIALIZER_CONTEXT_BLUEPRINT in self.context and password:
-            instance.set_password(password)
-            instance.save()
-        if len(instance.password) == 0:
+        string then use an unusable password. Supports both plaintext password and
+        pre-hashed password via password_hash parameter."""
+        if SERIALIZER_CONTEXT_BLUEPRINT in self.context:
+            # Fail if both password and password_hash are set
+            if password and password_hash and password_hash.strip():
```
gergosimonyi:
```suggestion
            if password is not None and password_hash is not None:
```

authentik/core/api/users.py
```diff
@@ -187,35 +188,54 @@ def create(self, validated_data: dict) -> User:
...
+                raise ValidationError("Cannot set both password and password_hash. Use only one.")
```
gergosimonyi:
```suggestion
                raise ValidationError(_("Cannot set both password and password_hash. Use only one."))
```

authentik/core/api/users.py
```diff
@@ -187,35 +188,54 @@ def create(self, validated_data: dict) -> User:
...
+            if password_hash and password_hash.strip():
```
gergosimonyi:
```suggestion
            if password_hash is not None:
```

website/docs/install-config/automated-install.mdx
```diff
@@ -4,9 +4,41 @@ title: Automated install
...
+# Original hash: pbkdf2_sha256$1000000$3emW1FZkrfjCKYpndtaYPQ$OZNNrcCc+cZ+paHV+kGRAOENTg5+Xny/de1/dNTgfg0=
+AUTHENTIK_BOOTSTRAP_PASSWORD_HASH="pbkdf2_sha256$$1000000$$3emW1FZkrfjCKYpndtaYPQ$$OZNNrcCc+cZ+paHV+kGRAOENTg5+Xny/de1/dNTgfg0="
+```
+
+This prevents Docker Compose from interpreting `$` as variable references.
+:::
+
+:::danger
+Do not use the example hashes shown in this documentation. Always generate your own hash using the commands above.
+:::
```
gergosimonyi: Is this needed? Anyone using this hash won't be able to log in, because they won't know the password corresponding to it.

Could also dramatically increase the number of iterations in the example to make sure no attacker will ever brute-force it if that's the issue.

website/docs/install-config/automated-install.mdx
```diff
@@ -4,9 +4,41 @@ title: Automated install
...
+Configure the default password for the `akadmin` user using a pre-hashed password. Only read on the first startup.
+
+To generate a hash, run this command before your initial deployment:
+
+```bash
+docker compose run --rm server hash_password 'your-password'
+```
+
+:::warning Docker Compose
+When using Docker Compose with a `.env` file, escape the `$` characters in the password hash by doubling them (`$$`):
```
gergosimonyi: What shell is this true in? `$$` is the [current process's PID](https://www.gnu.org/software/bash/manual/bash.html#index-_0024) in `bash`.

authentik/core/models.py
```diff
@@ -518,6 +518,18 @@ def set_password(self, raw_password, signal=True, sender=None, request=None):
         self.password_change_date = now()
         return super().set_password(raw_password)
 
+    def set_password_from_hash(self, password_hash: str):
```
gergosimonyi: This should emit signal similar to `set_password` above that can be used, for example, to create an event like `on_password_changed` in `authentik/events/signals.py`.

gergosimonyi (commented):
Setting password by hash should be possible via the API (as the title of this PR led me to believe), not only in blueprints.

This would enable workflows such as a user giving an admin their hash and logging in with the password that was used to generate that hash.

website/docs/install-config/automated-install.mdx
dominic-r: slack

website/docs/install-config/automated-install.mdx
dominic-r: the password was like "dont-use-this" so it should be fine, but I agree after sleeping on it that the notice isn't needed

dominic-r: Well, mostly what you said

authentik/core/api/users.py
dominic-r: is not None would accept white spaces and empty strings:

dominic-r: Take the following test which checks if the hash would get accepted:
```
❯ python3 none.py
Test: None
  Strip check: False (expected: False)
  None check: False

Test: ''
  Strip check: False (expected: False)
  None check: True

Test: '   '
  Strip check: False (expected: False)
  None check: True

Test: '\t'
  Strip check: False (expected: False)
  None check: True

Test: '\n'
  Strip check: False (expected: False)
  None check: True

Test: '  \t\n  '
  Strip check: False (expected: False)
  None check: True

Test: 'pbkdf2_sha256$1000000$abc'
  Strip check: True (expected: True)
  None check: True

Test: '  pbkdf2_sha256$1000000$abc  '
  Strip check: True (expected: True)
  None check: True


/tmp
❯ cat none.py
def test_password_hash_validation():
    test_cases = [
        (None, False),
        ("", False),
        ("   ", False),
        ("\t", False),
        ("\n", False),
        ("  \t\n  ", False),
        ("pbkdf2_sha256$1000000$abc", True),
        ("  pbkdf2_sha256$1000000$abc  ", True),
    ]

    for password_hash, expected in test_cases:
        if password_hash is not None:
            strip_result = bool(password_hash.strip())
        else:
            strip_result = False

        none_result = password_hash is not None

        print(f"Test: {repr(password_hash)}")
        print(f"  Strip check: {strip_result} (expected: {expected})")
        print(f"  None check: {none_result}")
        print()

test_password_hash_validation()

/tmp
❯
```

authentik/core/api/users.py
dominic-r: same as https://github.com/goauthentik/authentik/pull/18686#discussion_r2696323947

authentik/core/api/users.py
gergosimonyi: As far as I can tell, if `''` or `'    '` is passed, it will just result in a `ValueError` from `identify_hasher`, which would then be handled properly. That feels like correct behavior to me, would you not agree?

dominic-r: You're right, but it's a ValueError:
```
ValueError: Unknown password hashing algorithm '     '. Did you specify it in the PASSWORD_HASHERS setting?
ValueError: Unknown password hashing algorithm ''. Did you specify it in the PASSWORD_HASHERS setting?
```

kensternberg-authentik (approved):
Good fix.  Thanks.

authentik/core/models.py
```diff
@@ -518,6 +518,26 @@ def set_password(self, raw_password, signal=True, sender=None, request=None):
...
+        Raises ValueError if the hash format is not recognized.
+        """
+        identify_hasher(password_hash)  # Raises ValueError if invalid
+        if self.pk and signal:
+            from authentik.core.signals import password_changed
+
+            if not sender:
+                sender = self
+            password_changed.send(
+                sender=sender, user=self, password="", request=request
```
gergosimonyi: It seems to me that this would break [this LDAP receiver](https://github.com/goauthentik/authentik/blob/66342972ef7f2c76915e41954fbd96bd4658a308/authentik/sources/ldap/signals.py#L40) and [the analogous Kerberos one](https://github.com/goauthentik/authentik/blob/66342972ef7f2c76915e41954fbd96bd4658a308/authentik/sources/kerberos/signals.py#L20) (and maybe more)?

web/src/admin/users/UserPasswordHashForm.ts
```diff
@@ -0,0 +1,51 @@
...
+    public override async send(data: UserPasswordHashSetRequest): Promise<void> {
+        return new CoreApi(DEFAULT_CONFIG).coreUsersSetPasswordHashCreate({
+            id: this.instancePk || 0,
+            userPasswordHashSetRequest: data,
+        });
+    }
+
+    renderForm(): TemplateResult {
+        return html`
+            <ak-form-element-horizontal label=${msg("Password hash")} required name="passwordHash">
```
BeryJu: Add warning that using a hash will not sync to LDAP/Kerberos

dominic-r: Addressed

cheesegrits:
@dominic-r apologies for the ping.  I was wondering what the timeline might be for this?  I notice you moved it to "nice to have".

Not being able to set a hashed password through the API is a serious pain point for any usage where Authentik is not the source of truth.  For example, we are starting to deploy Authentik in place of OpenLDAP in a number US state and federal agency TAK environments (https://tak.gov).  In these scenarios, user and group management is done elsewhere and sync'ed to Authentik through the API, and Authentik is essentially just used as the LDAP layer for TAK to talk to.

At the moment, the only way to make this fully work is opening up Postgres to the backend management system, so it can write hashed passwords directly to the Authentik users table.  We're obviously not wild about doing that, so it'd be nice if we had some sense of when / if  the /set_password_hash endpoint might become a reality.

dominic-r:
Hey, Thanks for the ping. I'm aiming to get this in to 2025.5 which will release in the beginning~ of May.

cheesegrits:
> Hey, Thanks for the ping. I'm aiming to get this in to 2025.5 which will release in the beginning~ of May.

Thank you!

authentik/core/tests/test_hash_password_command.py
```diff
@@ -0,0 +1,383 @@
+"""Tests for hash_password management command and password_hash serializer functionality"""
```
BeryJu: Why are there 383 lines of tests for a 28 line management command

authentik/core/api/users.py
```diff
@@ -741,6 +781,51 @@ def set_password(self, request: Request, pk: int, body: UserPasswordSetSerialize
             update_session_auth_hash(self.request, user)
         return Response(status=204)
 
+    @permission_required("authentik_core.reset_user_password")
```
BeryJu, line 790: yeah Idk I'm still not a fan of this, having it in a blueprint is fine ig (even though I still think we don't need this at all but I can see the use in a bootstrap blueprint), however with this you circumvent any policies (which is true for the normal password set endpoint too but at least for that we _can_ add checks) and it won't get sync'd to any sources

authentik/core/api/users.py
```diff
@@ -390,6 +420,16 @@ class UserPasswordSetSerializer(PassiveSerializer):
     password = CharField(required=True)
 
 
+class UserPasswordHashSetSerializer(PassiveSerializer):
```
BeryJu, line 430: If we do include this I think we can just use `UserPasswordSetSerializer` instead of a duplicate serializer

BeryJu (changes requested):

authentik/core/tests/test_hash_password_command.py
dominic-r: Yea, I don't need some of these since they are testing some things like password which space which should already be covered by the hashing algo. I don't recall why I did this, but it can probably be reduced quite a bit

authentik/core/api/users.py
cheesegrits: @BeryJu sorry for jumping in here, but I wanted to add my perspective.  I totally understand your misgivings, in the context of the most common use-case for Authentik, where Authentik is the source of truth, and is where the users set their passwords.  But there are folk, like myself, that are using it more as a replacement for things like OpenLDAP, because Authentik is a lot easier to deploy and manage in modern cloud environments.  We manage our users elsewhere, and simply use Authentik as the LDAP layer in the middle.  All our group and user creation and management is handled through the API.

For example, I author the user management systems for a number of US State and Federal TAK service deployments (for emergency services and wildfire management).  We currently use OpenLDAP as the authentication and directory service layer, but I'm in the process of adding Authentik as an option for that layer, and migrating our clients to it.  Everything is working ... except password management.  There are lots of cases where I simply don't have the clear text password available when I'm adding or updating users, but I do have their PBKDF2-SHA256 or bcrypt hash.

Maybe adding an explicit "I know what I'm doing" permission to allow setting hashed passwords through the API would be appropriate, so those of us that do it are explicitly opting in and are aware of the potential issues?

dominic-r: @gergo and I discussed offline keeping this API-only and removing the button from the Admin interface. I'm leaning towards that, but looking into what you have to say @beryju

BeryJu: That's fair enough @cheesegrits, I think adding a separate permission for this is sensible @dominic-r

dominic-r: Perfect!

cheesegrits: > @gergo and I discussed offline keeping this API-only and removing the button from the Admin interface. I'm leaning towards that, but looking into what you have to say @BeryJu

What about things like service accounts?  Kinda need to be able to set passwords in the UI for those.

dominic-r: You would be able to create a token (or hash) for the service account, then bind it to the service account, like you would normally do, but it will stay API-only due to the nature of the feature

authentik/core/tests/test_hash_password_command.py
dominic-r: Fixed

authentik/core/api/users.py
dominic-r: Addressed

authentik/core/api/users.py
dominic-r: Applied

gergosimonyi: Disagree. This is accidental duplication, it should not be DRYed. Serializers are cheap.

authentik/core/api/users.py
gergosimonyi: Disagree with separate permission. This is now API-only functionality, it's not like non-technical admins will go around invoking this endpoint without understanding the gotchas. I see no benefit of a separate permission for this, but I see a drawback: already the number of in-code custom permissions is higher than we'd like and we should aim to limit (and afte #19643 maybe even lower) it.

BeryJu (approved):
I'd still rather not have this at all but I see the use-case, also @gergosimonyi convinced me that we dont need a separate permission and when we move the normal password change to use password policies we can change that to the change_user permission

authentik/core/api/users.py
```diff
@@ -181,47 +186,79 @@ def get_roles_obj(self, instance: User) -> list[RoleSerializer] | None:
...
-        directly setting a password. However should be done via the `set_password`
-        method instead of directly setting it like rest_framework."""
+        """Create a user, with blueprint-only password and permission writes."""
         password = validated_data.pop("password", None)
-        perms_qs = Permission.objects.filter(
-            codename__in=[x.split(".")[1] for x in validated_data.pop("permissions", [])]
-        ).values_list("content_type__app_label", "codename")
-        perms_list = [f"{ct}.{name}" for ct, name in list(perms_qs)]
+        password_hash = validated_data.pop("password_hash", None)
+        permissions = validated_data.pop("permissions", [])
```
gergosimonyi: Putting these in the `if` seems a bit cleaner, but that's subjective.

authentik/core/api/users.py
```diff
@@ -181,47 +186,79 @@ def get_roles_obj(self, instance: User) -> list[RoleSerializer] | None:
...
+        except ValueError as exc:
+            LOGGER.warning("Failed to identify password hash format", exc_info=exc)
+            raise ValidationError(INVALID_PASSWORD_HASH_MESSAGE) from exc
+
+    def _set_password(self, instance: User, password: str | None, password_hash: str | None = None):
+        """Set password from plain text or hash."""
+        if password_hash is not None:
+            instance.set_password_from_hash(password_hash)
+            instance.save()
+            return
```
gergosimonyi: The `return`s aren't needed, but that's subjective.

authentik/core/signals.py
```diff
@@ -22,7 +22,9 @@
 from authentik.flows.apps import RefreshOtherFlowsAfterAuthentication
 from authentik.root.ws.consumer import build_device_group
 
-# Arguments: user: User, password: str
+# Arguments: user: User, password: str | None, password_source: str | None
+# request: HttpRequest | None
```
gergosimonyi: This signal should not change. Instead, another signal should be called (say, `password_hash_changed`) that is listened to by the two places that actually need it.

authentik/core/api/users.py
dominic-r: `Clean blueprint password handling`

authentik/core/api/users.py
dominic-r: `Drop extra returns`

authentik/core/signals.py
dominic-r: `Split password hash signal`

gergosimonyi (approved):

cheesegrits:
@dominic-r just wanted to say a HUGE thank you for all your effort on this.  I just updated to 2026.5.0, tested the new set_password_hash endpoint, and it works perfectly.  I set a bcrypt hash from our user management server, was able to log in to Authentik with the new password, which then got correctly rehashed internally after login to pbkdf2_sha256.

Many thanks, this clears the way for us to move from OpenLDAP to Authentik in production next week.

PR #18727, opened 2025-12-10
web/flow: Fix spurious double submit  on ak-stage-autosubmit

### LINKED ISSUE #18655: Double POST with SAML on Chromium-based browsers
### Describe the bug

Given a SAML application with POST binding, the POST request back to the service provider carrying the SamlResponse is submitted twice on Chromium browsers in rapid succession. We have been able to replicate it with Edge, Chrome and Chromium on most recent versions.

### How to reproduce

1. Set up a SAML Provider with a POST binding
2. Execute authentication flow in a chromium based browser
3. Observe a double POST redirect in Network tab under developer tools

### Expected behavior

A single POST request being sent out

### Screenshots

Image

### Additional context

The root cause seems to be an incorrect use of a lit lifecycle hook `updated` in `ak-stage-autosubmit` which triggers twice under some conditions. 

https://github.com/goauthentik/authentik/blob/a0fe677efd696b26387efbccdaf0702146dec7a7/web/src/flow/stages/autosubmit/AutosubmitStage.ts#L28-L34

Depending on network latency and the DOM interaction that first POST may fully be transmitted before it gets cancelled by the second POST. Our service provider rejects authentication attempts when it sees a SamlResponse being replayed, so whenever the race condition is complete it results in failed authentication attempts sporadically for us.

### Deployment Method

Other (please specify)

### Version

2025.10.2

### Relevant log output

```shell
N/A
```

dminuoso:
## Details

Use correct lifecycle hook for ak-stage-autosubmit.

Fixes #18655 

---

## Checklist

If applicable

### DIFF
--- a/web/src/flow/stages/RedirectStage.ts
+++ b/web/src/flow/stages/RedirectStage.ts
@@ -41,9 +41,12 @@ export class RedirectStage extends BaseStage<RedirectChallenge, FlowChallengeRes
         return new URL(this.challenge.to, document.baseURI).toString();
     }
 
-    firstUpdated(changed: PropertyValues<this>): void {
-        super.firstUpdated(changed);
+    updated(changed: PropertyValues<this>): void {
+        super.updated(changed);
 
+        if (!changed.has("challenge")) {
+            return;
+        }
         if (this.promptUser) {
             document.addEventListener("keydown", (ev) => {
                 if (ev.key === "Enter") {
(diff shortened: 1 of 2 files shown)


BeryJu (approved):

web/src/flow/stages/RedirectStage.ts
dminuoso, line 1: @BeryJu This feels like a brittle alternative and I suspect this might regress in the future.

What’s wrong with firstUpdated? The lit documentation suggests its a reliable hook for exactly this purpose.

BeryJu: @dminuoso not applicable to this stage, but for other stages, there might be two stages of the same type in a row, in which case Lit will cache the element and only update the challenge, and as such will not trigger firstUpdated

PR #18728, opened 2025-12-10
Fix LDAP connection update on uniqueness change

### LINKED ISSUE #18726: LDAP sync fails after changing object_uniqueness_field/ldap_uniq
### Describe the bug

 When an LDAP source’s object_uniqueness_field is remapped (e.g., from objectSid to another attribute) and existing ldap_uniq attributes are
  migrated, the LDAP sync still tries to create new UserLDAPSourceConnection/GroupLDAPSourceConnection rows keyed by the new
  identifier. Existing rows retain the old identifier, so the insert collides with the unique constraint on (user_id, source_id), aborting sync
  with a duplicate-key error. Code paths: authentik/sources/ldap/sync/users.py and authentik/sources/ldap/sync/groups.py filter by identifier
  only, then insert, rather than update-or-create by (source, user)/(source, group).

### How to reproduce

  - Configure an LDAP source with object_uniqueness_field=objectSid (default); sync users/groups.
  - Change object_uniqueness_field to a different LDAP attribute (e.g., uidNumber); migrate existing Authentik users’ ldap_uniq attributes to
    match the new values.
  - Run LDAP sync again.
  - Observe sync failure with duplicate key value violates unique constraint "authentik_core_usersourc_user_id_source_id_ad1f5aa7_uniq" (user)
    or equivalent for groups.

### Expected behavior

After changing object_uniqueness_field and migrating ldap_uniq, LDAP sync should update the existing source connection’s identifier for that
    (user, source)/(group, source) instead of inserting a new row, completing without duplicate-key errors.

### Screenshots

Image

### Additional context

I am getting error log: ```
Failed to create user: duplicate key value violates unique constraint "authentik_core_usersourc_user_id_source_id_ad1f5aa7_uniq"
DETAIL:  Key (user_id, source_id)=(20, 7ae0e1c5-325c-41e3-9466-0d33b7282265) already exists. To merge new user with existing user, set the user's Attribute 'ldap_uniq' to 'S-1-5-21-2236801780-3321160328-3458063896-89160'
```

even tho the ldap_uniq is set exactly to the requested value, that can be seen in the attached screenshot

### Deployment Method

Docker

### Version

2025.8.5

### Relevant log output

```shell

```

skalahonza:
Closes https://github.com/goauthentik/authentik/issues/18726

## Details
  - Update user and group LDAP sync to update_or_create connections keyed by (source, user/group) and refresh identifier when the uniqueness
    value changes.
  - Prevents duplicate (user, source)/(group, source) constraint errors after remapping ldap_uniq/object_uniqueness_field.

---

## Checklist

If applicable


rissson:
cc. @krejcar25

krejcar25:
Just based off a quick read, correct me if it doesn't apply. Current state attempts to look up the user/group mapping with the current `ldap_uniq`, because that represents the identity in the directory. If either the identity or the lookup field changes, the lookup finds nothing and creates a new group. That's fine. Except when I change the `ldap_uniq` in such a way, that the identifiers clash with each other. That means we are trying to map `CN=group_a` identified by `newuniq=x` onto `CN=group_b` identified by `olduniq=x`.
I would be hesitant to believe this is actually intended behaviour. I see `ldap_uniq` in its current implementation as a handle for *import* and *statekeeping* operations, while the process described above would be *migration*?
The proposed change would update the pairing with the original group object in ak, to which roles, permissions, apps, entitlements, all of this is bound to, essentially changing the meaning of the group.

The current implementation of the `ldap_uniq` appearing twice is ... unfortunate ... If I could have until Sunday to let my pot of attributes-less sync finish brewing, we might see then if situation doesn't get simpler. It's 100% up to you if you want to maintain the hopefully-soon-to-be-deprecated lookup. What we discussed is, from my side, single weeks away from being confidently ready, as in, two tops.

skalahonza:
Do you need anything from me regarding this PR?

BeryJu:
For the time being we dont support changing the LDAP uniqueness field and changing it will require a new source. Once the migration way from the `ldap_uniq` field is done this migration might be possible

PR #17535, opened 2025-10-18
files: rework

rissson:
TODO:
- [x] frontend
- [x] authz for file backend
- [x] docs
- [x] used by endpoint works, but frontend doesn't show stuff, idk why

PR #18764, opened 2025-12-11
web: Fix background refreshing too frequently.

GirlBossRush:
## Details

This fixes a bug introduced in the #17535 where background images that differ by their query params trigger a repaint.


rissson:
We should re-evaluate if this is still necessary after https://github.com/goauthentik/authentik/pull/18784

kensternberg-authentik (approved):
Sweet!

PR #18768, opened 2025-12-12
web/admin: fix read-only provider selection for application form

dominic-r:
One of the ways to "bind" an application to a provider is to click "Create" under "Assigned to application" in the provider view. This PR fixes 2 issues:

* The value is now auto-filled, so the user doesn't need to do that anymore
* The value is now read-only, you don't need to change it since it's well for that provider. If that makes sense

### DIFF
--- a/web/src/elements/forms/SearchSelect/ak-search-select-view.ts
+++ b/web/src/elements/forms/SearchSelect/ak-search-select-view.ts
@@ -24,6 +24,7 @@ export interface ISearchSelectView {
     value?: string;
     open: boolean;
     blankable: boolean;
+    readOnly: boolean;
     caseSensitive: boolean;
     name?: string;
     placeholder: string;
@@ -125,6 +126,14 @@ export class SearchSelectView extends AKElement implements ISearchSelectView {
      */
     @property({ type: Boolean })
     public blankable = false;
+
+    /**
+     * Prevents user interaction while showing the current value.
+     *
+     * @attr
+     */
+    @property({ type: Boolean, attribute: "readonly" })
+    public readOnly = false;
 
     /**
      * If not managed, make the matcher case-sensitive during interaction.  If managed,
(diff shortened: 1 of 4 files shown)


web/src/elements/forms/SearchSelect/ak-search-select-view.ts
```diff
@@ -248,6 +257,8 @@ export class SearchSelectView extends AKElement implements ISearchSelectView {
     //#region Event Listeners
 
     #clickListener = (_ev: Event) => {
```
kensternberg-authentik, line 259: This repetition feels... wrong.  _Don't change it_.  I'm just saying that if we need this kind of procedural pass-through, there's something wrong with the architecture, but that's not on you.

kensternberg-authentik: You could do something like:  (This is off the top of my head, understand).  You'd have to remove the `#` and use Typescript's `private` keyword instead, so the name is referencable.  But this might work.  The idea is to remove all the `this.readOnly` there and replace them with a centralized "When the object is read-only, use the `readOnlyListener`, which does nothing, for all events."


```  
        const listeners = ["input", "click", "blur", "searchKeyup", "searchKeydown", "change"];
        const converter = this.readOnly
            ? (name) => [name, this.readOnlyListener]
            : (name) => [name, this[`${name}Listener`]];
        const eventHandlers = Object.fromEntries(listeners.map(converter));
```

And then...

```
                            @input=${eventHandlers.input}
                            @click=${eventHandlers.click}
                            @blur=${eventHandlers.blur}
```


If you were particularly clever, you might find a way to code this so that the listeners array isn't of strings, but must literally conform to the pattern, something like (again, this is off the top of my head):

```
type PrivateListener<T> = {
  [K in keyof T]: K extends `${infer Event}Listener` 
    ? Event
    : never;
}[keyof T];

const listeners: PrivateListener<this>[] = [...]
```

Let me emphasize that I am _not_ recommending you do this now.  :-)

kensternberg-authentik (approved):

PR #18809, opened 2025-12-14
web: fix slug input for capitalization

### LINKED ISSUE #17958: Application + Provider wizard slug naming
**Describe the bug**
If two or more capital letters are entered in an application name, the slug field doesn't pick up any lower case letters entered after those capital letters

**To Reproduce**

1. Go to **Applications > Applications** and click **Create with provider**
2. Enter a **Name** that starts with two or more capital letters followed by lowercase letters (e.g. "OAuth").
3. See that the slug value doesn't update beyond the capital letters

Can also be reproduced via the normal Application creation without provider (**Applications > Applications** and click **Create**)

**Expected behavior**
All letters to be used in slug

**Screenshots**

Image

Image

-   authentik version: [e.g. 2025.10.1]

dominic-r:
Closes: https://github.com/goauthentik/authentik/issues/17958

### DIFF
--- a/web/src/components/ak-slug-input.ts
+++ b/web/src/components/ak-slug-input.ts
@@ -2,13 +2,21 @@ import { HorizontalLightComponent } from "./HorizontalLightComponent.js";
 
 import { bound } from "#elements/decorators/bound";
 
-import { kebabCase } from "change-case";
-
 import { html } from "lit";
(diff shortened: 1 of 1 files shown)


GirlBossRush:
Tagging @kensternberg-authentik as he's had more experience with this element 🥂

PR #18830, opened 2025-12-15
web: add custom message with links for empty data export list

atereshkin:
## Details

This adds the text "To create a data export, navigate to Directory > Users or to Events > Logs." (with hyperlinks to respective pages) to the Data Exports page when there are no data exports instead of the default "No objects found.". 

---

## Checklist

### DIFF
--- a/web/src/admin/events/DataExportListPage.ts
+++ b/web/src/admin/events/DataExportListPage.ts
@@ -91,6 +91,20 @@ export class DataExportListPage extends TablePage<DataExport> {
             </div>
         </dl>`;
     }
+
+    protected renderEmpty(_inner?: TemplateResult): TemplateResult {
+        return super.renderEmpty(
+            html`<ak-empty-state icon=${this.pageIcon}
+                ><span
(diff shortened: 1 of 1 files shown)


kensternberg-authentik (approved):
Not rocket science℠.

web/src/admin/events/DataExportListPage.ts
```diff
@@ -91,6 +91,20 @@ export class DataExportListPage extends TablePage<DataExport> {
...
         </dl>`;
     }
+
+    protected renderEmpty(_inner?: TemplateResult): TemplateResult {
+        return super.renderEmpty(
+            html`<ak-empty-state icon=${this.pageIcon}
+                ><span
+                    >${msg(
+                        html`To create a data export, navigate to
+                            <a href="#/identity/users">Directory > Users</a> or to
```
kensternberg-authentik: Make sure this looks good on all browsers.  I think that '>' needs to be changed to '\&gt;'

PR #18840, opened 2025-12-15
tests/e2e: Add delay and serialized rollback to saml e2e test

melizeche:
## Details

Fix flaky SAML e2e test cleanup by waiting for background tasks and switching Selenium tests to serialized_rollback to avoid flush/TRUNCATE issues


tests/e2e/test_provider_saml.py
```diff
@@ -687,3 +687,5 @@ def test_sp_initiated_implicit_post_buffer(self):
             [self.user.email],
             f"Claim 'upn' mismatch at {self.driver.current_url}: {snippet}",
         )
+        # Wait for background tasks to complete before teardown to avoid TRUNCATE failures
+        sleep(3)
```
BeryJu, line 690: Even if a sleep "fixes" the issue I'd rather find a correct solution for waiting for the tasks so the rollback works cc @rissson

rissson: the thing is, background tasks are run in-line in tests. There's no thread created that could cause something to run in the background
https://github.com/goauthentik/authentik/blob/main/authentik/tasks/test.py#L46

BeryJu: yeah I still can't explain to myself why there are some exclusivelock errors in some of the tests sometimes

tests/e2e/test_provider_saml.py
```diff
@@ -687,3 +687,5 @@ def test_sp_initiated_implicit_post_buffer(self):
             [self.user.email],
             f"Claim 'upn' mismatch at {self.driver.current_url}: {snippet}",
         )
+        # Wait for background tasks to complete before teardown to avoid TRUNCATE failures
```
BeryJu:
```suggestion
```

BeryJu (approved):

PR #18863, opened 2025-12-16
website/docs: adjust RBAC-related details in 2025.12 release notes

gergosimonyi:


website/docs/releases/2025/v2025.12.md
```diff
@@ -18,6 +18,32 @@ To try out the release candidate, replace your Docker image tag with the latest
...
 ## Breaking changes
 
+### RBAC
+
+As a first step to overhaul authentik's access control system, much of how groups and roles work internally is altered in this release. We recommend you check any custom code (e.g. expression policies, property mappings) that deals with group/role memberships or access control.
+
+#### Group name uniqueness
+
+:::warning
+Make sure your group names are unique before attempting the upgrade.
```
tanberry:
```suggestion
Make sure your group names are unique before starting the upgrade.
```

tanberry (approved):
A nit. ;-)

melizeche (approved):

website/docs/releases/2025/v2025.12.md
```diff
@@ -18,6 +18,32 @@ To try out the release candidate, replace your Docker image tag with the latest
...
+
+#### Group name uniqueness
+
+:::warning
+Make sure your group names are unique before starting the upgrade.
+:::
+
+From `2024.6`, authentik enforced group name uniqueness through the API. However, groups created earlier or groups created by non-API mechanisms (e.g. a sync from a Source) may have left groups with duplicate names in your system. With `2025.12`, group name uniqueness will now be enforced on the database-level.
+
+We played with automatically renaming duplicates, but ultimately found it too confusing for admins. Instead, we made the migration fail loudly in case offending groups exist and now require manual renaming.
```
dominic-r, line 33: so the django migration fails loudly? then how do fix the groups cause authentik would be in a unhealthy state with the migration failing and more

dominic-r (approved):

website/docs/releases/2025/v2025.12.md
gergosimonyi: At that point people can still go back to `2025.10` if they really need to (but also they could fix stuff from django shell or right in the database).

dominic-r: we could have a quick doc on that on how to rename groups if not already done somewhere

tanberry: I believe the solution would be to rename any groups that are duplicates.. then run upgrade again.. ? @gergosimonyi ? but it would be good to get confirmation of this and add it to the rel notes here...

gergosimonyi: That is indeed the solution.

PR #18867, opened 2025-12-16
packages: akql

BeryJu:
Fork djangoql
Add support for variables
remove things we don't use
add typing (mostly)
remove code smells

things I want to look into for this:
- autocomplete support for variables
- also vendoring the frontend component


GirlBossRush (approved):

pyproject.toml
```diff
@@ -26,7 +26,6 @@ dependencies = [
   "django-storages[s3]==1.14.6",
   "django-tenants==3.10.1",
   "django==5.2.15",
-  "djangoql==0.19.1",
```
rissson, line 29: you'll need to add akql as a dep here

rissson:
packages/akql also needs to be added in lifecycle/container/Dockerfile

PR #18883, opened 2025-12-16
stages/identification: replace sleep with make_password

BeryJu:
Better implementation than sleeping randomly, this is closer to the actual authentication of a user


gergosimonyi:
I don't get this. `make_password(None)` does not hash, it [returns `'!' + random_string`](https://github.com/django/django/blob/5.2.9/django/contrib/auth/hashers.py#L103-L106). Timing is definitely off.

rissson:
`make_password(make_password(None))`: 0.11178956127166748s
`make_password("test_password")`: 0.11276108241081238s

LGTM

rissson (approved):

PR #18916, opened 2025-12-17
web/a11y: Modals

GirlBossRush:
## Details

This PR implements `<ak-modal>`, a custom element built on the native `<dialog>` element, with support for lazy-loading and screen-reader compatibility.

Given our heavy use of modals in the UI, the new component is fully backwards compatible with existing implementations. To illustrate the migration path, this PR applies `<ak-modal>` to three examples: **About authentik**, **Create Application**, and **Edit Application**.

- **About authentik** extends the base modal class with an updated template. The modal is added to the DOM only when the button is clicked and removed when closed — ideal for infrequently used or content-heavy modals.
- **Create Application** uses a `renderModal` utility function to trigger a modal from an event listener, with the same lazy-loading behaviour.
- **Edit Application** is similar, but also demonstrates passing data to the modal without a bound event listener per table row.

To keep this PR manageable, only these three modals have been migrated. The rest will follow in subsequent PRs. The new component is available for use in new modals immediately.


GirlBossRush:
Closed via #17812

PR #18929, opened 2025-12-18
packages/ak-guardian: cast safely

gergosimonyi:
`RoleObjectPermission.object_pk` is a `VARCHAR`, which needs to be cast to either `uuid` or an integer type before comparison. The resulting query (if the cast was to `uuid`) used include a subquery like

```
      SELECT 
        (X0."object_pk"):: uuid AS "obj_pk" 
      FROM 
        "guardian_roleobjectpermission" X0 
        INNER JOIN "auth_permission" X2 ON (X0."permission_id" = X2."id") 
      WHERE 
        (
          (... all_roles() query ...)
          AND X2."content_type_id" = 759
          AND X2."codename" IN ('view_certificatekeypair')
        )
```

Sometimes, the query planner would run the `:: uuid` cast before the `WHERE` clause, which would result in an error with "integer-in-varchar-form" `object_pk`s like

> ERROR:  invalid input syntax for type uuid: "30"

---

After this PR, that subquery is broken into two, as such:

```
    SELECT 
      (
        "permission_subquery"."object_pk"
      ):: uuid as "object_pk" 
    FROM 
      (
        SELECT 
          "guardian_roleobjectpermission"."object_pk" AS "object_pk" 
        FROM 
          "guardian_roleobjectpermission" 
          INNER JOIN "auth_permission" ON (
            "guardian_roleobjectpermission"."permission_id" = "auth_permission"."id"
          ) 
        WHERE 
            (... all_roles() query ...)
            AND "auth_permission"."content_type_id" = 58 
            AND "auth_permission"."codename" IN ('view_outpost')
          )
      ) "permission_subquery" OFFSET 0
  ) 
```

### DIFF
--- a/packages/ak-guardian/guardian/shortcuts.py
+++ b/packages/ak-guardian/guardian/shortcuts.py
@@ -296,7 +296,9 @@ def get_objects_for_user(  # noqa: PLR0912 PLR0915
     # `perms_queryset` is ran. Otherwise, the query planner may decide to cast every `object_pk`,
     # which breaks (for example) if it tries to cast an integer to a UUID. In such a case, the WHERE
     # of `perms_queryset` will remove any integer.
-    perms_subquery_sql, perms_subquery_params = perms_queryset.query.sql_with_params()
+    perms_subquery_sql, perms_subquery_params = perms_queryset.values_list(
+        pk_field, flat=True
+    ).query.sql_with_params()
     subquery = RawSQL(
         f"""
         SELECT ("permission_subquery"."object_pk")::{cast_type} as "object_pk"
--- a/packages/ak-guardian/guardian/shortcuts.py
+++ b/packages/ak-guardian/guardian/shortcuts.py
@@ -310,10 +310,10 @@ def get_objects_for_user(  # noqa: PLR0912 PLR0915
     ).query.sql_with_params()
     subquery = RawSQL(
         f"""
-        SELECT ("permission_subquery"."object_pk")::{cast_type} as "object_pk"
-        FROM ({perms_subquery_sql}) "permission_subquery"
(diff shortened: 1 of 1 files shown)


packages/ak-guardian/guardian/shortcuts.py
```diff
@@ -296,19 +297,37 @@ def get_objects_for_user(  # noqa: PLR0912 PLR0915
...
+        )
+        pk_field = "cast_object_pk"
 
     return queryset.filter(pk__in=perms_queryset.values_list(pk_field, flat=True))
 
 
-def _handle_pk_field(queryset):
+# See https://www.postgresql.org/docs/16/functions-info.html#FUNCTIONS-INFO-VALIDITY
+class PgInputIsValid(Func):
+    function = "pg_input_is_valid"
```
gergosimonyi: This depends on postgres version above 16.

packages/ak-guardian/guardian/shortcuts.py
```diff
@@ -296,19 +297,38 @@ def get_objects_for_user(  # noqa: PLR0912 PLR0915
...
+class RegexpLike(Func):
+    function = "regexp_like"
+    arity = 3
+    output_field = BooleanField()
+
+
+def _safe_cast(expression: F, *, output_field: Field, default: Value, regexp: str):
+    return Case(
+        When(
+            RegexpLike(expression, Value(regexp), Value("i")),
```
gergosimonyi: Still not in postgres v14 (current authentik minimum), but this is very close to a complete workaround.

BeryJu (approved):

dominic-r (approved):
!!

PR #18934, opened 2025-12-18
web/elements: progress-bar and table loading header

BeryJu:
Turn PFProgress into an element and add intermediate mode
image


kensternberg-authentik (approved):
Lovely.

PR #18941, opened 2025-12-18
core: add prettier failure on duplicate group names

gergosimonyi:
This does not eliminate #18901, but it may be a better way to be confronted by it.


authentik/core/migrations/0056_user_roles.py
```diff
@@ -1,10 +1,40 @@
...
+
+
+# This is just a nicer way to tell people about the new group name uniqueness requirement than them
+# facing a cold, hard `django.db.utils.IntegrityError`.
+def error_on_duplicate_group_name(apps: Apps, schema_editor: BaseDatabaseSchemaEditor):
+    db_alias = schema_editor.connection.alias
+    AuthentikGroup = apps.get_model("authentik_core", "Group")
+
+    duplicates = (
+        AuthentikGroup.objects.values("name")
```
BeryJu:
```suggestion
        AuthentikGroup.objects.using(db_alias).values("name")
```

gergosimonyi: I did this then copied over it 🤦

authentik/core/migrations/0056_user_roles.py
```diff
@@ -1,10 +1,41 @@
...
+        .filter(name__count__gt=1)
+        .order_by("-name__count", "name")
+    )
+    if len(duplicates):
+        for group in duplicates:
+            LOGGER.error(
+                "Group with duplicate name detected", name=group["name"], count=group["name__count"]
+            )
+        raise DuplicateNameError(
+            f"authentik 2025.12 forbids duplicate group names. For a list of duplicate groups, see logging output above. Please rename the offending groups and re-run the migration. For more information, see: https://version-2025-12.goauthentik.io/releases/2025.12/#group-name-uniqueness"
```
melizeche: Suggestion: "Since version 2025.12, authentik forbids duplicate group names..."

lifecycle/system_migrations/to_2025_12_group_duplicate.py
```diff
@@ -0,0 +1,41 @@
...
+            "select * from information_schema.tables where table_name = 'django_migrations';"
+        )
+        if not bool(self.cur.rowcount):
+            # No django_migrations table, no data to check
+            return False
+        # migration that introduces the uniqueness
+        self.cur.execute(
+            "select count(*) from django_migrations where app = 'authentik_core' and name = '0056_user_roles';"
+        )
+        return not bool(self.cur.rowcount)
```
rissson, line 30:
```suggestion
        # migration that introduces the uniqueness
        self.cur.execute(
            "select count(*) from django_migrations where app = 'authentik_core' and name = '0056_user_roles';"
        )
        return not bool(self.cur.rowcount)
```

rissson (changes requested):

lifecycle/system_migrations/to_2025_12_group_duplicate.py
rissson: I'm Tupid

rissson (approved):

PR #18944, opened 2025-12-18
web/maintenance: no unknown tag names

kensternberg-authentik:
web/maintenance: lint pass to fix broken or unrecognized tag names

# What

This code removes two places in the code that referenced obsolete tag names.

In AkWizardFormPage, the case was a tag that was defined but never used. It, in turn, referenced a tag that did not exist.

In AkApplicationWizard’s ProviderChoices, we referenced eight custom components that did not exist and were never defined anywhere in the code. The references to `renderers` were obsolete; despite being defined they were never used. (This lack of use was covered up by lots of `export`s discarding Typescript’s check against unused field.)

- [x] The code has been formatted

# Why

- WizardFormPage references ‘ak-wizard-form’, which does not exist
- No other component imports, inherits, or extends WizardFormPage. It only exists by itself.

``` shell
$ rg 'WizardFormPage'
src/elements/wizard/WizardFormPage.ts
39:export class WizardFormPage extends WizardPage {
```

- The objects referenced here in these renderers do not exist.
- Without them, the priority ordering code becomes much simpler
- No LocalTypeCreate calls are needed; just use the default API TypeCreate types now

    ./src/admin/applications/wizard/steps/ProviderChoices.ts

        Unknown tag <ak-application-wizard-authentication-by-oauth>. Did you mean <ak-application-wizard-application-step>?
        19:  html`<ak-application-wizard-authentication-by-oauth></ak-appl
        no-unknown-tag-name

        Unknown tag <ak-application-wizard-authentication-by-saml-configuration>. Did you mean <ak-application-wizard-application-step>?
        24:  html`<ak-application-wizard-authentication-by-saml-configuration></ak-appl
        no-unknown-tag-name


web/src/admin/applications/wizard/ak-application-wizard-main.ts
```diff
@@ -7,7 +7,7 @@ import "./steps/ak-application-wizard-provider-step.js";
 import "./steps/ak-application-wizard-submit-step.js";
 
 import { applicationWizardProvidersContext } from "./ContextIdentity.js";
-import { providerTypeRenderers } from "./steps/ProviderChoices.js";
+import { providerTypePriority } from "./steps/ProviderChoices.js";
```
kensternberg-authentik: I chose to re-name this object because this is all it encapsulates now: the priority order for our ProviderChoices.

web/src/admin/applications/wizard/ak-application-wizard-main.ts
```diff
@@ -50,21 +64,18 @@ export class AkApplicationWizardMain extends AKElement {
...
-                        ...providerType,
-                        renderer: providerTypeRenderers[providerType.modelName].render,
-                    }))
-                    .sort(
-                        (a, b) =>
-                            providerTypeRenderers[a.modelName].order -
-                            providerTypeRenderers[b.modelName].order,
-                    )
-                    .reverse(),
+            const providerNameToProviderMap = new Map(
```
kensternberg-authentik, line 69: 1. Map the *received* provider names to the providers.
2. In the priority order list, map the list of model names to their providers, creating an ordered list of providers.
3. Assert the priority order list does not have a misspelled model name.

web/src/admin/applications/wizard/ak-application-wizard-main.ts
```diff
@@ -32,6 +32,24 @@ const freshWizardState = (): ApplicationWizardState => ({
     errors: {},
 });
 
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const isTypeCreateArray = (v: any): v is TypeCreate[] => Array.isArray(v) && !v.includes(undefined);
```
GirlBossRush:
```suggestion
const isTypeCreateArray = (v: unknown): v is TypeCreate[] => Array.isArray(v) && !v.includes(undefined);
```

GirlBossRush: Alternatively, as a generic assertion function:

```ts

function assertEveryPresent<T>(
    input: (T | null | undefined)[],
    errorMessage: string,
): asserts input is T[] {
    if (!Array.isArray(input)) {
        console.debug("Input is not an array", input);
        throw TypeError(errorMessage);
    }

    for (const [index, item] of input.entries()) {
        if (item === null || typeof item === "undefined") {
            console.debug(`Item at index ${index} is null or undefined`, input);
            throw new TypeError(errorMessage);
        }
    }
}

...

assertEveryPresent(
    providersInOrder,
    "List includes name for which no provider model was returned.",
);

this.wizardProviderProvider.setValue(providersInOrder);

```

GirlBossRush (approved):

PR #18950, opened 2025-12-18
web/maintenance: no missing element type definitions

kensternberg-authentik:
web/maintenance: lint pass to add missing HTMLElementTagNameMap entries

# What

This code mechanically adds HTMLElementTagNameMap entries to those files that were missing it.

Every entry in the report is in this format:

    ./src/elements/ak-table/stories/ak-select-table.stories.ts

        'ak-select-table-test-sort' has not been registered on HTMLElementTagNameMap
        84:  export class SimpleTableSortTest extends LitElem
        no-missing-element-type-definition

It was trivial to create a Perl script that extracted the file name, the tag name, and the class name, and turn that into a “Open this file and append the HTMLElementTagNameMap definition to the end,” then run `prettier` and `build` to validate that nothing broke.

I also had to hand-edit the JSDoc for `Form`. It is not, by itself, an element. It is an abstract class from which you can derive elements. The `@element` tag there confused lit-analyze, and lit-analyze was correct to call it out.

# Why

These entries help Typescript & Lit-Analyze lint our product, validating that each element is being used correctly and that the types being passed to it are correct.


web/src/elements/forms/Form.ts
```diff
@@ -193,7 +193,7 @@ function reportInvalidFields(
  * produce the actual form, or include the form in-line as a slotted element. Bizarrely, this form
  * will not render at all if it's not actually in the viewport?[2]
  *
- * @element ak-form
+ * @class Form
```
kensternberg-authentik, line 196: The only hand-edit in the place.  This is not an `@element`; it is a `@class` from which elements may be derived, but it implements no tag-name of its own.  Lit-analyze was correct to call this out.

web/src/admin/events/EventMap.ts
```diff
@@ -164,3 +164,10 @@ export class EventMap extends AKElement {
         </div>`;
     }
 }
+
+declare global {
+    interface HTMLElementTagNameMap {
+        "ak-map": Map;
+        "ak-events-map": EventMap;
```
kensternberg-authentik, line 171: I had to write a special handler in my script because there are two components in one file!

BeryJu (approved):

GirlBossRush (approved):

PR #18953, opened 2025-12-18
web/maintenance: lint pass to add missing HTMLElementEventMap entries

kensternberg-authentik:
web/maintenance: lint pass to add missing HTMLElementEventMap entries

# What

For events where we had, through whatever mechanisms we used, abstracted event names enough that Typescript struggled with them, those names have been added to the respective elements’ JSDoc entry, so that Lit-Analyze could look a little harder.

In several places, outdated test harnesses were using old event names.

Finally, for RapiDoc and Openlayers, the event names provided did not have ElementEventMap entries. Since those events will always be listened for, from the contents within the shadowDOM, the Lit guidelines recommend listening for those on a customElement attached to `this`, and built with the constructor. This is no hardship; the listener always travels with the `:host`, so it does not need to be attached or detached, and the event handing logic is unchanged.

However, that change led to *Typescript* now complaining that there was no HTMLElementEventMap entry for those specific events. I have added `ts-expect-error` pragmas in those two places, with the appropriate comment. This seems like the better compromise, as Typescript is more robust.


web/src/admin/events/EventMap.ts
```diff
@@ -82,6 +82,23 @@ export class EventMap extends AKElement {
...
+                    composed: true,
+                    bubbles: true,
+                    detail: {
+                        eventId: eventId,
+                    },
+                }),
+            );
+        });
+    }
+
```
GirlBossRush, line 95: I ran into a similar situation while working on the LocaleContextController. Lit's own types inspired a TypeSafe fix.

- https://github.com/goauthentik/authentik/blob/main/web/src/elements/controllers/LocaleContextController.ts#L150-L178
- https://github.com/lit/lit/blob/main/packages/localize/src/internal/locale-status-event.ts

```suggestion

export const SELECTED_FEATURE_EVENT = "feature-selected";

declare global {
    interface HTMLElementEventMap {
        [SELECTED_FEATURE_EVENT]: CustomEvent<SelectedFeatureEventDetail>;
    }
}

export interface SelectedFeatureEventDetail {
    feature: Feature;
}

#dispatchSelectedFeature = (event: CustomEvent<SelectedFeatureEventDetail>) => {
        const eventId = event.detail.feature.getId();
        this.dispatchEvent(
            new CustomEvent("select-event", {
                composed: true,
                bubbles: true,
                detail: {
                   eventId,
                },
            }),
        );
    };

    public override connectedCallback() {
        super.connectedCallback();
        this.addEventListener(SELECTED_FEATURE_EVENT, this.#dispatchSelectedFeature);
```

web/src/components/stories/ak-page-navbar.stories.ts
```diff
@@ -40,9 +40,13 @@ declare global {
...
+    return html`
+        <story-ak-page-navbar open @ak-page-nav-menu-toggle=${() => {}}> </story-ak-page-navbar>
+    `;
 };
 
 export const PageNavbarWithIcon = () => {
-    return html` <story-ak-page-navbar open @sidebar-toggle=${() => {}}> </story-ak-page-navbar> `;
+    return html`
+        <story-ak-page-navbar open @ak-page-nav-menu-toggle=${() => {}}> </story-ak-page-navbar>
+    `;
```
GirlBossRush, line 51:
```ts
const noop = () => void 0

return html`
        <story-ak-page-navbar open @ak-page-nav-menu-toggle=${noop}> </story-ak-page-navbar>
    `;

File: web/src/standalone/api-browser/index.entrypoint.ts [hunk 3]
```diff
@@ -51,6 +51,23 @@ export class APIBrowser extends WithBrandConfig(Interface) {
...
+                    request: {
+                        headers: Headers;
+                    };
+                }>,
+            ) => {
+                e.detail.request.headers.append(CSRFHeaderName, getCookie("authentik_csrf"));
+            },
+        );
+    }
+
```
GirlBossRush:
```ts

export const BEFORE_TRY_EVENT = "before-try";

declare global {
    interface HTMLElementEventMap {
        [BEFORE_TRY_EVENT]: CustomEvent<BeforeTryEventDetail>;
    }
}

export interface BeforeTryEventDetail {
    request: {
        headers: Headers;
    };
}
```

```suggestion
    #appendCSRFHeader = (event: CustomEvent<BeforeTryEventDetail>) => {
        event.detail.request.headers.append(CSRFHeaderName, getCookie("authentik_csrf"));
    };

    public override connectedCallback(): void {
        super.connectedCallback();

        this.addEventListener(BEFORE_TRY_EVENT, this.#appendCSRFHeader);

        this.#synchronizeTheme();
        createUIThemeEffect(this.#synchronizeTheme);
    }

```

GirlBossRush (approved):
Minor tweaks to a major improvement 🥳

web/src/components/stories/ak-page-navbar.stories.ts
kensternberg-authentik: Not suuuuper concerned with that level of correctness in a story.  :-)

PR #18968, opened 2025-12-19
lib/sync/outgoing: handle deletions even if object does not exist in database

rissson:
## Details

See #18961

---

## Checklist

If applicable

### DIFF
--- a/authentik/enterprise/providers/microsoft_entra/clients/groups.py
+++ b/authentik/enterprise/providers/microsoft_entra/clients/groups.py
@@ -48,18 +48,13 @@ class MicrosoftEntraGroupClient(
         except TypeError as exc:
             raise StopSync(exc, obj) from exc
 
-    def delete(self, obj: Group):
+    def delete(self, identifier: str):
         """Delete group"""
-        microsoft_group = MicrosoftEntraProviderGroup.objects.filter(
-            provider=self.provider, group=obj
-        ).first()
-        if not microsoft_group:
-            self.logger.debug("Group does not exist in Microsoft, skipping")
-            return None
-        with transaction.atomic():
-            if self.provider.group_delete_action == OutgoingSyncDeleteAction.DELETE:
-                self._request(self.client.groups.by_group_id(microsoft_group.microsoft_id).delete())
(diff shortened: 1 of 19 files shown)


authentik/enterprise/providers/microsoft_entra/clients/groups.py
```diff
@@ -48,18 +48,13 @@ def to_schema(self, obj: Group, connection: MicrosoftEntraProviderGroup) -> MSGr
...
 
-    def delete(self, obj: Group):
+    def delete(self, identifier: str):
         """Delete group"""
-        microsoft_group = MicrosoftEntraProviderGroup.objects.filter(
-            provider=self.provider, group=obj
-        ).first()
-        if not microsoft_group:
-            self.logger.debug("Group does not exist in Microsoft, skipping")
-            return None
```
BeryJu, line 58: I am trying to remember if there was a reason this check exists

authentik/enterprise/providers/microsoft_entra/clients/users.py
```diff
@@ -43,28 +43,17 @@ def to_schema(self, obj: User, connection: MicrosoftEntraProviderUser) -> MSUser
...
-                )
-            elif self.provider.user_delete_action == OutgoingSyncDeleteAction.SUSPEND:
-                response = self._request(
-                    self.client.users.by_user_id(microsoft_user.microsoft_id).patch(
-                        MSUser(account_enabled=False)
-                    )
-                )
-            microsoft_user.delete()
-        return response
+        MicrosoftEntraProviderUser.objects.filter(
```
BeryJu, line 48: Do we even needs these deletes here? Cause when a user/group is deleted this should cascade..although I guess we need this to delete from the connection object?

BeryJu (approved):
Added some comments otherwise lgtm

authentik/enterprise/providers/microsoft_entra/clients/users.py
rissson: It might not be required, but it doesn't hurt for now I guess

PR #18969, opened 2025-12-19
website/glossary: improve

dominic-r:
Now passes:

```
authentik/website git:main*
(authentik) ❯ npm run lint-check

> @goauthentik/docs@0.0.0 lint-check
> eslint --max-warnings 0 .
```


website/docusaurus-theme/theme/DocCardList/GlossaryHelper.tsx
```diff
@@ -30,26 +30,11 @@ export interface GlossaryHelperProps {
...
-        [filteredTerms],
-    );
-
-    // Extract all unique tags from terms for the tag filter bar
-    const availableTags = React.useMemo(() => extractAvailableTags(terms), [terms]);
-
-    // Apply filters based on current view mode and user selections
-    useEffect(() => {
+    // Apply filters based on current view mode and user selections using useMemo instead of useEffect
+    const filteredTerms = React.useMemo(() => {
```
GirlBossRush: Import this to avoid `React.` repetition. 
```suggestion
    const filteredTerms = useMemo(() => {
```

website/docusaurus-theme/theme/DocCardList/GlossaryDocCardList.tsx
```diff
@@ -127,14 +127,14 @@ function GlossaryTermCard({ item, termCache }: { item: SidebarDocLike; termCache
                 <div className="card__header">
                     <h3 className="margin-vert--none" aria-label={termName || "Glossary term"}>
                         {termName || "Glossary term"}
-                        {isAuthentikSpecific && (
+                        {isAuthentikSpecific ? (
```
GirlBossRush: Avoid `is` as a prefix for booleans, instead reserving them for predicate functions which can be called e.g. `isSomething()`


```suggestion
                        {authentikSpecific ? (
```

website/docusaurus-theme/theme/DocCardList/GlossaryDocCardList.tsx
```diff
@@ -223,6 +223,7 @@ export default function GlossaryDocCardList({
                 return () => clearTimeout(timeoutId);
             }
         }
+        return undefined;
```
GirlBossRush: Direct use of `undefined` is almost always better conveyed with `null` to avoid an explicit return value being mistaken for an implicit return value.  

```suggestion
        return null;
```

website/docusaurus-theme/theme/DocCardList/GlossaryDocCardList.tsx
```diff
@@ -304,38 +305,37 @@ export default function GlossaryDocCardList({
...
-                            </div>
-                        ))}
-                    </>
-                );
             }
+            // Alphabetical view: group terms by first letter A-Z
+            const termsByAlphabet = groupByFirstLetter(filteredTerms);
+            return (
+                <>
+                    {termsByAlphabet.map(([letter, letterTerms]) => (
```
GirlBossRush: No need for fragments if the return value is a single component
```suggestion
termsByAlphabet.map(([letter, letterTerms]) => (
```

website/docusaurus-theme/theme/DocCardList/GlossaryHelper.tsx
```diff
@@ -30,26 +30,11 @@ export interface GlossaryHelperProps {
  */
 export const GlossaryHelper: React.FC<GlossaryHelperProps> = ({ terms, children, className }) => {
```
GirlBossRush: This file is a bit too busy to make sense of. Please split the components into their own files where possible

GirlBossRush (dismissed):
Needs another pass, also found a type error after running `npm run check-types`
```ts
/**
 * Generates stable React keys from sidebar items, with fallback to index
 */
function getStableKey(item: GlossaryItem | PropSidebarItem, idx: number) {
    if (typeof item === "object" && item !== null) {
        const match = ["docId", "id", "href", "label"].find(
            (name) => hasOwnProperty(item, name) && typeof item[name] === "string",
        );
        return match ? item[match] : idx;
    }
    return idx;
}
```

```ts
Element implicitly has an 'any' type because expression of type 'string' can't be used to index type 'PropSidebarItem | GlossaryItem'.
  No index signature with a parameter of type 'string' was found on type 'PropSidebarItem | GlossaryItem'.ts(7053)
 ```

kensternberg-authentik (approved):

kensternberg-authentik (approved):

PR #18970, opened 2025-12-19
web/maintenance/no unknown attributes (part 1)

kensternberg-authentik:
web/maintenance/no-unknown-attributes-1

# What

This commit is a collection of fixes and adaptations discovered while running lit-analyzer in a stricter role than usual. These fixes are to 9 of the existing issues; there are 16 more that will be addressed in the next two pull requests.

The following issues were uncovered.

- `ak-slug-input` does not take `autocomplete`.
- `ak-wizard-page-type-create` does not take, or use, the `name` attribute. It also has no `value` of its own, so it is not processed as a form object.
- `ak-endpoints-device-access-groups-form` does not take a `pk` attribute. It takes an `.instancePk` property.
- `ak-provider-oauth2-redirect-uri` is only used in one place, and that place uses the term `input-id` for the key. The component was expected `inputId`. Since it is a string and therefore an attribute, kebab-case is the appropriate fix here.
- `input-mode` is not a valid attribute. The attribute is `inputmode`, and the property is `inputMode`. It may not be undefined. If it is defined, the default is `text`. I have fixed this in the attribute and in the two Forms that used it.
- `form-associated-element` had both `name` and `type` as readonly. Since they are native attributes, they can be attributes or they can be readonly. They can’t be both. I have made them read-write.
- `user-source-settings-page` is only used in one place, and that place uses the term `input-id` for the key. The component was expected `inputId`. Since it is a string and therefore an attribute, kebab-case is the appropriate fix here.

These guideposts will be placed on the PR.


web/src/admin/admin-settings/stories/AdminSettingsFooterLinks.stories.ts
```diff
@@ -71,8 +71,5 @@ type Story = StoryObj;
 
 export const Default: Story = {
     render: () =>
-        html` <ak-admin-settings-footer-link
-            id="footer-link"
-            name="the-footer"
```
kensternberg-authentik, line 76: `name` is not used

web/src/admin/applications/ApplicationForm.ts
```diff
@@ -117,7 +117,6 @@ export class ApplicationForm extends WithCapabilitiesConfig(ModelForm<Applicatio
             ></ak-text-input>
             <ak-slug-input
                 name="slug"
-                autocomplete="off"
```
kensternberg-authentik, line 120: `ak-slug-input` does not understand this attribute.

web/src/admin/applications/wizard/steps/ak-application-wizard-provider-choice-step.ts
```diff
@@ -70,7 +70,6 @@ export class ApplicationWizardProviderChoiceStep extends WithLicenseSummary(Appl
                   <form class="pf-c-form pf-m-horizontal">
                       <ak-wizard-page-type-create
                           .types=${this.providerModelsList}
-                          name="selectProviderType"
```
kensternberg-authentik, line 73: `ak-wizard-page-type-create` is not a formAssociated element and does not have a `name` field of any kind.

web/src/admin/endpoints/DeviceAccessGroupsListPage.ts
```diff
@@ -2,6 +2,7 @@ import "#elements/forms/DeleteBulkForm";
 import "#elements/forms/ModalForm";
 import "#admin/endpoints/DeviceAccessGroupForm";
 import "#admin/policies/BoundPoliciesList";
+import "@patternfly/elements/pf-tooltip/pf-tooltip.js";
```
kensternberg-authentik, line 5: This component uses tooltip but did not import it.

web/src/admin/endpoints/DeviceAccessGroupsListPage.ts
```diff
@@ -42,7 +43,7 @@ export class DeviceAccessGroupsListPage extends TablePage<DeviceAccessGroup> {
             html`<ak-forms-modal>
                 <span slot="submit">${msg("Update")}</span>
                 <span slot="header">${msg("Update Group")}</span>
-                <ak-endpoints-device-access-groups-form slot="form" pk=${item.pbmUuid}>
+                <ak-endpoints-device-access-groups-form slot="form" .instancePk=${item.pbmUuid}>
```
kensternberg-authentik, line 46: `ak-endpoints-device-access-groups-form` is a standard `ModelForm`, which takes an `.instancePk` property to support taking both numbers and strings.

web/src/admin/providers/ProviderWizard.ts
```diff
@@ -57,7 +57,6 @@ export class ProviderWizard extends AKElement {
                 .finalHandler=${this.finalHandler}
             >
                 <ak-wizard-page-type-create
-                    name="selectProviderType"
```
kensternberg-authentik, line 60: `ak-wizard-page-type-create` is not a formAssociated element and does not have a `name` field of any kind.

web/src/admin/providers/oauth2/OAuth2ProviderRedirectURI.ts
```diff
@@ -42,6 +42,9 @@ export class OAuth2ProviderRedirectURI extends AkControlElement<RedirectURI> {
     };
 
     @property({ type: String })
+    public name?: string;
```
kensternberg-authentik: On the other hand, to make sense to receiving clients, this component requires a `name` field.

web/src/admin/providers/proxy/ProxyProviderFormForm.ts
```diff
@@ -1,3 +1,5 @@
+import "#components/ak-text-input";
+import "#components/ak-radio-input";
```
kensternberg-authentik, line 2: These were not included in the list of components used by this element.

web/src/admin/providers/proxy/ProxyProviderFormForm.ts
```diff
@@ -88,7 +90,7 @@ function renderProxySettings(provider: Partial<ProxyProvider> = {}, errors: Vali
             name="externalHost"
             label=${msg("External host")}
             input-hint="code"
-            input-mode="url"
+            inputmode="url"
```
kensternberg-authentik, line 93: `inputmode` (attribute) and `inputMode` (property) exist.  `input-mode` does not.

web/src/admin/providers/saml/SAMLProviderFormForm.ts
```diff
@@ -354,7 +353,6 @@ export function renderForm({
                     name="authnContextClassRefMapping"
                 >
                     <ak-search-select
-                        required
```
kensternberg-authentik, line 357: `ak-search-select` does not have a `required` attribute. (Should it?)

web/src/components/ak-text-input.ts
```diff
@@ -27,6 +27,9 @@ export class AkTextInput extends HorizontalLightComponent<string> {
     @property({ type: Boolean, attribute: "readonly" })
     public readOnly: boolean = false;
 
+    @property({ type: String, attribute: "inputmode" })
+    inputMode: string = "text";
```
kensternberg-authentik, line 31: This conforms with the standards in `[generated.dom.d.ts](https://github.com/microsoft/TypeScript/blob/main/src/lib/dom.generated.d.ts#L11471)` and the [WHATWG spec](https://html.spec.whatwg.org/multipage/interaction.html#attr-inputmode).  If specified, it is not optional.  If it is not optional, the default is `text`.

web/src/components/ak-text-input.ts
```diff
@@ -52,7 +55,7 @@ export class AkTextInput extends HorizontalLightComponent<string> {
             spellcheck=${ifPresent(code ? "false" : this.spellcheck)}
             aria-describedby=${this.helpID}
             placeholder=${ifPresent(this.placeholder)}
-            inputmode=${ifPresent(this.inputMode)}
+            inputmode=${this.inputMode}
```
kensternberg-authentik, line 58: Since it is not optional, `ifPresent` was no longer needed.

web/src/elements/forms/form-associated-element.ts
```diff
@@ -25,12 +25,12 @@ export interface FormAssociated extends Pick<
...
      * The name of the input, provided to the form.
      */
-    readonly name: string | null;
+    name: string | null;
 
     /**
      * The type of the input, provided to the form.
      */
-    readonly type: string;
+    type: string;
```
kensternberg-authentik, line 33: These can't be `readonly` and conform to the `formAssociated` standard.

web/src/elements/user/sources/SourceSettings.ts
```diff
@@ -25,7 +25,7 @@ import PFDataList from "@patternfly/patternfly/components/DataList/data-list.css
 export class UserSourceSettingsPage extends AKElement {
     static styles: CSSResult[] = [PFDataList, Styles];
 
-    @property({ type: Number })
+    @property({ type: Number, attribute: "user-id" })
```
kensternberg-authentik, line 28: Another case where the property name is given in camelCase, and the attribute referencing it is kebab-case.  The DOM does that for _itself_, but custom elements must implement that manually, and on their own terms. Polymer did this, but [the Lit team chose not to support it, citing DX concerns](https://github.com/lit/lit-element/issues/29#issuecomment-385569692).

web/src/admin/endpoints/DeviceAccessGroupsListPage.ts
GirlBossRush: Ironically, I also caught this in the modals clean up branch for a different reason. There seems to be some legacy expectation of `pk` as an attribute, but it's far back. 

https://github.com/goauthentik/authentik/pull/18916/files#diff-c5f9f5696f69be403d6f72aff40bcc0535fad0599f4139bee6349f1ceb00e4b2R28

(Writing this note to circle back in #18916)

web/src/admin/providers/oauth2/OAuth2ProviderRedirectURI.ts
GirlBossRush: Nit to align this with `input`'s default behavior of the `name` property being an empty string when no attribute is provided. 

```suggestion
    @property({ type: String, useDefault: true })
    public name = "";
```

web/src/admin/providers/saml/SAMLProviderFormForm.ts
GirlBossRush: I've attempted to refactor `ak-search-select` into a `FormAssociatedElement` on a few occasions but have always abandoned the task after discovering some edge-case. This might be fun after the Patternfly upgrade 😅

web/src/components/ak-text-input.ts
```diff
@@ -27,6 +27,9 @@ export class AkTextInput extends HorizontalLightComponent<string> {
     @property({ type: Boolean, attribute: "readonly" })
     public readOnly: boolean = false;
 
+    @property({ type: String, attribute: "inputmode" })
```
GirlBossRush: (If I understand this behavior correctly)

```suggestion
    @property({ type: String, attribute: "inputmode", useDefault: true })
```

web/src/elements/forms/form-associated-element.ts
```diff
@@ -25,12 +25,12 @@ export interface FormAssociated extends Pick<
     /**
      * The name of the input, provided to the form.
      */
-    readonly name: string | null;
+    name: string | null;
```
GirlBossRush, line 28: (if the `name` attribute/property works as I understand it)
 
```suggestion
    name: string;
```

web/src/elements/forms/form-associated-element.ts
```diff
@@ -112,9 +112,8 @@ export abstract class FormAssociatedElement<
         return this.internals.form;
     }
 
-    public get name() {
-        return this.getAttribute("name");
-    }
+    @property({ type: String, reflect: true })
+    name: string | null = null;
```
GirlBossRush, line 116:
```suggestion
    @property({ type: String, reflect: true, useDefault: true })
    public name = "";
```

GirlBossRush (approved):
Minor notes,

A marked improvement on our elements 🥳

web/src/admin/providers/saml/SAMLProviderFormForm.ts
kensternberg-authentik: Yeah, it's a beast to wrap your head around.  There's also a) we have `<dialog>` now to work around a lot of clipping issues, and b) patternfly-v4-elements has its own variant of `<select>` now that I should study in detail before attempting anything.

web/src/elements/forms/form-associated-element.ts
kensternberg-authentik: Rejecting this one because `useDefault` with an empty name and `reflect: true` isn't viable; that would just paint `name` into the tag, which isn't what we want, it's not a boolean attribute.

web/src/admin/providers/oauth2/OAuth2ProviderRedirectURI.ts
```diff
@@ -42,6 +42,10 @@ export class OAuth2ProviderRedirectURI extends AkControlElement<RedirectURI> {
     };
 
     @property({ type: String })
+    @property({ type: String, useDefault: true })
```
BeryJu: I dont think this should be duplicated

kensternberg-authentik: In the absence of `reflect: True`, `useDefault` doesn't do anything.

kensternberg-authentik: Oh! I'm glad I checked again. You're absolutely right, and I have no idea why all the tools didn't catch that.

PR #19014, opened 2025-12-22
web/maintenance: no unknown attributes part 2

kensternberg-authentik:
web/maintenance/no-unknown-attributes-2

# What

This commit is a collection of fixes and adaptations discovered while running lit-analyzer in a stricter role than usual.

- `src/admin/endpoints/connectors/agent/AgentConnectorSetup.ts`

After talking to @beryju, we determined that these labels aren’t shown and aren’t used.

- `src/admin/admin-overview/AdminOverviewPage.ts`
- `src/admin/admin-overview/DashboardUserPage.ts`
- `src/elements/cards/AggregatePromiseCard.ts`
- `src/elements/cards/stories/AggregatePromiseCard.stories.ts`

The `Promise` version of our card is not used by any client code. The Dashboard pages that were importing it want the vanilla `AggregateCard` instead.

- `./src/flow/stages/identification/IdentificationStage.ts`

Anchors do not have a `name` attribute, I cannot find any code using the name attributes as lookups, nor any CSS that might use the name attributes as guides. `ak-flow-password-input` is always required; the flag is unsupported and unnecessary.

- `./src/flow/stages/password/PasswordStage.ts`

Anchors do not have a `name` attribute, I cannot find any code using the name attributes as lookups, nor any CSS that might use the name attributes as guides. `ak-flow-password-input` is always required; the flag is unsupported and unnecessary.

- `src/user/user-settings/UserSettingsPage.ts`

This change to the `UserSettingsPage`:

``` diff
-                                userId=${ifPresent(currentUser?.pk)}
+                                user-id=${ifPresent(currentUser?.pk)}
```

… corresponds correctly with:

``` typescript
    @property({ type: Number, attribute: "user-id" })
    public userId?: number;
```

I find it odd (and remarkable) that nobody has complained about this yet. I even went so far as to [confirm my understanding](https://codepen.io/kensternberg-authentik/pen/raLNBwO) and, yes:

- when an attribute is truthy, property syntax does not set the field
- when an attribute is deliberately given a kebab-case name, using the camelCase variant does not set the field

However, when the attribute is truthy, attribute names are case-insensitive: ‘user-id’ and ‘User-Id’ in client code would work just fine.

## Note

A large enough number of warnings remain. Some of those are due to `lit-analyzer` not being updated to recognize newly Baseline global DOM properties like `inert` or `popover`. The rest are from RapiDoc and QrCode, which do not supply sufficient documentation or metadata for Lit-anaylzer to read correctly.


web/src/admin/endpoints/connectors/agent/AgentConnectorSetup.ts
```diff
@@ -149,7 +149,6 @@ export class AgentConnectorSetup extends AKElement {
                         <li>
                             <ak-endpoints-agent-connector-config
                                 class="pf-m-secondary"
-                                label=${msg("Windows")}
```
kensternberg-authentik: After talking to @beryju, we determined that these fields aren't used, and there's nothing in `ak-endpoints-agent-connector-config` to read the `label` attribute.

web/src/flow/stages/identification/IdentificationStage.ts
```diff
@@ -367,15 +367,13 @@ export class IdentificationStage extends BaseStage<
         const enrollmentItem = enrollUrl
             ? html`<div class="pf-c-login__main-footer-band-item">
                   ${msg("Need an account?")}
-                  <a name="enroll" href="${enrollUrl}">${msg("Sign up.")}</a>
```
kensternberg-authentik, line 368: Anchors don't take `name` field, and nothing else uses it for targeting or styling.

web/src/user/user-settings/UserSettingsPage.ts
```diff
@@ -165,7 +165,7 @@ export class UserSettingsPage extends WithSession(AKElement) {
                             </div>
                             <ak-user-settings-source
                                 allow-configuration
-                                userId=${ifPresent(currentUser?.pk)}
+                                user-id=${ifPresent(currentUser?.pk)}
```
kensternberg-authentik, line 168: See the comments in the PR intro.  This was... interesting.

GirlBossRush (approved):

melizeche (approved):

PR #19015, opened 2025-12-23
web: Fix stale flow background

dominic-r:
Closes: https://authentiksecurity.slack.com/archives/C07V5ETRFLK/p1766423965418069


kensternberg-authentik (approved):
Yep, "premature optimization is the root of all evil."

PR #19029, opened 2025-12-23
web/elements: hidden secrets not propagating

kensternberg-authentik:
web/bug/hidden-secrets-not-propagating

# What

This commit updates ak-secret-text-input, adding the `name` attribute to all valid input fields and updating the value writer to match those of known-working components, to ensure that either variety of the display is fully and correctly updated with the content of the hidden secret.

# Why

The hidden input field is the one that HorizontalFormElement was expecting to read its value from, but that field never received a `name` because it wasn’t present when the field was first updated.

HorizontalFormElement writes the `name` field to the first `<input>` it finds. That was the “dummy” input field, which has no working value.

Form ignored the input element because the value it read came with an undefined name.

Object-oriented state management sometimes bites.

closes https://github.com/goauthentik/authentik/issues/19023


web/src/components/ak-secret-text-input.ts
```diff
@@ -10,7 +10,7 @@ import { ifDefined } from "lit/directives/if-defined.js";
 
 @customElement("ak-secret-text-input")
 export class AkSecretTextInput extends HorizontalLightComponent<string> {
-    @property({ type: String, reflect: true })
+    @property({ type: String })
```
kensternberg-authentik, line 13: [Values should not be reflected.](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input#:~:text=When%20specified%20in%20the%20HTML%2C%20corresponds%20to%20the%20initial%20value)

web/src/components/ak-secret-text-input.ts
```diff
@@ -32,26 +42,32 @@ export class AkSecretTextInput extends HorizontalLightComponent<string> {
                 data-form-ignore="true"
                 value="**************"
             />
-            <input type="text" value="${ifDefined(this.value)}" ?required=${this.required} hidden />
+            <input
+                type="text"
+                value="${ifDefined(this.value)}"
+                ?required=${this.required}
+                name=${ifDefined(this.name)}
```
kensternberg-authentik, line 49: This is the actual fix.  On `blur`, this becomes the primary field with the name, which is the one `Form` will ultimately read.

web/src/components/ak-secret-text-input.ts
```diff
@@ -19,10 +19,20 @@ export class AkSecretTextInput extends HorizontalLightComponent<string> {
     @property({ type: String })
     public placeholder = "";
 
+    @property({ type: Number, attribute: "maxlength" })
+    public maxLength?: number;
```
kensternberg-authentik, line 23: Artifacts from comparison with `ak-text-input`.  I left them because, who knows, maybe specifying the min and max length of the input, once you've chosen to provide it, would be useful.

web/src/components/ak-secret-text-input.ts
```diff
@@ -19,10 +19,20 @@ export class AkSecretTextInput extends HorizontalLightComponent<string> {
...
+
+    @property({ type: Number, attribute: "minlength" })
+    public minLength?: number;
+
     #onReveal() {
         this.revealed = true;
     }
 
+    #inputListener = (ev: InputEvent) => {
```
kensternberg-authentik, line 32: This is a known-working implementation inputListener, so I'm keeping it.

GirlBossRush (approved):

PR #19033, opened 2025-12-24
blueprints: fix deadlock and task context error in MetaApplyBlueprint

dominic-r:
The change in fa65d4730 introduced two issues when applying blueprints that contain MetaApplyBlueprint entries (like default-brand.yaml):

1. Deadlock: Using apply_blueprint.send().get_result(block=True) sends a task to the worker and blocks waiting. Since the worker is already processing the parent blueprint, it cant process the nested task.
2. Task context error: Calling apply_blueprint() directly fails because the dramatiq actor requires CurrentTask context which doesn't exist when running from management commands or nested calls.

## Details

REPLACE ME

---

## Checklist

If applicable

### DIFF
--- a/authentik/blueprints/v1/meta/apply_blueprint.py
+++ b/authentik/blueprints/v1/meta/apply_blueprint.py
@@ -37,14 +37,20 @@ class ApplyBlueprintMetaSerializer(PassiveSerializer):
         return super().validate(attrs)
 
     def create(self, validated_data: dict) -> MetaResult:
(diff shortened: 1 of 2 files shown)


dominic-r:
following https://authentiksecurity.slack.com/archives/C04UGCE7BNV/p1766452020653469

dominic-r:
ok, confirmed resolution a second time with a reset then immediate check

BeryJu (approved):

PR #19086, opened 2025-12-29
core: fix read replica routing during transactions

dominic-r:
`FailoverRouter` was routing reads to replicas even inside active transactions. This caused the application wizard to fail with "Invalid pk - object does not exist" when using read replicas (mabye other things too, have not checked), as FK validation queries couldn't see uncommitted rows on the primary


dominic-r:
Can confirm resolution by testing on an instance which previously failed

gergosimonyi (commented):
@BeryJu I think this is the correct approach, but I'll leave the checkmark up to you.

BeryJu (approved):
LGTM, cc @rissson

Issue #9584, opened 2024-05-06
Initial setup link not working


melvinotieno:
**Describe the bug**

I am using the instructions on this link https://docs.goauthentik.io/docs/installation/docker-compose locally but when visiting the link http://localhost:9000//if/flow/initial-setup/ I get the error not found

**To Reproduce**

Followed the instructions exactly as indicated in the installation page using docker-compose without any modification. 

**Expected behavior**

Expect to see a setup page

**Screenshots**

Docker compose file as downloaded

![image](https://github.com/goauthentik/authentik/assets/46104684/3522c5df-d2da-4330-80a9-2374e430567e)

Screenshot of the setup link not working

![image](https://github.com/goauthentik/authentik/assets/46104684/24f662a1-d28e-4f8c-b936-287404be4902)

**Logs**
Output of docker-compose logs or kubectl logs respectively

**Version and Deployment (please complete the following information):**

-   authentik version: 2024.4.1
-   Deployment: docker-compose

**Additional context**

taoyx:
I also came into this issue with the latest release on Kubernetes (installed with helm)

Also, even I use this command:

kubectl exec -it deployment/authentik-worker  -- ak create_recovery_key 10 akadmin

and gained the accesss to authentik, I cannot add application and provider. 

and either worker and server pod don't report a error:(refer attached file)

The UI of administration area also looks strange

 it was reported worker is NOT connected, even though from kubernets, the worker pod is running properly. 
![image](https://github.com/goauthentik/authentik/assets/23740365/ebe4fec0-2e09-44fe-a77c-84c19378ebc9)

![image](https://github.com/goauthentik/authentik/assets/23740365/488af929-1eaa-43a8-9ef3-9de7650bad8b)

[server.log](https://github.com/goauthentik/authentik/files/15224508/server.log)
[worker.log](https://github.com/goauthentik/authentik/files/15224509/worker.log)

taoyx:
the issue about worker does NOT always occur:

![image](https://github.com/goauthentik/authentik/assets/23740365/a4ba0894-d1f7-4361-9618-0b67046734dc)

but I did encount a error when resetting akadmin password:

```
tyx@prod-k8s-master0:~$ k exec authentik-worker-6f98f5c69-dqsd4 -- ak create_recovery_key 10 akadmin
{"event": "Loaded config", "level": "debug", "logger": "authentik.lib.config", "timestamp": 1715018737.915451, "file": "/authentik/lib/default.yml"}
[...]

PR #6928, opened 2023-09-18
root: disable APPEND_SLASH

BeryJu:
## Details

Currently, with APPEND_SLASH enabled, requests to `core/users` will be redirected to `core/users/`. While this is fine for GET requests, with POST requests the redirect causes the request payload to get lost and the method turns into a GET, which causes a lot of hard-to-find errors. With this setting disabled, requests to `core/users` will simply return a 404.

---

## Checklist

-   [x] Local tests pass (`ak test authentik/`)
-   [x] The code has been formatted (`make lint-fix`)

If an API change has been made

-   [ ] The API schema has been updated (`make gen-build`)

If changes to the frontend have been made

-   [ ] The code has been formatted (`make web`)
-   [ ] The translation files have been updated (`make i18n-extract`)

If applicable

-   [ ] The documentation has been updated
-   [ ] The documentation has been formatted (`make website`)

PR #19091, opened 2025-12-29
core: make trailing slash optional for flow URLs

timothywashburn:
## Details

The required trailing slash is likely inconsequential in the majority of cases, however specifically when bootstrapping the project and using `/if/flow/initialsetup/` it seems there are many people including myself (see [this comment](https://github.com/goauthentik/authentik/issues/9584#issuecomment-2122152762)) that have mistakenly not included the trailing slash and are not used to the URLs `/if/flow/initialsetup` and `/if/flow/initialsetup/` being handled differently.

This change is made by swapping path() to re_path() to regex the pattern with and without the trailing slash.

It is worth mentioning that this is essentially the default behavior with APPEND_SLASH=True but this was previously disabled for other reasons (#6928).

---

## Checklist

### DIFF
--- a/authentik/core/urls.py
+++ b/authentik/core/urls.py
@@ -2,7 +2,7 @@
 
 from django.conf import settings
 from django.contrib.auth.decorators import login_required
-from django.urls import path
+from django.urls import path, re_path
 
 from authentik.core.api.application_entitlements import ApplicationEntitlementViewSet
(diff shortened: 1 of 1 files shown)


BeryJu:
We're looking into making it possible to use both URLs with and without trailing slashes while authentik continue to generate URLs with trailing slashes.

Regardless I dont want to add a special case just for the flow executor.

PR #19114, opened 2025-12-30
web: Fix Impersonation, Lit Reactive Controller Contexts

GirlBossRush:
## Details

This fixes several lifecycle issues within our Lit Controller Contexts that prevented consistent refresh behaviors:

- Ported the near duplicate lifecycle code from session, license, version, and config contexts into `ReactiveContextController`, fixing some logical disparities as each controller has matured
- Fixed race condition where rapid refresh events can trigger an abort controller with a string error, rather than an `Error` instance 
- Fixed race condition where references to abort controllers can become stale 
- Fixed issue where session mixin does not `requestUpdate` after impersonation
- Added abort controllers on `hostDisconnected` callbacks, fixing momentary errors logged during page reloads while impersonating.


kensternberg-authentik (approved):
Great work, seeing the common features and pulling them out.

PR #19121, opened 2025-12-30
web: Token Form Fixes

GirlBossRush:
## Details

This PR fixes a few issues discovered while testing the creation of authentication tokens for users, device enrollment, and similar interactions on forms with expiration fields.

- Fixed issue where autofocus is not consistently applied when creating a token or service user
- Fixed an issue where labels are misaligned with their sibling fields
- Fixed expiration date time fields not consistently disabling when expiration is toggled
- Fixed issue where editing a token may not show itself in a search select element if the paginated response does not include it
- Fixed field/label association for screen readers

### DIFF
--- a/web/src/admin/endpoints/connectors/agent/AgentConnectorForm.ts
+++ b/web/src/admin/endpoints/connectors/agent/AgentConnectorForm.ts
@@ -61,9 +61,11 @@ export class AgentConnectorForm extends WithBrandConfig(ModelForm<AgentConnector
     renderForm() {
         return html`<ak-text-input
                 name="name"
-                placeholder=${msg("Connector name...")}
+                placeholder=${msg("Type a connector name...")}
                 label=${msg("Connector name")}
                 value=${ifDefined(this.instance?.name)}
+                input-hint="code"
+                autofocus
                 required
             ></ak-text-input>
             <ak-text-input
--- a/web/src/admin/endpoints/connectors/agent/EnrollmentTokenForm.ts
+++ b/web/src/admin/endpoints/connectors/agent/EnrollmentTokenForm.ts
@@ -24,10 +24,10 @@ const EXPIRATION_DURATION = 30 * 60 * 1000; // 30 minutes
 
 @customElement("ak-endpoints-agent-enrollment-token-form")
 export class EnrollmentTokenForm extends WithBrandConfig(ModelForm<EnrollmentToken, string>) {
-    protected expirationMinimumDate: string = dateTimeLocal();
+    protected expirationMinimumDate = new Date();
 
     @state()
-    protected expiresAt: string | null = dateTimeLocal(Date.now() + EXPIRATION_DURATION);
+    protected expiresAt: Date | null = new Date(Date.now() + EXPIRATION_DURATION);
 
     @property({ type: String, attribute: "connector-id" })
     public connectorID?: string;
@@ -40,7 +40,7 @@ export class EnrollmentTokenForm extends WithBrandConfig(ModelForm<EnrollmentTok
         });
 
         this.expiresAt = token.expiring
-            ? dateTimeLocal(token.expires || Date.now() + EXPIRATION_DURATION)
+            ? new Date(token.expires || Date.now() + EXPIRATION_DURATION)
             : null;
 
         return token;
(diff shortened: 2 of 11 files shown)


web/src/admin/endpoints/connectors/agent/AgentConnectorForm.ts
```diff
@@ -61,9 +61,11 @@ export class AgentConnectorForm extends WithBrandConfig(ModelForm<AgentConnector
     renderForm() {
         return html`<ak-text-input
                 name="name"
-                placeholder=${msg("Connector name...")}
+                placeholder=${msg("Type a connector name...")}
```
kensternberg-authentik, line 64: "Enter", perhaps?

web/src/admin/tokens/TokenForm.ts
```diff
@@ -120,39 +150,46 @@ export class TokenForm extends ModelForm<Token, string> {
...
+                    {
+                        slot: "label",
+                        className: "pf-c-form__group-label",
+                        htmlFor: "expiration-date-input",
+                    },
+                    msg("Expires on"),
+                )}
+
                 <input
-                    type="text"
```
kensternberg-authentik, line 125: What shortcomings does `ak-text-input` have that makes it a poor choice here? (Not that I want you to use it if it's not valid, but I want to understand what I need to make it better.)

web/src/admin/users/ServiceAccountForm.ts
```diff
@@ -21,20 +23,17 @@ import {
 
 import { msg, str } from "@lit/localize";
 import { html, TemplateResult } from "lit";
-import { customElement, property } from "lit/decorators.js";
+import { customElement, property, state } from "lit/decorators.js";
 import { ifDefined } from "lit/directives/if-defined.js";
-import { createRef, ref } from "lit/directives/ref.js";
+
+const createExpirationValue = () => {
```
kensternberg-authentik: Maybe a utility?  Or 
`createExpirationValue = () => dateTimeLocal(datefns.addYears(Date.now(), 1))`

We already have datefns because chartjs needs it, and it's just so damn better than anything we had before Temporal.

kensternberg-authentik (approved):

web/src/admin/endpoints/connectors/agent/EnrollmentTokenForm.ts
```diff
@@ -10,28 +11,38 @@ import { dateTimeLocal } from "#common/temporal";
...
 import { customElement, property, state } from "lit/decorators.js";
 import { ifDefined } from "lit/directives/if-defined.js";
 
+const EXPIRATION_DURATION = 30 * 60 * 1000; // 30 minutes
+
 @customElement("ak-endpoints-agent-enrollment-token-form")
 export class EnrollmentTokenForm extends WithBrandConfig(ModelForm<EnrollmentToken, string>) {
-    @property()
-    connectorID?: string;
+    protected expirationMinimumDate: string = dateTimeLocal();
```
BeryJu: Any reason to not store this as a Date instance?

web/src/admin/tokens/TokenForm.ts
```diff
@@ -2,28 +2,40 @@ import "#elements/forms/FormGroup";
...
 
+const EXPIRATION_DURATION = 30 * 60 * 1000; // 30 minutes
+
 @customElement("ak-token-form")
 export class TokenForm extends ModelForm<Token, string> {
+    protected expirationMinimumDate: string = dateTimeLocal();
+
     @state()
-    showExpiry = true;
+    protected expiresAt: string | null = dateTimeLocal(Date.now() + EXPIRATION_DURATION);
```
BeryJu: Same as above

web/src/admin/users/ServiceAccountForm.ts
```diff
@@ -21,20 +23,15 @@ import {
...
-    #initialExpirationValue = dateTimeLocal(new Date(Date.now() + 1000 * 60 ** 2 * 24 * 360));
-
-    //#region Refs
-
-    #expiringInputRef = createRef<HTMLInputElement>();
-    #expirationDateInputRef = createRef<HTMLInputElement>();
-
-    //#endregion
+    @state()
+    protected expiresAt: string | null = dateTimeLocal(Date.now() + EXPIRATION_DURATION);
```
BeryJu: Same as above

web/src/admin/tokens/TokenForm.ts
GirlBossRush: One part motivation to reduce the changes in this form, another wanting to avoid adding numeric properties (e.g. `min`, `max`, `step`, etc).

Additionally, I'd like to revisit this as a text input to represent a _duration_ rather than a _timestamp_, e.g. `<ak-text-input value="minutes=30"/>`

PR #19141, opened 2026-01-01
web: Merge branch -- Stale notifications, synchronized context objects, rendering fixes

GirlBossRush:
## Details

This PR fixes several issues that contribute to stale, stalled, or otherwise incorrect notification counts in the page header.

The underlying issue stems from how we consume Lit's context objects for mutable values. At first glance it may appear as if the value of a context can be considered consistent across any component who applies a `With*` mixin. This is true...unless the data changes.

### How does this relate to notifications?

Previously, notifications depended on frequent checks for the `ak-refresh` event to get a clue as to when to fetch new data, both too often and too infrequently to maintain timely updates. This has been fixed through a combination of context-ification and adding lifecycle checks to coordinate the user's session context with a request for new notifications. A similar approach is used to align the session context with other authenticated contexts such as locale, version info, and license information. 

### Why hasn't this been an issue for the session context?

Session data has been more or less immutable in our approach of always applying page reloads after modifying its value. Additionally, we've quietly accepted some repetitive setting of the `session` property on some unusual situations, such as the AdminInterface element which both consumes and provides the same data. Because notification fetches depend on the session context, some changes are needed to synchronize their values. 

The end result for sessions is the same with fewer re-renders. 

## Dependencies

This PR is merge branch and includes several fixes composed to make notifications work consistently.

- #19152
- #19184
- #19185

## Fixes

- User library UI elements now layer as expected, such as in drawer content, modals, page headers, progress bars, and tooltips.
  - Combined `ak-interface-user` and `ak-interface-user-presentation`, fixing multiple session related re-renders and ARIA focus target headaches.
  - CSS parts of library elements are now exported to better expose custom CSS
  - API and notification drawers no longer overlap each other when both open.
- Several ARIA fixes for user library and user settings:
  - Table progress bars now include additional ARIA attributes and more accurately reflect the current visibility.
  - Added or revised ARIA labels for navigation header buttons, notification drawer, API drawer, user settings.
  - Color customizations more consistently use PatternFly's CSS variables over `!important` overrides.
- Better logging of user-driven events:
  - i.e. less duplication of `console.debug.bind(console, "some-component"), more `ConsoleLogger.prefix("some-component")`
- Fix issues surrounding global event dispatching:
  - Our pattern of `connectedCallback`/`disconnectedCallback` to manage events is now handled by our `@listen(Event)` decorator.
  - Event listeners on `window` are `passive: true` by default, fixing some UI lockups when notifications and API requests are frequently updating the page.
  - The Lit locale context now correctly dispatches its event to the CAPTCHA stage, fixing a race condition that often results in a mismatched language.


web/src/admin/AdminInterface/index.entrypoint.css
```diff
@@ -44,4 +44,12 @@ ak-sidebar-item:active ak-sidebar-item::part(list-item) {
 
 .pf-c-drawer__panel {
     z-index: var(--pf-global--ZIndex--xl);
+    flex-flow: row;
```
GirlBossRush: Small fix show the API and notification drawers simultaneously, instead of directly on top of each other.

web/src/common/api/middleware.ts
```diff
@@ -23,20 +23,23 @@ export interface RequestInfo {
 }
 
 export class LoggingMiddleware implements Middleware {
-    brand: CurrentBrand;
+    #logPrefix: string;
+
     constructor(brand: CurrentBrand) {
-        this.brand = brand;
+        this.#logPrefix = `%c[api/${brand.matchedDomain}]: `;
```
GirlBossRush: Log declutter while debugging the order of API calls

web/src/common/constants.ts
```diff
@@ -33,24 +33,14 @@ export const EVENT_REFRESH = "ak-refresh";
 export const EVENT_NOTIFICATION_DRAWER_TOGGLE = "ak-notification-toggle";
 export const EVENT_API_DRAWER_TOGGLE = "ak-api-drawer-toggle";
 export const EVENT_FLOW_INSPECTOR_TOGGLE = "ak-flow-inspector-toggle";
-export const EVENT_WS_MESSAGE = "ak-ws-message";
 export const EVENT_FLOW_ADVANCE = "ak-flow-advance";
 export const EVENT_LOCALE_REQUEST = "ak-locale-request";
 export const EVENT_REQUEST_POST = "ak-request-post";
 export const EVENT_MESSAGE = "ak-message";
-export const EVENT_THEME_CHANGE = "ak-theme-change";
```
GirlBossRush, line 41: Vestigial

web/src/common/logger.ts
```diff
@@ -0,0 +1,20 @@
...
+ */
+
+/**
+ * Create a logger function for a mixin.
+ *
+ * @param prefix The prefix to use for the logger.
+ * @param className The class name of the mixin.
+ * @returns A logger function.
+ */
+export function createDebugLogger(
```
GirlBossRush: I’d like to do a proper logging clean up in a follow up PR, using our Pino based logger which has some nice formatting and environment-specific log level features. 

In the short term, this function covers a lot of the duplicate console.log prefixing pattern.

web/src/common/messages.ts
```diff
@@ -1,6 +1,36 @@
+import type { TemplateResult } from "lit";
+
 export enum MessageLevel {
     error = "error",
     warning = "warning",
     success = "success",
     info = "info",
 }
+
```
GirlBossRush, line 9: Moved from Message component to avoid circular dependencies.

web/src/flow/FlowExecutor.ts
```diff
@@ -160,28 +157,33 @@ export class FlowExecutor
...
+
+        console.debug("authentik/ws: Reloading after session authenticated event");
+        window.location.reload();
     };
 
     public connectedCallback(): void {
         super.connectedCallback();
 
         window.addEventListener(EVENT_FLOW_INSPECTOR_TOGGLE, this.#toggleInspector);
-        window.addEventListener(EVENT_WS_MESSAGE, this.#websocketHandler as EventListener);
```
kensternberg-authentik, line 177: This looks like a good-place for an abort controller!

kensternberg-authentik (approved):
Better abstractions than the ones I came up with, at any rate...

web/logger/browser.js
```diff
@@ -0,0 +1,139 @@
+/**
```
GirlBossRush, line 1: See #19185

web/src/flow/stages/captcha/CaptchaStage.ts
```diff
@@ -245,15 +249,27 @@ export class CaptchaStage extends BaseStage<CaptchaChallenge, CaptchaChallengeRe
...
+     * Renders the Turnstile captcha frame.
+     *
+     * @remarks
+     *
+     * Turnstile will log a warning if the `data-language` attribute
+     * is not in lower-case format.
+     *
+     * @see {@link https://developers.cloudflare.com/turnstile/reference/supported-languages/ Turnstile Supported Languages}
+     */
     protected renderTurnstileFrame = () => {
```
GirlBossRush, line 262: See #19184

gergosimonyi (approved):

PR #19184, opened 2026-01-05
web: Fix CAPTCHA locale.

GirlBossRush:
## Details

This PR fixes a race condition that often results in a mismatched language in the CAPTCHA stage.

## Dependents

- #19141


GirlBossRush:
Resolved via #19141

PR #19185, opened 2026-01-05
web: Clean up prefixed debugging logs

GirlBossRush:
## Details

**tl;dr Fewer `console.debug.bind` and copy-pasted log prefix strings.**

This PR updates our logging approach in the frontend codebase by replacing multiple instances of `console.debug.bind(console, "some-component")` with an instance of browser-compatible `ConsoleLogger` static class. Similar to our Pino-based logger in Node.js, this class allows us to create prefixed loggers for different components without the ceremony of function binding.

The motivation behind this came about while debugging Lit's context event system. The goal was to have consistent logging across the codebase that can be easily toggled or filtered in the same API style as our Node.js logging. The browser `ConsoleLogger` is a step in that direction, providing a familiar interface for logging in the frontend.

## Dependents

- #19141


GirlBossRush:
Resolved via #19141

PR #19189, opened 2026-01-05
docs: fix missing command in `troubleshoot/login`

oliverrahner:
## Details

correct small docs mistake


oliverrahner:
Sorry, didn't think enough when making this change :-/

PR #19194, opened 2026-01-05
web: Defer table refresh, visibility checks.

GirlBossRush:
## Details

This PR defers table `fetch` calls after the element becomes visible, fixing excessive API requests on tabbed layouts.

### DIFF
--- a/web/src/elements/ak-progress-bar.ts
+++ b/web/src/elements/ak-progress-bar.ts
@@ -64,19 +64,19 @@ export class ProgressBar extends AKElement {
         `,
     ];
 
-    @property({ type: Number })
+    @property({ type: Number, reflect: true, useDefault: true })
     public min = 0;
 
-    @property({ type: Number })
+    @property({ type: Number, reflect: true, useDefault: true })
     public max = 100;
 
-    @property({ type: Number })
+    @property({ type: Number, reflect: true, useDefault: true })
     public value = 0;
 
-    @property({ type: Boolean })
+    @property({ type: Boolean, reflect: true, useDefault: true })
     public indeterminate = false;
 
-    @property({ type: String })
+    @property({ type: String, reflect: true, useDefault: true })
     public size: PFSize = PFSize.Medium;
 
     @property({ type: String })
@@ -105,7 +105,7 @@ export class ProgressBar extends AKElement {
                   `
                 : nothing}
             <div
-                class="pf-c-progress__bar"
+                class="pf-c-progress__bar ak-fade-in"
                 role="progressbar"
                 aria-valuemin=${this.min}
                 aria-valuemax=${this.max}
--- a/web/src/elements/table/Table.css
+++ b/web/src/elements/table/Table.css
@@ -146,6 +146,7 @@ time {
 }
 
 :host {
+    display: block;
     container-type: inline-size;
 }
 
--- a/web/src/elements/table/Table.ts
+++ b/web/src/elements/table/Table.ts
@@ -15,6 +15,7 @@ import { APIError, parseAPIResponseError, pluckErrorDetail } from "#common/error
 import { GroupResult } from "#common/utils";
 
 import { AKElement } from "#elements/Base";
+import { intersectionObserver } from "#elements/decorators/intersection-observer";
 import { WithLicenseSummary } from "#elements/mixins/license";
 import { WithSession } from "#elements/mixins/session";
 import { getURLParam, updateURLParams } from "#elements/router/RouteMatch";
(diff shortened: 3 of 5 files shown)


web/src/elements/ak-progress-bar.ts
```diff
@@ -64,19 +64,19 @@ export class ProgressBar extends AKElement {
...
-    @property({ type: Number })
+    @property({ type: Number, useDefault: true })
     public value = 0;
 
-    @property({ type: Boolean })
+    @property({ type: Boolean, useDefault: true })
     public indeterminate = false;
 
-    @property({ type: String })
+    @property({ type: String, useDefault: true })
```
GirlBossRush: Small fix to avoid a double render

web/src/elements/table/Table.css
```diff
@@ -146,6 +146,7 @@ time {
 }
 
 :host {
+    display: block;
```
GirlBossRush, line 149: Ensures that the intersection observer can measure the threshold accurately.

web/src/elements/table/Table.ts
```diff
@@ -76,7 +76,6 @@ export abstract class Table<T extends object>
     implements TableLike
 {
     static styles: CSSResult[] = [
-        PFBase,
```
GirlBossRush, line 79: Vestigial

web/src/elements/table/Table.ts
```diff
@@ -102,11 +103,55 @@ export abstract class Table<T extends object>
...
+    //#endregion
+
+    //#region Protected Properties
+
     /**
      * Customize the "No objects found" message.
      */
     protected emptyStateMessage = msg("No objects found.");
 
+    /**
```
GirlBossRush, line 117: Re-ordered to avoid "this.foobar used before assignment"

web/src/elements/table/Table.ts
```diff
@@ -284,15 +344,20 @@ export abstract class Table<T extends object>
         ) {
             this.#synchronizeColumnProperties();
         }
+
+        if (changedProperties.has("visible") && this.hasUpdated) {
+            this.#synchronizeRefreshSchedule();
+        }
```
GirlBossRush, line 346: Subtly different from the call in `firstUpdated`, only invoking here when `visible` changes.

web/src/elements/table/Table.ts
```diff
@@ -930,7 +1015,7 @@ export abstract class Table<T extends object>
             ${guard([this.paginated, this.lastRefreshedAt], renderBottomPagination)}`;
     }
 
-    render(): TemplateResult {
+    protected override render(): SlottedTemplateResult {
```
GirlBossRush, line 1014: We could hypothetically do something like this here:

```ts
if (this.visible) {
  return nothing;
}
```

However, the `@intersectionObserver` decorator doesn't handle this very well when combined with the `<ak-tabs>` element. We can achieve this with something like an `<ak-tabpanel>` element would handle visibility checks, plus all the attribute copy-pasting we do to make ARIA work.

web/src/styles/authentik/base/common.css
```diff
@@ -67,6 +67,29 @@
...
+
+.ak-fade-in {
+    opacity: 0;
+    animation-fill-mode: forwards;
+    animation-name: ak-fade-in;
+    animation-duration: 150ms;
+    animation-iteration-count: 1;
+    animation-timing-function: linear;
+    animation-delay: 500ms;
+}
```
GirlBossRush, line 89: Avoids UI flickering when quickly transitioning from loading state.

web/src/elements/table/Table.ts
```diff
@@ -133,32 +178,39 @@ export abstract class Table<T extends object>
     }
 
     /**
-     * Whether the table is currently fetching data.
+     * Timestamp of when a fetch was requested, but deferred due to the table not being visible.
      */
-    @state()
-    protected loading = false;
+    #deferredRefreshRequestAt: Date | null = null;
```
GirlBossRush, line 180: Timestamp vs boolean for future debug logging

web/src/elements/ak-progress-bar.ts
kensternberg-authentik: Doesn't mean anything without `reflect: true` there.  `useDefault` [controls what happens when an attribute is missing or removed](https://lit.dev/docs/components/properties/#:~:text=Set%20to%20true%20to%20prevent%20initial%20attribute%20reflection%20for%20the%20default%20value%20when%20reflect%20is%20set%20to%20true%2C) ("fall back to this default value"), and prevents the initial default value from being reflected (since it doesn't need to be, it's the default):

> useDefault: Set to true to prevent initial attribute reflection for the default value when reflect is set to true, and to reset the property to its default value when its corresponding attribute is removed.

kensternberg-authentik (approved):
The `useDefault` issue is harmless, so let's go for it.

PR #19222, opened 2026-01-07
Update OpenID redirect URI for Guacamole

Jeppedy:
I believe the bare redirect URI is incorrect, and needs to end with /guacamole.  At least, that's what was needed in my bare-bones setup, and I presume for all.

## Details

REPLACE ME

---

## Checklist

If applicable


dewi-tik:
I think that the current guide is correct. The redirect uri just needs to be the FQDN used to access guacamole as described here: https://guacamole.apache.org/doc/gug/openid-auth.html#required-configuration

`guacamole.company` is the FQDN of the Guacamole installation.

PR #19132, opened 2025-12-31
web: disable user settings fields when changes are not allowed

dominic-r:
Previously, when admins disabled the ability to change email, username, or name in System Settings, users would only see an error after attempting to submit the form.

Now, the field is marked as disabled.

## Details

REPLACE ME

---

## Checklist

-   [ ] Local tests pass (`ak test authentik/`)
-   [ ] The code has been formatted (`make lint-fix`)

If an API change has been made

-   [ ] The API schema has been updated (`make gen-build`)

If changes to the frontend have been made

-   [ ] The code has been formatted (`make web`)

If applicable

-   [ ] The documentation has been updated
-   [ ] The documentation has been formatted (`make docs`)

PR #19230, opened 2026-01-07
revert: web: disable user settings fields when changes are not allowed

BeryJu:
Reverts goauthentik/authentik#19132

cc @dominic-r 

regular users won't have permission to access this API, and also this setting can be configured on a per-user level with attributes which this doesn't check


dominic-r (approved):
I was just about to revert it lol Yea will reimplement. completely slipped my mind. thanks!

GirlBossRush (approved):

PR #19229, opened 2026-01-07
web: 2025.12 RC2 UI feedback

GirlBossRush:
## Details
```
 ╭┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄╮
 ┆ This PR is a merge branch! ┆
 ╰┄┄┄┄┄┄┄┄┄┬┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄╯

      　　　   ╱╲ ╱╲ 
      　   ╱￣(°㇀°⚟)￣╱╲
         ╱┌──じ㇀し───┐╲╱
          │          │
          │      ǝpoɔ│ ╱
          └─────┄───┄
``` 
---

### How to review 

This PR is split into separate topics to make the changes more digestable: 

- #19233
- #19234
- #19235
- #19309

PR #19235, opened 2026-01-07
web: UI Locale Fixes

GirlBossRush:
## Details

This PR addresses some feedback on 2025.12 RC2:

### Localization

- Fixed locale not updating the user interface after form submission.
- Updated "Auto-detect" label to include the detected language name.
- Fixed API-provided localizations from not refreshing button text after using the locale selector

### Compatibility 

The context controller registry gracefully switches from `WeakMap` to `Map` if browser support for `Symbol` keys is not supported.

### Translations

Starting in #18426, we revised front-end XLIFF file names to [Unicode BCP 47 Locale Identifiers](https://unicode.org/reports/tr35/#Unicode_locale_identifier). These files now include a region subtag and are delimited with a dash. The previous versions (with an underscore) were left in the tree during the clean up while we tested Transifex. 

We've since confirmed that the BCP 47 files are updating as expected, making these files ready for removal.

## Related Merge Branch

- #19229


web/src/flow/FlowExecutor.ts
```diff
@@ -169,29 +177,30 @@ export class FlowExecutor
...
 
+    @listen(LOCALE_STATUS_EVENT)
+    protected localeStatusListener = (event: CustomEvent<LocaleStatusEventDetail>) => {
+        if (event.detail.status !== "ready") {
+            return;
+        }
+
+        this.refresh();
+    };
+
```
BeryJu:
```suggestion
```

BeryJu (approved):

PR #19291, opened 2026-01-08
blueprints: fix creating blueprint instances created in tasks timing out

### LINKED ISSUE #18174: Add ability to instantiatate blueprints via metamodels

Currently, `metaaplyblueprint` allows applying already **instantiated** blueprints.

However, I have a different use case (related to #18161).

I would like to manage multiple blueprints from a single "main" one. The way I imagine this is to have something like this:

```yaml
version: 1
metadata:
    name: Main blueprint
entries:
  - model: authentik_blueprints.metaapplyblueprint # or different metamodel
    attrs:
      required: true # require `path` to be valid but create a new blueprint instance if one doesn't exist
      path: oci://registry.com/my/blueprint/1 
      context: {...} # this for example creates a flow with certain settings (see #18161) 
      ... 
    ... 
  ... # more entries
```

Have the ability to create new blueprint instances via metamodels

Load an instance of all blueprints manually first, then apply them via `metaaplyblueprint` with different contexts (see #18161). 

### Additional context

_No response_

BeryJu:
closes #18174

### DIFF
--- a/authentik/lib/utils/task.py
+++ b/authentik/lib/utils/task.py
@@ -1,12 +1,20 @@
+from functools import lru_cache
 from pathlib import Path
 from tempfile import gettempdir
 from typing import Any
 
 from dramatiq import Actor
 
-mode = (Path(gettempdir()) / "authentik-mode").read_text().strip()
+
+@lru_cache
+def current_mode():
+    try:
+        return (Path(gettempdir()) / "authentik-mode").read_text().strip()
+    except OSError:
+        return "server"
+
 
 def run_task[T: Any](t: Actor[(str), T], *args, **kwargs) -> T:
-    if mode == "worker":
+    if current_mode() == "worker":
         return callable(*args, **kwargs)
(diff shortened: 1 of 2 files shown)


authentik/lib/utils/task.py
```diff
@@ -0,0 +1,20 @@
...
+def current_mode():
+    try:
+        return (Path(gettempdir()) / "authentik-mode").read_text().strip()
+    except OSError:
+        return "server"
+
+
+def run_task[T: Any](t: Actor[(str), T], *args, **kwargs) -> T:
+    if current_mode() == "worker":
+        return callable(*args, **kwargs)
```
rissson, line 19: This may fail if the task calls `CurrentTask.get_task`

astro-stan:
Hello,

This has not seen any activity for a month. Any chance it can be merged soon?

astro-stan:
Hello,

Has this been fixed? Why close unmerged?

PR #19305, opened 2026-01-09
website/docs: update unique email policy

### LINKED ISSUE #19304: Ensure unique email addresses script improvement

Hi,
The suggested script to Ensure unique email addresses in the documentation seems to have a problem. Indeed, if you add this policy to the default user setting prompt stage, then it prevent users to update their informations (because ak_user_by(email__iexact=email) will match their own account and the policy will return False).

### Link

https://docs.goauthentik.io/customize/policies/expression/unique_email/

### Solution

The script should then check that the use with this email is different from the current user itself. The following code should do it.

```
# Ensure this matches the *Field Key* value of the prompt
field_name = "email"
email = request.context["prompt_data"][field_name]
# Search for users with this email
other_user = ak_user_by(email__iexact=email)
if other_user:
  # Ensure that it is another user
  if ( request.user.username != other_user.username):
    ak_message(other_user.username + " already use this email")
    return False
return True
```

### Additional context

_No response_

dewi-tik:
## Details

Closes #19304

---

## Checklist

If applicable

### DIFF
--- a/website/docs/customize/policies/expression/unique_email.md
+++ b/website/docs/customize/policies/expression/unique_email.md
@@ -17,7 +17,7 @@ other_user = ak_user_by(email__iexact=email)
 if other_user:
   # Ensure that it is another user
   if ( request.user.username != other_user.username):
-    ak_message(other_user.username + " already use this email")
+    ak_message("Email address in use")
     return False
 
 return True
--- a/website/docs/customize/policies/expression/unique_email.md
+++ b/website/docs/customize/policies/expression/unique_email.md
@@ -10,13 +10,14 @@ The snippet below can be used in an expression policy within enrollment flows. T
 # Ensure this matches the *Field Key* value of the prompt
 field_name = "email"
 email = request.context["prompt_data"][field_name]
+pending_user = request.context.get("pending_user")
 
-# Search for users with this email
-other_user = ak_user_by(email__iexact=email)
+from authentik.core.models import User
+query = User.objects.filter(email__iexact=email)
+if pending_user:
+    query = query.exclude(pk=pending_user.pk)
 
-if other_user:
-  # Ensure that it is another user
-  if ( request.user.username != other_user.username):
+if query.exists():
     ak_message("Email address in use")
     return False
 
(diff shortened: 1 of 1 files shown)


website/docs/customize/policies/expression/unique_email.md
```diff
@@ -2,16 +2,23 @@
...
-if ak_user_by(email__iexact=email):
-  ak_message("Email address in use")
-  return False
+
+# Search for users with this email
+other_user = ak_user_by(email__iexact=email)
+
+if other_user:
+  # Ensure that it is another user
+  if ( request.user.username != other_user.username):
```
dominic-r: this could be problematic in a registration context when there's no existing user? potentially

website/docs/customize/policies/expression/unique_email.md
```diff
@@ -2,16 +2,23 @@
...
+    ak_message(other_user.username + " already use this email")
```
dominic-r: don't think we should expose who else has that email

dominic-r: already use -> already uses

dominic-r (changes requested):
For user registration, a user that dooesn't exist yet:

```py
field_name = "email"

# Access to prompt data
prompt_data = request.context.get("prompt_data", {})
email = prompt_data.get(field_name)

if not email:
    return True

# Check if email already exists
if ak_user_by(email__iexact=email):
    ak_message("This email address is already in use")
    return False

return True
```

Any user updating their email address:
```py
field_name = "email"

# Access to prompt data
prompt_data = request.context.get("prompt_data", {})
email = prompt_data.get(field_name)

if not email:
    return True

existing_user = ak_user_by(email__iexact=email)

if existing_user:
    # Get the pending user (the user being modified)
    pending_user = request.context.get("pending_user")

    # If a different user owns this email, deny
    if not pending_user or existing_user.pk != pending_user.pk:
        ak_message("This email address is already in use")
        return False

return True
```

cc @BeryJu

BeryJu:
@dominic-r updated the policy

dominic-r (approved):

PR #19311, opened 2026-01-09
sources/oauth: add fallback for id_token when profile URL is not available

BeryJu:
For OpenID sources specifically, when a profile URL is not available, fallback to the id_token (if possible); if no id_token is available the same error is returned

https://github.com/goauthentik/internal-customer-ref/issues/2

### DIFF
--- a/authentik/sources/oauth/clients/oauth2.py
+++ b/authentik/sources/oauth/clients/oauth2.py
@@ -189,6 +189,8 @@ class UserprofileHeaderAuthClient(OAuth2Client):
         profile_url = self.source.source_type.profile_url or ""
         if self.source.source_type.urls_customizable and self.source.profile_url:
             profile_url = self.source.profile_url
+        if profile_url == "":
+            return None
         response = self.session.request(
             "get",
             profile_url,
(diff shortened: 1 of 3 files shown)


ryanpesek:
Not sure if this comment is relevant since your implementation differs from my quick testing, but the user identifier in an id_token is usually in `sub` instead of `id`.  I had to modify the get_user_id() function to also check for `sub`

https://github.com/goauthentik/authentik/blob/c88195853d814a6080010caa644cab21dc5a75f6/authentik/sources/oauth/views/callback.py#L64-L65

BeryJu:
> Not sure if this comment is relevant since your implementation differs from my quick testing, but the user identifier in an id_token is usually in `sub` instead of `id`. I had to modify the get_user_id() function to also check for `sub`
> 
> https://github.com/goauthentik/authentik/blob/c88195853d814a6080010caa644cab21dc5a75f6/authentik/sources/oauth/views/callback.py#L64-L65

OIDC always uses `sub`, https://github.com/goauthentik/authentik/blob/main/authentik/sources/oauth/types/oidc.py#L45-L46

@ryanpesek

PR #19341, opened 2026-01-12
web/admin: always retrieve selected provider when editing the application

kensternberg-authentik:
web/admin: always retrieve selected provider when editing the application

# What

Re-writes the `fetch` function for ak-provider-search-input so that, if there’s an assigned value and it does not appear in the currently retrieved list of providers, prepend it to the list so that it is always present and always selectable.

# Why

Our pagination windows can restrict the list of objects retrieved from the server, and when we’re chasing composite objects we have to retrieve the displayable elements of that object from their respective tables. This combination means that a paginated retrieval may not have the object indicated by the parent object’s PK for that object collection. We have to retrieve it separately if it’s not in the current collection.

This problem is probably endemic to some of our design decisions.


BeryJu (approved):
We probably need to do this for every search-select object

Issue #10768, opened 2024-08-05
/dev/shm volume is filling up


miminar:
**Describe the bug**

/dev/shm in authentik-server pod is quickly filling up. k8s' default volume size limit is 64Mi which is completely filled up within 6 hours.

Once filled, the container starts to restart, which doesn't really help, because the /dev/shm volume remains filled.

Bumping the size of /dev/shm volume to 512Mi just delays the symptoms to 48 hours.

**To Reproduce**

1. deploy the helm chart on k8s with metrics enabled:

    ```yaml
    authentik:
      server:
        metrics:
          enabled: true
          serviceMonitor:
            enabled: true
    ```
    
1. sync users with LDAP with large quantities of users (>10000) and groups
1. add a couple of applications+providers secured with the OAuth2/OpenID Provider for users to use

**Expected behavior**

`/dev/shm` does not ever run out of space

**Screenshots**

**Logs**
Output of docker-compose logs or kubectl logs respectively

**Version and Deployment (please complete the following information):**

- authentik version: [2024.6.2](https://goauthentik.io/docs/releases/2024.6)
- using helm chart [authentik-2024.6.2](https://github.com/goauthentik/helm/releases/tag/authentik-2024.6.2) 
- k8s version: [v1.30.2+k0s](https://github.com/k0sproject/k0s/releases/tag/v1.30.2%2Bk0s.0)

**Additional context**

The problematic files with unbounded growth are under `/dev/shm/authentik_prometheus_tmp` (showing the recent state shortly after a restart of the pod):

```sh
authentik@authentik-users-server-675b58969b-qjwcv:/$ du -ah /dev/shm/
0       /dev/shm/authentik-core.sock
4.0K    /dev/shm/authentik-gunicorn.3372311009.pid
1008K   /dev/shm/authentik_prometheus_tmp/gauge_all_1.db
8.8M    /dev/shm/authentik_prometheus_tmp/histogram_1.db
32K     /dev/shm/authentik_prometheus_tmp/counter_1.db
3.4M    /dev/shm/authentik_prometheus_tmp/gauge_all_2.db
29M     /dev/shm/authentik_prometheus_tmp/histogram_2.db
28K     /dev/shm/authentik_prometheus_tmp/counter_2.db
42M     /dev/shm/authentik_prometheus_tmp
0       /dev/shm/authentik_worker_tmp
4.0K    /dev/shm/authentik-mode
42M     /dev/shm/
```
Progress with 512Mi size limit:
[...]

PR #19369, opened 2026-01-13
lifecycle: bump shm size

BeryJu:
See https://github.com/goauthentik/authentik/issues/10768
and https://github.com/goauthentik/authentik/pull/19312

A temporary fix :tm:


rissson:
Should be backported if we want this to be released with 2025.12

rissson (approved):

PR #19403, opened 2026-01-14
outposts/ldap: fix build

BeryJu:
fix release build for LDAP outpost failing

closes #19396


.github/workflows/release-publish.yml
```diff
@@ -101,7 +101,7 @@ jobs:
       - name: make empty clients
         run: |
           mkdir -p ./gen-ts-api
-          mkdir -p ./gen-go-api
+          make gen-client-go
```
rissson, line 104: Duplicate with the addition below, no?

BeryJu: nah one of them is for container build the other is for binary build

rissson: ah yeah, fair enough

rissson (approved):

PR #19448, opened 2026-01-14
ci: fix checkout stable (for 2025.12)

gergosimonyi:
Fixes the fix at https://github.com/goauthentik/authentik/pull/18303

This fails on version branches that already have releases, because the version tag is named `version/${numbers}`, not just `${numbers}`.


.github/workflows/ci-main.yml
```diff
@@ -84,7 +84,7 @@ jobs:
           # Current version family based on
           current_version_family=$(cat internal/constants/VERSION | grep -vE -- 'rc[0-9]+$' || true)
           if [[ -n $current_version_family ]]; then
-              prev_stable=$current_version_family
+              prev_stable="version/$current_version_family"
```
BeryJu:
```suggestion
              prev_stable="version/${current_version_family}"
```

PR #19152, opened 2026-01-01
web: Fix user library colors, modal z-indexes, table progress bars

GirlBossRush:
## Dependents

- #19141

## Details

This PR covers a few small UI fixes discovered in testing:

- Fix z-index layering for drawer content, modals, page headers, progress bars
- Fix table progress bar ARIA
- Clean up color properties to better align with PatternFly's styling
- Export CSS parts of library elements to better expose custom CSS

PR #19460, opened 2026-01-15
web: Z-Index Fixes, Mobile Sidebar Behavior.

GirlBossRush:
## Details

This fixes several reported issues regarding z-index ordering in modals, headers, and the admin sidebar, along with a few variations on the issues for mobile and tablet viewports.This effort was initially introduced in #19152. This PR partially reverts to some of the Patternfly overrides we had in 2025.10, albeit with more targeted selection via CSS custom properties.

#### Z-indexing

Z-indexes for drawers, sidebars, headings, and modals use fewer explicit z-indexes to allow for rendering order to take precedence in a few select areas.

#### Behaviors

- The menu toggle button is moved into a slot to isolate the sidebar visibility state into a single element: the admin interface.
- The admin sidebar now collapses itself after route changes on tablet and mobile viewports, rather than obscuring the page and requiring the user to manually close the menu.
- The menu toggle button now only appears on viewport sizes when the sidebar sits on top of the content, matching our behavior on doc page sidebars. 

#### Accessibility

- A clickable backdrop area is added to the admin sidebar on mobile to match the modal backdrop behavior.


web/src/components/ak-page-navbar.css
```diff
@@ -0,0 +1,189 @@
+:host {
```
GirlBossRush, line 1: The bulk of this file is extracted verbatim from `ak-page-navbar.ts`, both for good housekeeping and to avoid Lit Analyzer's lack of [container queries](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Containment/Container_queries)

kensternberg-authentik (approved):
That's quite lovely work, and I appreciate pulling out the big CSS blobs into their own files.

PR #19476, opened 2026-01-15
tasks: threads instead of forks

BeryJu:
WIP


authentik/tasks/middleware.py
```diff
@@ -228,14 +241,15 @@ def run(addr: str, port: int):
...
+    def after_worker_boot(self, broker, worker):
+        self.thread = Thread(target=WorkerStatusMiddleware.run)
+        self.thread.start()
 
     @staticmethod
     def run():
+        setthreadtitle("authentik Worker status")
         status = WorkerStatus.objects.create(
             hostname=socket.gethostname(),
             version=authentik_full_version(),
```
rissson: I think this should actually go into the try/except because it can fail

rissson: And also it would actually fail, if the hostname doesn't change

rissson (approved):

PR #19485, opened 2026-01-15
root: revert enterprise loading behaviour

BeryJu:
extract reverted from https://github.com/goauthentik/authentik/commit/213cf4492869ade23d06606a4317d9cb5658d9d2


gergosimonyi (commented):
All in all this seems to work better than I expected. Couple of issues I ran into after `rm -rf authentik/enterprise`:

1. `make gen` fails on
```
ENUM_OVERRIDES: [
...
        "LicenseSummaryStatusEnum": "authentik.enterprise.models.LicenseUsageStatus",
...
]
```

2. Tests fail on many levels.
3. The frontend calls `/api/v3/enterprise/license/summary/` on first load which 404s (without further effect).
4. The frontend calls `/api/v3/providers/google_workspace/` on the overview page which 404s and therefore the Sync status widget fails.

PR #19498, opened 2026-01-15
providers/oauth2: Support login_hint

PeshekDotDev:
Closes #18290 


kensternberg-authentik (approved):
Front-end changes look sane and sensible.

authentik/flows/views/executor.py
```diff
@@ -66,6 +66,7 @@
 SESSION_KEY_GET = "authentik/flows/get"
 SESSION_KEY_POST = "authentik/flows/post"
 SESSION_KEY_HISTORY = "authentik/flows/history"
+SESSION_KEY_OVERRIDE_LOGIN_HINT = "authentik/flows/override_login_hint"
```
BeryJu: Should be flow context storage, not session

authentik/flows/views/executor.py
```diff
@@ -481,6 +482,8 @@ def get(self, request: HttpRequest) -> HttpResponse:
         if SESSION_KEY_PLAN in request.session:
             del request.session[SESSION_KEY_PLAN]
             LOGGER.debug("Canceled current plan")
+        # Set flag to prevent auto-skip on login_hint when user clicks "Not you?"
+        request.session[SESSION_KEY_OVERRIDE_LOGIN_HINT] = True
```
BeryJu: Would prefer if we can add this without adding extra logic here

authentik/flows/views/executor.py
BeryJu: Also the constant should be in the identification stage for plan context

BeryJu (approved):

PR #19504, opened 2026-01-16
web: Reduce Sentry Development Errors

GirlBossRush:
## Details

This PR address two sources of Sentry errors on page load when debugging is enabled:

1. When Sentry initializes a session
2. When the client-side router changes pages

Screenshot 2026-01-16 at 04 07 24

### DIFF
--- a/web/src/common/sentry/index.ts
+++ b/web/src/common/sentry/index.ts
@@ -31,6 +31,7 @@ function beforeSend(
     if (!hint) {
         return event;
     }
+
     if (hint.originalException instanceof SentryIgnoredError) {
         return null;
     }
@@ -40,6 +41,7 @@ function beforeSend(
     ) {
         return null;
     }
+
     return event;
 }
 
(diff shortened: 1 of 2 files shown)


web/src/common/sentry/index.ts
```diff
@@ -53,21 +55,25 @@ export function configureSentry(): void {
...
-    const integrations: Integration[] = [
-        browserTracingIntegration({
-            // https://docs.sentry.io/platforms/javascript/tracing/instrumentation/automatic-instrumentation/#custom-routing
-            instrumentNavigation: false,
-            instrumentPageLoad: false,
-            traceFetch: false,
-        }),
-    ];
+    const integrations: Integration[] =
+        process.env.NODE_ENV === "production"
```
GirlBossRush, line 59: Note that the repetition of `process.env.NODE_ENV` is how we hint to ESBuild that the value is a constant and the `browserTracingIntegration` import eligible for tree-shaking.

kensternberg-authentik (approved):
Yay! Quieter sentry on dev.

PR #19527, opened 2026-01-16
flows: add option for flow layout with frame background

BeryJu:
Add the option for sidebar left/right but use the background URL as an iframe


GirlBossRush (approved):

web/src/flow/FlowExecutor.ts
```diff
@@ -539,19 +537,45 @@ export class FlowExecutor
...
+
+            if (!src) return nothing;
+
+            return html`
+                <div class="ak-c-login__content" part="content">
+                    <iframe
+                        class="ak-c-login__content-iframe"
+                        part="content-iframe"
+                        name="flow-content-frame"
+                        src=${"https://y-n10.com"}
```
BeryJu:
```suggestion
                        src=${src}
```

PR #19536, opened 2026-01-17
root: add Nix support to Makefile for native dependency compilation

dominic-r:
Overview:

Add support for Nix-installed xmlsec, libxml2, libxslt, and libtool alongside existing Homebrew support. Nix detection takes precedence over Homebrew to prevent empty LDFLAGS/CPPFLAGS when brew exists but packages are installed via Nix.

Motivation:

Nix users with Homebrew also installed would get empty LDFLAGS/CPPFLAGS which caused xmlsec and lxml compilation/running the dev server to fail with missing library errors.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,33 +20,42 @@ GEN_API_TS = gen-ts-api
 GEN_API_PY = gen-py-api
 GEN_API_GO = gen-go-api
 
-BREW_LDFLAGS :=
-BREW_CPPFLAGS :=
-BREW_PKG_CONFIG_PATH :=
-
 UV := uv
 
-# For macOS users, add the libxml2 installed from brew libxmlsec1 to the build path
-# to prevent SAML-related tests from failing and ensure correct pip dependency compilation
+# For macOS users, configure native dependencies (xmlsec, libxml2, krb5) for pip compilation.
 ifeq ($(UNAME_S),Darwin)
-# Only add for brew users who installed libxmlsec1
-	BREW_EXISTS := $(shell command -v brew 2> /dev/null)
-	ifdef BREW_EXISTS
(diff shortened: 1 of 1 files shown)


gergosimonyi (commented):
I don't believe this is the way to go. The way nix is used is too varied for authentik to soft-enforce a single way in a command as high-level as `make install`.

The implementation itself is very happy in the happy case and very sad in the sad case. Currently I have 5 opposing points, but I'm almost certain I could come up with 5 more. 

1. Nix stores can be anywhere, not only at `/nix/store`.
2. `*-xmlsec-*-dev` (and others) can match _many_ things I'd not want it to match.
3. I may have used nix for installing xmlsec a year ago for a different project, but now use brew for developing authentik. A makefile can't make that determination based on the presence of some files.
4. `head -1` can result in a version I don't user anymore that hasn't yet been garbage collected. Once I run the nix garbage collector, everything is now broken.
5. Please don't make me `uv cache clean` every time I run `make install` :'(

dominic-r:
As Gergo highlighted offline, and a way better solution, just run the `uv sync` before `make install` in this format:

```
PKG_CONFIG_PATH=/nix/store/2127r3gwq8j88x3h0v8gd9l6f84m2kvv-xmlsec-1.3.7-dev/lib/pkgconfig:/nix/store/9pna2kdj9358d9k9z02vidlbl25ra6hw-libxml2-2.15.1-dev/lib/pkgconfig:/nix/store/qbgb0v3h662b204jfl5s2nrx06xi8jfi-libxslt-1.1.45-dev/lib/pkgconfig LIBRARY_PATH=/nix/store/2nn3qfanfwldhwkqjjdsphdch1583vir-libtool-2.5.4-lib/lib uv sync --frozen
```

PR #19542, opened 2026-01-17
web: update @goauthentik/api

Ma27:
## Details

Otherwise, e.g. the edit modal of Applications hangs infinitely on a loading spinner because `AdminFileListUsageEnum` is undefined and not an object.

cc @rissson @BeryJu 

---

## Checklist

If applicable

### DIFF
--- a/web/package-lock.json
+++ b/web/package-lock.json
@@ -23,7 +23,7 @@
                 "@floating-ui/dom": "^1.7.4",
                 "@formatjs/intl-listformat": "^8.1.2",
                 "@fortawesome/fontawesome-free": "^7.1.0",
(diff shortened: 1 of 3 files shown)


BeryJu:
I was gonna say this should only make a difference for development as we generate the client directly for the release build, but then I saw the reference nix install issue - fair enough.

FYI I'd recommend also switching that build generate the client on the fly if possible, we're probably gonna switch to having 2 api clients, one on a rolling version for our webinterface and one on the release cycle for everyone else to use.

Ma27:
Thanks, that's good to know @BeryJu !
In the meantime, I think this is still a valid fix to merge and backport, no?

BeryJu (approved):

PR #19547, opened 2026-01-18
admin/files: fix manageable check blocking file creation on fresh installs

### LINKED ISSUE #19546: Configured file backend does not support file management.
### Describe the bug

I cannot seem to get the new Files page to work.

The error is:
Configured file backend does not support file management.

### How to reproduce

1. Install using docker-compose.yml
2. navigate to https://authentik.mydomain.net/if/admin/#/files
3. Configured file backend does not support file management.

### Expected behavior

The new file manager should be available to use.

### Screenshots

_No response_

### Additional context

Ubuntu 22.04

Authentik defaults (/data directory, local filesystem, etc.)

### Deployment Method

Docker

### Version

2025.12.1

### Relevant log output

```shell
I haven't been able to find any errors in the docker logs.
```

dominic-r:
Overview:

The manageable property in `FileBackend` checked `base_path.exists()` which includes the usage and schema subdirectories (e.g. `/data/media/public`).

On fresh installs these directories don't exist yet, which causes the check to return False and blocking file uploads under Customization > Files (and the corresponding API routes).

Changed to check `_base_dir.exists()` (e.g. `/data`) instead, since that's the actual mount point. The subdirectories are created automatically by `save_file()` via `mkdir(parents=True)` when the first file is uploaded.

An explanation has also been added above the changed line.

Testing:

Spun up a Docker installation using the docs. Then, made the fix and built an image locally. Ran `docker compose down -v` and deleted volume directories, then ran the compose file again to ensure a fresh set up, and confirmed files can be created.

Motivation:

Closes: https://github.com/goauthentik/authentik/issues/19546

### DIFF
--- a/authentik/admin/files/backends/file.py
+++ b/authentik/admin/files/backends/file.py
@@ -45,8 +45,13 @@ class FileBackend(ManageableBackend):
 
     @property
     def manageable(self) -> bool:
+        # Check _base_dir (the mount point, e.g. /data) rather than base_path
+        # (which includes usage/schema subdirs, e.g. /data/media/public).
+        # The subdirectories are created on first file write via mkdir(parents=True)
+        # in save_file(), so requiring them to exist beforehand would prevent
+        # file creation on fresh installs.
         return (
-            self.base_path.exists()
+            self._base_dir.exists()
             and (self._base_dir.is_mount() or (self._base_dir / self.usage.value).is_mount())
             or (settings.DEBUG or settings.TEST)
         )


BeryJu (approved):

PeteKersker:
I pulled the "cherry pick" version using the instructions above (changing my .env file) and it solved the [#19546](https://github.com/goauthentik/authentik/issues/19546) problem.
I was able upload a file to the server using the UI and verified the file using command line "ls" command.
Thank you!

General Note: This is my first time installing and learning the Authentik product. I'm really impressed. Well done!

dominic-r:
Thanks for those kind words @PeteKersker !

PR #19548, opened 2026-01-18
web/admin: fix file upload not preserving extension for custom names with dots

dominic-r:
Overview:

The `hasBasenameExtension()` function in `FileUploadForm.ts` incorrectly determined whether a custom filename already had an extension by checking if it contained any dot at position > 0.

This caused filenames like "`e._.e`" to be treated as having an extension, so the original file's extension was not appended. The file would be saved as "`e._.e`" instead of "e._.e.jpg", which caused `mimetypes.guess_type()` to return `None` (since ".e" is not a recognized extension) and the backend to fall back to "application/octet-stream".

Removed `hasBasenameExtension()` entirely. Since the UI explicitly states "Optionally rename the file (without extension)", we now always append the original file's extension when a custom name is provided.

Testing:

1. Upload a JPG file with custom name "e" --> saves as "e.jpg", and is detected as "image/jpeg"
2. Upload a JPG file with custom name "`e._.e`" --> now saves as "e._.e.jpg",and is detected as "image/jpeg"

Motivation:

Fixes incorrect MIME type detection for uploaded files when users provide custom filenames containing dots.

### DIFF
--- a/web/src/admin/files/FileUploadForm.ts
+++ b/web/src/admin/files/FileUploadForm.ts
@@ -45,7 +45,6 @@ function getFileExtension(fileName: string): string {
     return fileName.slice(lastDot);
 }
 
-
 @customElement("ak-file-upload-form")
 export class FileUploadForm extends Form<Record<string, unknown>> {
     @property({ type: String, useDefault: true })
--- a/web/src/admin/files/FileUploadForm.ts
+++ b/web/src/admin/files/FileUploadForm.ts
@@ -45,11 +45,6 @@ function getFileExtension(fileName: string): string {
     return fileName.slice(lastDot);
 }
 
-function hasBasenameExtension(fileName: string): boolean {
-    const baseName = fileName.split("/").pop() ?? fileName;
-    const lastDot = baseName.lastIndexOf(".");
-    return lastDot > 0;
-}
 
 @customElement("ak-file-upload-form")
 export class FileUploadForm extends Form<Record<string, unknown>> {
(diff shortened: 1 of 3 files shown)


web/src/admin/files/FileUploadForm.ts
```diff
@@ -83,14 +77,13 @@ export class FileUploadForm extends Form<Record<string, unknown>> {
...
         let finalName = this.selectedFile.name;
         if (customName) {
             assertValidFileName(customName);
             const ext = getFileExtension(this.selectedFile.name);
-            finalName =
-                ext && !hasBasenameExtension(customName) ? `${customName}${ext}` : customName;
+            finalName = ext ? `${customName}${ext}` : customName;
         } else {
             assertValidFileName(this.selectedFile.name);
         }
```
kensternberg-authentik: Since `getFileExtension` always returns a string, inccluding an empty string if there is no extension, you could re-write this entire block:

```
const finalName = customName
    ? `${customName}${getFileExtension(this.selecteFile.name)}`
    : this.selectedFile.name;

assertValidFileName(finalName);
```

... and remove the assertion that follows it, since the `else` clause only duplicates exactly what the assertion that follows does.  You want to validate that the filename _produced_ by these decisions is valid for storage.  The code and comments as they stand obscure that.

kensternberg-authentik (approved):
It's valid and it does what it says it should do, but the custom filename logic is a little convoluted and obscures intent; I left a suggestion on how to clean it up.

PR #19554, opened 2026-01-18
providers/oauth2: add `logout+jwt` token type for oidc logout token.

jeroen1602:
## Details

The oidc back-channel logout spec recommends using explicitly typed JWTs using the `typ` parameter in the JWT's header.

[spec](https://openid.net/specs/openid-connect-backchannel-1_0.html#CrossJWT)

This may be a breaking change for some implementations if they were already checking the type of the token to be `JWT` (the default value).

---

## Checklist

If applicable

### DIFF
--- a/authentik/providers/oauth2/utils.py
+++ b/authentik/providers/oauth2/utils.py
@@ -259,4 +259,4 @@ def create_logout_token(
     if session_key:
         payload["sid"] = hash_session_key(session_key)
     # Encode the token
-    return provider.encode(payload, "logout+jwt")
+    return provider.encode(payload, jwt_type="logout+jwt")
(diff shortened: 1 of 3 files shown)


authentik/providers/oauth2/utils.py
```diff
@@ -259,4 +259,4 @@ def create_logout_token(
     if session_key:
         payload["sid"] = hash_session_key(session_key)
     # Encode the token
-    return provider.encode(payload)
+    return provider.encode(payload, "logout+jwt")
```
BeryJu:
```suggestion
    return provider.encode(payload, jwt_type="logout+jwt")
```

BeryJu (approved):

PR #19562, opened 2026-01-19
web/forms: fix forms not resetting state when modal closes

dominic-r:
Overview:

Forms were not properly resetting their state when closing modals, which caused stale values to persist when reopening forms. This affected all forms with `@state()` decorated properties.

Testing:

1. Create any item (user, token, application, etc.), close modal
2. Click Create again, form should show default/empty values
3. Edit an item, cancel, click Create - form should be empty
4. Edit an item, cancel, edit same item - should show correct data

Motivation:

Form inputs retained values from previous create/edit operations.

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -72,6 +72,7 @@
         "#tests/*": "./tests/*.js",
         "#e2e": "./e2e/index.ts",
         "#e2e/*": "./e2e/*.ts",
+        "#types/*": "./types/*/index.d.ts",
         "#*/browser": {
             "types": "./out/*/browser.d.ts",
             "import": "./*/browser.js"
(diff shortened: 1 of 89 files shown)


GirlBossRush:
@dominic-r This is a great improvement over the current behavior. #19395, and its follow up in #18916 solves this by completely clearing the modal from the page, but that'll take some time to merge.

This PR touches a lot of places so I've pushed up a few type changes which allow these changes to merge into #19395:

- Added `protected` and `public` where appropriate
- Added Lit Analyzer doc strings
- Enabled additional analyzer checks 
- More consistent usage of `null` over `undefined` for stateful properties

dominic-r:
Thanks for these great updates @GirlBossRush !

GirlBossRush (approved):

PR #19588, opened 2026-01-19
rbac: clean up roles and permissions

gergosimonyi:
This was purposefully not included in `2025.12` to split the changes up.

The main content of this patch is in the migrations. Everything else follows more or less automatically.


blueprints/schema.json
```diff
@@ -11284,8 +11283,7 @@
                     "type": "string",
                     "enum": [
                         "MIT",
-                        "Heimdal",
-                        "other"
```
gergosimonyi: This change (and its analogue over at `schema.yml`) was automatically included when I ran `make gen`. I'll leave it in.

authentik/core/models.py
```diff
@@ -355,13 +355,17 @@ def exclude_anonymous(self) -> QuerySet:
...
+    user_permissions = None
+
     uuid = models.UUIDField(default=uuid4, editable=False, unique=True)
     name = models.TextField(help_text=_("User's display name."))
     path = models.TextField(default="users")
     type = models.TextField(choices=UserTypes.choices, default=UserTypes.INTERNAL)
 
     sources = models.ManyToManyField("Source", through="UserSourceConnection")
-    ak_groups = models.ManyToManyField("Group", related_name="users")
+    groups = models.ManyToManyField("Group", related_name="users")
```
BeryJu, line 368: We might want some sort of proxy for the old field as its probably used in policies quite frequently, something that'll log an event when used.

gergosimonyi: I thought this deserved its own `EventAction`, see https://github.com/goauthentik/authentik/pull/19588/commits/6fc2469695c92af26c66e712b9f0d51af472a06f for my first pass on implementing this.

authentik/events/models.py
```diff
@@ -112,6 +112,7 @@ class EventAction(models.TextChoices):
     SYSTEM_EXCEPTION = "system_exception"
 
     CONFIGURATION_ERROR = "configuration_error"
+    DEPRECATION_USED = "deprecation_used"
```
```suggestion
    CONFIGURATION_WARNING = "configuration_warning"
```

Should make it a bit more re-usable for other things

dominic-r (approved):
lgtm for docs

website/docs/releases/2026/v2026.2.md
```diff
@@ -12,7 +12,11 @@ To try out the release candidate, replace your Docker image tag with the latest
 
 ## Highlights
 
-<!-- ## Breaking changes -->
+## Breaking changes
+
+### RBAC
+
+`User.ak_groups` was renamed to `User.groups`. We recommend you check any custom code (e.g. expression policies, property mappings) that deals with group memberships to update them if necessary.
```
BeryJu:
```suggestion
`User.ak_groups` has been deprecated. Users' groups are now accessed through `User.groups`. Usage of `.ak_groups` will continue to function, but will create a configuration warning event. We recommend you check any custom code (e.g. expression policies, property mappings) that deals with group memberships to update them if necessary.
```

BeryJu (approved):
Looks good, added one small comment to the release notes to tell people the old one is deprecated but will still work.

One other thing I suggest for this is splitting out the changes to integrations since those are not deployed from the version branch and as such will go live as soon as this is merged, so we should separate that and only merge it once this is out. (Or maybe even later since .ak_groups will still work)

PR #19623, opened 2026-01-20
web/elements: remove pfbase everywhere

kensternberg-authentik:
web/maintenance: deprecate PFBase in favor of an adopted stylesheet

# What !?!?!?

Removes `PFBase` from all components.

# Why !?!?!?

In `AkElement`, there was this code:

        protected static override finalizeStyles(styles?: CSSResultGroup): CSSResultOrNative[] {
            if (!styles) return [$PFBase, $AKBase];

            if (!Array.isArray(styles)) return [$PFBase, createCSSResult(styles), $PFBase, $AKBase];

            return [
                $PFBase,
                // ---
                ...(styles.flat() as CSSResultOrNative[]).map(createCSSResult),
                $AKBase,
            ];
        }

I’ve refined this:

        protected static override finalizeStyles(styles: CSSResultGroup = []): CSSResultOrNative[] {
            const elementStyles = [
                $PFBase,
                // Route around TSC`s known-to-fail typechecking of `.flat(Infinity)`. Removes types.
                ...([styles] as Array<unknown>).flat(Infinity),
                $AKBase,
                // Restore types. Safe: we control AKBase and PFBase in this file, and `styles` are
                // typed on function signature.
            ] as CSSResultOrNative[];

            // Remove duplicates in reverse order to preserve last-insert-wins semantics of CSS.
            // See: https://github.com/lit/lit/blob/main/packages/reactive-element/src/reactive-element.ts#L945
            const elementSet = new Set(elementStyles.reverse());
            // Reverse again because the return type is an array, and process as a CSSResult
            return Array.from(elementSet).reverse().map(createCSSResult);
        }

… with the duplication removal documented in Lit 3.0. `styles` defaults to an array, is cast to an array, then automatically flattented before the deduplication is run.

With this, both PFBase and AKBase are automatically included with each and every component that inherits from `AKElement`. At that point, the inclusion of `PFBase` interface-wide made no sense.

So they had to be removed:

    $ for i in $(rg -t typescript -l PFBase | rg -v 'elements/Base\.ts') ; do \
        perl -pi.bak -e 's{import PFBase from ".patternfly/patternfly/patternfly-base.css";}{}' "$i" ; \
        done
    $ for i in $(rg -t typescript -l PFBase | rg -v 'elements/Base\.ts') ; do \
        perl -pi.bak -e 's/PFBase,//' "$i" ; \
        done

This commit removes 131 `import` statements from the source code. As a result, the bundle is about 27K smaller… which admittedly is about 0.2% smaller than before. Ah, well. “Every little bit helps,” right?


web/src/elements/Base.ts
```diff
@@ -34,17 +34,21 @@ export class AKElement extends LitElement implements AKElementProps {
...
-        ];
+            // Restore types. Safe: we control AKBase and PFBase in this file, and `styles` are
+            // typed on function signature.
+        ] as CSSResultOrNative[];
+
+        // Remove duplicates in reverse order to preserve last-insert-wins semantics of CSS. See:
+        // https://github.com/lit/lit/blob/main/packages/reactive-element/src/reactive-element.ts#L945
+        const elementSet = new Set(elementStyles.reverse());
+        // Reverse again because the return type is an array, and process as a CSSResult
+        return Array.from(elementSet).reverse().map(createCSSResult);
```
kensternberg-authentik, line 51: This, I say, this is the only major change, and it's just cleanup to install the dedupe semantics.

GirlBossRush:
Looks great! 🥳

Can this be combined with #18618?

GirlBossRush (approved):

PR #19657, opened 2026-01-22
admin/files: add centralized theme variable support for file URLs

dominic-r:
Overview:

Adds support for `%(theme)s` placeholder in file paths, which allows theme-specific assets (like logos, backgrounds, icons) to be served based on the user's current theme (light/dark).

This replaces the previous implementation (reverted in this PR) which only handled theme substitution in the Go file backend and instead uses the new approach which centralizes theme logic and works across both backends.

Testing:

Try out the following for the file and s3 backend:

* Ensure themed images load
* Ensure non-themed images load

Motivation:

Internal


authentik/admin/files/backends/passthrough.py
```diff
@@ -46,3 +46,15 @@ def file_url(
     ) -> str:
         """Return the URL as-is for passthrough files."""
         return name
+
+    def themed_urls(
```
rissson, line 50: I don't think we necessarily need to deny theme variables for external URLs.

authentik/brands/api.py
```diff
@@ -90,7 +96,9 @@ class CurrentBrandSerializer(PassiveSerializer):
     matched_domain = CharField(source="domain")
     branding_title = CharField()
     branding_logo = CharField(source="branding_logo_url")
+    branding_logo_themed_urls = DictField(child=CharField(), allow_null=True)
```
rissson: We should have a custom serializer expliciting which fields are set in here

rissson (changes requested):

web/src/elements/mixins/branding.ts
```diff
@@ -81,10 +91,18 @@ export const WithBrandConfig = createMixin<BrandingMixin>(
                 return this.brand.brandingLogo ?? DefaultBrand.brandingLogo;
             }
 
+            public get brandingLogoThemedUrls(): ThemedUrls | null | undefined {
+                return this.brand.brandingLogoThemedUrls;
```
kensternberg-authentik, line 95: `null | undefined` is a bit of a code smell.  Can you pick one or the other, and prefer `null`?

kensternberg-authentik (approved):
I don't see anything radically broken on the front-end.  I just wonder if we have _documentation_ that describes all of these ways that logos and such get into the product.

rissson (approved):

BeryJu (approved):

PR #19670, opened 2026-01-22
web/a11y: CAPTCHA Stage Form

GirlBossRush:
## Details

This PR is a follow up on #19555:

- Added ARIA labels and placeholders to form inputs
- Fixed missing ARIA attributes in numeric and secret input components
- Rephrased order of field instructions for a more screen-reader friendly flow.


kensternberg-authentik (approved):
Looks functional.  Don't see any issues.

PR #19697, opened 2026-01-23
providers/scim: Purge inconsistent remote resources for AWS provider

### LINKED ISSUE #20411: Authentik doesn't delete SCIM users/group when uses LDAP source with 'deferred' sync trigger mode
### Describe the bug

We use LDAP (FreeIPA) as a data source for Authentik. When the 'Outgoing Sync Trigger Mode' parameter is set to 'Deferred until end', Authentik does not delete users and groups provisioned through the SCIM provider.

The current cleanup logic is [based on pre_delete hooks](https://github.com/goauthentik/authentik/blob/version/2025.12.3/authentik/lib/sync/outgoing/signals.py#L60-L61), but these hooks [are disabled in deferred mode](https://github.com/goauthentik/authentik/blob/version/2025.12.3/authentik/sources/ldap/tasks.py#L160-L161). The SCIM provider does not implement any alternative cleanup logic for users and groups.

As a workaround, enabling 'Immediate' mode resolves the issue, but in our setup, it triggers thousands of tasks.

Some time ago, I submitted a PR that addresses this behavior: https://github.com/goauthentik/authentik/pull/19697

### How to reproduce

1. Configure an LDAP/FreeIPA source for users and groups.
2. Set the 'Outgoing Sync Trigger Mode' for the LDAP source to 'Deferred until end'.
3. Enable 'Delete Not Found Objects' for the LDAP source.
4. Add a SCIM provider (AWS in our case).
5. Provision a few groups and users with the SCIM provider.
6. Delete some of these groups/users from the LDAP source.
7. The users/groups are deleted from the Authentik database but remain in the SCIM provider. Moreover, the users stay active and can continue using their access if an alternative authentication method is available.

### Expected behavior

Users and groups provisioned with SCIM, have to be eventually deleted when they are deleted from an LDAP source. Regardless of the 'Outgoing sync trigger mode' value.

### Screenshots

_No response_

### Additional context

_No response_

### Deployment Method

Docker

### Version

2025.12.3

### Relevant log output

```shell

```

bitpavel-l25:
## Details

Authentik doesn't purge (clean up) SCIM resources in two cases:

When a resource isn't found in the local database. We've observed this when a user is removed from the source LDAP catalog.

When a resource no longer matches the configured filters—for example, when a user is removed from the corresponding filtering group.

This PR introduces new logic for purging SCIM resources.

I've added a "Purge objects" flag to the SCIM provider configuration, allowing administrators to control this behavior. For backward compatibility, it's disabled by default.

Since AWS's SCIM implementation differs slightly from the official RFC (particularly in pagination), and we need this for our AWS integration, the PR implements purging only for the AWS provider. The code includes placeholders for other providers, so support for additional cases can be contributed by interested parties.

This PR fixes: https://github.com/goauthentik/authentik/issues/20411

## Checklist

-   [+] Local tests pass (`ak test authentik/`)
-   [+] The code has been formatted (`make lint-fix`)

-   [+] The API schema has been updated (`make gen-build`)

-   [+] The code has been formatted (`make web`)

If applicable

-   [+] The documentation has been updated
-   [+] The documentation has been formatted (`make docs`)


authentik/providers/scim/clients/users.py
```diff
@@ -118,3 +123,50 @@ def update(self, user: User, connection: SCIMProviderUser):
...
+        if len(remote_user_ids) < 1:
+            return
+        local_user_ids = {}
+        for i in SCIMProviderUser.objects.filter(provider=self.provider).values_list(
+            "scim_id", "user_id"
+        ):
+            local_user_ids[i[0]] = i[1]
+        for id in remote_user_ids:
+            if id not in local_user_ids.keys():
+                self._request("DELETE", f"/Users/{id}")
```
ImmanuelVonNeumann, line 150: This seems like you are deleting any user from the remote that is not represented inside authentik.

I am unsure if this is the right approach. 
In multiple use-cases there exist local accounts on the remote that should be kept (for example emergency accounts if the Auth Provider fails).

Having had a brief look at the SCIM Documentation ([RFC7643](https://www.rfc-editor.org/rfc/rfc7643.html) f.) it seems there is no clear way on how to approach this.

I see three possibilities to implement this within authentik
1. When a user loses access to the scim application, delete them on the remote (same for groups unselected from ``filter_groups``)
2. Only compare remote users and groups with lingering authentik scim users and groups and only delete those on the remote where the lingering authentik user/group does not have access anymore
3. Only purge remote objects that have an ``externalId`` set and aren't elligible for the authentik application anymore.

I believe approach 2 to be the best approach for authentik's architecture.
The code change to your current approach in that case should be minimal too.

What do others think?

ImmanuelVonNeumann (commented):
I believe the approach to deleting remote users is too open (see comment).

PS: I think there could be some formatting to the code to improve readability. 
I'll hold off on those until there is a clear approach on the former topic in case the code needs to be refactored anyways.

authentik/providers/scim/clients/users.py
bitpavel-l25: @ImmanuelVonNeumann I'm agree with the point that remote users and groups that don't match Authentik internal objects shouldn't be deleted _by default_. That's why I've added a [`Purge Objects` switch input](https://github.com/goauthentik/authentik/pull/19697/changes#diff-ff588ec8b0658d7fdd84d64f5e91ae57ea123d6cdd24eb34436257897551d46aR266) which is disabled by default. 
Besides that, Authentik has at least [one known  bug](https://github.com/goauthentik/authentik/issues/20411) with users and groups deletion. The current MR would allow to resolve such issues.
I'm able to test this functionality only with AWS atm, this provider doesn't expose non-SCIM users and groups through SCIM API. It would be interesting to get details about various providers indeed.

bitpavel-l25: Hello @ImmanuelVonNeumann 
This PR fixes the bug described at https://github.com/goauthentik/authentik/issues/20411
When I set  'Outgoing sync trigger mode: Deferred until end' and run LDAP sync, Authentik behaves this way:
1. Users and groups that have been deleted at LDAP directory, are automatically deleted at Authentik
2. Related DB objects are also deleted by [cascade delete](https://github.com/goauthentik/authentik/blob/version/2025.12.3/authentik/providers/scim/models.py#L30). Including DB records for SCIM provisioned users and groups.
3. SCIM sync tasks are [skipped](https://github.com/goauthentik/authentik/blob/version/2025.12.3/authentik/sources/ldap/tasks.py#L161)
4. Full SCIM sync is started after the LDAP sync is completed, but DB doesn't contain any records related to the previously provisioned users/groups anymore.

Thus, your points 1 and 2 are not achievable with current logic because Authentik DB doesn't contain any records for provisioned users/groups at the full SCIM sync run. We either have to reconsider the behaviour of 'Outgoing sync trigger mode' and run some tasks synchronously, or remove cascade delete and clean up provisioned users/groups on the application side. Both options will cause much bigger changes and more questions to discuss. 
The 3rd option you proposed (to delete only remote objects with `externalId`) sounds reasonable. If it unblocks the current PR, I'll add it.

PR #19714, opened 2026-01-23
ci: make API client publish retryable

BeryJu:
we frequently run into race conditions when upgrading the API client after publishing. This separates the publish and the upgrade to make the upgrade retry-able

### DIFF
--- a/.github/workflows/api-ts-publish.yml
+++ b/.github/workflows/api-ts-publish.yml
@@ -16,6 +16,8 @@ permissions:
 jobs:
   build:
     runs-on: ubuntu-latest
+    outputs:
+      new_version: ${{ steps.publish.outputs.version }}
     steps:
       - id: generate_token
         uses: actions/create-github-app-token@29824e69f54612133e76f7eaac726eef6c875baf # v2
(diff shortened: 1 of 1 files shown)


dominic-r (approved):
lgtm but potential need for backoff/retries as discussed on slack

.github/workflows/api-ts-publish.yml
```diff
@@ -33,19 +35,33 @@ jobs:
         run: make gen-client-ts
       - name: Publish package
         working-directory: gen-ts-api/
+        id: publish
         run: |
           npm i
           npm publish --tag generated
+          export VERSION=`node -e 'console.log(require("../gen-ts-api/package.json").version)'`
```
dominic-r, line 42:
```suggestion
          export VERSION="$(node -p "require('./package.json').version")"
```

dominic-r: potential enhancement

dominic-r: * removes the ../gen-client-ts  as we can see the working-directory above 
* adds quotes instead of backticks for potential edge cases 
* uses node -p as it means basically "evaluate this expression and print the result"

rissson:
I'm moving this to its own repo today.

PR #19720, opened 2026-01-24
web: display custom attributes on admin view pages

### LINKED ISSUE #18625: Display custom atrributes under User Info in admin interface

Not necessarily a problem. You need to click edit user to be able to view attributes and it is not suitable for checking at a glance information that may be stored in a custom attribute like department, manager, etc.

When viewing a users info in the admin interface, it would be nice for it to list custom attributes without needing to click into edit user.

I would even add the option myself if there was an easy way to customize the admin interface.

### Additional context

_No response_

dominic-r:
Overview:

Add a reusable ak-object-attributes-card component that displays custom attributes on User, Group, and Device admin view pages.

This allows admins to see custom attributes directly on the overview tab without needing to open the edit form.

The component:
- Filters out system attributes (goauthentik.io/* prefixed keys)
- Optionally excludes the notes attribute
- Renders values based on type: booleans as status labels, arrays as comma-separated lists, objects as formatted JSON

Testing:

1. Navigate to Admin > Identity > Users > [any user]
2. Verify "Custom Attributes" card appears below Changelog
3. Add custom attributes via Edit form:
```
{
  "department": "Engineering",
  "employee_id": 12345,
  "is_contractor": false,
  "is_manager": true,
  "skills": ["Python", "TypeScript", "Go"],
  "office_location": {
    "building": "HQ",
    "floor": 3,
    "desk": "A-42"
  },
  "notes": "This should NOT appear in Custom Attributes card",
  "goauthentik.io/user/sources": ["should-be-filtered"]
}
```
4. Confirm they appear in the card, system attributes are hidden
5. Repeat for Groups and Devices

Screenshot:

image

image

Motivation:

Admins frequently need to view custom attributes on users, groups, and devices. Currently this requires clicking Edit and scrolling to the attributes field.

Closes: https://github.com/goauthentik/authentik/issues/18625

### DIFF
--- a/web/src/components/ak-object-attributes-card.ts
+++ b/web/src/components/ak-object-attributes-card.ts
@@ -4,12 +4,36 @@ import { AKElement } from "#elements/Base";
 
 import { type DescriptionPair, renderDescriptionList } from "#components/DescriptionList";
 
+import { isMatching, match, P } from "ts-pattern";
+
 import { msg } from "@lit/localize";
 import { CSSResult, html, TemplateResult } from "lit";
 import { customElement, property } from "lit/decorators.js";
 
 import PFCard from "@patternfly/patternfly/components/Card/card.css";
 import PFDescriptionList from "@patternfly/patternfly/components/DescriptionList/description-list.css";
+
+/**
+ * Formats a value for display based on its type.
+ */
+function formatValue(value: unknown): TemplateResult | string {
+    const jsonify = (v: Record<string | number | symbol, unknown> | unknown[]) =>
+        html`<code>
+            <pre style="margin: 0; white-space: pre-wrap;">${JSON.stringify(v, null, 2)}</pre>
+        </code>`;
+
+    const Stringable = P.union(P.string, P.number);
+    const isStringable = (s: unknown) => isMatching(Stringable, s);
+    const isStringableArray = (a: unknown[]) => a.every(isStringable);
+
+    return match(value)
+        .with(P.nullish, () => "-")
+        .with(P.boolean, (v) => html`<ak-status-label ?good=${v}></ak-status-label>`)
+        .with(Stringable, (v) => String(v))
+        .with(P.array(), (v) => (isStringableArray(v) ? v.join(", ") : jsonify(v)))
+        .with(P.record(P.string, P.any), (v) => jsonify(v))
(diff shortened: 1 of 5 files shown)


BeryJu (commented):
LGTM in screenshots

web/src/components/ak-object-attributes-card.ts
```diff
@@ -0,0 +1,112 @@
...
+            if (key.startsWith("goauthentik.io/")) return false;
+            if (this.excludeNotes && key === "notes") return false;
+            return true;
+        });
+    }
+
+    /**
+     * Formats a value for display based on its type.
+     */
+    private formatValue(value: unknown): TemplateResult | string {
```
kensternberg-authentik: This method never calls `this`, so it's a function, not a method. Every time you create this object, you create a copy of this function despite the fact that it never changes, which wastes memory and time. (Not a lot, I know). To avoid that, move it out of the class and make it a standalone function elsewhere in the file, then call it without `this`.  (Also, dereferencing `this` for something that never uses it wastes CPU.)

It's also one long conditional with some repetition.  We have an excellent conditional matching library.  Consider (warning, this is a rough draft off the top of my head before I've finished my first cup of coffee; it may be incomplete or buggy):

```
import { match, P, isMatching } from "ts-pattern";

function formatValue(value: unknown): TemplateResult | string {
    const jsonify = (s: Record<string | number | symbol, unknown> | Array<unknown>) =>
        html`<code>
            <pre style="margin: 0; white-space: pre-wrap;">${JSON.stringify(value, null, 2)}</pre>
        </code>`;

    const Stringable = P.union(P.string, P.number);
    const isStringable = (s: unknown) => isMatching(Stringable, s);
    const isStringableArray = (a: Array<unknown>) => a.every(isStringable);

    match(value)
        .with(P.nullish, () => "-")
        .with(P.boolean, (v) => html`<ak-status-label ?good=${v}></ak-status-label>`)
        .with(Stringable, (v) => String(v))
        .with(P.array, (v) => (isStringableArray(v) ? v.join(", ") : jsonify(v)))
        .with(P.object, (v) => jsonify(v))
        .otherwise((v) => String(v));
}
```

kensternberg-authentik (approved):
Approved, but with a suggestion.

Issue #16644, opened 2025-09-05
Feature Request: Only Show Relevant Sections in Admin Interface


ErikAhlund:
# Feature request: Show only admin sections the current user can view/edit (hide inaccessible sections)

## Summary  
When a user is granted **“Can access admin interface”**, the admin UI currently shows *all* top-level sections (Applications, Dashboards, Brands, etc.), even if the user lacks view/edit permissions for those objects. This results in a cluttered admin for delegated admins and often a page full of permission errors.  

I’m requesting that the admin interface only display sections (menu items, cards, tabs, list pages) for which the current user has at least one meaningful permission (e.g. `view`, `change`, `add`, `delete`). If the user has no permissions for a section, that section should be hidden from the UI.

---

## Motivation / Why this is important
- **Cleaner UX for delegated admins:** Client or tenant admins should see only the resources they manage. Showing everything is confusing and makes the admin experience noisy.  
- **Reduced errors & noise:** Right now the interface shows pages that then fail with permission errors — hiding inaccessible sections prevents those empty/errorful pages.  
- **Security by minimising information exposure:** Hiding sections avoids leaking the existence of admin sections or resources that the user shouldn’t know about.  
- **Lower support burden:** Users get fewer confusing options to click, reducing helpdesk tickets and misconfiguration attempts.  
- **Multi-tenant friendliness:** For setups that separate clients into groups, scoped admin views are essential for delegating safely.  

---

## Current behavior (problem)
1. Grant `Can access admin interface` to a delegated admin group (e.g. `client_admin`).  
2. The delegated admin can load the admin UI.  
3. The UI shows **all** admin sections (Applications, Dashboards, Brands, etc.), even for sections where the user has no `view`/`change`/`add`/`delete` permissions.  
4. Clicking into these sections results in permission errors or pages that cannot render correctly. The top-level UI remains cluttered and misleading.  

---

## Proposed behavior
- When rendering the admin UI, **filter top-level sections and navigation** to include only sections for which the current user has **at least one** of the following permissions:  
  - `view`  
  - `add`  
  - `change`  
  - `delete`  
  - or other actionable permissions (e.g. `manage_group_members`)  

- The permission check should be *model-based* or *section-based*:  
  - If the user has any permission on the model/resource represented by the section, show it.  
  - Otherwise hide it.  

- For custom sections that map to multiple permissions, show the section if the user has any permission required to meaningfully interact with that section.  

[...]

Issue #19232, opened 2026-01-07
Feature Request: Configurable Sidebar Navigation (Hide, Reorder, Pin Menus)


DNA-Bryan:
### Is your feature request related to a problem?

It would be very useful to have a way to customize the Authentik sidebar navigation, specifically:

Hide certain menu items

Rearrange the order of sidebar menus

Pin frequently used menus to a dedicated section

This would allow creating lighter and role-focused admin experiences, especially for delegated or limited admins.

### Describe the solution you'd like

1. Hide Sidebar Menus (Per Role or Group)

     Allow administrators to:

     Hide specific sidebar menu items based on:

           User role

           Group membership

           Permission set

           Optionally apply this globally or per tenant.

2. Rearrange Sidebar Menu Order

      Allow customization of the sidebar order:

           Drag-and-drop UI in admin settings or

           Configuration via role/group settings

3. Pin / Favorite Menus

      Add a “Pinned” or “Favorites” section at the top of the sidebar where users (or admins) can:

           Pin frequently used menu items

           Quickly access important sections without scrolling      

### Describe alternatives that you've considered

Currently, permissions are used to prevent certain users from accessing records, creating new ones, or modifying existing data. That said, having the ability to hide these sections altogether would greatly improve the experience for users who are not meant to interact with all features.

dewi-tik:
Due to how authentik works, 1 and 2 are not feasible. However, we're looking into the ability to PIN and considering it for a future release.

dominic-r:
Assigning for pinning. Will clean up and link existing pr

dominic-r:
We're closing this in favor of this parent issue: #25323

PR #19756, opened 2026-01-26
Filter admin sidebar items on load

sheyaln:
Provides a more refined and less noisy experience for non-superadmins.
addresses: https://github.com/goauthentik/authentik/issues/16644

Only requirement introduced is to add any new tabs and/or permissions to AdminSidebar.ts and ui_permissions.py

---

## Checklist

If applicable

### DIFF
--- a/authentik/admin/api/ui_permissions.py
+++ b/authentik/admin/api/ui_permissions.py
@@ -1,0 +1,126 @@
+"""Admin UI Permissions API"""
+
+from rest_framework.permissions import IsAuthenticated
+from rest_framework.request import Request
+from rest_framework.response import Response
+from rest_framework.views import APIView
+
+from authentik.core.api.utils import PassiveSerializer
+from rest_framework.fields import BooleanField
+
+
+class UIPermissionsSerializer(PassiveSerializer):
+    """Serializer for UI permissions"""
+
+    can_view_admin_overview = BooleanField(read_only=True)
+    can_view_system_tasks = BooleanField(read_only=True)
+
+    can_view_applications = BooleanField(read_only=True)
+    can_view_providers = BooleanField(read_only=True)
+    can_view_outposts = BooleanField(read_only=True)
+
+    can_view_devices = BooleanField(read_only=True)
+    can_view_device_groups = BooleanField(read_only=True)
+    can_view_connectors = BooleanField(read_only=True)
+
+    can_view_events = BooleanField(read_only=True)
+    can_view_notification_rules = BooleanField(read_only=True)
+    can_view_notification_transports = BooleanField(read_only=True)
+    can_view_data_exports = BooleanField(read_only=True)
+
+    can_view_policies = BooleanField(read_only=True)
+    can_view_property_mappings = BooleanField(read_only=True)
+    can_view_blueprints = BooleanField(read_only=True)
+    can_view_files = BooleanField(read_only=True)
+    can_view_reputation = BooleanField(read_only=True)
+
+    can_view_flows = BooleanField(read_only=True)
+    can_view_stages = BooleanField(read_only=True)
+    can_view_prompts = BooleanField(read_only=True)
+
+    can_view_users = BooleanField(read_only=True)
+    can_view_groups = BooleanField(read_only=True)
+    can_view_roles = BooleanField(read_only=True)
+    can_view_initial_permissions = BooleanField(read_only=True)
+    can_view_sources = BooleanField(read_only=True)
+    can_view_tokens = BooleanField(read_only=True)
+    can_view_invitations = BooleanField(read_only=True)
+
+    can_view_brands = BooleanField(read_only=True)
+    can_view_certificates = BooleanField(read_only=True)
+    can_view_outpost_integrations = BooleanField(read_only=True)
+    can_view_settings = BooleanField(read_only=True)
+
+    can_view_licenses = BooleanField(read_only=True)
+
+
+class UIPermissionsView(APIView):
+    """Return UI permissions for the current user"""
+
+    permission_classes = [IsAuthenticated]
+
+    def get(self, request: Request) -> Response:
+        """Get UI permissions for current user"""
+        user = request.user
+
+        permissions = {
+            "can_view_admin_overview": True,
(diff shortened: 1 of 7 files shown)


dominic-r:
cc @gergosimonyi

dewi-tik:
Thank you for the PR. We agree with the idea of this and think that it would be an improvement but we're not going to use this implementation because it doesn't take object permissions into account and statically defines all permissions.

We'll look into creating our own PR for this in future. Thanks again.

sheyaln:
Thanks for the response @dewi-tik. 
Anything I can do to encourage this feature? Any additional feedback that I can take into account and try again?

dewi-tik:
> Thanks for the response @dewi-tik. Anything I can do to encourage this feature? Any additional feedback that I can take into account and try again?

@gergosimonyi and @BeryJu can give you answers on this.

gergosimonyi:
Hi sheyaln, thank you for the PR! This feature was discussed internally recently, you can see another implementation attempt [here](https://github.com/goauthentik/authentik/pull/19317). I voiced my opinions against this, my argument was different from Dewi's summary above, so I'll record it here.

TL;DR: my argument is that the potential drawbacks of this feature far outweigh the potential benefits.

---

**Benefits**

I am mostly convinced that one benefit of this feature is that it provides a cleaner user interface. I see how this can improve a user's experience and lessen confusion. However, I believe the benefits stop there. I did take a look at #16644 as well, I remain unconvinced that the 4 other items claimed as benefits are really beneficial, _especially_ the safety/security points.

How great is this benefit? I'm not sure. I would believe that most users would quickly learn not to bother clicking on a page on which they got "access denied". From then on, the benefit is, in my opinion, purely visual. ((Of course my view is probably biased in my favor. I am trying very hard to reduce that bias, so please bear with me. I'm happy to hear counterarguments.))

**Drawbacks**

The first drawback I'll explore is the big one. This should convince anyone to, at the very least, make this feature optional instead of forcing it on the user.

1. I believe it is inevitable that we'll make a programmer error when implementing this and a programmer error in this feature is unacceptable.

I believe a programmer error is inevitable because there are many ways to gain a permission and there are many permission combinations that should result in a page being shown, such as:
- in addition to `view_user`, the `add_user` permission should also enable `Directory > Users`
- in addition to global permissions, object permissions should also enable pages
- deciding whether to show pages may require nontrivial logic (e.g. `System > Settings` or even inheritence-based resources like `Applications > Providers`)
- https://github.com/goauthentik/authentik/issues/19643 will create custom permissions that will further alter the rules in ways the code can't predict
- (but even currently, [custom code](https://docs.goauthentik.io/install-config/configuration/#custom-python-settings) may alter how permissions behave -- anyone using this currently would have to edit the frontend as well to accommodate for such changes. I put this in parentheses because we don't officially support `user_settings.py`.)

((I'm not trying to chew you out, but I'd like to note that neither this PR or the linked PR considers any of these, which illustrates how easy it is to make an error.))

The impact of a programmer error is that someone can't navigate to a page where they need to do their work. The only recourse is to patch authentik (or wait for the authentik team for a patch). That risk in my opinion is unacceptable.

I won't go into this much detail for the rest of my list, but here they are.

On the technical side:

2. I don't see an easy way to do this without significant computational cost (even with aggressive caching).
3. The maintenance burden seems extremely large. I see no way to reasonably automate tests for this feature. This, over time, results in drawback 1, a programmer error.

And on the subjective side:

4. This feature results in unknown unknowns. Personally, I hate working with software that doesn't let me know what I _could_ do if I just asked my admin for permissions.
5. This feature results in shifting menus when permissions are added/removed, which can break muscle memory. Personally, I view this drawback to be roughly on par with the benefit.

---

Rant over, thank you for reading however much you read of it. Now let me present an alternative feature set that might result in lower cognitive load for the user.

- Pinned items. Each user should have the option to mark any navigation item as "pinned". These items show up at the top of the navigation bar (perhaps in addition to their already occupied space). (Perhaps they are also orderable?)
- Hidden items. Each user should have the option to mark any navigation item as "hidden". These items only show up in a (perhaps visually quieter) collection at the bottom of the navigation bar.
- Store the menu state. The accordion should keep closed/open state information so users can further customize their clutter level across page loads.
- Better error messages. The current "failed to fetch objects" message is not very helpful. We should make it more friendly and maybe even present a hint as to which permission(s?) the user should request from their admin to see the contents.

---

That was a lot. What do you think?

sheyaln:
Thanks for the thorough response @gergosimonyi!

Makes a lot of sense. Your suggested feature set takes us most of the way there, imo. Only thing I would like to add to your suggested feature set is admin-defined _starter_ hide/show configurations. Should I cheer on the team from the sidelines, or take a few unexperienced stabs at it?


Anecdotal opinions:
> in addition to global permissions, object permissions should also enable pages

That's a feature in Authentik that I haven't seen in many other IdPs. This makes this _much less_ of a solved issue than I originally thought.

>((I'm not trying to chew you out, but I'd like to note that neither this PR or the linked PR considers any of these, which illustrates how easy it is to make an error.))

Makes perfect sense. Chewing out can be necessary.

> I did take a look at https://github.com/goauthentik/authentik/issues/16644 as well, I remain unconvinced that the 4 other items claimed as benefits are really beneficial, especially the safety/security points.

As far the safety/security points in 16644, I think it's security through obscurity, which is a debate I'm unqualified to participate in. I do think leaving the user with unknown unknowns can be a feature in some circumstances. I can show you 10 slack messages this month saying "hey @sheyaln Authentik admin is broken, says failed to fetch object". 

I think your bias is that you're a **_very_** technical and technically literate person. Other technical people are not always as technically literate, as oxymoronic as that sounds. 
Based on anecdotal experience training users on other pinned/hidden items UIs: Even with pinned items, the clutter can still be enough of a cognitive load to cause enduser mistakes. Not catastrophic, like a programmer mistake would be, but it can be operationally expensive over time.

gergosimonyi:
https://github.com/goauthentik/authentik/issues/19232 also requested something like this (including pinning) so it seems like visual clutter is starting to affect multiple parties. We'll definitely keep this on the agenda as well.

We gladly welcome any contribution if you feel like it and have the time :)

PR #19759, opened 2026-01-26
revert: website/integrations: wazuh: Change exchange key generation to 64 bytes

dewi-tik:
Reverts goauthentik/authentik#18769

See https://github.com/goauthentik/authentik/pull/18769#issuecomment-3799000848


dominic-r (approved):
How did I miss that 😭

PR #19768, opened 2026-01-26
web: Enforce `challenge` nullish types.

GirlBossRush:
## Details

This PR enforces the actual type of the `challenge` property in flow stages, i.e. consistently checking for nullish values.

Closes:  https://github.com/goauthentik/authentik/issues/19774


kensternberg-authentik (approved):
Routine. I've seen a lot of shotgun surgery already this week, and it's already Monday.  What can we do to bring these patterns into code?

PR #19775, opened 2026-01-27
web/flows: fix stray "}" rendered in AccessDeniedStage

dominic-r:


dominic-r:
Pulled in with https://github.com/goauthentik/authentik/pull/19768/commits

PR #19788, opened 2026-01-27
web: Session UI Config Lifecycle

GirlBossRush:
## Details

This PR fixes an issue which prevented the UI config from synchronizing after the session context is refreshed: 

- Added a generic `applyThemeChoice` function to validate and assign values to `document.dataset.themeChoice`. This can also be used in a future UI element to toggle the preferred scheme

### DIFF
--- a/web/src/elements/mixins/session.ts
+++ b/web/src/elements/mixins/session.ts
@@ -29,6 +29,18 @@ export const SessionContext = createContext<APIResult<SessionUser>>(
 export type SessionContext = typeof SessionContext;
 
 /**
+ * The Lit context for the user interface configuration.
+ *
+ * @category Context
+ * @see {@linkcode WithSession}
+ */
+export const UIConfigContext = createContext<Readonly<UIConfig>>(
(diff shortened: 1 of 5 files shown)


web/src/elements/mixins/session.ts
```diff
@@ -116,7 +126,16 @@ export const WithSession = createMixin<SessionMixin>(
             })
             public [kAKLocale]!: LocaleContextValue;
 
-            #uiConfig: Readonly<UIConfig> = DefaultUIConfig;
+            @consume({
+                context: UIConfigContext,
+                subscribe,
+            })
+            @property({ attribute: false, useDefault: true })
```
GirlBossRush, line 133: @kensternberg-authentik After a few rounds of testing I can confirm that `useDefault: true` without `reflect` appears to have a desired effect, preventing the host's `updated` callback from invoking on the first render.

kensternberg-authentik: Cool.  Leave as is.  :-)

kensternberg-authentik (approved):

PR #19844, opened 2026-01-29
root: fix NPM_VERSION in Makefile

gergosimonyi:
Some of us only have `python` through `uv` :)


BeryJu (commented):
should probably move it down to use `$(UV)` otherwise it might try to sync packages and break?

rissson (approved):

dominic-r (approved):

PR #19864, opened 2026-01-29
web: add delete button to user detail page

### LINKED ISSUE #18191: Ability to delete a user from a user's detailed page
A small QoL of not having to go back to the user list, select the user, and click Delete.

Could go hand in hand with the task for a user panic button. yk since we're adding more buttons

dominic-r:
Overview:

Add a "Delete" button directly on the user detail page. Extracts common user deletion logic (usedBy check and delete API calls) into shared utility functions.

Testing:

1. Create a test user
2. Navigate to Directory > Users > [your user]
3. Click Delete

Motivation:

Closes: https://github.com/goauthentik/authentik/issues/18191


kensternberg-authentik (approved):
Straightforward.

web/src/admin/users/UserViewDeleteForm.ts
```diff
@@ -0,0 +1,18 @@
...
+import { customElement } from "lit/decorators.js";
+
+@customElement("ak-user-view-delete-form")
+export class UserViewDeleteForm extends UserDeleteForm {
+    onSuccess(): void {
+        super.onSuccess();
+        // Navigate to user list after successful deletion
+        window.location.hash = "#/identity/users";
+    }
+}
```
GirlBossRush:
```ts
import { navigate } from "#elements/router/RouterOutlet";
```

```suggestion
    protected override onSuccess(): void {
        super.onSuccess();

        navigate("/identity/users")
    }
}
```

web/src/elements/user/utils.ts
```diff
@@ -40,3 +42,22 @@ export class UserDeleteForm extends DeleteForm {
...
+}
+
+/**
+ * Delete a user via the API.
+ */
+export function deleteUser(user: User) {
+    return new CoreApi(DEFAULT_CONFIG).coreUsersDestroy({
+        id: user.pk,
+    });
+}
```
GirlBossRush, line 51: There's some changes to the API client that will make this brittle in the near future.

```suggestion

```

web/src/admin/users/UserListPage.ts
```diff
@@ -179,16 +180,8 @@ export class UserListPage extends WithBrandConfig(
...
-                        id: item.pk,
-                    });
-                }}
-                .delete=${(item: User) => {
-                    return new CoreApi(DEFAULT_CONFIG).coreUsersDestroy({
-                        id: item.pk,
-                    });
-                }}
+                .usedBy=${(item: User) => userUsedBy(item)}
+                .delete=${(item: User) => deleteUser(item)}
```
GirlBossRush:
```ts
class SomeElement extends AKBase {
  #api = new CoreApi(DEFAULT_CONFIG)
}
```

```suggestion
                .usedBy=${(item: User) => this.#api.coreUsersUsedByList({ id: item.pk })}
                .delete=${(item: User) => this.#coreUsersDestroy({ id: item.pk })}
```

GirlBossRush (changes requested):

web/src/admin/users/UserViewDeleteForm.ts
```diff
@@ -0,0 +1,18 @@
+import { UserDeleteForm } from "#elements/user/utils";
+
+import { customElement } from "lit/decorators.js";
+
+@customElement("ak-user-view-delete-form")
+export class UserViewDeleteForm extends UserDeleteForm {
+    onSuccess(): void {
+        super.onSuccess();
+        // Navigate to user list after successful deletion
+        window.location.hash = "#/identity/users";
```
BeryJu: This is the main reason we dont have these buttons. Ideally we'd navigate back in the history, but that might be a different tab. Else you might've gotten to this page from any other URL that links to a user. Even with this simplified approach you'd loose any search queries/pagination/user paths selection

BeryJu (changes requested):

web/src/admin/users/UserViewDeleteForm.ts
BeryJu: (Also I dont like implementing these things in a specific way for just one object, if anything they should be added to all objects that have a *view page)

PR #19867, opened 2026-01-29
providers/oauth2: allow cross-provider token introspection

### LINKED ISSUE #11597: 2024.8.3 /o/application/introspect on OIDC tokens broken
**Describe the bug**
When I call into the OIDC provider's `/o/application/introspect` endpoint, or trigger it via the traefik-forward-auth outpost, it always return `{"isActive":"false"}`.
The bug happens in `2024.8.3` but is fine in 2024.8.2

**To Reproduce**
For distilled version:
1. Create an OIDC provider that can login via Google or Facebook
2. Login through the login with your UI until you hit your protected web page, get the `access_token` in the chrome Developer tools Network tab when you call into the authentik endpoints to get user information;
3. Call into your `introspect` endpoint like this: `curl -XPOST "https://authentik.<your_domain>/application/o/introspect/" -d '{"client_id":"<your_client_id>","client_secret":"<your_client_secret>","token":"<access_token_from_ui>"}'`

The above step is good enough to reproduce the bug.  The <client_secret> is a bit hard to find sometimes, we may need to go into authentik database to get it but I believe the devs know how to get it even if it is a "public" OIDC that does not show secrets.

**Expected behavior**
The response returns information json and "is_active" equals to true

**Screenshots**

**Logs**
No logs

-   authentik version: [2024.8.3] - but not a bug in 2024.8.2
-   Deployment: [kubernetes]

**Additional context**

### LINKED ISSUE #11618: Allow cross provider oauth2 token introspection and restrict to confidential clients

harryyoud:
## Details

closes #11597
closes #11618

Since #11537, you can no longer introspect tokens created by another provider. This creates problems when using one OAuth2 provider to authenticate to an API protected by another provider.

My example is Stalwart mail server (which supports OAuth2 authentication), with Roundcube mail server. Currently, I have to use the userinfo endpoint, which is no longer recommended because Stalwart cannot check the audience of the token (see https://www.reddit.com/r/stalwartlabs/comments/1q8otw8/comment/nzfigmr/).

With this patch, you can now allow other providers tokens to be used at the introspection endpoint on a whitelist basis. This is in keeping with the security patch for #11537, but allows the flexibility of cross-provider token introspection.

---

## Checklist

If applicable


authentik/providers/oauth2/views/introspection.py
```diff
@@ -46,13 +46,14 @@ def from_request(request: HttpRequest) -> TokenIntrospectionParams:
...
-        access_token = AccessToken.objects.filter(token=raw_token, provider=provider).first()
+        access_token = AccessToken.objects.filter(token=raw_token, provider__in=providers).first()
         if access_token:
-            return TokenIntrospectionParams(access_token, provider)
-        refresh_token = RefreshToken.objects.filter(token=raw_token, provider=provider).first()
+            return TokenIntrospectionParams(access_token, access_token.provider)
+        refresh_token = RefreshToken.objects.filter(token=raw_token, provider__in=providers).first()
         if refresh_token:
-            return TokenIntrospectionParams(refresh_token, provider)
+            return TokenIntrospectionParams(refresh_token, refresh_token.provider)
```
harryyoud, line 56: Please may I have some advice on whether we should be returning the authenticated provider or the original token's provider in the introspection response - this only influences the `client_id` part of the token JSON response

authentik/providers/oauth2/views/introspection.py
```diff
@@ -46,13 +46,14 @@ def from_request(request: HttpRequest) -> TokenIntrospectionParams:
...
         if not provider:
             raise TokenIntrospectionError
+        providers = [provider, *provider.allowed_provider_tokens.all()]
 
-        access_token = AccessToken.objects.filter(token=raw_token, provider=provider).first()
+        access_token = AccessToken.objects.filter(token=raw_token, provider__in=providers).first()
         if access_token:
-            return TokenIntrospectionParams(access_token, provider)
-        refresh_token = RefreshToken.objects.filter(token=raw_token, provider=provider).first()
+            return TokenIntrospectionParams(access_token, access_token.provider)
```
harryyoud, line 53: Here also, see comment below

harryyoud:
Tested; working well for me. I've published a testing server docker image at https://github.com/harryyoud/authentik/pkgs/container/authentik with these changes.

harryyoud:
@BeryJu are you happy to review this since you commented: https://github.com/goauthentik/authentik/issues/11597#issuecomment-3308564774

dewi-tik:
Closed by #21513

PR #13124, opened 2025-02-19
admin: clean up Recovery process by admin

gergosimonyi:
## Details

Closes #9671
Closes #12445 (Therefore, in this regard supersedes #12591)

---

## Checklist

-   [ ] Local tests pass (`ak test authentik/`)
-   [ ] The code has been formatted (`make lint-fix`)

If an API change has been made

-   [ ] The API schema has been updated (`make gen-build`)

If changes to the frontend have been made

-   [ ] The code has been formatted (`make web`)

If applicable

-   [ ] The documentation has been updated
-   [ ] The documentation has been formatted (`make website`)

PR #19875, opened 2026-01-30
core: ask for token duration on recovery link/email by admin

gergosimonyi:
An updated version of #13124


authentik/core/api/users.py
```diff
@@ -564,18 +566,24 @@ def _create_recovery_link(self, for_email=False) -> tuple[str, Token]:
...
+                {"non_field_errors": _("Recovery flow not applicable to user")}
             ) from None
         _plan = FlowToken.pickle(plan)
         if for_email:
             _plan = pickle_flow_token_for_email(plan)
+        if token_duration:
+            timedelta_string_validator(token_duration)
+            expires = now() + timedelta_from_string(token_duration)
+        else:
+            expires = default_token_duration()
```
BeryJu:
```suggestion
        expires = default_token_duration()
        if token_duration:
            timedelta_string_validator(token_duration)
            expires = now() + timedelta_from_string(token_duration)
```

authentik/core/api/users.py
BeryJu: use `@validate` instead of raw parameters and then we can get rid of manual validation

authentik/core/api/users.py
```diff
@@ -752,27 +773,33 @@ def recovery(self, request: Request, pk: int) -> Response:
...
+            email_stage := EmailStage.objects.filter(pk=email_stage_uuid).first()
+        ):
+            LOGGER.debug("Email stage does not exist")
+            raise ValidationError({"non_field_errors": _("Email stage does not exist.")})
+        if not request.user.has_perm("authentik_stages_email.view_emailstage", email_stage):
+            LOGGER.debug("User has no view access to email stage")
+            raise ValidationError(
+                {"non_field_errors": _("User has no view access to email stage.")}
+            )
+        token_duration = request.query_params.get("token_duration", "")
```
BeryJu: with serializer this isn't required

authentik/lib/utils/uuid.py
```diff
@@ -0,0 +1,9 @@
+import uuid
+
+
+def is_uuid_valid(str: str):
+    try:
+        uuid.UUID(str)
+        return True
+    except ValueError:
```
BeryJu: should also not be needed with validator

BeryJu (approved):

PR #19882, opened 2026-01-30
Update Audiobookshelf doc

Jammrock:
## Details

The Audiobookshelf guide assumes subdomains are used and subfolders are not.

Newer ABS OIDC settings defaults expect the `/audiobookshelf` subfolder. Following this doc and not catching the ABS subfolder setting, or understanding what it does, causes a Redirect URI Error in Authentik.

This change clarifies subdomain versus subfolder steps to prevent future headaches for new users.

---

## Checklist

-   [x ] Local tests pass (`ak test authentik/`)

This is a documentation change. No code was updated.

If an API change has been made - N/A

If changes to the frontend have been made - N/A

If applicable

-   [x ] The documentation has been updated
-   [ x] The documentation has been formatted (`make docs`)

### DIFF
--- a/website/integrations/media/audiobookshelf/index.md
+++ b/website/integrations/media/audiobookshelf/index.md
@@ -15,6 +15,9 @@ support_level: community
 The following placeholders are used in this guide:
 
 - `audiobookshelf.company` is the FQDN of the Audiobookshelf installation.
+  - Most steps assume this FQDN includes a sub-domain, when used.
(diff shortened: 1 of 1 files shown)


dominic-r:
Thank you for this contribution. Unfortunately, we won't be proceeding with it due to the notice on L16 regarding the FQDN already taking care of the changes suggested in this PR.

PR #19887, opened 2026-01-31
web/admin: source forms not rendering

kensternberg-authentik:
web/admin: source-forms-not-rendering

# What

Replaces the logic for determining types in the `StrictUnsafe` directive such that all types are assessed for `isProperty` first, and if it’s not a property, `String()` types are passed as an attribute, not a property. Just checking the type for `Boolean` is not sufficient.

Replaces the logic for rendering the SourceForms to ensure that forms that do not need a model name are not passed a model name. Run-time type-checking was failing for forms that do not take a model name because they already know it.

# Why

This looks like a case of excessive cleverness and insufficient testing. Trying to abstract the creation of the models down to a single call without breaking the code is an admirable goal, but this is fragile code because of the demands of the different models, especially the OAuth2 models which have different names depending on the uniqueness of the source (Discord vs Azure vs Mailcow, etc.).

# Incomplete

The code also suffers from a second level of cleverness in that it delays the render of the form until the modal is made visible. This works for the modal for creating new sources, and it seems to work fine on the “View One Source -\> \[Edit\]” case, but the edit button on the SourcesList page does not work.


web/src/admin/sources/SourceWizard.ts
```diff
@@ -63,14 +63,14 @@ export class SourceWizard extends AKElement {
                 >
                 </ak-wizard-page-type-create>
                 ${this.sourceTypes.map((type) => {
+                    const { modelName } = type;
+                    const props = modelName.includes("oauthsource") ? { modelName } : {};
```
kensternberg-authentik, line 67: Only Oauth needs modelName; the rest have singular model names and know it internally.

web/src/elements/utils/unsafe.ts
```diff
@@ -35,6 +35,21 @@ export function isAKElementConstructor(input: CustomElementConstructor): input i
     return Object.prototype.isPrototypeOf.call(AKElement, input);
 }
 
+function getPrefix(type: unknown, isProperty: boolean) {
```
kensternberg-authentik, line 38: You must check if a property _is a Property_, and only then perform some judgement for Boolean.  Also, there's no way this code understands events.  Suggestion: emulate `spread` in its entirely and do the annotations yourself, since you, the human, are the only one who knows if a function being passed in is a property (like a formatter or something) or an event handler.

GirlBossRush (approved):

PR #19999, opened 2026-02-04
web/flow: refactor flow executor so component selection is in an easy-to-maintain table

kensternberg-authentik:
web/flow: refactor FlowExecutor so that client-side stage selection is separate from stage execution

# What

Extracts and normalizes the *massive* switch/case statement into a table, eliminating as much repetition as possible. Where the server-side stage token and the client-side component have the same tag, only one is required. There were three different patterns for prop definitions, and those have been regularized into an expression with a compile-time type check, and the most common one can be omitted from the stage definition table.

# Why

1.  Because it’s hella cleaner. Stages are clear and easy to spot in the table (especially when it’s alphabetically ordered, OMG). Stages that disagree in name with their components, stages that take props different from the “standard” set, and stages that need `import` statements, are all easy to identify.
2.  Because identifying what we *do* with our web components is critical to their success, and to the success of the styling system the authentik web team envisions. FlowExecutor provides selection and execution of stages, but it also provides the inspector, the locale selector, headers, footers, customizations, and branding. Clearing away clutter to make that easier to see makes future refactoring for compatibility mode and dark theme handling much easier.


web/src/flow/FlowExecutor.ts
```diff
@@ -219,7 +216,7 @@ export class FlowExecutor
         WebsocketClient.close();
     }
 
-    protected refresh = (): Promise<void> => {
+    protected refresh = async () => {
```
kensternberg-authentik, line 218: I just do what the type-checker tells me, man.

GirlBossRush (approved):

PR #20029, opened 2026-02-04
web/flow: dedupe the set error flow state

kensternberg-authentik:
# What

Extracts the logic for setting the flow state to FlowError.

# Why

It was just duplication. Trying to clean up state management is easier when special state handling is isolated into a single method.

### DIFF
--- a/web/src/flow/FlowExecutor.ts
+++ b/web/src/flow/FlowExecutor.ts
@@ -8,7 +8,7 @@ import Styles from "./FlowExecutor.css" with { type: "bundled-text" };
 import { stages } from "./FlowExecutorSelections";
 
 import { DEFAULT_CONFIG } from "#common/api/config";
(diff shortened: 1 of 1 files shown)


web/src/flow/FlowExecutor.ts
```diff
@@ -210,19 +218,13 @@ export class FlowExecutor
             })
             .then((challenge) => {
                 this.challenge = challenge;
+                return !!this.challenge;
             })
             .catch(async (error) => {
                 const parsedError = await parseAPIResponseError(error);
```
kensternberg-authentik, line 224: Because I am a bear of very little brain, I wonder about this.  In the API, these two methods have the same return signature.  So why do we `await` on one error, and another?  I strongly suspect there is a mistake here.

```
FlowsApi.flowsExecutorSolve(...): Promise<ChallengeTypes>;
FlowsApi.flowsExecutorGet(...): Promise<ChallengeTypes>;
```

PR #20030, opened 2026-02-04
web/flow: generate a single API object for network transactions and use it for the lifetime of the FlowExecutor

kensternberg-authentik:
web/flow: dedupe the creation of fresh FlowApi instances

# What

Generates a single instance of FlowApi() that the FlowExecutor can use over the course of its lifetime.

# Why

Looking at the code generated by OpenApi, it’s clear that the parameters with which the API commits network transactions are immutable after construction; likewise, our particular invocation of `DEFAULT_GLOBALS` is also immutable with respect to a single instance of the FlowExecutor. With that in mind, there’s no reason to keep rebuilding the same network transaction object over and over; just instantiate it and live with it. In the conflict between rules-of-thumb “Never store what you can express” and “Extract repetitious expressions into instances,” the latter rule wins here.


web/src/flow/FlowExecutor.ts
```diff
@@ -196,7 +200,7 @@ export class FlowExecutor
         WebsocketClient.close();
     }
 
-    protected setFlowErrorChallenge(error: APIError) {
+    private setFlowErrorChallenge(error: APIError) {
```
GirlBossRush:
```suggestion
    #setFlowErrorChallenge(error: APIError) {
```

GirlBossRush (approved):

PR #20060, opened 2026-02-05
internal: define package variables for paths

tacerus:
## Details
During distribution packaging, files are typically installed in standardized paths (for example, adhering to the FHS). Avoid the need for patching the source code to facilitate this by allowing the paths needed by the server to be overwritten at build time.

The defaults are kept as they are to not affect existing behavior.

Example usage (simplified):
```
$ go build \
  -ldflags '-X "goauthentik.io/internal/gounicorn.manageCmd=/usr/bin/ak-manage" -X "goauthentik.io/internal/gounicorn.gunicornCmd=/usr/bin/gunicorn" -X "goauthentik.io/internal/gounicorn.gunicornConf=/etc/authentik/gunicorn.conf.py" -X "goauthentik.io/internal/gounicorn.pidDir=/run/authentik" -X "goauthentik.io/web.staticDir=/usr/share/authentik/web/dist"' ./cmd/server
```

Ideally, similar functionality would also exist for the various path references in the Python code. However, those require further consideration as there is no equivalent concept of build time options like there is in Go.

---

## Checklist

If applicable

### DIFF
--- a/internal/gounicorn/gounicorn.go
+++ b/internal/gounicorn/gounicorn.go
@@ -17,6 +17,13 @@ import (
 	"goauthentik.io/internal/utils"
 )
 
+var (
+	manageCmd    string = "./manage.py"
+	gunicornCmd  string = "gunicorn"
+	gunicornConf string = "./lifecycle/gunicorn.conf.py"
+	pidDir       string = ""
+)
+
 type GoUnicorn struct {
 	Healthcheck      func() bool
 	healthyCallbacks []func()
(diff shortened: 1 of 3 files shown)


internal/gounicorn/gounicorn.go
```diff
@@ -58,16 +65,16 @@ func New(healthcheck func() bool) *GoUnicorn {
 }
 
 func (g *GoUnicorn) initCmd() {
-	command := "./manage.py"
+	command := manageCmd
```
BeryJu, line 68: I dont think this is required for this, since manage.py is only used during development where you're probably running `go run ./cmd/server`

internal/gounicorn/gounicorn.go
```diff
@@ -58,16 +65,16 @@ func New(healthcheck func() bool) *GoUnicorn {
...
 	args := []string{"dev_server"}
 	if !config.Get().Debug {
-		pidFile, err := os.CreateTemp("", "authentik-gunicorn.*.pid")
+		pidFile, err := os.CreateTemp(pidDir, "authentik-gunicorn.*.pid")
```
BeryJu, line 71: Might not be needed since that'll just use TMP by default

BeryJu (commented):
Looks good for the most part; if you dont mind me asking what's the idea behind adjusting these paths?

internal/gounicorn/gounicorn.go
tacerus: Thanks for the pointer, I didn't realize it was overwritten after `!config.Get().Debug`. I think it might still be useful though if trying to debug against the system Python environment?

internal/gounicorn/gounicorn.go
tacerus: That was actually the reason, because I try to avoid storing files in world-writable directories. Indeed it could be set using `TMPDIR`, but it's not ideal because it would still default to `/tmp` on most systems if not set.

dewi-tik:
This code is on its way out - as of 2026.8. Therefore closing this PR. But thank you for the contribution

PR #20063, opened 2026-02-05
web/flow: separate flow inspector lifecycle from flow executor lifecycle

kensternberg-authentik:
web/flow: extract inspector into standalone lifecycle

# What

Removes all of the code from `FlowExecutor` related to the inspector and isolates it into its own component. The lifecycle of FlowExecutor’s inspector handling has been adjusted to maintain the existing behavior.

# How

FlowExecutor is reduced to merely presenting the button:

- In `FlowExecutor`:
  - Remove all the controls and references to FlowInspector
    - Remove the capabilities check
    - Remove the inspector listener
    - Remove the render guards
    - Remove the “inspect” PropVariant (and remove it from `FlowExecutorSelections`)
    - Remove the inspector toggle
    - Remove the inspector renderer
  - Always dispatch FlowAdvance events (if the inspector is not present they will be ignored)
  - Adjust `ak-stage-redirect` to not take “promptUser” as an attribute
  - Replace the whole render-inspector-button clause with `ak-flow-inspector-button`
  - Adjust CSS to use `ak-flow-inspector-button` instead of `.inspector-button`

RedirectStage now queries the context for inspector availability and state:

- In `stages/RedirectStage`:
  - Change `promptUser` from a property accessor to a simple accessor that queries the parent context for the inspector state
  - Remove the `@property` clause

`FlowInspectorButton` takes over these responsibilities, isolating this separate concern into a single file:

- Manages loaded, available, and open states
- Does the capabilities check
- Listens for FlowInspectorChangeEvents on the window object
- Renders nothing at all if the inspector is inaccessible or if the inspector is present and covering up the button
- On connectedCallback checks if the URL indicate the inspector should already be open
- Manages loading the FlowInspector on demand and toggles the drawer on state change

To my great surprise, `FlowInspector` itself required no changes.


web/src/flow/stages/RedirectStage.ts
```diff
@@ -41,6 +38,12 @@ export class RedirectStage extends BaseStage<RedirectChallenge, FlowChallengeRes
         return new URL(this.challenge?.to || "", document.baseURI).toString();
     }
 
+    get promptUser() {
```
kensternberg-authentik, line 43: Replace the attribute with a check on the actual state.  Removes the need for special code to "forward" the button state during execution of the flow.

web/src/flow/FlowExecutor.ts
```diff
@@ -363,7 +326,6 @@ export class FlowExecutor
             match(variant)
                 .with("challenge", () => challengeProps)
                 .with("standard", () => ({ ...challengeProps, ...litParts }))
-                .with("inspect", () => ({ ...challengeProps, "?promptUser": inspectorOpen }))
```
kensternberg-authentik, line 384: Special case to forward inspector state to the redirect state, mentioned above, removed.

web/src/flow/FlowExecutor.ts
```diff
@@ -313,10 +279,7 @@ export class FlowExecutor
                 flowChallengeResponseRequest: payload,
             })
             .then((challenge) => {
-                if (this.inspectorOpen) {
-                    window.dispatchEvent(new AKFlowAdvanceEvent());
-                }
-
+                window.dispatchEvent(new AKFlowAdvanceEvent());
```
kensternberg-authentik, line 288: The check isn't necessary; if the inspector is not present, these events will just be ignored.

web/src/flow/FlowExecutorSelections.ts
```diff
@@ -7,7 +7,7 @@ import "#flow/stages/RedirectStage";
 
 import { isMatching, match, P } from "ts-pattern";
 
-export const propVariants = ["standard", "challenge", "inspect"] as const;
+export const propVariants = ["standard", "challenge"] as const;
```
kensternberg-authentik: Removing the "inspect" special case.

web/src/flow/FlowExecutor.ts
```diff
@@ -66,10 +65,7 @@ import PFTitle from "@patternfly/patternfly/components/Title/title.css";
  * @prop {ChallengeTypes | null} challenge - The current challenge to render.
  */
 @customElement("ak-flow-executor")
-export class FlowExecutor
-    extends WithCapabilitiesConfig(WithBrandConfig(Interface))
-    implements StageHost
-{
+export class FlowExecutor extends WithBrandConfig(Interface) implements StageHost {
```
kensternberg-authentik, line 79: The only capability being checked for was `CanDebug`, and that's been moved into the FlowInspector code.

web/src/flow/FlowExecutorStageFactory.ts
```diff
@@ -18,7 +18,7 @@ import type { FlowChallengeComponentName, StageModuleCallback } from "#flow/type
 import { match, P } from "ts-pattern";
 
 export type { FlowChallengeComponentName, StageModuleCallback };
-export const propVariants = ["standard", "challenge", "inspect"] as const;
+export const propVariants = ["standard", "challenge"] as const;
```
kensternberg-authentik, line 21: Inspect is no longer a special case of stages; it's a special case of _one_ stage, the RedirectStage, which can now detect the need to trigger the Delayed Redirect by checking with the browser if an inspector is open and visible, so this case and the specialized attribute it signaled are no longer necessary.

GirlBossRush (approved):

PR #20083, opened 2026-02-06
web/flow: Type safe stage imports

GirlBossRush:
## Details

This PR builds on Ken's clean up, adding exhaustive type checking to the flow executor stage mappings. Similar to our [locale definitions](https://github.com/goauthentik/authentik/blob/main/web/src/common/ui/locale/definitions.ts#L33-L51), each flow challenge name is mapped to...

- A promise which resolves to a stage element constructor
- A variant (standard, challenge, inspect), followed by the promise constructor
- A variant, with a tag name matching the challenge name
- A variant, to a different tag name

For the majority of our stage imports, the imported module can be used to look up the tag name, avoiding an area of the code which can become unsynchronized. The stage mapping is also type checked to match all entries, effectively preventing a situation where the API can return a component name for which we do not have a corresponding element.

## How to review this PR

The `FlowExecutorSelections.ts` and `FlowExecutor.ts` files is where the action happens.The bulk of this PR is updating the existing flow stages to have a default export, ensuring that each import has the correct type.


web/src/flow/FlowExecutorStages.ts
```diff
@@ -0,0 +1,87 @@
...
+// @remarks
+// The different ways a stage can be associated with its server-side component are listed in the
+// type declaration above. The variants are meant to reduce the amount of information you have to
+// provide:
+//
+// - If the server-side component and the client-side tag are the same, only provide the component.
+// - Variants describe the attribute needs. There are only two variant: "standard" and "challenge."
+//   The "challenge" variant is for components that immediately issue redirects. "standard" is the
+//   default; you don't need to specify it.
+// - If the stage needs to be live immediately, import it above. Otherwise, provide an import
```
BeryJu, line 35: certain things qualify a stage for being instant imported, such as opening windows (like plex and apple) and certain webauthn interactions (like webauthnregister)

web/src/flow/FlowExecutorStages.ts
```diff
@@ -0,0 +1,87 @@
...
+// |---|,---|,---|    |\  |,---.. . .    `---.|--- ,---.,---.,---.,---.    |---|,---.,---.,---.
+// |   ||   ||   |    | \ ||---'| | |        ||    ,---||   ||---'`---.    |   ||---'|    |---'
+// `   '`---'`---'    `  `'`---'`-'-'    `---'`---'`---^`---|`---'`---'    `   '`---'`    `---'
+//                                                      `---'
+// @remarks
+// The different ways a stage can be associated with its server-side component are listed in the
+// type declaration above. The variants are meant to reduce the amount of information you have to
+// provide:
+//
+// - If the server-side component and the client-side tag are the same, only provide the component.
```
BeryJu, line 31: This should always be the case, is there anywhere where this isn't the case?

kensternberg-authentik: Yes, and they're documented right here in this file:

```
    ["ak-source-oauth-apple", "ak-flow-source-oauth-apple"],
    ["ak-source-plex", "ak-flow-source-plex"],
    ["ak-source-telegram", "ak-flow-source-telegram"],
    ["xak-flow-redirect", "ak-stage-redirect"],
```

web/src/flow/FlowExecutorStages.ts
kensternberg-authentik: Is this list sufficient guidance?  Or is there more?  If we can give guidance for when to supply a dynamic vs static import, that would be awesome.

PR #20096, opened 2026-02-07
sources/oauth: Fix InvalidAudienceError in id_token fallback

ryanpesek:
## Details
When using an OIDC source without a profile URL (relying on id_token claims fallback), authentication fails with InvalidAudienceError if the id_token contains an aud claim. The decode() call in OpenIDConnectClient.get_profile_info() doesn't pass an expected audience parameter. PyJWT validates the aud claim by default when present, causing the decode to fail.

---

## Checklist

If applicable

### DIFF
--- a/authentik/sources/oauth/types/oidc.py
+++ b/authentik/sources/oauth/types/oidc.py
@@ -49,8 +49,9 @@ class OpenIDConnectClient(UserprofileHeaderAuthClient):
             raw = get_unverified_header(id_token)
             jwk = PyJWKSet.from_dict(self.source.oidc_jwks)
             key = [key for key in jwk.keys if key.key_id == raw["kid"]][0]
-            return decode(id_token, key=key, algorithms=raw["alg"])
-        except PyJWTError, IndexError, ValueError:
+            return decode(id_token, key=key, algorithms=[raw["alg"]], audience=self.get_client_id())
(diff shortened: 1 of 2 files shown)


authentik/sources/oauth/types/oidc.py
```diff
@@ -49,8 +49,9 @@ def get_profile_info(self, token: dict[str, str]) -> dict[str, Any] | None:
             raw = get_unverified_header(id_token)
             jwk = PyJWKSet.from_dict(self.source.oidc_jwks)
             key = [key for key in jwk.keys if key.key_id == raw["kid"]][0]
-            return decode(id_token, key=key, algorithms=raw["alg"])
-        except PyJWTError, IndexError, ValueError:
+            return decode(id_token, key=key, algorithms=[raw["alg"]], audience=self.get_client_id())
```
BeryJu: Good point, we should probably also add `options={"verify_iss": False}` since we don't currently save the issuer of the OIDC provider and the id_token will likely include an `iss` claim

ryanpesek: It seems like this might already be the default behavior as the id_token from the provider I am testing with does return an `iss` claim. Happy to make the change though.

BeryJu (approved):

ryanpesek:
@BeryJu thanks for taking a look! Can this also be back-ported to 2025.12?

PR #20134, opened 2026-02-10
web: Flow Executor layout fixes

GirlBossRush:
## Details

This PR addresses feedback on the flow executor in a few scenarios:

- Styles for site footer links no longer depend on a shadow root, fixing alignment on error pages.
- The loading overlay's styles now have our latest compatibility mode hints, fixing an issue where the spinner appeared __on top__ of the content instead overlayed __atop__ the content.
- The flow card's min height now only applies when placeholder content is slotted, fixing _some_ oddly sized cards, such as on the error page.

### DIFF
--- a/web/src/flow/FormStatic.css
+++ b/web/src/flow/FormStatic.css
@@ -1,0 +1,41 @@
+:host,
+ak-form-static.style-scope {
+    margin-block-start: var(--pf-global--spacer--sm);
+    display: flex;
+    align-items: center;
+    justify-content: space-between;
+    flex-flow: wrap;
+    gap: var(--pf-global--spacer--sm);
+}
+
+.pf-c-avatar {
+    flex: 0 0 auto;
+}
+
+.primary-content {
+    display: flex;
+    align-items: center;
+    flex: 1 1 auto;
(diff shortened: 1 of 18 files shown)


web/src/flow/FormStatic.css
```diff
@@ -0,0 +1,41 @@
+:host,
+ak-form-static.style-scope {
```
GirlBossRush, line 2: Ported into separate file to clarify behavior. The `ak-form-static.style-scope` selector here is the primary change for this element.

BeryJu (approved):

web/src/flow/components/ak-brand-footer.ts
```diff
@@ -37,7 +38,9 @@ export class BrandLinks extends AKElement {
                 const children = sanitizeHTML(BrandedHTMLPolicy, link.name);
 
                 if (link.href) {
-                    return html`<li><a href="${link.href}">${children}</a></li>`;
+                    return html`<li part="list-item">
+                        <a part="list-item-link" href=${link.href}>${children}</a>
```
tacerus, line 42: Hi! Is the removal of the quotation marks around the `href`value here intentional?

PR #20141, opened 2026-02-10
website/docs: add email verification scope doc

dewi-tik:
## Details

Adds a section to the oauth scopes section about email verification. This can then be linked to when required by integration docs. Rather than duplicating the same info across the docs in different locations.

---

## Checklist


website/docs/add-secure-apps/providers/oauth2/frontchannel_and_backchannel_logout.mdx
```diff
@@ -1,5 +1,5 @@
 ---
```
dewi-tik, line 1: Changed this because the text is cut off in the sidebar.

website/docs/add-secure-apps/providers/oauth2/index.mdx
```diff
@@ -183,6 +183,28 @@ This does _not_ apply to special scopes, as those are not configurable in the pr
 - `user:email`: Allows read-only access to `/user`, including email address
 - `read:org`: Allows read-only access to `/user/teams`, listing all the user's groups as teams.
 
+### Email scope verification
+
+In authentik releases before 2025.10 the `email` scope set the `email_verified` claim to true. As we don't have a single source of whether a users' email is verified or not, and claiming that it is verified could lead to security implications, this claim now defaults to false.
```
dominic-r:
```suggestion
In authentik releases before 2025.10, the `email` scope set the `email_verified` claim to `true`. As we don't have a single source of whether a users' email is verified or not, and claiming that it is verified could lead to security implications, this claim now defaults to `false`.
```

dominic-r: my first impressions but gimme a bit more time to refine the wording pls

dominic-r: ok let's try this @dewi-tik 

```suggestion
In authentik releases prior to 2025.10, the email scope always set the `email_verified` claim to `true`. Since authentik does not have a single authoritative source to determine whether a user's email is actually verified, asserting this claim could have security implications. As of 2025.10, `email_verified` now defaults to `false`.
```

website/docs/add-secure-apps/providers/oauth2/index.mdx
```diff
@@ -183,6 +183,28 @@ This does _not_ apply to special scopes, as those are not configurable in the pr
...
+
+To successfully authenticate users, some applications may require this claim to be true, in which case you can create a custom `email` scope mapping that always returns `email_verified` as true:
```
dominic-r:
```suggestion
Some applications require this claim to be `true` in order to authenticate users. In those cases, you can define a custom email scope mapping that always returns `email_verified` as `true`:
```

website/docs/add-secure-apps/providers/oauth2/index.mdx
```diff
@@ -183,6 +183,28 @@ This does _not_ apply to special scopes, as those are not configurable in the pr
...
+
+```python
+return {
+    "email": request.user.email,
+    "email_verified": True
+}
+```
+
+For greater security, verify users' email addresses and store email verification status as a user attribute (e.g. `email_verified` (True/False)). Then create this scope mapping to return it dynamically:
```
dominic-r:
```suggestion
For greater security guarantees, verify users' email addresses and store the verification status as a user attribute (for example, `email_verified` set to `true` or `false`). You can then configure the scope mapping to return this value dynamically:
```

dominic-r (changes requested):

dominic-r (approved):

website/docs/add-secure-apps/providers/oauth2/index.mdx
```diff
@@ -183,6 +183,28 @@ This does _not_ apply to special scopes, as those are not configurable in the pr
 - `user:email`: Allows read-only access to `/user`, including email address
 - `read:org`: Allows read-only access to `/user/teams`, listing all the user's groups as teams.
 
+### Email scope verification
+
+In authentik releases prior to 2025.10, the email scope always set the `email_verified` claim to `true`. Since authentik does not have a single authoritative source to determine whether a user's email is actually verified, asserting this claim could have security implications. As of 2025.10, `email_verified` now defaults to `false`.
```
tanberry: We do already document this setting, over in Property mappings, but we could minimize that note and link to this one.

 https://docs.goauthentik.io/add-secure-apps/providers/property-mappings/#scope-mappings-with-oauth2

website/docs/add-secure-apps/providers/oauth2/index.mdx
```diff
@@ -183,6 +183,28 @@ This does _not_ apply to special scopes, as those are not configurable in the pr
...
+
+Some applications require this claim to be `true` in order to authenticate users. In those cases, you can define a custom email scope mapping that always returns `email_verified` as `true`:
+
+```python
```
tanberry, line 192: Can we say where to add this? In their "configuration file", wherever that might be? Anyhints?

website/docs/add-secure-apps/providers/oauth2/index.mdx
dewi-tik: Added that link

website/docs/add-secure-apps/providers/oauth2/index.mdx
dewi-tik: Added a note about where this is done

dominic-r (approved):

PR #20144, opened 2026-02-10
website/docs: endpoint/devices: add authentik version tags

dewi-tik:
## Details

Adds authentik version tags to all endpoint device docs - some users had issues with trying to run this on 2025.10

---

## Checklist

If applicable


BeryJu (commented):
These should not have a plus at the end, otherwise LGTM

dewi-tik:
> These should not have a plus at the end, otherwise LGTM

Fixed

dominic-r (approved):

PR #20175, opened 2026-02-11
core: option to reset reputation score to 0 on successful login

### LINKED ISSUE #20148: Option to automatically set reputation score to 0 upon successful login
An option/flag to automatically set reputation score back to 0 for a user/IP upon successful login. Rather than just increasing reputation score by one.

dominic-r:
Adds a tenant-level setting to reset reputation scores to 0 on successful login instead of incrementing by 1.

Closes: https://github.com/goauthentik/authentik/issues/20148


authentik/policies/reputation/signals.py
```diff
@@ -62,4 +62,11 @@ def handle_identification_failed(sender, request, uid_field: str, **_):
 @receiver(user_logged_in)
 def handle_successful_login(sender, request, user, **_):
     """Raise score for successful attempts"""
-    update_score(request, user.username, 1)
+    tenant = getattr(request, "tenant", get_current_tenant())
+    if tenant.reputation_reset_on_login:
+        remote_ip = ClientIPMiddleware.get_client_ip(request)
+        Reputation.objects.filter(identifier=user.username, ip=remote_ip).update(
+            score=0, expires=reputation_expiry()
+        )
```
gergosimonyi, line 70: This seems to set reputation to 0 even if the score was already positive. I don't think that's intended.

gergosimonyi (commented):
Is there any reason for reputation to go up 1 by 1 in the first place? What are we protecting against here? An attacker behind the same (real or cgNATted) IP that benefits from 5 retries per successful login instead of 1 retry?

Can't this behavior just be the new behavior instead of putting it behind yet-another-flag™?

dominic-r:
cc @dewi-tik and others for ^^

GirlBossRush (approved):
Looks good front-end wise. 

@gergosimonyi ?

dewi-tik:
> Is there any reason for reputation to go up 1 by 1 in the first place? What are we protecting against here? An attacker behind the same (real or cgNATted) IP that benefits from 5 retries per successful login instead of 1 retry?
> 
> Can't this behavior just be the new behavior instead of putting it behind yet-another-flag™?

That's actually a good point. Hadn't considered that.

authentik/policies/reputation/signals.py
dominic-r: The linked issue says:

> set reputation score back to 0 for a user/IP upon successful login

so I interpret it as "set to 0 always", but I can change the logic if you want?

dewi-tik:
Spoke with @BeryJu about this and he's happy for the default behavior to be changed as [suggested by Gergo](https://github.com/goauthentik/authentik/pull/20175#pullrequestreview-3782354065) rather than this be added as an option.

PR #20196, opened 2026-02-11
api: fix `test_build_schema`

gergosimonyi:
When already existing schema files are read-only, such as in `make test-docker`.

### DIFF
--- a/authentik/api/tests/test_schema.py
+++ b/authentik/api/tests/test_schema.py
@@ -1,6 +1,7 @@
 """Schema generation tests"""
 
 from pathlib import Path
+from tempfile import gettempdir
 from uuid import uuid4
 
 from django.core.management import call_command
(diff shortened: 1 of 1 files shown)


rissson (approved):

authentik/api/tests/test_schema.py
```diff
@@ -29,15 +30,13 @@ def test_browser(self):
 
     def test_build_schema(self):
         """Test schema build command"""
-        blueprint_file = Path("blueprints/schema.json")
-        api_file = Path("schema.yml")
-        blueprint_file.unlink()
-        api_file.unlink()
+        blueprint_file_name = f"/tmp/{str(uuid4())}.json"
+        api_file_name = f"/tmp/{str(uuid4())}.yml"
```
rissson:
```suggestion
        tmp = Path(gettempdir())
        blueprint_file_name = tmp / f"str(uuid4()).json"
        api_file_name = tmp / f"str(uuid4()).yml"
```
with the proper imports as well
```
from pathlib import Path
from tempfile import gettempdir
```

rissson (changes requested):

rissson (approved):

PR #20207, opened 2026-02-11
ci: fix release testing

gergosimonyi:


scripts/test_docker.sh
```diff
@@ -2,7 +2,7 @@
 set -e -x -o pipefail
 hash="$(git rev-parse HEAD || openssl rand -base64 36 | sha256sum)"
 
-AUTHENTIK_IMAGE="xghcr.io/goauthentik/server"
+AUTHENTIK_IMAGE="ghcr.io/goauthentik/server"
```
rissson:
```suggestion
AUTHENTIK_IMAGE="xghcr.io/goauthentik/server"
```

this shouldn't be changed. We are building the image here, and we *don't* want a valid name, to prevent accidental pushes.

rissson (changes requested):

scripts/test_docker.sh
rissson:
```suggestion
AUTHENTIK_IMAGE="authentik.invalid/goauthentik/server"
```

rissson (approved):

PR #20283, opened 2026-02-13
enterprise/lifecycle: use datetime instead of date to track review cycles

atereshkin:
## Details

This fixes the incorrect review Opened date display for new reviews described in #20265
The root cause of the problem was the use of a DateField for storing review opening date which made it ambiguous across time zones  (which combined with some questionable behavior of openapi-generator led to incorrect to incorrect dates in the UI). The fix is to switch to a DateTimeField. Extra precaution had to be taken because the task that applies lifecycle rules runs once a day and direct comparison of datetimes could lead to 24h delays in opening new reviews and reviews going overdue because of millisecond fluctuation in the task execution time.

 
---

## Checklist


GirlBossRush (approved):

authentik/enterprise/lifecycle/api/iterations.py
```diff
@@ -67,13 +68,15 @@ def get_object_verbose(self, iteration: LifecycleIteration) -> str:
     def get_object_admin_url(self, iteration: LifecycleIteration) -> str:
         return admin_link_for_model(iteration.object)
 
-    @extend_schema_field(DateField())
+    @extend_schema_field(DateTimeField())
     def get_grace_period_end(self, iteration: LifecycleIteration) -> date:
```
BeryJu:
```suggestion
    def get_grace_period_end(self, iteration: LifecycleIteration) -> datetime:
```

with this you shouldn't need `@extend_schema_field()`

gergosimonyi (approved):

PR #20301, opened 2026-02-14
rbac: fix role creation in permissions migration

gergosimonyi:


gergosimonyi:
No, this doesn't matter, there's a check for `group_id` on the first line of the function.

PR #20307, opened 2026-02-15
policies: handle lists in messages

tacerus:
## Details

So far, passing a list to ak_manage() used in a policy in a user settings flow would cause the UI to not only not print any messages, but also to "glitch" (sometimes the form disappears, sometimes the user is redirected to a different flow interface).
It is useful to support returning multiple messages from a policy - for example, with multiline prompts where individual lines yield different validation results, individual alerts are easier to parse for user than a long single line of text.
Hence solve the problem by enhancing the functionality to split list elements across messages, while keeping the existing behavior for strings.

---

## Checklist

If applicable

### DIFF
--- a/authentik/policies/expression/evaluator.py
+++ b/authentik/policies/expression/evaluator.py
@@ -33,9 +33,12 @@ class PolicyEvaluator(BaseEvaluator):
         self._context["ak_message"] = self.expr_func_message
         self._context["ak_user_has_authenticator"] = self.expr_func_user_has_authenticator
 
-    def expr_func_message(self, message: str):
+    def expr_func_message(self, message: list[str] | str):
         """Wrapper to append to messages list, which is returned with PolicyResult"""
-        self._messages.append(message)
+        if isinstance(message, list):
+            self._messages.extend(message)
+        else:
+            self._messages.append(message)
 
     def set_policy_request(self, request: PolicyRequest):
         """Update context based on policy request (if http request is given, update that too)"""
--- a/website/docs/customize/policies/expression.mdx
+++ b/website/docs/customize/policies/expression.mdx
@@ -40,14 +40,21 @@ Expression policies with sample Python that we have documented:
 
 ## Available functions
 
-### `ak_message(message: str)`
+### `ak_message(message: list[str] | str)`
 
 Add a message, visible by the end user. This can be used to show the reason why they were denied.
 
-Example:
+Example for a single message:
 
 ```python
 ak_message("Access denied")
+return False
+```
+
+Example for multiple messages:
+
+```python
+ak_message(["Line 1: Not authorized.", "Line 2: Incorrect length."])
 return False
 ```
 
--- a/authentik/policies/expression/tests.py
+++ b/authentik/policies/expression/tests.py
@@ -45,14 +45,23 @@ class TestEvaluator(TestCase):
         evaluator.set_policy_request(self.request)
         self.assertEqual(evaluator.evaluate(template).passing, True)
 
-    def test_messages(self):
-        """test expression with message return"""
+    def test_message(self):
+        """test expression with single message return"""
         template = 'ak_message("some message");return False'
         evaluator = PolicyEvaluator("test")
         evaluator.set_policy_request(self.request)
         result = evaluator.evaluate(template)
         self.assertEqual(result.passing, False)
         self.assertEqual(result.messages, ("some message",))
+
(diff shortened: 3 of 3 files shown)


website/docs/customize/policies/expression.mdx
```diff
@@ -40,17 +40,24 @@ Expression policies with sample Python that we have documented:
 
 ## Available functions
 
-### `ak_message(message: str)`
+### `ak_message(message: list | str)`
```
rissson:
```suggestion
### `ak_message(message: list[str] | str)`
```

authentik/policies/expression/evaluator.py
```diff
@@ -35,7 +35,10 @@ def __init__(self, policy_name: str | None = None):
 
     def expr_func_message(self, message: str):
```
rissson:
```suggestion
    def expr_func_message(self, message: list[str] | str):
```
might even be better to use Iterable? I'm not sure what list.extend accepts as argument

rissson (changes requested):

website/docs/customize/policies/expression.mdx
tacerus: Adjusted.

authentik/policies/expression/evaluator.py
tacerus: Adjusted - as the condition is specific to `list` I think `list[str]` makes more sense than `Iterable`. It seems `extend()` itself does use `Iterable`, however`.

authentik/policies/expression/evaluator.py
```diff
@@ -33,9 +33,12 @@ def __init__(self, policy_name: str | None = None):
...
         self._context["ak_user_has_authenticator"] = self.expr_func_user_has_authenticator
 
-    def expr_func_message(self, message: str):
+    def expr_func_message(self, message: list[str] | str):
         """Wrapper to append to messages list, which is returned with PolicyResult"""
-        self._messages.append(message)
+        if isinstance(message, list):
+            self._messages.extend(message)
+        else:
+            self._messages.append(message)
```
BeryJu, line 41:
```suggestion
    def expr_func_message(self, *messages: str):
"""Wrapper to append to messages list, which is returned with PolicyResult"""
        self._messages.extend(messages)
```

website/docs/customize/policies/expression.mdx
```diff
@@ -40,17 +40,24 @@ Expression policies with sample Python that we have documented:
...
 
 ```python
 ak_message("Access denied")
 return False
 ```
 
+Example for multiple messages:
+
+```python
+ak_message(["Line 1: Not authorized.", "Line 2: Incorrect length."])
```
BeryJu, line 57:
```suggestion
ak_message("Line 1: Not authorized.", "Line 2: Incorrect length.")
```

authentik/policies/expression/evaluator.py
tacerus: test_execution_logging fails with this, need to investigate

tacerus: Figured the tests out, but the bug at hand still exists with this change, passing a list will cause the settings UI to bug out with no alerts printed.

tacerus:
Just in case my explanation is confusing, here is a demonstration with the policy 

```
d = context.get('prompt_data', {}).get('attributes', {}).get('sshPublicKey', '').strip()
if d == 'valid':
  context['prompt_data']['attributes']['sshPublicKey'] = [d]
  return True
if d == 'err_list':
  ak_message(['there is a', 'problem'])
if d == 'err_str':
  ak_message('there is a problem!')
return False
```

_without_ my patch:


https://github.com/user-attachments/assets/85ed4595-5912-410c-a2bc-a6cea88a25ae

Background is having a box for SSH keys, and validating each line. For each invalid line, an error should be appended, and then returned as an alert.

(However, besides the desired functionality, I think the UI response to this case is not quite right.)

authentik/policies/expression/evaluator.py
BeryJu: @tacerus with that change above, you shouldn't pass a list but rather multiple arguments `ak_message("foo", "bar")` or spread the list `ak_message(*msgs)`

tacerus: Yes, with your suggestion I tried that variant, and if I recall correctly, that showed the same broken effect as when passing a list in the current state of the code.

PR #20325, opened 2026-02-16
core: recursion to get deepest child

gergosimonyi:
@dominic-r

### DIFF
--- a/authentik/lib/utils/inheritance.py
+++ b/authentik/lib/utils/inheritance.py
@@ -1,0 +1,41 @@
+from django.db.models import Model, OneToOneField, OneToOneRel
+
+
+def get_deepest_child(parent: Model) -> Model:
+    """
+    In multiple table inheritance, given any ancestor object, get the deepest child object.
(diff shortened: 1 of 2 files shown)


dominic-r (approved):

authentik/lib/utils/inheritance.py
```diff
@@ -0,0 +1,41 @@
...
+    submodel_accessors = [submodel_field.get_accessor_name() for submodel_field in submodel_fields]
+    # End Copy
+
+    child = None
+    for submodel in submodel_accessors:
+        try:
+            child = getattr(parent, submodel)
+            break
+        except AttributeError:
+            continue
```
BeryJu, line 37: should be a `hasattr` probably, this isn't directly in the hotpath but could be quite likely involved in it and throwing exceptions just for that is wasteful

gergosimonyi: I wager we'll save about a microsecond per execution with

```suggestion
        if hasattr(parent, submodel):
            child = parent.submodel
            break
```

:)

PR #20330, opened 2026-02-16
web: unify footer styles

tacerus:
## Details

So far, the footer in pages rendered in the server side templates was not formatted the same way as the footer defined in the JS parts. This was mainly noticeable by one footer being aligned centered, while the other was aligned to the left of the page.
Solve this by applying the footer styles in the login component as well.

Note this duplicates the "styles" block from
web/src/flow/components/ak-brand-footer.ts. I could not find where to place the block to have it be included from both places.

Please let me know if there's a good way to deduplicate these.

---

## Checklist

If applicable

### DIFF
--- a/authentik/core/templates/login/base_full.html
+++ b/authentik/core/templates/login/base_full.html
@@ -45,7 +45,7 @@
                             </div>
                         </main>
                         <footer aria-label="Site footer" class="pf-c-login__footer pf-m-dark">
-                            <ul class="pf-c-list pf-m-inline">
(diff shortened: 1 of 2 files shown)


GirlBossRush:
Hi @tacerus,

Thank you for your submission. Your intuition was correct and the fix makes sense. However, this _should_ be fixed in #20134. If you rebase this PR and can spot a visual improvement, I can give this another look :)

tacerus:
Hi @GirlBossRush, indeed, with your change, mine is no longer necessary. Thank you!

PR #20371, opened 2026-02-17
web/flow: separation of concerns, autoredirect, captchas and webauthn

kensternberg-authentik:
web/flow: re-arrange IdentificationStage for maintainability, part 2: AutoRedirect, Webauthn & Captcha

## What

These three subsystems have been removed from the IdentificationStage and put into their own Lit-Controllers.

### AutoRedirect

AutoRedirect is a _rule_ that triggers when a challenge arrives that meets certain conditions.  The existing code stated the rules is a mix of negative and positive checks, when really there's only one condition that has to be met, that these all be true:

        const onlyOneSource = sources.length === 1;
        const noUserAccessibleInputs = (userFields || []).length === 0;
        const noAlternativeMethods = !passwordlessUrl;

Also, it was checked _after_ the initial render, which seemed oddly wasteful; by the time you get to `willUpdate`, you know if you have a challenge that meets these conditions.  Moving the rule into a Controller and issuing a "change the challenge" event when this condition was met on `hostUpdate` makes this happen before rendering and removes a lot of Demeter violation.

### Webauthn

[Webauthn](https://webauthn.io/) is a mechanism that retrieves a stored secret key from something the user has, such as a Yubi-key or a vault on their browser; the standard DH-style cryptographic challenge compares a generated authentication request token to a corresponding public key. A Javascript component queries the browser for Webauth compatibility and, if it’s found, triggers a Webauthn transaction with which the user must interact independent of the stages’s HTML or presentation. If the transaction succeeds, the user is automatically sent to the redirect stage.

None of the code associated with Webauthn has anything to do with the stage rendering or event monitoring; it even has its own `submit()` phase, triggered automatically after the user grants the browser access to the passkey.

Therefore, I have placed it into a [Lit Controller](https://lit.dev/docs/composition/controllers/). The IdentificationStage activates the controller, and after the page has “settled” the controller checks if passkey use is authorized and that the browser has a passkey mechanism available, and if both are true, it sends the browser a request for the passkey transaction to begin and, once it has retrieved the credentials, transacts with the server on its own to see if redirection can proceed.

None of that needs to be in the IdentificationStage. That’s about 60 lines of code moving out of IdentificationStage.

### Captcha

Captcha is a little more invasive. It needs to track Captcha “freshness” (how old the last Captcha is), the current Captcha state and, because Captchas are loaded from third-parties, whether or not the Captcha loaded at all.

In the IdentificationStage, we need to conditionally render the Captcha, we need to handle if the Captcha failed, and we need to disable the button until the Captcha has loaded.

By moving “is there a captcha?” into a controller, the only references left in IdentificationStage are:

- activation of the controller at construction
- use a named accessor to delay enabling the submit button while the Captcha loads
- use a named method to handle the Captcha refresh-on-failure cycle
- use a named method to render the Captcha, which will return `nothing` when Captcha is not enabled

That’s about 40 lines of code moving out of IdentificationStage.

### Conclusion

That’s 100 lines of IdentificationStage replaced with 6 lines of controller references. “That’s a lot of pain moving out of the world.”

I did noticed that IdentificationPasskey shares a *lot* of code with the Webauthn Stage, but I haven’t looked into how to push that further up. (Yet).


web/src/flow/stages/identification/IdentificationStage.ts
```diff
@@ -93,138 +99,44 @@ export class IdentificationStage extends BaseStage<
...
 
-    public updated(changedProperties: PropertyValues<this>) {
+    constructor() {
+        super();
+        this.#captcha = new IdentificationCaptchas(this);
+        this.addController(this.#captcha);
+        this.#passkey = new IdentificationPasskey(this);
+        this.addController(this.#passkey);
+        this.#autoredirect = new IdentificationAutoRedirect(this);
+        this.addController(this.#autoredirect);
```
kensternberg-authentik, line 128: This particularly wordy variant of controller instantiation is the only one that convinces Typescript these private fields have been both initialized _and_ used.

web/src/flow/stages/identification/IdentificationStage.ts
kensternberg-authentik, line 77: The two above were static, so I've moved them out of the class; no need to reinstantiate them every time.

PR #20376, opened 2026-02-18
website/docs: move providers page under applications docs

gergosimonyi:
The providers overview page lived at `/providers`, outside the `/add-secure-apps/` hierarchy every other provider page uses. This moves its slug to `/add-secure-apps/providers` and adds `_redirects` rules so old `/providers/...` links — including deep links to individual provider pages — keep working.

While extending the redirect handling, the `docusaurus-theme/redirects` module was refactored: the rewrite logic is now built around a single compiled rule list, the parse → `RewriteIndex` → plugin-pair wiring that all three sites duplicated moved into a shared `createRedirectPlugins` helper, `finalDestination` stops on cyclic redirect chains instead of looping forever in the browser, and the module gained unit tests. The trailing-slash normalization CodeQL flagged as polynomial (`js/polynomial-redos`) is replaced with a plain string scan, here and in the same pattern in the llms-txt plugin.

This also carries a small cleanup pass over the docs build infrastructure: the API site now reuses the shared Algolia and footer config instead of a hardcoded copy (fixing its `externalUrlRegex` to exclude `api.goauthentik.io` rather than `docs.goauthentik.io`), and latent ESLint errors in the llms-txt module are fixed. Validated by running the website test suite, ESLint, and full builds of all three sites; the built output contains the expected `/providers/*` redirect pages.


dewi-tik:
We're in sync: https://github.com/goauthentik/authentik/pull/20379

dominic-r (approved):

dominic-r (approved):

PR #20396, opened 2026-02-18
web: Fix element property names with custom attributes.

GirlBossRush:
## Details

A follow up on #20373, this PR allows for properties which differ from their attribute e.g.

```ts
    @property({ attribute: "pk" })
    public instancePk: PKT;
```


kensternberg-authentik (approved):
A solid cleanup.

PR #20400, opened 2026-02-18
ci: Consistent NPM versions via Corepack

GirlBossRush:
## Details

**tl;dr Fewer "works on my machine" issues caused by NPM version drift between developer environments, CI runners, and Docker builds. Lockfile linting also prevents accidental lockfile mutations from slipping through CI.**

This PR integrates [Corepack](https://nodejs.org/api/corepack.html) into the CI pipeline to enforce a consistent, pinned version of NPM across all GitHub Actions workflows, Dockerfiles, and local development environments. Previously, workflows relied on whatever version of NPM happened to be bundled with the Node.js runner, which could vary between environments and lead to subtle inconsistencies.

### `actions/setup-node`

A composite action derived from our use of the published `setup-node` action, our action bootstraps Corepack first (without caching to avoid version mismatches) and lints the runtime environment. Finally, NPM install occurs for both the monorepo root and any relevant working-directories.

### Linting and bootstrap scripts

...to manage the Corepack lifecycle. Similar to `npm ci`'s checks for package-lock.json changes, these scripts give clarity to what might be causing CI errors and if those errors stem from compatible versions of NPM which may output different optional peer dependencies.

- `setup-corepack.mjs`: installs and pins the correct Corepack version
- `lint-runtime.mjs`: asserts that Node.js and NPM versions match what's declared in package.json
- `lint-lockfile.mjs`: validates lockfile integrity before installs

### Workflow updates

- Similar to our use of `uv`, all instances of `npm` are now prefixed with `corepack`  ensure the pinned version is always used
- The lint job in ci-web uses steps instead of matrix commands, avoiding repeated API builds for what is a series of linting commands

### CI warnings about mismatched Node.js and NPM versions

A warning may occur within the published `setup-node` action. This is expected and can be ignored, as our own action will install the correct Corepack version and enable it for subsequent steps.


dominic-r (approved):

kensternberg-authentik (approved):

scripts/node/setup-corepack.mjs
```diff
@@ -0,0 +1,94 @@
...
+import * as fs from "node:fs/promises";
+import { parseArgs } from "node:util";
+
+import { ConsoleLogger } from "../../packages/logger-js/lib/node.js";
+import { $, parseCWD, reportAndExit } from "./utils/commands.mjs";
+import { corepack, pullLatestCorepack } from "./utils/corepack.mjs";
+import { resolveRepoRoot } from "./utils/git.mjs";
+import { findNPMPackage, loadJSON, npm } from "./utils/node.mjs";
+
+const FALLBACK_NPM_VERSION = "11.11.0";
```
dominic-r: i think it would be interesting to hook this up and the package manager field with dependabot

scripts/node/setup-corepack.mjs
```diff
@@ -0,0 +1,94 @@
...
+    logger.info(`corepack ${corepackVersion || "not found"}`);
+
+    if (corepackVersion && !parsedArgs.values.force) {
+        logger.info("Corepack is already installed, skipping download (use --force to override)");
+        return;
+    }
+
+    await pullLatestCorepack(cwd);
+
+    await npm`install --force -g corepack@latest`({ cwd });
```
dominic-r, line 55: shouldn't the corepack version be pinned here, for security

GirlBossRush: There was enough pushback on vendoring and version pinning in the first few iterations. Pushing it to next release

scripts/node/setup-corepack.mjs
GirlBossRush: It's and interesting idea but given Dependabot's track record and our...4-dimensional build system, the fewer third-party integrations the better IMO.

scripts/node/setup-corepack.mjs
```diff
@@ -13,7 +13,7 @@ import { corepack, pullLatestCorepack } from "./utils/corepack.mjs";
 import { resolveRepoRoot } from "./utils/git.mjs";
 import { findNPMPackage, loadJSON, npm } from "./utils/node.mjs";
 
-const FALLBACK_NPM_VERSION = "11.11.0";
+const FALLBACK_NPM_VERSION = "11.13.0";
```
dominic-r: can we also pin the sha here

gergosimonyi (approved):
Approving backend.

.github/actions/setup/action.yml
```diff
@@ -18,19 +18,24 @@ runs:
   using: "composite"
   steps:
     - name: Cleanup apt
-      if: ${{ contains(inputs.dependencies, 'system') || contains(inputs.dependencies, 'python') }}
+      if: ${{ contains(inputs.dependencies, 'system') || contains(inputs.dependencies,
+        'python') }}
```
rissson: could you please revert all those string changes that don't actually change the string?

rissson (dismissed):

.github/workflows/api-ts-publish.yml
```diff
@@ -0,0 +1,65 @@
+---
+name: API - Publish Typescript client
```
rissson: This shouldn't be here. We don't publish the typescript API client from this repo.

PR #20402, opened 2026-02-19
enterprise: monkey patch pyjwt to accept mismatching key

gergosimonyi:
Internal ref: https://authentiksecurity.slack.com/archives/C08ABMS9TRP/p1771472226439349

### DIFF
--- a/authentik/enterprise/license.py
+++ b/authentik/enterprise/license.py
@@ -110,6 +110,7 @@ class LicenseKey:
             intermediate.verify_directly_issued_by(get_licensing_key())
         except InvalidSignature, TypeError, ValueError, Error:
             raise ValidationError("Unable to verify license") from None
+        _validate_curve_original = ECAlgorithm._validate_curve
         try:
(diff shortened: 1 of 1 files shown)


authentik/enterprise/license.py
```diff
@@ -110,6 +111,7 @@ def validate(jwt: str, check_expiry=True) -> LicenseKey:
         except InvalidSignature, TypeError, ValueError, Error:
             raise ValidationError("Unable to verify license") from None
         try:
+            ECAlgorithm._validate_curve = lambda *_: True
```
BeryJu, line 121:
```suggestion
            # See https://github.com/goauthentik/authentik/pull/20402
            ECAlgorithm._validate_curve = lambda *_: True
```

authentik/enterprise/license.py
BeryJu, line 132: We should add a `finally` here that undoes the monkeypatch

BeryJu (approved):

PR #20408, opened 2026-02-19
website/integrations: update jellyseerr docs to include enable proxy step

antholeole:
I ran into this

https://github.com/seerr-team/seerr/pull/1505/changes#r2311308711

will convert to ready to review need to run `make docs` and whatnot


dominic-r (approved):
The conflict is resolved against current main, and the proxy guidance now follows the current Seerr integration page. `make integrations` passes.

PR #20429, opened 2026-02-19
web/admin/bugfix: Edit Stage not working. Invoking IdentificationStageForm not working

kensternberg-authentik:
## What

1.  Fix the field being referenced by Flows -\> \[One Flow\] -\> StageBindings -\> \[Edit Stage\] to use the PK for the *stage*, rather than the *binding*.
2.  Added a check in `StrictUnsafe`: if the property is “wrapped” and untyped, treat it as an attribute, not a property.
3.  Edit the `ak-bound-stages-list` target attribute to be an attribute, not a property.

## Why

1.  This looks like a simple typo. To avoid this in the future, *we need tests*.
2.  `ModelForm` uses both a converter and get/set accessors to manage the pk (primary key) of the object it is being invoked to edit: the first because Django primary keys can be either strings or numbers, and the latter because we have special transactional requirements when a primary key changes. Lit’s magic for handling this creates some weirdness around JavaScript prototyping (`wrapped` becomes the only key on the object; all the other keys become delegated to a prototype object), so `hasOwn()` can’t be used; we just have to check for `wrapped` and `!type`.
3.  PKs are either strings or numbers, and ModelForm has a smart converter. There’s no need to shove the values around as properties and, in fact, that’ll break some things because there’s a working `attribute` field on `ModelForm.pk`! Removing the `.` property marker both avoids this issue and makes visible exactly what item-id is being referenced.


web/src/admin/flows/BoundStagesList.ts
```diff
@@ -89,7 +89,7 @@ export class BoundStagesList extends Table<FlowStageBinding> {
             html` <ak-forms-modal>
                     ${StrictUnsafe<CustomFormElementTagName>(item.stageObj?.component, {
                         slot: "form",
-                        instancePk: item.pk,
+                        instancePk: item.stageObj?.pk,
```
kensternberg-authentik, line 92: This looks like a simple typo.  To avoid this in the future, *we need tests*.

web/src/admin/flows/FlowViewPage.ts
```diff
@@ -272,7 +272,7 @@ export class FlowViewPage extends AKElement {
                 >
                     <div class="pf-c-card">
                         <div class="pf-c-card__body">
-                            <ak-bound-stages-list .target=${this.flow.pk}> </ak-bound-stages-list>
+                            <ak-bound-stages-list target=${this.flow.pk}> </ak-bound-stages-list>
```
kensternberg-authentik, line 275: PKs are either strings or numbers, and ModelForm has a smart converter.  There's no need to shove the values around as properties and, in fact, that'll break some things because there's a working `attribute` field on `ModelForm.pk`!  Removing the `.` property marker both avoids this issue and makes visible exactly what item-id is being referenced.

web/src/elements/utils/unsafe.ts
```diff
@@ -43,19 +43,23 @@ export const Prefix = {
...
-): Prefix {
+function resolvePrefix<T extends WrappedPropertyDeclaration>(propDeclaration: T): Prefix {
     if (!propDeclaration.attribute) {
         return Prefix.Property;
     }
 
+    if ("wrapped" in propDeclaration && propDeclaration.wrapped && !propDeclaration.type) {
+        return Prefix.Attribute;
+    }
+
```
kensternberg-authentik, line 62: `ModelForm` uses both a converter and get/set accessors to manage the pk (primary key) of the object it is being invoked to edit: the first because Django primary keys can be either strings or numbers, and the latter because we have special transactional requirements when a primary key changes. Lit's magic for handling this creates some weirdness around JavaScript prototyping (`wrapped` becomes the only key on the object; all the other keys become delegated to a prototype object), so `hasOwn()` can't be used; we just have to check for `wrapped` and `!type`.

BeryJu (approved):

PR #20431, opened 2026-02-20
core: users/groups reduce number of database queries

ryanpesek:
## Summary
Optimizations for the users and groups API endpoints to reduce N+1 queries and database load.

## Changes
Users API 
1. Annotate is_superuser to avoid N+1 queries
Changed is_superuser from a direct BooleanField to a SerializerMethodField that uses a database annotation. The get_queryset method now annotates each user with _annotated_is_superuser using an Exists subquery that checks superuser group membership (including inherited membership via group ancestry).

Before: Each user's is_superuser property triggered a separate query to check group membership.
After: A single annotated query determines superuser status for all users in the response.

2. Always prefetch groups/roles with ID-only optimization
When include_groups=false or include_roles=false, the queryset now uses Prefetch with .only() to fetch just the primary keys instead of skipping the prefetch entirely.

Why: The groups and roles PKs are always serialized in the response. Previously, when include_groups=false, no prefetch occurred, causing N+1 queries when serializing the PK list. Now we always prefetch, but use ID-only queries when full objects aren't needed.

Groups API 
1. Always prefetch parents and children
Changed get_queryset to always prefetch parents and children alongside roles, removing the conditional prefetch logic.

Why: The parents and children PKs are always serialized in the response regardless of the include_parents/include_children query parameters (those control whether the full objects are included). Always prefetching avoids N+1 queries when serializing the PK lists.

2. Bulk validation for users field
Added BulkPrimaryKeyRelatedField and BulkManyRelatedField classes that validate all user PKs in a single COUNT query instead of one query per PK.

Before: Updating a group with 1000 users triggered 1000 individual SELECT queries to validate each user PK exists.
After: A single SELECT COUNT(*) query validates all PKs exist. Per-PK queries only run when there's an actual validation error (to report which specific PK is invalid).

---

## Checklist

If applicable


gergosimonyi (commented):
Hi Ryan!

Re:

> 2. Bulk validation for users field
> Added BulkPrimaryKeyRelatedField and BulkManyRelatedField classes that validate all user PKs in a single COUNT query instead of one query per PK.
> 
> Before: Updating a group with 1000 users triggered 1000 individual SELECT queries to validate each user PK exists.
> After: A single SELECT COUNT(*) query validates all PKs exist. Per-PK queries only run when there's an actual validation error (to report which specific PK is invalid).

How did you produce this issue? I tried on current `main` (ea4848c7c680e2d3d0e57260bd9b046135de10d4 at the time of writing) with a group that has exactly 1000 users, from the admin interface, and found that whether I

- update the group object (I toggled`is_superuser`), or
- update group memberships (I added a 1001st user)

the number of SQL queries was way below 1000. Scanning through them I didn't find a section that seemed to validate current users.

Am I misunderstanding something? Or is this only an issue in a very specific case I didn't try?

ryanpesek:
> Hi Ryan!
> 
> Re:
> 
> > 2. Bulk validation for users field
> >    Added BulkPrimaryKeyRelatedField and BulkManyRelatedField classes that validate all user PKs in a single COUNT query instead of one query per PK.
> > 
> > Before: Updating a group with 1000 users triggered 1000 individual SELECT queries to validate each user PK exists.
> > After: A single SELECT COUNT(*) query validates all PKs exist. Per-PK queries only run when there's an actual validation error (to report which specific PK is invalid).
> 
> How did you produce this issue? I tried on current `main` ([ea4848c](https://github.com/goauthentik/authentik/commit/ea4848c7c680e2d3d0e57260bd9b046135de10d4) at the time of writing) with a group that has exactly 1000 users, from the admin interface, and found that whether I
> 
> * update the group object (I toggled`is_superuser`), or
> * update group memberships (I added a 1001st user)
> 
> the number of SQL queries was way below 1000. Scanning through them I didn't find a section that seemed to validate current users.
> 
> Am I misunderstanding something? Or is this only an issue in a very specific case I didn't try?

Hey @gergosimonyi thanks for taking a look. This fix is for the PUT endpoint `/core/groups/{group_uuid}/` as we update group membership in bulk. I am pretty sure when you use the admin UI, that is calling the POST `/core/groups/{group_uuid}/add_user/` or POST `/core/groups/{group_uuid}/remove_user/` which only accept one user pk at a time. The PUT endpoint has to validate that all user pk in the `users` attribute are valid

gergosimonyi:
Ohhh, I see. We don't allow the viewset's default `POST`/`PUT`/`PATCH` action to change many-to-many on [some other API endpoint] and I automatically assumed it was not allowed here either. My bad.

This seems like a reasonable workaround until such time that the issue is fixed in DRF. They already know about it: https://github.com/encode/django-rest-framework/issues/9607

Thank you for this contribuiton!

gergosimonyi (approved):

PR #20440, opened 2026-02-20
Show previously configured NameID and AuthnContextClassRef parameters for SAML

### LINKED ISSUE #20146: When I open edit page for a provider, some UI fields and not filled properly
### Describe the bug

When I open a provider for Edit, some UI fields are not filled properly. When I click save, the default value is saved instead of the previously configured parameters.
Only 'select' elemets are affected. I can see the issue for:
- SCIM provider: filter_group attribute
- SAML provider: name_id_mapping attribute

API returns the correct provider configuration with object ID. But when UI is trying to convert the object ID to the name using extra API requests, instead of sending the request to read a specific object, it tries to read all objects with limit of 20 items. When we have more than 20 objects, there's a chance to not to receive details for the required one.

`GET https://domain.com/api/v3/core/groups/?include_users=false&ordering=name`
```
{
    "pagination": {
        "next": 2,
        "previous": 0,
        "count": 3109,
        "current": 1,
        "total_pages": 156,
        "start_index": 1,
        "end_index": 20
    },
    "results": [
...
```

`GET https://domain.com/api/v3/propertymappings/provider/saml/?ordering=saml_name&page=1&page_size=20&search=`
```
{
    "pagination": {
        "next": 2,
        "previous": 0,
        "count": 28,
        "current": 1,
        "total_pages": 2,
        "start_index": 1,
        "end_index": 20
    },
    "results": [
...
```

### How to reproduce

For SCIM provider: 
1. Create 40 groups
2. Create a SCIM provider and fill the 'Group' field (at 'User filtering' section)
3. Save
4. Open the provider for Edit. WIth 50% probability the field 'Group' is empty. Depends on the group name since the API request sorts group by name.
5. You can see the limit of 20 items at the API requests from the browser developer tools

For SAML provider:
1. Create 40 'SAML Provider Property Mapping' at Customization section
2. Create SAML provider and fill the 'NameID property mapping' field (at 'Advanced protocol settings' section)
3. Save
4. Open the provider for Edit. WIth 50% probability the field 'NameID property mapping' empty.
5. You can see the limit of 20 items at the API requests from the browser developer tools

### Expected behavior

All provider parameters have to be displayed properly when I open the provider for edit. Regardless of the name of the object.

### Screenshots

_No response_

### Additional context

_No response_

### Deployment Method

Other (please specify)

### Version

2025.12.3

### Relevant log output

```shell

```

bitpavel-l25:
## Details

Resolves https://github.com/goauthentik/authentik/issues/20146
Since users filter has been refactored for SCIM provider at the new RC versions, only SAML provider has to be fixed. It's done at the current PR.
It would be better to refactor the whole `ak-search-select` component to accept current value as an argument instead of implementing the logic for every component use. But it would require much broader testing and refactoring. I let Authentik team consider other cases. 

---

## Checklist

-   [+] The code has been formatted (`make web`)

If applicable

### DIFF
--- a/web/src/admin/providers/saml/SAMLProviderFormForm.ts
+++ b/web/src/admin/providers/saml/SAMLProviderFormForm.ts
@@ -27,6 +27,7 @@ import {
     KeyTypeEnum,
     PropertymappingsApi,
     PropertymappingsProviderSamlListRequest,
+    PropertymappingsProviderSamlRetrieveRequest,
     SAMLBindingsEnum,
     SAMLLogoutMethods,
     SAMLNameIDPolicyEnum,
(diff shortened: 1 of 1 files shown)


dewi-tik:
We're working in improvements to ak-search-select that will cover this issue. We'll leave this open to track the problem.

PR #20483, opened 2026-02-23
enterprise/stages/mtls: fix traefik cert encoding

BeryJu:
https://github.com/traefik/traefik/issues/9669
https://doc.traefik.io/traefik/reference/routing-configuration/http/middlewares/passtlsclientcert/#pem


gergosimonyi (commented):
Technically a breaking change. Please call it out in the release notes.

PR #20490, opened 2026-02-24
blueprints: Add support for dynamic blueprint entries

astro-stan:
## Details

Until now, blueprint entries were purely static, and no custom YAMLTags could be used for creating them. 

This had several drawbacks:

1. Repetitiveness: A lot of times, a blueprint author had to create 2 or more entries that are almost identical but differ by 1 or 2 fields (think having to create 2 or more groups or having to bind several policies to the same stage). Because the entries were static, the blueprint author was forced to copy and paste what is essentially the same entry several times. This was repetitive and error-prone
2. Impossible to create entries depending on context data (or other factors). Think of a blueprint that allows creating user groups. Its context might include a list of group names to be created. Without the dynamic entries support, this was impossible to achieve because there was no way to generate an entire blueprint entry dynamically

---

## Checklist

If applicable


BeryJu:
The idea and the code are good but I'm quite unsure about how much more logic should be added to the blueprints language, in an attempt to not make it a full on template language.

If people want this, sure, if not then maybe not

astro-stan:
Hi @BeryJu.

Thanks for taking a look at this. I understand the desire to keep the blueprint language clean and avoid over-complicating it. 

However, this would be a powerful feature, enabling advanced automation, such as dynamically generating groups, stages, policy bindings, etc. Moreover, it could be used to reduce mistakes  and maintainance effort caused by duplication by making blueprints DRY(-er). 

It's a feature which I've needed for quite a while and one that has prevented me from creating some very flexible and useful blueprints, which I hope could one day be used by not just me but the whole authentik community. 

Moreover this feature is completely optional and backwards-compatible, so authors who prefer simpler, static blueprints don't have to use it or change their workflows at all. It simply provides another tool for those managing complex, scalable setups who really need it.

I'm happy to address any concerns you might have regarding the implementation.

PR #20508, opened 2026-02-24
website/docs: fix linux setup docs

BeryJu:
found by @PeshekDotDev 
added some other small fixes from my side
might need some slight re-writing @dewi-tik


dewi-tik (approved):
Good catch @PeshekDotDev. I've updated the wording a bit and made it clear that it's optional.

PR #20530, opened 2026-02-24
fixup(scim): Adjust SCIM client AWS compatibility implementation

josegomezr:
## Details

AWS IAM Identity Center SCIM server does support PATCH operations as it implemented, however the Compatibility Mode was forcing the `patch.supported` flag to false and thus never sending patch requests for user memberships.

When bypassing this by setting `compatibility_mode = "default"`, then memberships could be forwarded but groups couldn't be updated.

This patch aligns the behavior of `compatibility_mode = "aws"` to properly send PATCH requests for group updates as well as user membership updates.

Additionally, dumped the `ServiceProviderConfig` objects with `mode="json"`, it raised JSONSerializationError's without `mode="json"`.

AWS SCIM limitations: https://docs.aws.amazon.com/singlesignon/latest/developerguide/limitations.html

---

## Checklist

If applicable (maybe?)

---

Cherry picked from SUSE/authentik#13

### DIFF
--- a/authentik/providers/scim/clients/base.py
+++ b/authentik/providers/scim/clients/base.py
@@ -104,8 +104,6 @@ class SCIMClient[TModel: "Model", TConnection: "Model", TSchema: "BaseModel"](
 
         try:
             config = ServiceProviderConfiguration.model_validate(self._request("GET", path))
-            if self.provider.compatibility_mode == SCIMCompatibilityMode.AWS:
-                config.patch.supported = False
             if self.provider.compatibility_mode == SCIMCompatibilityMode.SLACK:
                 config.filter.supported = True
         except (ValidationError, SCIMRequestException, NotFoundSyncException) as exc:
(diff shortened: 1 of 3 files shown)


josegomezr:
Took the liberty to graft externalId handling from SUSE/authentik#15 onto this PR as well. Hope it's not an issue :)

marcispauls:
@BeryJu — wanted to flag this for your attention. We've been hitting this exact bug in production with the AWS Identity Store SCIM provider, and can share a secondary symptom not yet described in this PR:

The failed group PUTs don't just fail silently — they **jam the Dramatiq worker queue**. Retries accumulate and starve user-creation tasks entirely, so new users never get provisioned even while the sync reports `last_sync_status: done`. The contradictory state (`is_running: true` + `last_sync_status: done`) makes it look stuck rather than broken. The only workaround is bypassing SCIM entirely and provisioning users directly via the target API.

The fix in this PR (adding `patch.supported` properly for the AWS compat mode) would resolve both the 501 errors and the queue starvation downstream. Would appreciate a review when you get a chance.

PR #20532, opened 2026-02-24
website/inteegrations: fix redirect URI in Seafile integration

DystopianRescuer:
If you just follow orders exactly as said here it'll lead you to an error because of the strict policy having a slash at the end but in the seahub_settings.py in the docs tells you to put it without it. This minor change can help people to not encounter this minimal error sometimes difficult to see.


dewi-tik (approved):
Thanks for reporting this @DystopianRescuer

PR #20541, opened 2026-02-24
web: Packagify Logger

GirlBossRush:
Pretty much copied from `web/logger`, which should be removed in a later PR.


gergosimonyi (approved):
My `prettier` wants to change ordering in `package.json`.

kensternberg-authentik (approved):
The new parts are pretty straightforward, and the rest is familiar boilerplate.

PR #20574, opened 2026-02-25
core: add concrete model for group membership

BeryJu:
allows us to save other details for a group membership

first step of #20573


authentik/core/migrations/0058_groupmembership.py
```diff
@@ -0,0 +1,80 @@
...
+    db_alias = schema_editor.connection.alias
+
+    Group = apps.get_model("authentik_core", "group")
+    GroupMembership = apps.get_model("authentik_core", "GroupMembership")
+
+    print("Re-creating group memberships, this might take a couple of minutes...")
+    for group in progress_bar(Group.objects.using(db_alias).all()):
+        GroupMembership.objects.bulk_create(
+            [GroupMembership(group=group, user=user) for user in group.users.all()]
+        )
```
gergosimonyi, line 21: I can think of 2 ways to not have to wait for this:

1. Not doing the copy, adding a [`db_table`](https://docs.djangoproject.com/en/5.2/ref/models/options/#django.db.models.Options.db_table) field to `GroupMembership`.
2. Doing the copy in SQL, which would make it pretty much instantaneous even for very large membership numbers.

(I prefer the first.)

gergosimonyi: Offline resolution: we should do solution 2, copy with an SQL command. Can this break something? We believe the answer is no. There are currently no permissions or object lifecycle management defined for m2m through objects and we can't immediately think of other ways this could break things.

Reasoning: we want to keep using "stock django" anywhere we can. We feel `db_table` should only be manually set in cases more special than this one.

gergosimonyi:
Offline discussion: we would like to make all m2m through relations explicit. We should probably do them all at once in this PR instead of one by one for economies of scale and to minimize number of migrations.

BeryJu:
superseeded by #23315

PR #20577, opened 2026-02-25
blueprints: Add support for nested `!Context` values

astro-stan:
## Details

This PR adds a backwards- compatible support for resolving nested values within a blueprint's context.

Example:

```yaml
context:
 goauthentik.io: priority-value
 goauthentik:
   io: non-priority-value
  foo:
    bar:
      - baz: value
entries:
  - ...
    key1: !Context foo.bar.0.baz # Resolves to `value`
    key2: !Context goauthentik.io # Resolves to `priority-value`, even though a nested dot notation key with the same name exists
```

Why is this needed?

When writing more complex blueprints, I frequently want to group context values by "category" or take part of another key's value. Until now I had to write all keys using their full paths i.e. `authentication.something.setting1`, `authentication.something.setting2`, etc instead of writing them as a dictionary like so:

```yaml
authentication:
  something:
    setting1: ...
    setting2: ...
```

Moreover, sometimes I've had use cases where I want to take the value of 1 element in a list, for example:

```yaml
context:
  some-list:
     - value 1
     - value 2  # I want this value only 
```

This was impossible to do until now

## Checklist

If applicable

-   [X] The documentation has been formatted (`make docs`)


website/docs/customize/blueprints/v1/tags.mdx
```diff
@@ -104,15 +104,44 @@ Example:
...
+      configure_flow: !Context foo # Resolves to `some-value`
+      authentication_flow: !Context some.nested.0 # Resolves to `value`
 ```
 
 {/* prettier-ignore-end */}
 
 Find values from the context. Can optionally be called with a default like `!Context [foo, default-value]`.
 
+:::note
+The nested value dot notation (`some.nested.0`) requires authentik 2026.5.
```
rissson:
```suggestion
The nested value dot notation (`some.nested.0`) requires authentik 2026.11.
```

rissson (approved):

dominic-r (approved):

rissson:

PR #20578, opened 2026-02-25
web/flow: separate out independent behavior tracks from IdentificationStage (autoredirect, webauthn, captcha, rememberme)

kensternberg-authentik:
web/flow: re-arrange IdentificationStage for maintainability, part 2: AutoRedirect, Webauthn & Captcha

## What

These three subsystems have been removed from the IdentificationStage and put into their own Lit-Controllers.

### AutoRedirect

AutoRedirect is a _rule_ that triggers when a challenge arrives that meets certain conditions.  The existing code stated the rules in a mix of negative and positive checks, when really there's only one condition that has to be met, that these all be true:

        const onlyOneSource = sources.length === 1;
        const noUserAccessibleInputs = (userFields || []).length === 0;
        const noAlternativeMethods = !passwordlessUrl;

Also, it was checked _after_ the initial render, which seemed oddly wasteful; by the time you get to `willUpdate`, you know if you have a challenge that meets these conditions.  Moving the rule into a Controller and issuing a "change the challenge" event when this condition was met on `hostUpdate` makes this happen before rendering and removes a lot of Demeter violation.

### Webauthn

[Webauthn](https://webauthn.io/) is a mechanism that retrieves a stored secret key from something the user has, such as a Yubi-key or a vault on their browser; the standard DH-style cryptographic challenge compares a generated authentication request token to a corresponding public key. A Javascript component queries the browser for Webauth compatibility and, if it’s found, triggers a Webauthn transaction with which the user must interact independent of the stages’s HTML or presentation. If the transaction succeeds, the user is automatically sent to the redirect stage.

None of the code associated with Webauthn has anything to do with the stage rendering or event monitoring; it even has its own `submit()` phase, triggered automatically after the user grants the browser access to the passkey.

Therefore, I have placed it into a [Lit Controller](https://lit.dev/docs/composition/controllers/). The IdentificationStage activates the controller, and after the page has “settled” the controller checks if passkey use is authorized and that the browser has a passkey mechanism available, and if both are true, it sends the browser a request for the passkey transaction to begin and, once it has retrieved the credentials, transacts with the server on its own to see if redirection can proceed.

None of that needs to be in the IdentificationStage. That’s about 60 lines of code moving out of IdentificationStage.

### Captcha

Captcha is a little more invasive. It needs to track Captcha “freshness” (how old the last Captcha is), the current Captcha state and, because Captchas are loaded from third-parties, whether or not the Captcha loaded at all.

In the IdentificationStage, we need to conditionally render the Captcha, we need to handle if the Captcha failed, and we need to disable the button until the Captcha has loaded.

By moving “is there a captcha?” into a controller, the only references left in IdentificationStage are:

- activation of the controller at construction
- use a named accessor to delay enabling the submit button while the Captcha loads
- use a named method to handle the Captcha refresh-on-failure cycle
- use a named method to render the Captcha, which will return `nothing` when Captcha is not enabled

That’s about 40 lines of code moving out of IdentificationStage.

### Conclusion

That’s 100 lines of IdentificationStage replaced with 6 lines of controller references. “That’s a lot of pain moving out of the world.”

I did noticed that IdentificationPasskey shares a *lot* of code with the Webauthn Stage, but I haven’t looked into how to push that further up. (Yet).


web/src/flow/stages/identification/controllers/AutoRedirectController.ts
```diff
@@ -0,0 +1,28 @@
...
+        }
+
+        const { userFields, passwordlessUrl, sources = [] } = challenge;
+
+        // The rules for auto-redirect:
+        const onlyOneSource = sources.length === 1;
+        const noUserAccessibleInputs = (userFields || []).length === 0;
+        const noAlternativeMethods = !passwordlessUrl;
+
+        if (onlyOneSource && noUserAccessibleInputs && noAlternativeMethods) {
```
kensternberg-authentik, line 33: Reminder: This is what the original looked like. I hope the new one is more readable!

```
        // We only want to auto-redirect to a source if there's only one source.
        if (this.challenge.sources?.length !== 1) return;

        // And we also only do an auto-redirect if no user fields are select
        // meaning that without the auto-redirect the user would only have the option
        // to manually click on the source button
        if ((this.challenge.userFields || []).length !== 0) return;

        // We also don't want to auto-redirect if there's a passwordless URL configured
        if (this.challenge.passwordlessUrl) return;

        const source = this.challenge.sources[0];
        this.host.challenge = source.challenge;
```

web/src/flow/stages/identification/controllers/RememberMeController.ts
```diff
@@ -22,11 +22,33 @@ export class AkRememberMeController implements ReactiveController {
...
+        }
+        if (!this.usernameField) {
+            return;
+        }
+        localStorage?.setItem("authentik-remember-me-user", this.usernameField.value);
+        localStorage?.setItem("authentik-remember-me-session", this.localSession);
+        this.usernameField.addEventListener("keyup", this.trackRememberMe);
+    };
+
+    constructor(private host: RememberMeHost) {}
```
kensternberg-authentik, line 66: Mostly, this just gets rid of the `.bind()` clauses.

web/src/flow/stages/identification/controllers/WebauthnController.ts
```diff
@@ -0,0 +1,89 @@
...
+            this.passkey = (this.host.challenge as PasskeyChallenge)?.passkeyChallenge ?? null;
+        }
+        if (this.passkey) {
+            this.#startConditionalWebAuthn(this.passkey);
+        }
+    }
+
+    public hostDisconnected() {
+        this.#abortController?.abort();
+        this.#abortController = null;
```
kensternberg-authentik, line 67: I think this code is somewhat suspect.  This is the same logic as the original: we automatically create a new credential request whenever the host updates, but does the host _updating_ coincide with the host _reattaching_ to a DOM? For that matter, in what scenario does the IdentificationStage ever detach and reattach to the DOM?  Suspect, but it probably gets away with it.

web/src/flow/types.ts
```diff
@@ -58,6 +59,19 @@ export interface SubmitOptions {
     invisible: boolean;
 }
 
+// Make the "component" field optional, since the Executor controls what component type is being
+// manipulated.
+type PartialComponent<T> = T extends { component: infer C } & infer Rest
+    ? { component?: C } & Omit<Rest, "component">
+    : never;
+
```
kensternberg-authentik, line 67: This type allows us to maintain the same protocol as before, only passing the payload as an event rather than all the `this.host.host.submit()` [Demeter violations](https://en.wikipedia.org/wiki/Law_of_Demeter) we were living with before. This will let us nest sub-stages without having to hack our way all up the chain of responsibilities to the FlowExecutor; any stage that speaks the protocol will now Just Work.

GirlBossRush (approved):

PR #20606, opened 2026-02-26
packages/django-dramatiq-postgres: use fork

BeryJu:
fixes a big memory issue since upgrading to python 3.14

3.14 switches the default (in our case on linux) to forkserver, which seems to bring with it a much higher memory usage

closes #20537


rissson (approved):

ishmatovashahnoza782-create (commented):
#20537

PR #20627, opened 2026-02-27
crypto: fix kid legacy signal

BeryJu:
without this in the container you'd get `UnsupportedDigestmodError [digital envelope routines] unsupported` due to FIPS


gergosimonyi (approved):
Absolutely hate this

PR #20662, opened 2026-03-02
providers/proxy: move search path to query instead of runtime parameter

### LINKED ISSUE #19939: unsupported startup parameter: search_path when using PGBouncer
Our embedded outpost connects to postgres using search_path in the startup parameter where as django runs a query for this, and pg bouncer doesn't support this startup parameter

ref https://github.com/goauthentik/internal-customer-ref/issues/5

BeryJu:
runtime parameter is incompatible with pgbouncer

closes #19939

### DIFF
--- a/internal/outpost/proxyv2/postgresstore/connpool.go
+++ b/internal/outpost/proxyv2/postgresstore/connpool.go
@@ -25,8 +25,6 @@ type RefreshableConnPool struct {
 	currentDSN string
 	gormConfig *gorm.Config
 
-	SearchPath string
-
 	// Connection pool settings (stored for reapplication after reconnection)
 	maxIdleConns    int
 	maxOpenConns    int
@@ -58,10 +56,6 @@ func NewRefreshableConnPool(initialDSN string, gormConfig *gorm.Config, maxIdleC
 		connMaxLifetime: connMaxLifetime,
 	}
 
-	if err := pool.setSearchPath(context.Background(), db); err != nil {
-		return nil, err
-	}
-
 	return pool, nil
 }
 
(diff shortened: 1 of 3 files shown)


internal/outpost/proxyv2/postgresstore/connpool.go
```diff
@@ -121,6 +138,11 @@ func (p *RefreshableConnPool) refreshCredentials(ctx context.Context) error {
 	newDB.SetMaxOpenConns(p.maxOpenConns)
 	newDB.SetConnMaxLifetime(p.connMaxLifetime)
 
+	if err := p.setSearchPath(ctx, newDB); err != nil {
```
dominic-r: one ExecContext on *sql.DB does not guarantee search_path for all pooled connections. database/sql DB is a pool and per-connection state is only reliably observed in Tx/Conn. https://pkg.go.dev/database/sql#DB https://www.postgresql.org/docs/current/sql-set.html

dominic-r (approved):

internal/outpost/proxyv2/postgresstore/postgresstore.go
```diff
@@ -187,23 +187,50 @@ func BuildConnConfig(cfg config.PostgreSQLConfig) (*pgx.ConnConfig, error) {
...
 		connOpts, err := parseConnOptions(cfg.ConnOptions)
 		if err != nil {
 			return nil, fmt.Errorf("failed to parse connection options: %w", err)
 		}
+		// Handle search_path outside of startup RuntimeParams for PgBouncer compatibility.
+		// ConnOptions should still override DefaultSchema when both are set.
+		if connOptionsSearchPath, ok := connOpts["search_path"]; ok {
+			effectiveSearchPath = connOptionsSearchPath
+			delete(connOpts, "search_path")
+		}
```
rissson: I don't agree with this. We override it always in django, so it should just be removed.

internal/outpost/proxyv2/postgresstore/postgresstore.go
```diff
@@ -187,23 +187,47 @@ func BuildConnConfig(cfg config.PostgreSQLConfig) (*pgx.ConnConfig, error) {
...
+		delete(connConfig.RuntimeParams, "search_path")
+	}
+
+	// Set search_path after connection startup to avoid startup-parameter issues with PgBouncer.
+	if effectiveSearchPath != "" {
+		connConfig.AfterConnect = func(ctx context.Context, pgConn *pgconn.PgConn) error {
+			result := pgConn.ExecParams(
+				ctx,
+				"select pg_catalog.set_config('search_path', $1, false)",
```
rissson, line 221: so django does `SET search_path = $1`. Any difference we should consider?

rissson (approved):

PR #20663, opened 2026-03-02
website/docs: kerberos: add note about caching

dewi-tik:
## Details

Adds note about issues related to caching

---

## Checklist

If applicable

### DIFF
--- a/website/docs/users-sources/sources/protocols/kerberos/index.md
+++ b/website/docs/users-sources/sources/protocols/kerberos/index.md
@@ -155,4 +155,4 @@ if localpart == "username":
 
 You can start authentik with the `KRB5_TRACE=/dev/stderr` environment variable for Kerberos to print errors in the logs.
 
-Reverse proxy caching can occasionally interfere with the Kerberos authentication flow. Specifically, the reverse proxy may cache the second 401 Unauthorized response, which includes the WWW-Authenticate header required for the client’s Kerberos negotiation, instead of forwarding it to Authentik. This behavior disrupts the authentication handshake. To resolve such issues, disable response caching on the reverse proxy for routes involved in Kerberos authentication.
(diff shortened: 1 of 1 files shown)


website/docs/users-sources/sources/protocols/kerberos/index.md
```diff
@@ -154,3 +154,5 @@ if localpart == "username":
 ## Troubleshooting
 
 You can start authentik with the `KRB5_TRACE=/dev/stderr` environment variable for Kerberos to print errors in the logs.
+
+Reverse proxy caching can occasionally interfere with the Kerberos authentication flow. Specifically, the reverse proxy may cache the second 401 Unauthorized response, which includes the WWW-Authenticate header required for the client’s Kerberos negotiation, instead of forwarding it to Authentik. This behavior disrupts the authentication handshake. To resolve such issues, disable response caching on the reverse proxy for routes involved in Kerberos authentication.
```
dominic-r:
```suggestion
Reverse proxy caching can break the Kerberos authentication flow because, during negotiation, the server sends a second `401 Unauthorized` response containing the `WWW-Authenticate` header that the client needs to continue the handshake; if the reverse proxy caches that `401` instead of forwarding it to authentik, the authentication process fails, so response caching should be disabled on any reverse proxy routes involved in Kerberos authentication.
```

tanberry (approved):
Thanks @dewi-tik for doing this. On;ly suggestion would be to put that last sentence first ("response caching should be disabled on any reverse proxy routes involved in Kerberos authentication") to first call attention to that, and _then_ explain the why part. But it's also fine as it is. :-)

r3pek:
Thanks @dewi-tik.... 2 days debugging this :sweat_smile:

Keep up the good work guys :muscle:

dominic-r (approved):

PR #20698, opened 2026-03-04
web: build system had some legacy stuff that I found confusing while working on the CSS ordering

kensternberg-authentik:
web: replace deepmerge with more explicit configuration; 'build --proxy' misnamed, changed to 'build -styles-only


web/scripts/build-web.mjs
```diff
@@ -84,60 +130,28 @@ const BASE_ESBUILD_OPTIONS = {
...
     define: bundleDefinitions,
     format: "esm",
-    logOverride: {
-        /**
-         * HACK: Silences issue originating in ESBuild.
-         *
-         * @see {@link https://github.com/evanw/esbuild/blob/b914dd30294346aa15fcc04278f4b4b51b8b43b5/internal/logger/msg_ids.go#L211 ESBuild source}
-         * @expires 2025-08-11
-         */
-        "invalid-source-url": "silent",
```
kensternberg-authentik, line 137: This is part of what inspired the check. I saw this "expiration" note and dug in.  It's not needed anymore.

web/scripts/build-web.mjs
```diff
@@ -84,60 +130,28 @@ const BASE_ESBUILD_OPTIONS = {
...
+ * @returns {BuildOptions}
+ */
+export function createESBuildOptions(entryPoints, plugIns = []) {
+    const plugins = [...BASE_ESBUILD_PLUGINS, ...plugIns];
+
+    return {
+        ...BASE_ESBUILD_OPTIONS,
+        entryPoints,
+        plugins,
+    };
```
kensternberg-authentik, line 156: These are the things that we modify with our current ESBuildOptions: entrypoints, and plugins. We should be explicit about that. Until our mods get so large we can't keep track of them, in which case we need a different build system.

web/scripts/build-web.mjs
```diff
@@ -52,8 +51,55 @@ const assets = [
     [path.resolve(PackageRoot, "icons"), "./assets/icons"],
 ];
 
+const entryPointNames = Object.keys(EntryPoint);
+const entryPoints = Object.values(EntryPoint);
+const entryPointsDescription = entryPointNames.join("\n\t");
```
kensternberg-authentik, line 56: In "store vs. calculate" thinking, the repeated use of these expressions had the potential to lead to Last Line problems, so I pulled them out and gave them names.

web/scripts/build-web.mjs
```diff
@@ -52,8 +51,55 @@ const assets = [
...
+
 /**
- * @type {Readonly<BuildOptions>}
+ * @type {Plugin[]}
+ */
+const BASE_ESBUILD_PLUGINS = [
```
kensternberg-authentik, line 61: Pulled this out and gave it its own identity so that `ESBuildOptions()` could be explicit about annotation.  The big win here is that `deepmerge` will perform concatenation, but that's not clear; a dev in the future who assumes it does something more clever may end up with duplicate plugins.  By bringing the concatenation into view, we reduce that possibility.

.gitignore
```diff
@@ -213,3 +213,13 @@ source_docs/
 ### Docker ###
 tests/openid_conformance/exports/*.zip
 compose.override.yml
+
+### LLMs and "coding assistants" ###
```
kensternberg-authentik: This was just necessary.  Don't store slop sessions in our code.

GirlBossRush (approved):

PR #20715, opened 2026-03-04
feat: add API endpoint to trigger LDAP source sync

borutmrak:
## Details

This PR adds an API endpoint POST `/sources/ldap/{slug}/sync` that triggers a source synchronization.

We use multiple systems that sync with our LDAP directory, but we have to do manual syncs when adding new users etc (or wait until next scheduled sync) and it is tedious. We made a watcher that can send webhooks when users are added/removed or their group membership changes. Since Authentik did not contain a way to trigger the sync, I added one.

---

## Checklist

If applicable


rissson:
You should use the schedule triggering endpoint for this I think

borutmrak:
> You should use the schedule triggering endpoint for this I think

You're right, I found `/api/v3/tasks/schedules/:id/send/` now. It didn't occur to me that this is a way to do it. It seemed wierd that there is a LDAP sync status endpoint but not a sync trigger one so I just went ahead and created one.

Do you know if this is documented somewhere? Otherwise I think it should be added and I'm willing to do it, just point me to where it should be.

borutmrak:
There is however a small usability issue. I don't see a way to get the task ID used in that call from the UI, you have to use the API for that (`curl -X GET "https://<hostname>/api/v3/tasks/schedules/?actor_name=authentik.sources.ldap.tasks.ldap_connectivity_check&search=<slug>"
 -H 'accept: application/json' | jq -r '.results.[0].id' `). With this PR you can directly use the source slug, which I consider nicer - this is however not a reason enough to accept this PR.

Oh well :)

rissson:
Yes, you'd have to fetch the schedule ID from the API. I'm gonna go ahead and close this then.

PR #20718, opened 2026-03-04
core: Block non-super users from managing superuser groups

josegomezr:
## Details

Any user with a global "add user to group" permission is able to add himself to a super user group and gain undesired permissions.

We spotted this subtle condition when designing directory management roles.

---

## Checklist

### DIFF
--- a/authentik/core/api/groups.py
+++ b/authentik/core/api/groups.py
@@ -343,6 +343,12 @@ class GroupViewSet(UsedByMixin, ModelViewSet):
     def add_user(self, request: Request, body: UserAccountSerializer, pk: str) -> Response:
         """Add user to group"""
         group: Group = self.get_object()
+
+        if group.is_superuser and not request.user.is_superuser:
+            raise ValidationError(
+                _("User does not have permission to add users to a superuser group.")
+            )
+
         user: User = (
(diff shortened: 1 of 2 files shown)


gergosimonyi:
The current behavior is intended. The permission system, in its current form, is very broad and includes many permissions (or combinations of permissions) that can result in privilege escalation. This is one of them. Other easy ones include `assign_role_permissions`, `change_role`, `change_expressionpolicy`, `change_flowstagebinding` etc.

A possible solution for your use case is to create object-level `add_user_to_group` permissions. That way you can precisely control which groups' membership you allow to be changed.

josegomezr:
> A possible solution for your use case is to create object-level add_user_to_group permissions. That way you can precisely control which groups' membership you allow to be changed.

That's a solution we're discussing too. In our labs we're assigning front line support group management permissions to handle +1800 groups.

To make it work with object level permission we'd have to keep track of the +1800 assignments and add/remove accordingly so the group management role can still perform group management task to all groups but the super user ones.

We also considered adding an extra permission to the system, something like: `authentik_rbac.add_user_to_superuser_group`.

To make it "easy" i started with a trimmed down implementation without extra permissions.

dewi-tik:
This is covered by the last security release.

PR #20736, opened 2026-03-05
lifecycle: make gunicorn --max-requests configurable

severin:
## Details

Authentik uses _gunicorn_'s `max-request` setting which restarts the server after ~1000 requests. This is currently hardcoded.

I find this behaviour disruptive, i.e. in an environment where only one worker is run (I know it is advised against doing so, but for staging or test environments we still do it) this setting causes periodic downtime, since restarting the server is not instant.

As I understand the setting is only meant as a temporary last line of defense against memory leaks (see also https://github.com/benoitc/gunicorn/discussions/3042); and I hope authentik is actually not leaking memory 😉 

I personally would like to get completely rid of this, but as an intermediate step I think this should be at least configurable. 

This PR makes the setting configurable (by providing a value of `0` it can also be disabled) through environment variables.

Let me know what you think: I'm wondering if I'm missing something and the setting is crucial? I'm also wondering if you think that it should be removed completely?

Also let me know if I should do any changes to the PR; I tried to follow the guidelines and PR template but maybe I missed something.

---

## Checklist

### DIFF
--- a/lifecycle/gunicorn.conf.py
+++ b/lifecycle/gunicorn.conf.py
@@ -42,8 +42,8 @@ os.environ.setdefault("DJANGO_SETTINGS_MODULE", "authentik.root.settings")
 
 preload_app = True
 
-max_requests = 1000
-max_requests_jitter = 50
+max_requests = CONFIG.get_int("web.max_requests", 1000)
+max_requests_jitter = CONFIG.get_int("web.max_requests_jitter", 50)
 
 logconfig_dict = get_logger_config()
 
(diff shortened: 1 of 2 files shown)


rissson (approved):

rissson:
I'm actually thinking we might want to bump this to an even higher value like 10k. @BeryJu thoughts?

BeryJu (approved):

severin:
Thanks @rissson and @BeryJu! 

Would be great if this is released with the next patch of 2026.2; I assume this is going to happen based on the label applied?

rissson:
That is correct

PR #20770, opened 2026-03-06
website: override DocSearch button colors in light mode

dominic-r:
After:

image

image

Before:

image

image

### DIFF
--- a/website/docusaurus-theme/config.js
+++ b/website/docusaurus-theme/config.js
@@ -42,7 +42,10 @@ const CommonPresetOptions = {
         anonymizeIP: true,
     },
     theme: {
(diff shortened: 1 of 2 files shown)


tanberry (approved):
To my limited CSS skills, it looks great, and thank you very much @dominic-r !

kensternberg-authentik (approved):
This looks correct, but long term this should probably be in docusaurus-config.

dominic-r:
I agree, reasoning as to why:

> the changed file consumes the npm package rather than the source tree of ./packages/whatever, so I opted for the easier fix rather than putting it in docusaurus-config, patching node modules locally, making sure a new version is published and so on

PR #20785, opened 2026-03-08
blueprints: Add support for generating flat and unique sequences in blueprint enumeration tags

astro-stan:
## Details

I would like to be able to generate flattened and possibly unique sequences during blueprint templating.

The specific use case I have is something like this:

```yaml
...
context:
  authentication.backends: # Used for all password stages
    - authentik.core.auth.InbuiltBackend
entries:
 ...
  - model: authentik_stages_password.passwordstage
    identifiers:
      name:  ldap-password-stage
    attrs:
      backends:
       - !Enumerate
         - !Enumerate 
           # Use all configured backends as for the other password stages but ensure the token backend is included too
           - [!Context authentication.backends, authentik.core.auth.TokenBackend]
           - FLAT_SEQ
           - !Value 0
         - UNIQ_SEQ
         - !Value 0
```

Consequently, this also fixes a bug that went unnoticed for quite some time - If there are YAML tags nested more then one level in the `iterables` section of the `Enumerate` tag - they would not have been resolved and would have caused an error.

---

## Checklist

If applicable


astro-stan:
Remaining errors on CI seem unrelated as far as I can tell. I get the same errors when I apply the `tags.yaml` blueprint on `main`.

dewi-tik:
Same answer as here: https://github.com/goauthentik/authentik/pull/20490#issuecomment-5621974511

astro-stan:
Hi @dewi-tik, I've responded to the comment in the other PR, so I won't repeat the same here.

However, I also don't think this PR is making blueprints that much more complicated? It simply adds more modes to the Enumerate tag I originally authored. I've outlined, what I think is, a pretty reasonable use case in the PR description.

I'm happy to address any concerns you might have regarding the implementation.

PR #20802, opened 2026-03-09
web/flow: reset stale authenticator selection between consecutive validate stages

### LINKED ISSUE #14941: Available MFA device is not shown properly
**Describe the bug**
If you have multiple MFA validation stages after each other, as one stage with let's say Webauthn and TOTP, once the user passes this stage ang goes to the next stage that uses for example SMS and email, the flow will remember the last MFA device the user used and show the prompt for this even though the stage does not allow that specific method.

**To Reproduce**

Configure a following flow:

 Identification stage -> MFA validation stage with email or SMS -> MFA validation stage with TOTP or Webauthn, the user will now be presented with email if they used that in the previous stage, and have to manually click Change authenticator to a valid one in order to continue -> Login

**Expected behavior**
Show the user the only MFA option that is allowed in the current stage.

-   authentik version: 2025.4.1

Oluwatobi-Mustapha:
Fixes #14941

When the authenticator-validate stage component is reused across consecutive MFA stages, the previous selected challenge can persist into the next stage even if that challenge is no longer allowed.

This change resets the selected challenge when `challenge` updates and the previous selection is not present in the new `deviceChallenges` set, then re-runs the existing auto-selection initialization path for the new challenge.

Also adds unit coverage for stale-selection reset behavior.

### DIFF
--- a/web/src/flow/stages/authenticator_validate/AuthenticatorValidateStage.ts
+++ b/web/src/flow/stages/authenticator_validate/AuthenticatorValidateStage.ts
@@ -84,10 +84,6 @@ const createDevicePickerPropMap = () =>
 export function resolveAuthenticatorComponentTag(
     deviceClass: DeviceClassesEnum | null | undefined,
 ) {
-    if (deviceClass === null || deviceClass === undefined) {
-        return null;
-    }
-
     switch (deviceClass) {
         case DeviceClassesEnum.Static:
         case DeviceClassesEnum.Totp:
--- a/web/src/flow/stages/authenticator_validate/AuthenticatorValidateStage.ts
+++ b/web/src/flow/stages/authenticator_validate/AuthenticatorValidateStage.ts
@@ -81,7 +81,13 @@ const createDevicePickerPropMap = () =>
         },
     }) as const satisfies Record<DeviceClassesEnum, DevicePickerProps>;
 
-export function resolveAuthenticatorComponentTag(deviceClass?: DeviceClassesEnum | null) {
+export function resolveAuthenticatorComponentTag(
+    deviceClass: DeviceClassesEnum | null | undefined,
+) {
+    if (deviceClass === null || deviceClass === undefined) {
+        return null;
+    }
+
     switch (deviceClass) {
         case DeviceClassesEnum.Static:
         case DeviceClassesEnum.Totp:
@@ -92,9 +98,9 @@ export function resolveAuthenticatorComponentTag(deviceClass?: DeviceClassesEnum
             return "ak-stage-authenticator-validate-webauthn";
         case DeviceClassesEnum.Duo:
             return "ak-stage-authenticator-validate-duo";
+        default:
+            return null;
     }
-
-    return null;
 }
 
 @customElement("ak-stage-authenticator-validate")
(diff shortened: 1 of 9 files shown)


web/src/flow/stages/base.ts
```diff
@@ -46,8 +46,7 @@ export abstract class BaseStage<Tin extends StageChallengeLike, Tout = unknown>
 
     protected logger = ConsoleLogger.prefix(`flow:${this.tagName.toLowerCase()}`);
 
-    // TODO: Should have a property but this needs some refactoring first.
-    // @property({ attribute: false })
+    @property({ attribute: false })
```
kensternberg-authentik: `type: Object` is probably a better flag here than `attribute: false`.

web/src/flow/stages/authenticator_validate/AuthenticatorValidateStage.ts
```diff
@@ -127,7 +157,7 @@ export class AuthenticatorValidateStage
 
         // We don't use this.submit here, as we don't want to advance the flow.
         // We just want to notify the backend which challenge has been selected.
-        new FlowsApi(DEFAULT_CONFIG).flowsExecutorSolve({
+        this.#api.flowsExecutorSolve({
```
kensternberg-authentik, line 162: Hee hee!

web/src/flow/stages/authenticator_validate/AuthenticatorValidateStage.ts
```diff
@@ -77,6 +81,22 @@ const createDevicePickerPropMap = () =>
         },
     }) as const satisfies Record<DeviceClassesEnum, DevicePickerProps>;
 
+export function resolveAuthenticatorComponentTag(deviceClass?: DeviceClassesEnum | null) {
+    switch (deviceClass) {
```
kensternberg-authentik, line 87: The `deviceClass?` in the function signature and the lack of a `default` here add up to some kind of default behavior, but it's not clear when or why that would happen.  What would this function look like if "Sometimes we get nothing and that's okay?" was clearly communicated?

web/src/flow/stages/authenticator_validate/AuthenticatorValidateStage.ts
```diff
@@ -264,48 +304,31 @@ export class AuthenticatorValidateStage
...
     }
 
     protected renderAuthenticatorSelection(): TemplateResult {
         return html`<form class="pf-c-form">
             ${this.renderUserInfo()}${this.renderStagePicker()}${this.renderDevicePicker()}
         </form>`;
     }
-    render(): TemplateResult {
+
+    protected override render(): TemplateResult {
```
kensternberg-authentik, line 333: I know this isn't your circus and these aren't your monkeys, but while I was reviewing this I had that "This is doing two things isn't it?" and... sure enough, I get down to the renderer and it's doing two things.

What if there were two different classes that represented the two different things, and you could isolate their use cases?  Would that be possible here?  (I don't actually know enough to know if `renderStagePicker` and `renderDeviceChallenge` have independent existences, but it does look like, independent of `this.challenge`, they don't share a lot of business logic).

Oluwatobi-Mustapha: Hi @kensternberg-authentik, thanks for the review. I’ll take a closer look at this and get back to you tomorrow. It’s a little late here.

Oluwatobi-Mustapha: Hi again @kensternberg-authentik 

I have addressed the first two comments in this PR. 

for the third point (splitting the two responsibilities into separate classes), I agree it would improve clarity, but it’s a larger refactor.

would you prefer I do that in this PR now, or open a focused follow-up PR after this merges?

GirlBossRush (approved):

GirlBossRush:
@Oluwatobi-Mustapha Thank you sending such a detailed PR! 🥂 

The changes here look great and align with an ongoing effort to make the flow stages easier to test and reason about. I can merge this now and the changes should appear in the next release, as well as our next patch release of 2026.2

PR #20823, opened 2026-03-10
events: add option to configure webhook CA

BeryJu:
closes https://github.com/goauthentik/authentik/pull/10845


authentik/events/models.py
```diff
@@ -409,21 +421,29 @@ def send_webhook(self, notification: Notification) -> list[str]:
...
+                    exc.response.text if exc.response else str(exc)
+                ) from exc
+            return [
+                response.status_code,
+                response.text,
+            ]
+
+        if self.webhook_ca:
+            with DockerInlineTLS(self.webhook_ca) as tls:
+                return send(verify=tls.verify)
```
rissson: Looking at the code from `docker.tls.TLSConfig`, this will only do `verify=None`, no?

BeryJu: should be fixed now

rissson (approved):

website/docs/sys-mgmt/events/transports.md
```diff
@@ -43,6 +43,10 @@ This will send a POST request to the given URL with the following contents:
 
 The `Content-Type` header is set to `text/json`.
 
+##### Webhook Certificate Authority
+
+If the server the webhook URL points to does not have a certificate issued by a public certificate authority, you can select a Certificate-Keypair here which will be used to validate the server's certificate. If no keypair is selected, authentik uses the Root certificates from [mkcert.org](https://mkcert.org/).
```
dewi-tik:
```suggestion
If the server in the webhook URL doesn’t have a certificate issued by a public certificate authority, you can select a Certificate-Keypair here to validate the server’s certificate. If no keypair is selected, authentik will use the Root certificates from [mkcert.org](https://mkcert.org/).
```

dewi-tik (approved):
Minor suggestion, not essential. Otherwise LGTM from docs perspective.

PR #20834, opened 2026-03-10
web/flow: provide labels for the stage import-and-invoke table

kensternberg-authentik:
## What

Provide labels for the different parts of a stage import-and-invoke record. @BeryJu found the table-oriented set-up hard to read and, having revised the code, I think he was right. This is still more explicit: ‘switch/case/break’ statements are machinery, the *how* and not the *what*; these labels are all the “what” and the “how” is neatly tucked away in the constructor.

### DIFF
--- a/web/src/flow/FlowExecutorStageFactory.ts
+++ b/web/src/flow/FlowExecutorStageFactory.ts
@@ -21,19 +21,7 @@ export type { FlowChallengeComponentName, StageModuleCallback };
 export const propVariants = ["standard", "challenge"] as const;
 export type PropVariant = (typeof propVariants)[number];
 
-// The first type supports "import only."
-type StageEntryMetadata =
-    | []
-    | [tag: string]
-    | [variant: PropVariant]
-    | [tag: string, variant: PropVariant];
-
 const STANDARD = propVariants[0];
-const isImport = (x: unknown): x is StageModuleCallback => typeof x === "function";
-const PVariant = P.when(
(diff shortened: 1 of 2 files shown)


web/src/flow/FlowExecutorStageFactory.ts
```diff
@@ -60,12 +48,22 @@ function resolveStageTag(module: ResolvedDefaultESModule<BaseStageConstructor>):
...
 
+// The first type supports "import only."
+type StageEntryMetadata = {
+    tag: string | undefined;
+    variant: PropVariant | undefined;
+};
+
 interface StageMappingInit {
-    token: FlowChallengeComponentName;
+    stage: FlowChallengeComponentName;
```
kensternberg-authentik, line 58: Language change: `stage` is what the server calls it. Let's stick to that.

web/src/flow/FlowExecutorStageFactory.ts
```diff
@@ -60,12 +48,22 @@ function resolveStageTag(module: ResolvedDefaultESModule<BaseStageConstructor>):
     return tag;
 }
 
+// The first type supports "import only."
+type StageEntryMetadata = {
+    tag: string | undefined;
+    variant: PropVariant | undefined;
```
kensternberg-authentik, line 54: `ts-pattern` is surprisingly strict, making a distinction between "key present but value undefined" and "key missing" in object dereferencing.

GirlBossRush (approved):

BeryJu (approved):
❤️

PR #20880, opened 2026-03-13
web/applications: add wsfed to app wizard

PeshekDotDev:
## Details

Add wsfed to the application provider combo wizard

---

## Checklist

If applicable


web/src/admin/providers/wsfed/WSFederationProviderFormForm.ts
```diff
@@ -0,0 +1,334 @@
...
+                    required
+                    .errorMessages=${errors.sessionValidNotOnOrAfter}
+                    help=${msg("Session not valid on or after current time + this value.")}
+                ></ak-text-input>
+                <ak-form-element-horizontal
+                    label=${msg("Default NameID Policy")}
+                    required
+                    name="defaultNameIdPolicy"
+                >
+                    <select class="pf-c-form-control">
```
kensternberg-authentik, line 263: Suggestion:

At the top of your file:

```
const samlNameIDPolicyAndLabel = [
    [SAMLNameIDPolicyEnum.UrnOasisNamesTcSaml20NameidFormatPersistent, msg("Persistent")],
    [SAMLNameIDPolicyEnum.UrnOasisNamesTcSaml11NameidFormatEmailAddress, msg("Email address")],
    [SAMLNameIDPolicyEnum.UrnOasisNamesTcSaml20NameidFormatWindowsDomainQualifiedName, msg("Windows")],
    [SAMLNameIDPolicyEnum.UrnOasisNamesTcSaml11NameidFormatX509SubjectName, msg("X509 Subject")],
    [SAMLNameIDPolicyEnum.UrnOasisNamesTcSaml20NameidFormatTransient, msg("Transient")],
];
```

Then you can replace all of this with:

```
                    ${samlNameIDPolicyAndLabel.map(
                        ([policy, label]) =>
                            html`<option value=${policy} ?selected=${provider?.defaultNameIdPolicy === policy}>
                                ${label}>/option>
                            </option>`
                    )}
```

web/src/admin/providers/wsfed/WSFederationProviderFormForm.ts
```diff
@@ -0,0 +1,334 @@
...
+                            "Configure how the NameID value will be created. When left empty, the NameIDPolicy of the incoming request will be respected.",
+                        )}
+                    </p>
+                </ak-form-element-horizontal>
+                <ak-form-element-horizontal
+                    label=${msg("AuthnContextClassRef Property Mapping")}
+                    name="authnContextClassRefMapping"
+                >
+                    <ak-search-select
+                        .fetchObjects=${async (query?: string): Promise<SAMLPropertyMapping[]> => {
```
kensternberg-authentik: Really don't understand why people are still using this long form when ak-search-select-ez and arrow _expression_ functions exist and the Typescript inference engine is so excellent:

```
// outside the class definition. (This really should be in a utility file somewhere)
export const withQuery = <T>(search: string | undefined, args: T) =>
    search ? { ...args, search } : args;

   // At the top of your render function:
    const propertyMappingSearch: ISearchSelectConfig<SAMLPropertyMapping> = {
        fetchObjects: async (query?: string) =>
            (
                await new PropertymappingsApi(DEFAULT_CONFIG).propertymappingsProviderSamlList(
                    withQuery(query, { ordering: "saml_name" })
                )
            ).results,
        renderElement: (item: SAMLPropertyMapping) => item.name,
        value: (item: SAMLPropertyMapping | undefined) => item?.pk,
        selected: (item: SAMLPropertyMapping) => provider.authnContextClassRefMapping === item.pk,
    };

   // In your template:
   <ak-search-select-ez .config=${propertyMappingSearch}></ak-search-select-ez>
```

As general rules:
1. Don't supply the return type unless you absolutely need the client code to check if they're using it correctly.  You almost never need to supply a return type for code that isn't exported. _Do_ check with your IDE to affirm that the inferred type is what you expect, but then let the inferrer do its job.
2. If an arrow function is returning a single expression, skip the `{ return ... }` boilerplate and use the arrow expression function syntax.
3. It's rarely necessary to declare your args and arg type. If you pass a field or value the Api doesn't take, the type checker will tell you!

kensternberg-authentik (approved):
It looks solid.  I made a pair of suggestions.  If your code is working to your satisfaction, I don't recommend changing it; I just want to point out that there are cleaner ways of writing some things.

BeryJu (approved):

web/src/admin/providers/wsfed/WSFederationProviderFormForm.ts
PeshekDotDev: Much cleaner, I updated this

PR #20898, opened 2026-03-14
web/flow: extract lifecycle events peripheral to stage management into their own controllers

kensternberg-authentik:
## What

Three features embedded in FlowExecutor, Iframe message handling (from captchas), Multitab message handling, and Websocket message handling, have been extracted from the FlowExecutor and placed into their own controllers.

The `renderFrameBackground()` method has been removed.

# Why

The three features mentioned are all *peripheral* to the task of coordinating challenges. The Iframe message handling may result in a challenge being returned, but there’s a bit of set-up and tear down that doesn’t really correspond well to the central concern of the FlowExecutor; it’s more like a sub-stage of IdentificationStage. By being attached to the executor as Controllers they participate in the executor’s lifecycle and have access to it, but their own internal logic is separated out, making them easier to understand and maintain. As a result, all of the associated machinery– attaching to `window`, disconnecting the websocket client, and so on– can be removed from the FlowExecutor.

The `renderFrameBackground()` method is not used.


web/src/flow/FlowExecutor.ts
```diff
@@ -394,32 +358,6 @@ export class FlowExecutor extends WithBrandConfig(Interface) implements StageHos
         return html`<slot name="placeholder"></slot>`;
     }
 
-    protected renderFrameBackground(): SlottedTemplateResult {
```
kensternberg-authentik, line 418: This method is not called from anywhere. An analysis of the `./dist` folder shows that it's _so_ unused that the tree-shaker deletes it from the deployment.

web/src/flow/controllers/FlowIframeMessageController.ts
```diff
@@ -0,0 +1,56 @@
...
+}
+
+/**
+ * Coordinate with authentication events across iframe boundaries
+ *
+ * @remarks
+ *
+ * Third-party assistants in the authentication process, notably CAPTCHAs, are loaded in iframes.
+ * This controller listens for iframe boundary-crossing events and, if appropriate, forwards them to
+ * the FlowExecutor.
```
BeryJu: I dont think this comment is correct, this is for stages like gdtc that submit to the host from a frame

web/src/flow/controllers/FlowMultitabController.ts
```diff
@@ -0,0 +1,75 @@
...
+        /* no op */
+    }
+
+    onMultitab = () => {
+        const { challenge } = this.host;
+        if (!challenge) {
+            return;
+        }
+
+        document.title = "continued";
```
BeryJu:
```suggestion
```

This is a debugging leftover, cant believe it's still in the code 🙈

web/src/flow/FlowExecutor.ts
BeryJu: This not being used doesn't sit right, this should be used somewhere, ref https://github.com/goauthentik/authentik/pull/19527

BeryJu: https://github.com/goauthentik/authentik/pull/20063/changes#diff-4ed6c41b430eb15195ace2424e99a073db6c9e588f4627cf0f899ef9fad0f084L501 that's the culprit

BeryJu (approved):

PR #20900, opened 2026-03-14
core: redirect service accounts away from main UI like external users

fheisler:
Service account and internal service account users are now redirected to the brand's default application (or shown access denied) when accessing the homepage, consistent with external user behavior. Adds interface view tests covering all user types.


rissson (approved):
A fletcher PR! With tests nonetheless!

BeryJu (approved):

PR #20921, opened 2026-03-16
packages/django-dramatiq-postgres: scheduler: only dispatch tasks if they're not running yet

### LINKED ISSUE #20885: Add uniqueness constraint to periodic tasks

Currently periodic tasks get enqueued strictly based on their schedule. This can lead to unnecessary overhead if queue processing is slow by having multiple instances of the same task scheduled.

An example:
The cleanup task is enqueued every 5 minutes. Suppose there is an issue with queue processing (maybe a spike of tasks or a worker process becoming unresponsive for some time) and the task is not completed within 5 minutes. The scheduler now enqueues another instance of the cleanup task and there are now two tasks in the queue. 

The scheduler should check the queue before enqueuing a periodic task: if there is a task of the same class already enqueued then it should not enqueue another instance of the same task.

Coming from the Ruby world I would like to see something like _sidekiq_'s unique jobs: https://sidekiq.org/wiki/Ent-Unique-Jobs

I don't know if _dramatiq_ has a task uniqueness functionality already built-in or if there is a plugin that offers it. But since authentik uses postgres as a backend it should be relatively straightforward to implement a custom middleware that queries the tasks table for an existing task before enqueueing a new one.

The current implementation works, it just potentially causes overhead and unnecessary load on the task system.

### Additional context

_No response_

rissson:
## Details

closes https://github.com/goauthentik/authentik/issues/20885

---

## Checklist

If applicable

### DIFF
--- a/packages/django-dramatiq-postgres/django_dramatiq_postgres/scheduler.py
+++ b/packages/django-dramatiq-postgres/django_dramatiq_postgres/scheduler.py
@@ -1,7 +1,6 @@
 from typing import Any, cast
 
 import pglock
-from django.db import router, transaction
 from django.db.models import QuerySet
 from django.utils.functional import cached_property
 from django.utils.module_loading import import_string
@@ -47,7 +46,7 @@ class Scheduler:
         count = 0
         for schedule in self.query_set.filter(next_run__lt=now()):
             if (
-                schedule.tasks.using(self.db_alias)
+                schedule.tasks.using(self.db_alias)  # type: ignore[attr-defined]
                 .exclude(state__in=(TaskState.DONE, TaskState.REJECTED))
                 .exists()
             ):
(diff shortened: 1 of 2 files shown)


packages/django-dramatiq-postgres/django_dramatiq_postgres/scheduler.py
```diff
@@ -28,12 +28,12 @@ def model(self) -> type[ScheduleBase]:
...
     @property
     def query_set(self) -> QuerySet[ScheduleBase]:
-        return self.model._default_manager.filter(paused=False)
+        return self.model._default_manager.using(self.db_alias).filter(paused=False)
 
     def process_schedule(self, schedule: ScheduleBase) -> None:
         schedule.next_run = schedule.compute_next_run()
         schedule.send(self.broker)
-        schedule.save()
+        schedule.save(update_fields=["next_run"])
```
BeryJu, line 36: Does this need a similar fallback as in the other place where we do a save with `update_fields` in case the field hasn't changed?

rissson: It shouldn't. compute_next_run will always return a value in the future, which will be different from the current one, which will always be in the past

PR #19217, opened 2026-01-06
website/docs: add a new page to help people get started after install is complete 

tanberry:
This PR will add a tutorial style doc to help people get going after they finish installing authentik.

-   [x] The documentation has been updated
-   [x] The documentation has been formatted (`make docs`)

PR #20961, opened 2026-03-18
website/docs: update kubernetes install guide for Gateway API

dominic-r:
Replace the old `server.gateway` example with the current `server.route.main` configuration and clarify that the Gateway resource must be created separately.

Also: keep an Ingress fallback example for clusters without Gateway API support.

### DIFF
--- a/website/docs/install-config/install/kubernetes.md
+++ b/website/docs/install-config/install/kubernetes.md
@@ -43,17 +43,33 @@ authentik:
         password: "ThisIsNotASecurePassword"
 
 server:
-    gateway:
-        # Specify kubernetes gateway controller class name
-        GatewayClassName: nginx | traefik | kong
-        enabled: true
-        hosts:
-            - authentik.domain.tld
+    route:
+        main:
+            enabled: true
+            hostnames:
+                - authentik.domain.tld
+            parentRefs:
+                - name: shared-gateway
+                  namespace: default
 
 postgresql:
     enabled: true
     auth:
         password: "ThisIsNotASecurePassword"
+```
+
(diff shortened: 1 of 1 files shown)


BeryJu:
I'd rather keep both of them for people that haven't migrated away from ingress yet, but seems that our ingress docs got broken in https://github.com/goauthentik/authentik/pull/19217/changes#diff-8d59f1f5bad5e5cb3c2e4967c920f514a8486edabecb788cfc62a2897c279dc6R45-R48

so need to revert that gateway to ingress

website/docs/install-config/install/kubernetes.md
```diff
@@ -43,19 +43,35 @@ authentik:
...
 
 postgresql:
     enabled: true
     auth:
         password: "ThisIsNotASecurePassword"
 ```
 
+The Helm chart creates an `HTTPRoute`, but it does not create `Gateway` or `GatewayClass` resources. Create the `Gateway` separately, then set `server.route.main.parentRefs` to that `Gateway` resource's name and namespace. In the example above, `name: shared-gateway` and `namespace: default` must match the manually created `Gateway`.
+
+If your cluster or controller does not support the Gateway API yet, replace the `server.route` section above with this Ingress configuration:
```
dewi-tik:
```suggestion
If your cluster or controller does not support the Gateway API, replace the `server.route` section above with this Ingress configuration:
```

dewi-tik (approved):
One tiny change otherwise LGTM

rissson (approved):

BeryJu (approved):

PR #20978, opened 2026-03-18
enterprise/endpoints/connectors: Fleet conditional access stage

### LINKED ISSUE #20959: Investigate fleetdm connector with endpoint stage
- Requires Okta conditional access configured in fleet?
  (not sure)
- controller should fetch Conditional access root ca public key into authentik
- when selected, proxy to mtls stage with cert from above
- Wrap stage to do the device lookup after cert has been validated

That would make it easier but still require the admin to setup mtls requirements as in reverse proxy, `AutoSelectCertificateForUrls`, etc

BeryJu:
Previously the fleet connector was only there to ingest data from fleet's devices into authentik, but you still needed the authentik agent or for example google chrome device trust to associate the device to the session.

However since fleet has support for doing this with certificates with okta (in which case fleet acts as an intermediate SMAL IDP, which we can also support), we can use our mTLS stage to effectively do the same thing.

This PR adds logic to
- Pull in the Conditional access Root CA Cert from fleet
- Add a fleet stage which uses the mTLS stage logic to find a device for the given certificate
- A bunch of tests for the above

closes #20959


website/docs/endpoint-devices/device-compliance/fleet-conditional-access.md
```diff
@@ -0,0 +1,37 @@
...
+
+authentik automatically retrieves the Conditional Access Root CA certificate from Fleet via the Fleet connector. The Endpoint stage then verifies the device’s Fleet-issued certificate against this Root CA. If validation succeeds, the device is bound to the user’s current authentik session.
+
+## Prerequisites
+
+- You must have [configured compliance](./configuration.md) in authentik
+- The [Fleet connector](./connectors/fleetdm.md) must be configured in authentik
+- Conditional access Root CA Certificate must be pulled from Fleet via the Fleet connector. This is an automatic process
+- A Fleet Enterprise license is required
+
```
BeryJu, line 18: The other thing we should note here is that the fleet conditional access mobileconfig profile must be applied to your machines to make this work.

BeryJu (commented):
LGTM docs wise aside from one point

PR #20979, opened 2026-03-18
providers/oauth2: evaluate property mappings in client credentials JWT flow

### LINKED ISSUE #18767: client_credentials grant does not apply oauth2 source property mappings
### Describe the bug

I've set up an oauth2 provider, google as an oauth2 source, have added a property mapping to the google source `return {"username": info["email"], "email: info["email"]}`. I can get a jwt from the oauth2 provider when passing a google-generated jwt, but the username+email claims in my returned authentik jwt are never populated.  Is that expected?  I see here that the autogenerated users from a jwt login never have their email field set... https://github.com/goauthentik/authentik/blob/main/authentik/providers/oauth2/views/token.py#L537

Does this grant simply bypass all the oauth2 customization configuration?

Also logging calls in the property mappings are not present in the server logs... is that expected?

### How to reproduce

1. Set up an oauth2 provider
2. Added a property mapping to the google source `return {"username": info["email"], "email: info["email"]}`
3. Set up google as a federated oauth2 source, using the property mapping from 2
4. request a token via the `client_credentials` grant, passing a google-issued jwt access token
5. inspect the returned token - no username/email claims.

### Expected behavior

I'd see those claims with values mapped from the passed google token.

### Screenshots

_No response_

### Additional context

_No response_

### Deployment Method

Docker

### Version

2025.10.0

### Relevant log output

```shell

```

Rishabh-git10:
## Details

Currently, the `client_credentials` flow using JWT assertions bypasses property mappings, hardcoding the generated user's attributes based solely on the `sub` claim and application name.

This PR updates `__create_user_from_jwt` to leverage the core `SourceMapper` engine. It now properly evaluates configured user property mappings and applies them to the generated service account, aligning this flow's behavior with standard Authentik logins. 

- Replaced manual user attribute assignment with `SourceMapper().build_object_properties()`.
- Added unit tests to verify correct mapping extraction from the incoming JWT payload.

closes #18767

---

## Checklist

If applicable


BeryJu (approved):
Looks good, made a couple small changes:

- Always top level imports unless they're needed inline to avoid circular imports
- clamp the username at its max length (technically an unrelated change)
- keep the original test case and change your modification into a new test case

authentik/providers/oauth2/views/token.py
```diff
@@ -530,18 +531,30 @@ def __post_init_device_code(self, request: HttpRequest):
...
         self.device_code = code
 
-    def __create_user_from_jwt(self, token: dict[str, Any], app: Application, source: OAuthSource):
+    def __create_user_from_jwt(
+        self, token: dict[str, Any], app: Application, source: OAuthSource, request: HttpRequest
+    ):
         """Create user from JWT"""
         with audit_ignore():
+            # Run the JWT payload through the core mapping engine
+            mapped = SourceMapper(source).build_object_properties(
```
rissson, line 540: This is weird to me, using the SourceMapper in a provider. While the semantics currently are somewhat the same, they might change in the future. I'd rather use the same code as in `get_claims` in userinfo.py

rissson: disregard, I can't read

rissson (approved):

authentik/providers/oauth2/views/token.py
BeryJu: SourceMapper is actually correct here since this is the codepath for authenticating using a JWT created by a configured federated source

PR #20982, opened 2026-03-18
website/docs: fix notification transport etc

BeryJu:
before: 
image
after:
image

### DIFF
--- a/website/docs/sys-mgmt/events/notification_rule_expression_policies.mdx
+++ b/website/docs/sys-mgmt/events/notification_rule_expression_policies.mdx
@@ -1,5 +1,6 @@
 ---
 title: Expression Policies
+sidebar_label: Notification Rule Expression Policies
 ---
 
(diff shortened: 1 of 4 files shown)


dominic-r (approved):

website/docs/sys-mgmt/events/notification_rule_expression_policies.mdx
```diff
@@ -1,5 +1,5 @@
 ---
-title: Notification Rule Expression Policies
+title: Expression Policies
```
tanberry, line 2: We already have a separate page about Expression policies, so let's not use the exact same title (bad SEO mojo).

BeryJu: yeah maybe I should only updated the sidebar label

tanberry: That would work well, and not mess with SEO... thanks!

PR #20984, opened 2026-03-18
web/admin: fix missing OSM referrerPolicy header

BeryJu:
OSM started enforcing this, and due to our usage of openlayers-elements which is on OpenLayers 7.5 instead of 10.x this didn't get fixed by a dependency update.

see https://operations.osmfoundation.org/policies/tiles/


kensternberg-authentik (approved):
After reading the OpenStreetmap API docs, this looks straightforward and correct.  Builds cleanly and runs without apparent error.

PR #21010, opened 2026-03-19
providers/saml: skip unsupported bindings when importing metadata

### LINKED ISSUE #21009: SAML Provider metadata import selects first endpoint even if binding is unsupported
### Describe the bug

When creating a new SAML Provider via “SAML Provider from Metadata”, authentik sometimes selects an endpoint binding that authentik does not support.
This happens when the metadata contains multiple endpoints and the first entry uses an unsupported binding; the current selection logic ends up applying that first entry anyway, resulting in an invalid configuration (see screenshot).

This behavior appears to be “by design” in the current implementation (it is even mentioned in comments), but it leads to a poor default and a broken provider setup out-of-the-box.

I am preparing a PR to change the selection logic so that authentik skips unsupported bindings and chooses a supported default (e.g., HTTP-POST / HTTP-Redirect) instead. This issue is to track that change and link the PR.

### How to reproduce

1. Go to Admin console.
2. Click Applications / Provides
3. Click "Create" button
4. Choose SAML Provider from Metadata, then click Next
5. Input and select arbitrary Name and flows. And upload attached metadata.
6. Click finish.
7. Looks like successfully configured Provider, but appears URL of unsupported binding, such as SOAP.

[multi-bindings.xml](https://github.com/user-attachments/files/26109226/multi-bindings.xml)

### Expected behavior

Expected behavior
- When importing metadata, authentik should prefer supported bindings and never auto-select an unsupported binding as the provider’s default.

Actual behavior
- The provider is configured with the first binding from metadata even if it is unsupported, causing misconfiguration.

### Screenshots

Image

### Additional context

_No response_

### Deployment Method

Docker

### Version

2026.5.0-rc1

### Relevant log output

```shell

```

ikob:
## Details

The "SAML Provider from Metadata" flow could select the first ACS/SLS endpoint even when its
Binding is not supported by authentik, resulting in an invalid provider configuration.

Prefer supported bindings (HTTP-POST / HTTP-Redirect) and skip unsupported entries when creating SAML provider from metadata.

This patch will close #21009

Note: Split metadata extraction from endpoint selection policy to keep parsing and decisions separate.

---

## Checklist

If applicable

### DIFF
--- a/authentik/providers/saml/processors/metadata_extract.py
+++ b/authentik/providers/saml/processors/metadata_extract.py
@@ -1,0 +1,111 @@
+"""SAML metadata XML extractors (no policy): parse, walk, and collect candidates."""
+
+from typing import Any
+
+from lxml import etree  # nosec
+from structlog.stdlib import get_logger
+
+from authentik.common.saml.constants import (
+    NS_MAP,
+    SAML_BINDING_POST,
+    SAML_BINDING_REDIRECT,
+)
+
+LOGGER = get_logger()
+
+
+def extract_sp_descriptor(entity: etree._Element) -> etree._Element:
+    """Return the first SPSSODescriptor element."""
+    sp = entity.xpath("./md:SPSSODescriptor", namespaces=NS_MAP)
+    if not sp:
+        raise ValueError("EntityDescriptor has no SPSSODescriptor")
+    return sp[0]
+
+
+# ============================================================
+# Candidate extraction
+# ============================================================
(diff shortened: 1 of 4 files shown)


ikob:
Updated this branch onto the latest `main`.

dewi-tik:
Thanks @ikob 

We've created a new PR to do this but in a slightly different way: #24716

Thank you regardless.

PR #21019, opened 2026-03-19
fix(core/api): normalize oversized JSON integers in dict fields

Oluwatobi-Mustapha:
Fixes #20868

When user/group attributes contain integers outside the signed 64-bit range, `JSONDictField` accepts them, but API response rendering can fail with `TypeError: Integer exceeds 64-bit range` from `orjson`.

This change normalizes out-of-range integers to strings in `JSONDictField` on both input (`to_internal_value`) and output (`to_representation`). Values inside int64 range remain unchanged, and previously stored oversized values can now be rendered safely.

Also adds unit coverage for large-integer normalization and API coverage for listing users with oversized integer attributes.

### DIFF
--- a/authentik/core/api/utils.py
+++ b/authentik/core/api/utils.py
@@ -29,10 +29,34 @@ def is_dict(value: Any):
     raise ValidationError("Value must be a dictionary, and not have any duplicate keys.")
 
 
+INT_64_MIN = -(2**63)
+INT_64_MAX = 2**63 - 1
+
+
+def normalize_large_integers(value: Any) -> Any:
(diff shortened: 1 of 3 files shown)


rissson:
we're actually looking into using a serializer that supports oversized ints. cc. @melizeche

Oluwatobi-Mustapha:
do you want me to rework this pr to use that serializer approach? @rissson

rissson:
I'll let marcelo comment on this depending on what his plan is

BeryJu:
not needed anymore with https://github.com/goauthentik/authentik/pull/23528

PR #21041, opened 2026-03-20
internal/outpost/ak: fix ws URL on outpost restart

rissson:
## Details

REPLACE ME

---

## Checklist

If applicable

### DIFF
--- a/internal/outpost/ak/api_event.go
+++ b/internal/outpost/ak/api_event.go
@@ -32,9 +32,11 @@ func (ac *APIController) getWebsocketURL(akURL url.URL, outpostUUID string, quer
 	return wsUrl
 }
 
-func (ac *APIController) initEvent(akURL url.URL, outpostUUID string) error {
+func (ac *APIController) initEvent(outpostUUID string, attempt int) error {
+	akURL := ac.akURL
 	query := akURL.Query()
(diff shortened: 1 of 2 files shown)


internal/outpost/ak/api_event.go
```diff
@@ -106,18 +108,10 @@ func (ac *APIController) recentEvents() {
 		return
 	}
 	ac.wsIsReconnecting = true
-	u := url.URL{
-		Host:   ac.Client.GetConfig().Host,
-		Scheme: ac.Client.GetConfig().Scheme,
-		Path:   strings.ReplaceAll(ac.Client.GetConfig().Servers[0].URL, "api/v3", ""),
```
BeryJu, line 112: this is actually a good point, do we account for subpaths in the unix socket request anywhere?

rissson: mmh we don't actually

rissson: incorrect, we do let me find the line

rissson: https://github.com/goauthentik/authentik/blob/main/cmd/server/server.go#L58 there we go

BeryJu (approved):

PR #21046, opened 2026-03-20
enterprise/lifecycle: remove one review per object limitation

atereshkin:
## Details

This lifts the  requirement that only one Lifecycle Rule can apply to a single object, and, therefore, allows multiple concurrent reviews per obejct. 

Additionally, this removes task list from the lifecycle rule list page and attaches the tasks to the schedule instead of rule.

---

## Checklist

If applicable


authentik/enterprise/lifecycle/tasks.py
rissson: this should be attached to the schedule.

rissson (changes requested):

authentik/enterprise/lifecycle/tasks.py
atereshkin: Sorry, missed this one.

rissson (approved):

website/docs/sys-mgmt/object-lifecycle-management.md
```diff
@@ -22,9 +22,9 @@ You can create and configure Lifecycle rules via the **Events** > **Lifecycle Ru
 A lifecycle rule can be scoped to:
 
 - **A specific object**: The rule applies only to that individual Application, Group, or Role.
-- **An entire object type**: The rule applies to all objects of that type that don't have their own specific rule, e.g., all applications.
+- **An entire object type**: The rule applies to all objects of that type, e.g., all applications.
```
dominic-r:
```suggestion
- **An entire object type**: The rule applies to all objects of that type (e.g., all applications).
```

dominic-r (approved):

authentik/enterprise/lifecycle/models.py
```diff
@@ -214,9 +199,15 @@ def _get_event_args(self) -> dict:
         }
 
     def initialize(self):
+        if (self.content_type.app_label, self.content_type.model) == ("authentik_core", "group"):
```
kensternberg-authentik, line 202: Like, this is just an _or_ statement, right?

authentik/enterprise/lifecycle/tasks.py
```diff
@@ -4,14 +4,17 @@
...
-@actor(description=_("Dispatch tasks to validate lifecycle rules."))
+@actor(description=_("Dispatch tasks to apply lifecycle rules."))
 def apply_lifecycle_rules():
     for rule in LifecycleRule.objects.all():
         apply_lifecycle_rule.send_with_options(
             args=(rule.id,),
-            rel_obj=rule,
+            rel_obj=Schedule.objects.get(
+                actor_name="authentik.enterprise.lifecycle.tasks.apply_lifecycle_rules"
+            ),
```
kensternberg-authentik, line 17: _Ahem_.  Lifecycle Rule Scheduling (here and the one above) appears to be unrelated to the "one review per object limitation" description.  Just sayin'.

web/src/admin/lifecycle/ObjectLifecyclePage.ts
```diff
@@ -58,237 +44,67 @@ export class ObjectLifecyclePage extends Table<Review> {
...
-            <div class="pf-c-page__main-section pf-m-no-padding-mobile">
-                <div class="pf-l-grid pf-m-gutter">
-                    ${this.renderReviewSummary()}
-                    <div class="pf-c-card pf-l-grid__item pf-m-9-col">
-                        <div class="pf-c-card__title">${msg("Reviews")}</div>
-                        ${super.render()}
-                    </div>
-                </div>
+        return html`
+            <ak-lifecycle-preview-banner slot="header"></ak-lifecycle-preview-banner>
```
kensternberg-authentik: I'm a little baffled by the `slot="header"` addition here.  The `<ak-lifecycle-preview-banner>` component is being rendered inside the `<ak-object-lifecycle-page>` object's shadow; it'll have no effect at all.  The purpose of a slot is to _position_ something you put in the element's _lightDOM_ with respect to the element's template of HTML; that's not happening here.  In one respect, it's probably harmless, but it's a place where maintenance confusion can creep in.

web/src/admin/lifecycle/ObjectReviewIteration.ts
```diff
@@ -0,0 +1,281 @@
...
+
+    //#region Protected Properties
+
+    protected override emptyStateMessage = msg("No reviews yet.");
+
+    protected columns: TableColumn[] = [
+        [msg("Reviewed on"), "timestamp"],
+        [msg("Reviewer"), "reviewer"],
+        [msg("Note"), "note"],
+    ];
```
kensternberg-authentik, line 60: A point of, well, education. First, _this is correct_.  But you may not understand why.  :-). Normally, for a static table like this, I would be tempted to either mark it `static` or move it up and out of the component. In both of those cases, though, the `msg()` function would run just after the bundle is pulled from the server, which could be before the locale is determined, which would then cause the labels to be provided in the wrong language.  By rebuilding them with every new instance, you guarantee they're populated with the locale that was live when the page is loaded, not when the bundle is pulled.

web/src/admin/lifecycle/ObjectReviewIteration.ts
```diff
@@ -0,0 +1,281 @@
...
+        return match(this.iteration?.state)
+            .with(P.nullish, LifecycleIterationStateEnum.UnknownDefaultOpenApi, () => nothing)
+            .with(
+                LifecycleIterationStateEnum.Pending,
+                () => html`${this.renderOpenedOn()}${this.renderGracePeriodTill()}`,
+            )
+            .with(LifecycleIterationStateEnum.Reviewed, () => this.renderNextReviewDate())
+            .with(LifecycleIterationStateEnum.Overdue, () => this.renderOpenedOn())
+            .with(LifecycleIterationStateEnum.Canceled, () => this.renderOpenedOn())
+            .exhaustive();
```
kensternberg-authentik, line 188: 👍 `match` is awesome, especially when paired with `exhaustive`.

kensternberg-authentik (approved):
I have some nitpicking (mostly about the Python), but overall this is solid. My biggest concern is the `slot="header"` no-op.  It can stay for now, since we're time-compressed, but I'd like to schedule removing it since it will complicate future maintenance.

authentik/enterprise/lifecycle/tasks.py
atereshkin: Yeah, it's just something that came up while I was working on the main thing. Should have been a separate PR. Too late to untangle now, but added a mention of this change that to the description.

web/src/admin/lifecycle/ObjectReviewIteration.ts
atereshkin: Noted.

authentik/enterprise/lifecycle/models.py
atereshkin: True. But it's merely a coincidence that the then-clauses are identical. The objects in `self.object` in respective branches are different types independently having a field called `name`. So unless  you insist I'd leave it like this.

web/src/admin/lifecycle/ObjectLifecyclePage.ts
atereshkin: Sorry, must have made sense at some point but then was moved out of the context where it did in a series of refactorings. Fixed.

PR #21054, opened 2026-03-20
website/integrations: fix Microsoft 365 federation guide

dominic-r:
Based on me setting it up in my tenant last night

### DIFF
--- a/website/integrations/platforms/microsoft/index.md
+++ b/website/integrations/platforms/microsoft/index.md
@@ -42,7 +42,7 @@ If you are using an [Active Directory source](/docs/users-sources/sources/direct
 
 #### For cloud-only users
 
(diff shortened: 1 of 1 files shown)


website/integrations/platforms/microsoft/index.md
```diff
@@ -41,29 +42,51 @@ If you are using an [Active Directory source](/docs/users-sources/sources/direct
...
 
         ```python showLineNumbers
         # For users synchronized from Active Directory with the 'entra_immutable_id' attribute.
         # Replace 'entra_immutable_id' with whatever you set this attribute's name to.
         return user.attributes.get("entra_immutable_id", "")
 
-        # OR for cloud-only users with email address as their immutable ID
-        # return user.email
+        # OR for cloud-only users whose authentik username matches their Entra UPN
+        return user
```
BeryJu: I dont think thats right

dominic-r:
```suggestion
        return user.email
```

dewi-tik (approved):

PR #21073, opened 2026-03-23
fix: guard signal handlers and is_known_device against unsaved user

themavik:
## Summary
Fixes #21036.
**Root cause:** The identification stage creates a dummy unsaved User (pk=None) to prevent timing-based user enumeration. When this user propagates to the login stage, `is_known_device()` passes it to `AuthenticatedSession.objects.filter(user=user)`, which raises `ValueError` because Django requires saved model instances in related filters.
**Fix:** Added `if not user.pk: return` guards before DB queries that receive the user object.

## Changes
- `authentik/stages/user_login/stage.py`: return False from `is_known_device()` when user.pk is None
- `authentik/core/signals.py`: skip `user_logged_in_session` when user.pk is None
- `authentik/events/signals.py`: skip `on_user_logged_in` when user.pk is None

## Testing
- Verified all three call sites that receive `user` from the `user_logged_in` signal now guard against unsaved instances

### DIFF
--- a/authentik/core/signals.py
+++ b/authentik/core/signals.py
@@ -47,6 +47,8 @@ def post_save_application(sender: type[Model], instance, created: bool, **_):
 @receiver(user_logged_in)
 def user_logged_in_session(sender, request: HttpRequest, user: User, **_):
     """Create an AuthenticatedSession from request"""
+    if not user.pk:
+        return
 
     session = AuthenticatedSession.from_request(request, user)
     if session:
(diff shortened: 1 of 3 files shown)


authentik/core/signals.py
```diff
@@ -47,6 +47,8 @@ def post_save_application(sender: type[Model], instance, created: bool, **_):
 @receiver(user_logged_in)
 def user_logged_in_session(sender, request: HttpRequest, user: User, **_):
     """Create an AuthenticatedSession from request"""
+    if not user.pk:
+        return
```
BeryJu, line 51: These signals should not be reachable without a `pk` on the user

PR #21074, opened 2026-03-23
providers/scim: add compatibility mode for qualified.com

flx5:
## Details
Qualified does not have a service provider config endpoint but supports filtering (to a degree).
Enable filtering here so that we can sync existing users.

---

## Checklist

If applicable


authentik/providers/scim/models.py
```diff
@@ -81,6 +81,7 @@ class SCIMCompatibilityMode(models.TextChoices):
     AWS = "aws", _("AWS")
     SLACK = "slack", _("Slack")
     SALESFORCE = "sfdc", _("Salesforce")
+    QUALIFIED = "qualified", _("Qualified")
```
BeryJu:
```suggestion
    QUALIFIED = "qualified", _("Qualified.com")
```

web/src/admin/providers/scim/SCIMProviderFormForm.ts
```diff
@@ -198,6 +198,11 @@ export function renderForm({ provider = {}, errors = {}, update }: SCIMProviderF
                             value: CompatibilityModeEnum.Sfdc,
                             description: html`${msg("Altered behavior for usage with Salesforce.")}`,
                         },
+                        {
+                            label: msg("Qualified"),
```
BeryJu:
```suggestion
                            label: msg("Qualified.com"),
```

BeryJu (approved):
LGTM, could you also either run `makemigrations` or update the previous SCIM provider migration with the options change

flx5:
Sure. Migrations generated

PR #21111, opened 2026-03-24
providers/proxy: Add a default maxResponseBodySize to Traefik Middleware

wiiaam:
## Details

In response to CVE-2026-26998, Traefik added a `maxResponseBodySize` configuration option and is recommending this be set, otherwise a warning will show in the traefik logs.

https://doc.traefik.io/traefik/reference/routing-configuration/http/middlewares/forwardauth/#maxresponsebodysize

Added a default maxResponseBodySize of 4MB to the `traefik middleware` proxy outpost component. Also fixed an issue where the reconciler would throw an error when trying to run a json patch on a non openapi object.

---

## Checklist

If applicable

### DIFF
--- a/authentik/outposts/controllers/k8s/base.py
+++ b/authentik/outposts/controllers/k8s/base.py
@@ -185,8 +185,10 @@ class KubernetesObjectReconciler[T]:
 
         patch = self.get_patch()
         if patch is not None:
-            current_json = ApiClient().sanitize_for_serialization(current)
-
+            try:
+                current_json = ApiClient().sanitize_for_serialization(current)
+            except AttributeError:
+                current_json = asdict(current)
             try:
                 if apply_patch(current_json, patch) != current_json:
                     raise NeedsUpdate()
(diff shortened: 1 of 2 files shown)


BeryJu (approved):
LGTM

@wiiaam have you checked what happens if someone were to run this but hadn't upgraded their CRDs in their cluster?

wiiaam:
The API server will just prune the field. The only situation where this would cause an error is if field_validation was set to Strict when calling create_namespaced_custom_object, currently this will revert to the Warn default

https://github.com/kubernetes-client/python/blob/master/kubernetes/docs/CustomObjectsApi.md#create_namespaced_custom_object

I'm happy to do a follow up PR to catch these warnings (for all k8s objects) and show it in the task status?

BeryJu:
> The API server will just prune the field. The only situation where this would cause an error is if field_validation was set to Strict when calling create_namespaced_custom_object, currently this will revert to the Warn default
> 
> https://github.com/kubernetes-client/python/blob/master/kubernetes/docs/CustomObjectsApi.md#create_namespaced_custom_object
> 
> I'm happy to do a follow up PR to catch these warnings (for all k8s objects) and show it in the task status?

Fair enough, yeah that would be nice to add if you have resources for that

PR #21115, opened 2026-03-24
events: add helper to log deprecation configuration_warning message

BeryJu:
extracted from https://github.com/goauthentik/authentik/pull/18274

### DIFF
--- a/authentik/core/models.py
+++ b/authentik/core/models.py
@@ -545,21 +545,9 @@ class User(SerializerModel, AttributesMixin, AbstractUser):
             cause=cause,
             stacktrace=stacktrace,
         )
-        if not Event.objects.filter(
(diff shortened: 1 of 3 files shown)


authentik/core/models.py
```diff
@@ -1138,24 +1126,17 @@ def expire_action(self, *args, **kwargs):
...
-                    ".filter_not_expired() is deperecated as the default lookup now excludes "
-                    "expired objects."
-                ),
-            )
-            event.expires = datetime.now() + timedelta(days=30)
-            event.save()
+        Event.log_deprecation(
+            deprecation_id,
+            message=(
+                ".filter_not_expired() is deperecated as the default lookup now excludes "
```
rissson:
```suggestion
                ".filter_not_expired() is deprecated as the default lookup now excludes "
```

PR #21117, opened 2026-03-24
packages/client-rust: init

rissson:
## Details

REPLACE ME

---

## Checklist

If applicable


Makefile
```diff
@@ -239,10 +239,13 @@ gen-client-go: gen-clean-go  ## Build and install the authentik API for Golang
 	make -C ${PWD}/${GEN_API_GO} build version=${NPM_VERSION}
 	go mod edit -replace goauthentik.io/api/v3=./${GEN_API_GO}
 
+gen-client-rs:
+	make -C packages/client-rust build version=${NPM_VERSION}
```
```suggestion
	make -C "${PWD}/packages/client-rust" build version=${NPM_VERSION}
```

BeryJu (approved):
we maybe should also add a CI step to validate that the generated client(s) match the schema

BeryJu:
oh and also since lint is failing, add rustfmt to the client makefile

Makefile
rissson: updated, and in a few other places too where we referenced .cargo/...

rissson:
> we maybe should also add a CI step to validate that the generated client(s) match the schema

it's there!

rissson:
> oh and also since lint is failing, add rustfmt to the client makefile

yeah, I added that so that `make gen-client-rs` will format after calling down to the client. I'm not sure if we should do the `cargo fmt` in the client directly since we rely on global configuration for that.

Either way, the entry point for developers should be `make gen-client-rs`, so that shouldn't be an issue

PR #21120, opened 2026-03-24
packages/client-ts: init

rissson:
## Details

REPLACE ME

---

## Checklist

-   [ ] Local tests pass (`ak test authentik/`)
-   [ ] The code has been formatted (`make lint-fix`)

If an API change has been made

-   [ ] The API schema has been updated (`make gen-build`)

If changes to the frontend have been made

-   [ ] The code has been formatted (`make web`)

If applicable

-   [ ] The documentation has been updated
-   [ ] The documentation has been formatted (`make docs`)

PR #21141, opened 2026-03-25
root: optimise api client generation speed

BeryJu:
(With rust and go client, this is before #21120)

Before:

```
________________________________________________________
Executed in   89.26 secs    fish           external
   usr time   29.53 secs    0.27 millis   29.53 secs
   sys time    8.69 secs    1.65 millis    8.69 secs
```

After:

```
________________________________________________________
Executed in   58.25 secs    fish           external
   usr time   32.49 secs  194.00 micros   32.49 secs
   sys time   10.58 secs  995.00 micros   10.58 secs
```

(In addition to that, after adding the ts client, the time should stay roughly the same)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,7 @@
 
 SHELL := /usr/bin/env bash
 .SHELLFLAGS += ${SHELLFLAGS} -e -o pipefail
+CPUS ?= $(shell (nproc --all || sysctl -n hw.ncpu) 2>/dev/null || echo 1)
 PWD = $(shell pwd)
 UID = $(shell id -u)
 GID = $(shell id -g)
@@ -239,7 +240,10 @@ gen-client-py: gen-clean-py ## Build and install the authentik API for Python
 gen-dev-config:  ## Generate a local development config file
 	$(UV) run scripts/generate_config.py
 
-gen: gen-build gen-client-go gen-client-rust gen-client-ts
+gen: # gen-build
+	$(MAKE) _gen -j $(CPUS)
+
+_gen: gen-client-go gen-client-rust gen-client-ts
 
 #########################
 ## Node.js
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -239,7 +240,10 @@ gen-client-py: gen-clean-py ## Build and install the authentik API for Python
 gen-dev-config:  ## Generate a local development config file
 	$(UV) run scripts/generate_config.py
 
-gen: gen-build gen-client-go gen-client-rust gen-client-ts
+gen: # gen-build
+	$(MAKE) _gen -j $(CPUS)
```
rissson:
```suggestion
	$(MAKE) _gen -j
```

you could just do this, and it'll use all CPUs without needing to compute this here

BeryJu: technically it may use more than the number of CPUs available, but that should still be fine tbh

Makefile
```diff
@@ -239,7 +239,10 @@ gen-client-py: gen-clean-py ## Build and install the authentik API for Python
 gen-dev-config:  ## Generate a local development config file
 	$(UV) run scripts/generate_config.py
 
-gen: gen-build gen-client-go gen-client-rust gen-client-ts
+gen: gen-build
+	$(MAKE) _gen -j
+
+_gen: gen-client-go gen-client-rust gen-client-ts
```
rissson:
```suggestion
gen-clients: gen-client-go gen-client-rust gen-client-ts  ## Build and install API clients used by authentik

gen: gen-build  ## Build and install API schema and clients used by authentik
	$(MAKE) _gen -j
```

and then in the CI step checking for whether something changed, call `make -j gen-clients`

BeryJu: sure but I'd still like to have just a `gen` target that does `gen-build` and then the clients

rissson: yes I agree

rissson (approved):

GirlBossRush (approved):

PR #21174, opened 2026-03-26
core, web: Vendored client follow-ups

GirlBossRush:
- Updated packages for use with TypeScript 6
- Fix search results including symlinks.


packages/client-ts/src/runtime.ts
```diff
@@ -261,20 +261,38 @@ export class ResponseError extends Error {
     override name: "ResponseError" = "ResponseError";
     constructor(public response: Response, msg?: string) {
         super(msg);
+
+        // restore prototype chain
+        const actualProto = new.target.prototype;
+        if (Object.setPrototypeOf) {
+            Object.setPrototypeOf(this, actualProto);
+        }
```
kensternberg-authentik, line 269: These three constructors look more or less the same; you couldn't, oh, I dunno, have a parent class that does this?  :-)

rissson: this is generated code

BeryJu (approved):

PR #21198, opened 2026-03-27
fix: mask username in reputation logs to prevent accidental password logging

cyphercodes:
## Summary

When users accidentally enter their password as the username during failed login attempts, the password was being logged in plain text in the reputation system's log output.

## Changes

- Added `mask_identifier()` function that masks identifiers in logs (shows first 2 and last 1 character, masks the rest)
- Added `mask_for_log` parameter to `update_score()` function
- Failed login and identification attempts now mask the username in logs
- Successful logins remain unmasked (since the username is validated)
- Added tests for the masking functionality

## Security Impact

This prevents accidental password exposure in logs when users mistakenly enter their password in the username field.

Fixes #21138

### DIFF
--- a/authentik/policies/reputation/signals.py
+++ b/authentik/policies/reputation/signals.py
@@ -20,7 +20,18 @@ from authentik.tenants.utils import get_current_tenant
 LOGGER = get_logger()
 
 
-def update_score(request: HttpRequest, identifier: str, amount: int):
+def mask_identifier(identifier: str) -> str:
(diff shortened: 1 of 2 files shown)


BeryJu:
even with this there are still many other places where the identifier is logged and persisted. It'll be in the reputation table which can be viewed by the admin and the event log too. There isn't a way to solve this issue as we dont know whether an identifier should be censored or not.

PR #21206, opened 2026-03-27
web/a11y: Modal revisions, behavior fixes

GirlBossRush:
## Details

- Less flakey tests
- Fixes for table scroll behavior
- Fixes for visibility checks 
- Better chip group rendering, placeholders, etc
- Fix update settings form ARIA, labels, button placement
- Orange color scheme for command palette
- Safari rendering fixes (z-indexing, artifacts, font weights)
- Split dropdown buttons for application list


kensternberg-authentik (commented):
1.  The first thing in the PR, “NavigatorFixture,” is just an upgrade to the flexibility of the fixture. Nothing that depends on it now will break. It could be its own PR.

2.  The fixes for ChipGroup.ts, Chip.ts, RelatedUserList.ts (which uses those) and common.css (which adds `ak-m-plceholder` as a class) are independent of all the other stuff in the PR. Make that a separate PR, please.

3.  The changes to `intersectionObserver` (`./elements/decorators/intersection-observer`) are pretty important, but they’re only used in the patch to `Table` later. Could `intersectionObserver` be its own PR without breaking anything?

4.  The [Safari CSS fix](https://github.com/goauthentik/authentik/pull/21206/changes#diff-9da7ab69a76a8cf2d8f48e952b33c3cf5e0f39bf5d5d4071b9cc14282853972eR62) is a thing by itself. It should be its own PR.

5.  The Table.css changes could be a PR by themselves.

6.  The Wizard background removal in `./components/Wizard/wizard.css` could be its own PR.

7.  So could the “orange color scheme” stuff.

8.  The places where you use the dropdown button could be their own PR. I love this new addition, but it’s just drowned in this massive PR. Please split this into a PR of its own– the new button, and where you use it.

9.  The Form Header and Form Submit button (`./elements/forms/Form.ts`, `./elements/forms/Form.css`, and the parts of AdminSettingsForm (`./admin/admin-settings/AdminSettingsForm.ts`) are all adjustments to make ARIA work better and be more visible in the code. They could be their own PR.

After this, break out the "modal to dialog" stuff into its own PR.  I'm not sure how you'd break that up, it's huge, but aside from the changes _inside_ it, it's a mechanical fix that someone can eyeball with ripgrep, and at least it's one thing on its own.

Consider if the changes to `AKElement` (`@localized`, `hostStyles`, the big rename of `hasSlot` to `findSlot`) can be their own PR.

There's a bunch of other stuff in there that makes me go "Why is this here?" like the skeleton stuff, the change to test timeouts, and so on that I haven't even looked at carefully enough yet, but I bet they could be separated out as well. Even if you only use Skeleton in the dialogs, having that as a separate PR would be a hell of a good thing.

And for some of this, please add Storybooks so we can see what we're up against.

GirlBossRush:
Resolved in #21336

PR #21224, opened 2026-03-29
providers/saml, sources/saml: Import SP/IdP metadata to update existing providers/sources

ikob:
## Details
Feature: Import SP/IdP metadata to update an existing SAML Provider/Source.

Authentik’s SAML Provider (IdP) side previously had no good way to update an existing provider from refreshed metadata. In addition, SAML Source side does not have any interface to import IdP metadata.

This PR adds a workflow to import new SP/IdP metadata and apply it to an existing SAMLProvider/Source, updating the provider configuration accordingly.

To support real-world metadata that can include multiple certificates (e.g., rollover), this PR also introduces a Keyring model and uses it during import/apply to manage certificate sets derived from metadata.

### What's Included
- Import SP metadata and apply to an existing provider (import → update), not only import → create.
- Add Certificate KeyPair Ring model and API to represent an ordered set of certificates.
- Wire import/apply behavior so multi-cert metadata can be represented without losing candidates.

- Import IdP metadata and apply to an existing provider both import → create, and import → update.

### Notes / current limitations
- This implementation does not include “overwrite protection” / merge policy to avoid updating locally customized settings.
- Periodic pulling/refresh is out of scope here; this provides the manual update mechanism needed when metadata changes. But, this is a good start point to periodic metadata pulling such with URL.

Hopefully close a part of #20536.

---

## Checklist

-   [X] The API schema and clients have been updated (`make gen`)

If applicable


ikob:
Force-pushed a rewritten branch. Feature scope unchanged from the description above; this is base/structure/migration hygiene plus one web fix.

- Rebased onto latest main, and split into 3 commits by area: crypto (keyring) → sources/saml → providers/saml.
- Keyring moved to authentik/crypto/ (was under providers/saml/utils/) — it's shared by both source and provider import.
- Migrations linearized — one per app, direct child of each app's upstream tip, no more merge migrations; fresh installs migrate cleanly.
- Web fix: keyring binding form now loads PatternFly form styles (the Order input was rendering unstyled in Shadow DOM).
- Tested unit + end-to-end against a live Shibboleth IdP/SP (multi-cert keyring verification, IdP- and SP-side logins).

PR #21360, opened 2026-04-02
website/docs: document Grafana OAuth admin sync workaround

### LINKED ISSUE #21249: Mention single admin edge case in Grafana integration docs

So I struggled a bit with setting up Grafana OAuth with Authentik following the official integration docs. Once the flow was in place, I could authenticate with Authentik, but upon redirecting back to Grafana I was getting **User sync error**.

Grafana logs showed this:
```
 grafana  | logger=user.sync t=2026-03-29T21:24:34.648239877Z level=error msg="Failed to update user attributes" error="cannot remove last grafana admin" id= isProvisioned=false login=<my login> email= name=<my authentik name> isGrafanaAdmin=false emailVerified=null
grafana  | logger=user.sync t=2026-03-29T21:24:34.648275239Z level=error msg="Failed to update user" error="cannot remove last grafana admin" auth_module=oauth_generic_oauth auth_id=dd750379cc609871e224a531d470b56b20b445b764c23e45ef3d68fc93ed2cee
grafana  | logger=authn.service t=2026-03-29T21:24:34.648290282Z level=error msg="Failed to run post auth hook" client=auth.client.generic_oauth id= error="[user.sync.internal] unable to update user"
grafana  | logger=context userId=0 orgId=0 uname= t=2026-03-29T21:24:34.650614575Z level=info msg="Request Completed" method=GET path=/login/generic_oauth status=302 remote_addr=172.19.0.1 time_ms=1184 duration=1.184828409s size=29 referer= handler=/login/:name status_source=server 
```

I figured the cause is because of my setup:
- had a Grafana admin created
- tried to login and sync an Authentik admin with the same credentials
- Grafana may have some logic that for syncing existing users, triggers a check for removing last admin and fails

### Link

https://integrations.goauthentik.io/monitoring/grafana/

### Solution

In the Grafana OAuth settings I disabled `Allow assign Grafana admin` and changed the email address to other than in authentik. I then created the account via OAuth which worked! Subsequently, I logged back into the non-OAuth Grafana account, gave my new OAuth user admin rights and then deleted the non-OAuth account.

I realize it's a bit hacky, but maybe it's worth to add this workaround to the docs?

### Additional context

_No response_

dominic-r:
Clarify a Grafana Generic OAuth failure mode when an existing local admin account overlaps with the first OAuth login, and document where the Grafana admin assignment toggle lives.

Also fix the broken Grafana integration link in the first-steps guide.

Closes #21249

### DIFF
--- a/website/integrations/monitoring/grafana/index.mdx
+++ b/website/integrations/monitoring/grafana/index.mdx
@@ -214,3 +214,9 @@ If you get `user does not belong to org` error when trying to log into grafana f
 auto_assign_org = true
 auto_assign_org_id = <id-of-your-default-organization>
 ```
+
+If your first OAuth login fails with a `user.sync` error similar to `cannot remove last grafana admin`, check whether the OAuth user matches an existing local Grafana server admin account by login or email. Grafana can adopt the existing account into the OAuth flow, and if the mapped OAuth role is not a server admin, the login can fail while Grafana tries to remove the last remaining server admin.
+
+For OSS Grafana, the simplest workaround is to keep the existing local admin temporarily, create a separate OAuth-backed user first, grant that new user admin access, and only then remove the old local admin account.
+
+If you are configuring Generic OAuth from the Grafana UI instead of a config file, the `Allow assign Grafana admin` toggle is under `Administration > Authentication > Generic OAuth`. Its config-file equivalent is `allow_assign_grafana_admin`, or `GF_AUTH_GENERIC_OAUTH_ALLOW_ASSIGN_GRAFANA_ADMIN` in Docker-based deployments. This setting is only required if your `role_attribute_path` can return `GrafanaAdmin`; the example role mapping on this page returns `Admin`, `Editor`, or `Viewer`.
--- a/website/integrations/monitoring/grafana/index.mdx
+++ b/website/integrations/monitoring/grafana/index.mdx
@@ -215,7 +215,7 @@ auto_assign_org = true
 auto_assign_org_id = <id-of-your-default-organization>
 ```
 
-If your first OAuth login fails with a `user.sync` error similar to `cannot remove last grafana admin`, check whether the OAuth user matches an existing local Grafana server admin account by login or email. Grafana can adopt the existing account into the OAuth flow, and if the mapped OAuth role is not a server admin, the login can fail while Grafana tries to remove the last remaining server admin.
+If your first OAuth login fails with a `user.sync` error such as `cannot remove last grafana admin`, verify whether the OAuth user's username or email matches that of an existing local Grafana server admin account. In such cases, Grafana attempts to link the local account to the OAuth user. If the OAuth role mapped to that user is not a server admin, the login can fail because Grafana tries to remove the last remaining server admin account.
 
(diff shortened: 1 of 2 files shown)


website/integrations/monitoring/grafana/index.mdx
```diff
@@ -214,3 +214,9 @@ If you get `user does not belong to org` error when trying to log into grafana f
 auto_assign_org = true
 auto_assign_org_id = <id-of-your-default-organization>
 ```
+
+If your first OAuth login fails with a `user.sync` error similar to `cannot remove last grafana admin`, check whether the OAuth user matches an existing local Grafana server admin account by login or email. Grafana can adopt the existing account into the OAuth flow, and if the mapped OAuth role is not a server admin, the login can fail while Grafana tries to remove the last remaining server admin.
```
dewi-tik:
```suggestion
If your first OAuth login fails with a `user.sync` error such as `cannot remove last grafana admin`, verify whether the OAuth user's username or email matches that of an existing local Grafana server admin account. In such cases, Grafana attempts to link the local account to the OAuth user. If the OAuth role mapped to that user is not a server admin, the login can fail because Grafana tries to remove the last remaining server admin account.
```

website/integrations/monitoring/grafana/index.mdx
```diff
@@ -214,3 +214,9 @@ If you get `user does not belong to org` error when trying to log into grafana f
...
+
+For OSS Grafana, the simplest workaround is to keep the existing local admin temporarily, create a separate OAuth-backed user first, grant that new user admin access, and only then remove the old local admin account.
```
dewi-tik:
```suggestion
For OSS Grafana, the simplest workaround is to create a separate OAuth-backed user first, grant that new user admin access, and only then remove the old local admin account.
```

website/integrations/monitoring/grafana/index.mdx
```diff
@@ -214,3 +214,9 @@ If you get `user does not belong to org` error when trying to log into grafana f
...
+
+If you are configuring Generic OAuth from the Grafana UI instead of a config file, the `Allow assign Grafana admin` toggle is under `Administration > Authentication > Generic OAuth`. Its config-file equivalent is `allow_assign_grafana_admin`, or `GF_AUTH_GENERIC_OAUTH_ALLOW_ASSIGN_GRAFANA_ADMIN` in Docker-based deployments. This setting is only required if your `role_attribute_path` can return `GrafanaAdmin`; the example role mapping on this page returns `Admin`, `Editor`, or `Viewer`.
```
dewi-tik:
```suggestion
If you are configuring Generic OAuth from the Grafana UI instead of a config file, the `Allow assign Grafana admin` toggle is under **Administration** > **Authentication** > **Generic OAuth**. 

Its config-file equivalent is `allow_assign_grafana_admin`, or `GF_AUTH_GENERIC_OAUTH_ALLOW_ASSIGN_GRAFANA_ADMIN` in Docker-based deployments. This setting is only required if your `role_attribute_path` can return `GrafanaAdmin`; the example role mapping on this page returns `Admin`, `Editor`, or `Viewer`.
```

dewi-tik (approved):

website/integrations/monitoring/grafana/index.mdx
dominic-r: aint no way I used backticks for that

PR #19069, opened 2025-12-26
sources/ldap: implement nested group parentship sync

krejcar25:
## Details

Release 2025.12 brought many-to-many group inheritance. This PR builds on that, and adds true nested group synchronisation from LDAP. `LDAPSource` model fields were renamed to reflect this, and membership sync now runs through both user and group objects in LDAP reflects directory inheritance structure in Authentik.

If accepted, this PR resolves #9460.

---

## Checklist

-   [x] Local tests pass (`ak test authentik/`)
-   [x] The code has been formatted (`make lint-fix`)

If an API change has been made

-   [x] The API schema has been updated (`make gen-build`)

If changes to the frontend have been made

-   [x] The code has been formatted (`make web`)

If applicable

-   [x] The documentation has been updated
-   [x] The documentation has been formatted (`make docs`)

PR #13796, opened 2025-04-07
sources/ldap: add source connections

rissson:
## Details

Needs #13797

---

## Checklist

-   [ ] Local tests pass (`ak test authentik/`)
-   [ ] The code has been formatted (`make lint-fix`)

If an API change has been made

-   [ ] The API schema has been updated (`make gen-build`)

If changes to the frontend have been made

-   [ ] The code has been formatted (`make web`)

If applicable

-   [ ] The documentation has been updated
-   [ ] The documentation has been formatted (`make website`)

PR #21392, opened 2026-04-05
sources/ldap: Switch to new connection tracking, deprecated attribute-based connection

BeryJu:
this is a requirement for https://github.com/goauthentik/authentik/pull/19069

We've been tracking user/group-source connections with proper connection objects since https://github.com/goauthentik/authentik/commit/7fd35b1dfcf550d414da13cf205d98823c76445b (https://github.com/goauthentik/authentik/pull/13796), but we still use `ldap_uniq` in the attributes of the user/group to match users for sync.

This switches the logic around to use the connection object first, with a fallback for the old setup (this shouldn't really occur). The goal is to eventually get rid of the `LDAP_UNIQUENESS` constant.

This also contains the UI to show user/group connections and the ability to create those connections manually, mainly so we can adjust the error message to point to that new form, instead of the old attribute.

cc @krejcar25

### DIFF
--- a/authentik/sources/ldap/sync/membership.py
+++ b/authentik/sources/ldap/sync/membership.py
@@ -8,7 +8,11 @@ from ldap3 import SUBTREE
 from ldap3.utils.conv import escape_filter_chars
 
 from authentik.core.models import Group, User
-from authentik.sources.ldap.models import LDAP_DISTINGUISHED_NAME, LDAP_UNIQUENESS, LDAPSource
+from authentik.sources.ldap.models import (
+    LDAP_DISTINGUISHED_NAME,
+    GroupLDAPSourceConnection,
+    LDAPSource,
+)
 from authentik.sources.ldap.sync.base import BaseLDAPSynchronizer
 from authentik.tasks.models import Task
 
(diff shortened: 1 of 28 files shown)


web/src/admin/sources/ldap/LDAPSourceGroupList.ts
```diff
@@ -0,0 +1,87 @@
+import "#elements/forms/DeleteBulkForm";
```
rissson, line 1: I think this is something we want to do for all sources actually.

authentik/sources/ldap/sync/membership.py
```diff
@@ -104,13 +108,15 @@ def get_group(self, group_dict: dict[str, Any]) -> Group | None:
...
+            )
             if not groups.exists():
                 if self._source.sync_groups:
                     self._task.info(
                         f"Group does not exist in our DB yet, run sync_groups first: '{group_dn}'",
                         group=group_dn,
                     )
                 return None
-            self.group_cache[group_uniq] = groups.first()
+            self.group_cache[group_uniq] = groups.first().group
```
rissson, line 121:
```suggestion
            group_source_connection = GroupLDAPSourceConnection.objects.filter(identifier=group_uniq).select_related(
                "group"
            ).first()
if group_source_connection is None:
if self._source.sync_groups:
self._task.info(
f"Group does not exist in our DB yet, run sync_groups first: '{group_dn}'",
group=group_dn,
)
return None
            self.group_cache[group_uniq] = group_source_connection.group
```
saves a db query, but this is really nitpicking at this point

rissson (approved):
overall looks good, I haven't read the tests though.

web/src/admin/sources/ldap/LDAPSourceGroupList.ts
BeryJu: well, we kinda do, ish

for sources like kerberos/ldap/scim it makes sense to have here, but for OAuth/SAML I think it makes more sense on the individual users' page? or Maybe both I suppose

rissson: Yeah probably both for all sources

PR #21393, opened 2026-04-06
outposts/k8s: Add logging for kubernetes API calls

wiiaam:
## Details

Added a wrapper function to handle Kubernetes API calls and output any warnings from the headers. Also tidies up the error output for Kubernetes API errors to make them more human readable.

With field validation set to "Warn" (default):

If field validation was set to "Strict":

---

## Checklist

If applicable


authentik/outposts/controllers/k8s/base.py
```diff
@@ -29,6 +30,8 @@
 
 T = TypeVar("T", V1Pod, V1Deployment)
 
+K8S_RESPONSE_LEN = 3
```
rissson, line 33: I assume this is the length of a response that has errors?
If so, let's name it `K8S_RESPONSE_WITH_ERRORS_MIN_LENGTH` and potentially add a comment next to it.

authentik/outposts/controllers/k8s/base.py
```diff
@@ -201,6 +204,28 @@ def reconcile(self, current: T, reference: T):
...
+            kwargs.setdefault("field_validation", "Warn")
+        try:
+            result = func(*args, **kwargs)
+            if isinstance(result, tuple) and len(result) >= K8S_RESPONSE_LEN:
+                data, _, headers, *_ = result
+                if raw_warning := (headers or {}).get("Warning"):
+                    clean_msg = raw_warning.replace('\\"', '"')
+                    self.logger.warning(f"Kubernetes API Warning: {clean_msg}")
+                return data
+            self.logger.error()
```
rissson, line 219: Shouldn't something be logged here? This results in an empty log message.

rissson (changes requested):

authentik/outposts/controllers/k8s/base.py
```diff
@@ -201,6 +204,28 @@ def reconcile(self, current: T, reference: T):
...
+        if "field_validation" in inspect.signature(func).parameters:
+            kwargs.setdefault("field_validation", "Warn")
+        try:
+            result = func(*args, **kwargs)
+            if isinstance(result, tuple) and len(result) >= K8S_RESPONSE_LEN:
+                data, _, headers, *_ = result
+                if raw_warning := (headers or {}).get("Warning"):
+                    clean_msg = raw_warning.replace('\\"', '"')
+                    self.logger.warning(f"Kubernetes API Warning: {clean_msg}")
+                return data
```
BeryJu, line 218: I agree with the idea of doing this but I think this is the incorrect way of doing it. Instead of inspecting function arguments and response types, it should be a subclass of the k8s `ApiClient` class that handles this correctly.

BeryJu (changes requested):

PR #21395, opened 2026-04-06
flows: add warning message for expired password reset links

bkoragan:
Fixes #21306

## Details

When a user clicks an expired password reset link, the flow executor silently filtered out the expired token (via `ExpiringManager`) and restarted the flow with no explanation, leaving the user confused.

This PR adds a check in `_check_flow_token` to detect expired tokens using `including_expired()` and displays a clear error message: **"This link has expired. Please request a new one."** The expired token is also cleaned up upon detection.

**Changes:**
- `authentik/flows/views/executor.py`: Added `FlowTokenExpiredError` exception, expired token detection in `_check_flow_token`, and error handling in `dispatch`
- `authentik/flows/tests/test_executor.py`: Added `test_expired_flow_token` to verify the warning message and token cleanup

Closes #21306

## Checklist

-   [ x] Local tests pass (`ak test authentik/`)
-   [ x] The code has been formatted (`make lint-fix`)

-   [] The API schema and clients have been updated (`make gen`)

If applicable


bkoragan:
@BeryJu  - would you mind taking a look at this change when you have a chance and letting me know if you have any concerns.

BeryJu:
The downside of this implementation is that it discloses whether a token exists or not. The other issue being that if the token has expired and already been cleaned up from the database, this won't show a helpful error message either.

My idea for fixing this would be an extra flow authentication requirement (require_token), check for that in the flow planner and executor and if the flow has this requirement set but no token is available (either expiry or invalid), it will show an error message stating that.

bkoragan:
> The downside of this implementation is that it discloses whether a token exists or not. The other issue being that if the token has expired and already been cleaned up from the database, this won't show a helpful error message either.
> 
> My idea for fixing this would be an extra flow authentication requirement (require_token), check for that in the flow planner and executor and if the flow has this requirement set but no token is available (either expiry or invalid), it will show an error message stating that.

@BeryJu Thanks for the feedback. You're right on both points — the current implementation leaks token existence via the including_expired() lookup, and tokens that have already been purged by cleanup jobs won't show a helpful message at all.

Here's how I'm planning to implement it:

1. New FlowAuthenticationRequirement.REQUIRE_TOKEN choice alongside REQUIRE_OUTPOST, REQUIRE_REDIRECT, etc.
2. Enforcement in FlowPlanner._check_authentication: If the flow has authentication = REQUIRE_TOKEN and PLAN_CONTEXT_IS_RESTORED is not set in the context, raise FlowNonApplicableException with a generic message: "This link is invalid or has expired. Please request a new one."
This gives us the same error regardless of whether the token was never valid, already expired, or cleaned up from the DB - no information disclosure.
3. Recovery and invite flows that want this behavior would need authentication set to REQUIRE_TOKEN. Should this be the default for new recovery/enrollment flows, or left as opt-in via admin/blueprint config?

Let me know if this aligns with what you had in mind, or if there's anything you'd adjust before I start the implementation.

BeryJu:
@bkoragan for the default, I dont think we should make it default, it generally depends on the flow, some might be used directly by people from just the URL, some might only be used by admins to generate links, some might be used by both

bkoragan:
> @bkoragan for the default, I dont think we should make it default, it generally depends on the flow, some might be used directly by people from just the URL, some might only be used by admins to generate links, some might be used by both

Agreed — the default remains NONE as-is. REQUIRE_TOKEN is opt-in per flow via the admin UI or blueprints. No flow designation automatically gets it, so admins can decide based on their use case (direct URL access vs. generated links vs. both).

bkoragan:
> > @bkoragan for the default, I dont think we should make it default, it generally depends on the flow, some might be used directly by people from just the URL, some might only be used by admins to generate links, some might be used by both
> 
> Agreed — the default remains NONE as-is. REQUIRE_TOKEN is opt-in per flow via the admin UI or blueprints. No flow designation automatically gets it, so admins can decide based on their use case (direct URL access vs. generated links vs. both).

@BeryJu  I've incorporated your review comments/approach. Please review and share if any feedback. Thanks

bkoragan:
> > > @bkoragan for the default, I dont think we should make it default, it generally depends on the flow, some might be used directly by people from just the URL, some might only be used by admins to generate links, some might be used by both
> > 
> > 
> > Agreed — the default remains NONE as-is. REQUIRE_TOKEN is opt-in per flow via the admin UI or blueprints. No flow designation automatically gets it, so admins can decide based on their use case (direct URL access vs. generated links vs. both).
> 
> @BeryJu I've incorporated your review comments/approach. Please review and share if any feedback. Thanks

@BeryJu please have a review when you get a chance and add other reviewers as required. Thanks

BeryJu (approved):

PR #21420, opened 2026-04-07
packages/django-dramatiq-postgres: fix consumer crash on empty message payload

bkoragan:
Fixes #21410

## Details

After upgrading from 2025.10 to 2026.02, workers flood with `DecodeError: Failed to decode message` / `EOFError: Ran out of input`, causing a continuous consumer restart loop every 3 seconds.

**Root cause:** `_post_process_message` had changes to store `message=b""` for completed (DONE) tasks to save space. However,`requeue()` only updated the task `state` back to `QUEUED` withoutre-encoding the message payload. If a task with an emptied message re-entered a consumable state (e.g. via retry middleware, race condition, or stale pre-upgrade tasks with incompatible pickle format), `_consume_one` would call `Message.decode(b"")`, triggering the `EOFError` crash loop.

**Fix (two changes in `broker.py`):**

1. **`_consume_one`**: Guard against empty/null message payloads before decoding. The corrupt task is **deleted** (not rejected) since a task with no payload can never be executed and has no recoverable data. This follows the same pattern as `flush()` and `_auto_purge()` which already delete tasks directly. Associated `TaskLog` entries are cascade-deleted, but these only contain stale "Task has been queued" entries from a prior run..

2. **`requeue`**: Re-encode the in-memory `Message` object when updating the task state to `QUEUED`, ensuring the DB row always has a valid decodable payload. This prevents the root cause from recurring.

**Why delete instead of reject or just log?**
- Just logging would cause a log-spam loop: `_consume_one` atomically sets `state=CONSUMED` before the payload check, so `_fetch_pending_messages` would keep picking it up every iteration.
- Rejecting would work but leaves a tombstone row with no useful data. Since `_auto_purge` already deletes DONE/REJECTED tasks on a schedule, immediate deletion is cleaner and consistent with existing patterns.

Closes #21410

---

## Checklist

If applicable

### DIFF
--- a/authentik/tasks/tests/test_broker.py
+++ b/authentik/tasks/tests/test_broker.py
@@ -1,8 +1,8 @@
 """Tests for django-dramatiq-postgres broker fixes related to empty/corrupt message payloads and requeue behavior. Issue #21410"""
 
 from django.test import TestCase
-from dramatiq import Message, actor, get_broker
 from django_dramatiq_postgres.models import TaskState
+from dramatiq import Message, actor, get_broker
 
 from authentik.tasks.models import Task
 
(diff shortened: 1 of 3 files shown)


bkoragan:
seems, my commits were conflicted with other chnages , will open a fresh PR.

PR #21421, opened 2026-04-07
core: fix policy binding objects not being nullable

gergosimonyi:
#21275 part 2

Closes #21416


rissson:
You're missing a `make gen-clients`, otherwise LGTM

rissson (approved):

BeryJu (approved):

PR #21443, opened 2026-04-07
ci: avoid running setup before docker build

rissson:
## Details

REPLACE ME

---

## Checklist

If applicable

### DIFF
--- a/.github/actions/docker-push-variables/push_vars.py
+++ b/.github/actions/docker-push-variables/push_vars.py
@@ -1,7 +1,6 @@
 """Helper script to get the actual branch name, docker safe"""
 
 import os
-import re
 from json import dumps
 from pathlib import Path
 from sys import exit as sysexit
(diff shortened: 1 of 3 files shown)


.github/actions/docker-push-variables/push_vars.py
```diff
@@ -1,11 +1,21 @@
 """Helper script to get the actual branch name, docker safe"""
 
 import os
+import re
 from json import dumps
+from pathlib import Path
 from sys import exit as sysexit
 from time import time
 
-from authentik import authentik_version
```
BeryJu, line 8: surely we can still import that file without needing a regex to extract the version?

rissson: yeah, but we'd need to create __init__.py files everywhere in that tree otherwise you end up with:
```
  File "/home/risson/goauthentik/authentik/.github/actions/docker-push-variables/push_vars.py", line 10, in <module>
    from .... import authentik
ImportError: attempted relative import with no known parent package
```

BeryJu: I mean we could just find the file and eval it ig, still better than using regex to extract a variable name imo

rissson: yeah, fair enough

rissson: updated

BeryJu: or actually, maybe we should edit sys.path to include the correct dir?

anyways I'll stop nit picking on this

rissson: never mind, it's not happy for some reason

BeryJu (approved):

PR #21445, opened 2026-04-07
root: include relative time for each test case in logs

BeryJu:
in an effort to figure out where CI jobs that time out hang, this adds relative time between each test case both for start and end of test case

### DIFF
--- a/authentik/root/test_plugin.py
+++ b/authentik/root/test_plugin.py
@@ -1,9 +1,12 @@
 import math
 from os import environ
 from ssl import OPENSSL_VERSION
+from time import monotonic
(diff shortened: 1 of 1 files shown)


rissson:
I mean yes, but also, the current thing that is slow for me is `apt install` hanging on some http request to azure, sooooo

BeryJu:
> I mean yes, but also, the current thing that is slow for me is `apt install` hanging on some http request to azure, sooooo

@rissson yeah more so cases like this https://github.com/goauthentik/authentik/actions/runs/24089928668/job/70273201411?pr=20483

PR #21475, opened 2026-04-08
website/integrations: update FortiGate SSLVPN doc

leodlsrt:
## Details

This PR will add more details in the "Integrate with FortiGate SSLVPN" documentation.

"Slash" at the end in the Issuer and in the idp-entity-id is required to that's works.


dewi-tik (approved):
Thank you for the fixes @leodlsrt!

dominic-r (approved):

PR #21482, opened 2026-04-08
web/flows: prevent leader tab deadlock in continuous login flow

ryanpesek:
## Details

When the flowsContinuousLogin flag is enabled, the leader tab (the one completing login) discovers other open tabs via BroadcastChannel, tells them to continue, and waits for each to confirm exit before proceeding. Two issues could cause the leader to hang indefinitely:

1) Flag guard on responder side: multiTabOrchestrateLeave() checked flowsContinuousLogin before sending the exit broadcast. Tabs that loaded before the flag was enabled (or before a deploy with updated code) would receive the continue signal and navigate away successfully, but never send the exit confirmation—leaving the leader stuck in an infinite polling loop with localStorage lock never cleaned up. Removed the flag guard from multiTabOrchestrateLeave(); the flag check in multiTabOrchestrateResume() on the leader side is sufficient to gate the feature.

2) No timeout for unresponsive tabs: Tabs old enough to predate the flag guard fix, or those that crash/close during orchestration, would never send an exit message. Added a 3-second per-tab timeout so the leader logs a warning and moves on rather than waiting forever.

Additionally, added a pagehide event listener in Broadcast that automatically sends the exit message when a page unloads. This provides defense-in-depth for any edge case where a tab navigates away without explicitly calling multiTabOrchestrateLeave().

---

## Checklist

If applicable

### DIFF
--- a/web/src/flow/tabs/broadcast.ts
+++ b/web/src/flow/tabs/broadcast.ts
@@ -52,9 +52,14 @@ export class Broadcast extends BroadcastChannel {
         }
     };
 
+    #onPageHide = () => {
+        this.akExitTab();
+    };
+
     constructor() {
         super(BROADCAST_CHANNEL_NAME);
         this.addEventListener("message", this.#onMessage);
+        window.addEventListener("pagehide", this.#onPageHide);
         this.#logger = ConsoleLogger.prefix("mtab/broadcast");
     }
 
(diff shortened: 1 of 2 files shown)


ryanpesek:
Closing this in favor of https://github.com/goauthentik/authentik/pull/21583

PR #21507, opened 2026-04-09
web/e2e: accept options in NavigatorFixture.waitForPathname

GirlBossRush:
## Summary
- Forwards an optional `options` parameter from `NavigatorFixture.waitForPathname` to Playwright's `waitForURL`, allowing callers to configure timeout, wait-until strategy, etc.

## Test plan
- [ ] Existing e2e tests pass unchanged (the new parameter is optional with no default change)


kensternberg-authentik (approved):
See? Now that's super-easy to review!  :-)

PR #21508, opened 2026-04-09
web/elements: add viewport helpers and extend intersection observer

GirlBossRush:
## Summary
- Add `elements/utils/viewport.ts` with `findNearestBoxTarget` and `isInViewport` helpers for elements that render with `display: contents`
- Extend the `@intersectionObserver` decorator with an opt-in `useAncestorBox` mode that walks the DOM to find a parent with an actual layout box, fixing false-negative visibility for `display: contents` hosts

## Test plan
- [ ] Existing `@intersectionObserver()` call sites (Tabs, Table, ak-timestamp, FlowExecutor, LibraryPage) continue to work — the new parameter defaults to `false`
- [ ] Verify `display: contents` elements report correct intersection when `useAncestorBox` is enabled

### DIFF
--- a/web/src/elements/utils/viewport.ts
+++ b/web/src/elements/utils/viewport.ts
@@ -1,0 +1,22 @@
+export function findNearestBoxTarget(element?: Element | null): Element {
+    if (!element) {
+        return document.documentElement;
+    }
+
+    if (element.getClientRects().length) {
+        return element;
+    }
+
+    return findNearestBoxTarget(element.parentElement);
+}
+
+export function isInViewport(element: Element): boolean {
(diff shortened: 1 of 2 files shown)


web/src/elements/utils/viewport.ts
```diff
@@ -0,0 +1,22 @@
+export function findNearestBoxTarget(element?: Element | null): Element {
+    if (!element) {
```
kensternberg-authentik, line 2: Is this different from or equivalent to `element === document.documentElement`?

kensternberg-authentik (approved):
Minor question, but I assume the answer is yes, they're equivalent.  Approved!

PR #21512, opened 2026-04-09
web/elements: default @listen target to host element and add split-button Dropdown

GirlBossRush:
## Summary
- Change the `@listen` decorator default target from `window` to `this` — host-scoped listeners are the more common case for Lit components
- Add explicit `target: window` to the five existing call sites that dispatch on the window (`ak-interface-admin`, `APIDrawer`, `SidebarItem`, `FlowExecutor`, and Dropdown's own refresh listener)
- Add split-button support to the Dropdown component with `SplitButtonSelector` / `ToggleButtonSelector` statics and corresponding CSS padding variables

## Test plan
- [ ] Verify admin drawer toggle, locale status, API drawer, sidebar hash navigation, and flow session authentication listeners still fire correctly
- [ ] Verify dropdown menus open/close as before
- [ ] Test split-button dropdown rendering (if any admin pages use it)

### DIFF
--- a/web/src/elements/buttons/Dropdown.ts
+++ b/web/src/elements/buttons/Dropdown.ts
@@ -9,6 +9,9 @@ import { customElement } from "lit/decorators.js";
 
 @customElement("ak-dropdown")
 export class DropdownButton extends AKElement {
+    public static SplitButtonSelector = `.pf-c-dropdown__toggle.pf-m-split-button .pf-c-dropdown__toggle-button:last-child`;
+    public static ToggleButtonSelector = `.pf-c-dropdown__toggle:not(.pf-m-split-button)`;
+
     public static override shadowRootOptions: ShadowRootInit = {
         ...AKElement.shadowRootOptions,
         delegatesFocus: true,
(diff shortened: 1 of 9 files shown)


web/src/elements/buttons/Dropdown.ts
```diff
@@ -72,45 +74,45 @@ export class DropdownButton extends AKElement {
...
+        const menuItemButtons = this.querySelectorAll<HTMLElement>(".pf-c-dropdown__menu-item");
 
-        // for (const menuItemButton of menuItemButtons) {
-        //     menuItemButton.addEventListener("click", this.hide, {
-        //         capture: true,
-        //         signal: this.#abortController.signal,
-        //     });
-        // }
+        for (const menuItemButton of menuItemButtons) {
+            menuItemButton.addEventListener("click", this.hide, {
```
kensternberg-authentik, line 106: All right, gonna be a butthead here and ask: _why_ can't we do this as a top-level listener, get the event target and ask for an ID or some other referent? Then you could just build the click listener on construction, and not have to worry about abort controllers at all.

kensternberg-authentik (approved):
Tsk! This could still have been two PRs (@listen and Dropdown), but it's small enough.

One thing I'm not clear on is what's different between the `@listen` listener and all the individual listeners?  Also, one thing that I've learned recently is that if you have a listener on window that triggers a method in your component, that component will _never_ be garbage collected if removed from the DOM until the listener is removed, since it has a hard reference to your component.  If we need to track "clicks outside," we should probably look at how other people have done it and find a technique with weaker references.

PR #21517, opened 2026-04-09
website/docs: add another sentence to First Steps about restricting access to apps

### LINKED ISSUE #20710: Difficulty with what I would say the most basic use case

Hello there.
I'm struggling to get something to work which is in my opinion the most basic thing and I know is supported but can seem to find it anywhere.

I want to be able to have a group/roll when assigned, a user has access to one specific application. E.g. cloud group -> nextcloud application access. 

I'm sorry if I sound spoiled but even after reading through the "groups", "roles" and "access control" docs, I can't be able to find something like:

Application xy -> Policy / Group / User Bindings -> Assign Policy / Group / User to that application.

ChatGPT tells me to do this:

"Attach ONE Policy to the ONE App

Go to:

Applications → MyApp → Policies → Bind policy

Create an Expression Policy:
```python
return request.user.groups.filter(name="myapp_users").exists()
```
Bind it."

Are you telling me, I have to create a python script for the most important thing this app has to support?
I doubt it. But please help me.

It's somewhat of a rant over here because it's bothering me for months and I can't seem to figure it out and I really don't think, an application as sophisticated as this would force me to write a custom script for roles based application access. I'm either too stupid to find the right documentation page or it is so complicated that I'm not willing to put that much work into a simple access restriction (and I shouldn't really have to right?).

### Link

https://docs.goauthentik.io/users-sources/groups/, https://docs.goauthentik.io/users-sources/roles/, https://docs.goauthentik.io/users-sources/access-control/

### Solution

Show in the doc how I can allow users on a Role/Group based system to access specific applications. Not roles, not authentic settings, just applications.

### Additional context

I'm an authentik noob. But After months of going "well I need to implement this" into reading the documentation several times, I still don't have a clue, how to do (in my opinion) the simplest of tasks. If I overlooked it, please provide the link to me.

Thank you.

tanberry:
This PR adds another tip about using bindings to restrict access to an app. 
Closes #20710 

If applicable


website/docs/install-config/first-steps/index.mdx
```diff
@@ -195,6 +195,7 @@ Now that you can access the authentik Admin interface, and you have added an app
 Now that you have added your first application, and a new user, here are some typical next steps:
 
 - Assign your new user to appropriate [groups](../../users-sources/user/user_basic_operations.md#add-a-user-to-a-group) and [roles](../../users-sources/user/user_basic_operations.md#add-a-user-to-a-role).
+    - Learn more about [RBAC in authentik](../../users-sources/access-control/index.mdx), including how to [restrict access](../../add-secure-apps/applications/manage_apps.mdx#use-bindings-to-control-access) to an application. See more below about [using bindings to restrict access](#using-bindings-to-allow-or-restrict-access-to-applications).
```
dewi-tik:
```suggestion
- [Restrict access](../../add-secure-apps/applications/manage_apps.mdx#use-bindings-to-control-access) to an application. See more below about [using bindings to restrict access](#using-bindings-to-allow-or-restrict-access-to-applications).
- Learn more about [Role Based Access Control (RBAC) in authentik](../../users-sources/access-control/index.mdx). 
```

dewi-tik: I think that this looks and reads better than:
image

Also, in authentik RBAC is technically different to restricting access to applications.

BeryJu: agree, RBAC and access to apps are two quite different parts

tanberry: Yes, definitely two different things; I was trying to answer concerns of the fellow who opened the issue.. he did not know that the word "roles" was to do with role-based access.

dewi-tik (approved):

PR #21518, opened 2026-04-09
web/admin: include avatar in user list page

BeryJu:
idk why we never did this


kensternberg-authentik (approved):
Short and sweet and readable.

PR #21529, opened 2026-04-09
Enable cross-compilation from arm64 to amd64 in Dockerfile

joaocfernandes:
## Details

Add symmetric cross-compilation support to the Go builder stage in the container Dockerfile, enabling multi-platform builds (linux/amd64,linux/arm64) from both arm64 and x86_64 build hosts.

## Problem

The existing Dockerfile only supports cross-compiling Go from x86_64 → arm64. Building with --platform linux/amd64 from an arm64 host (e.g., Apple Silicon Mac) fails because:
- No amd64 cross-compilation toolchain is installed
- No CC override is set for the amd64 target when the build host is arm64


rissson:
We're actually removing cross compile. I don't have the PR on hand, I'll link it tomorrow

joaocfernandes:
I found it https://github.com/goauthentik/authentik/pull/21324 . Will follow that one

PR #21531, opened 2026-04-09
web/admin: fix user list avatar

BeryJu:
clearly should've tested the alignment in the previous PR with more setups


kensternberg-authentik (approved):
Not very rocket-sciency.

Issue #3934, opened 2022-11-03
expand user information


badsmoke:
is it possible to enter more user information in the internal directory of authentik?

e.g. 
title
position
supervisor
personnel number
address
telephone number

richardokonicha:
I'd like to work on this

dewi-tik:
#21544 Will cover this

dewi-tik:
At the moment this is achievable with custom attributes. The linked PR will give more flexibility.

PR #21544, opened 2026-04-10
core: Object attributes

BeryJu:
better UX for dealing with attributes on various objects.

Instead of just exposing a raw JSON field with YAML input, add the ability to create a "schema" of attributes which is used to generate a UI for the configured objects.

The other idea with this is to ship some standard object attributes for common attributes, which then means we can use them in property mappings by default

todos:
- [x] define and re-visit default object attributes and integrate them
- [x] fix serialization error when applying blueprint 2nd time
- [x] validation
- [ ] implement UI for array
- [ ] docs

no array support for initial version

default keys:
- given_name
- family_name
- settings.locale
- address_street
- address_region
- address_locality
- address_postal_code
- phone_number
- unix_shell
- employee_number
- employee_job_title
- employee_departmenet


dewi-tik:
Closes #3934

dewi-tik:
Closes https://github.com/goauthentik/authentik/issues/23231

GirlBossRush (approved):

web/src/elements/ak-array-input.ts
```diff
@@ -1,20 +1,19 @@
 import { AKControlElement } from "#elements/ControlElement";
```
kensternberg-authentik, line 1: There's a storybook for this component.  Does it still work?

[Edit] I can answer my own question: it looks like it works okay.

authentik/core/api/object_attributes.py
```diff
@@ -0,0 +1,94 @@
...
+        fields = [
+            "pk",
+            "object_type",
+            "object_type_obj",
+            "enabled",
+            "created",
+            "key",
+            "label",
+            "last_updated",
+            "regex",
```
kensternberg-authentik, line 74: I know we've had discussed this before when regex was added as a user-supplied check for RedirectURIs in oauth2, but for completeness I have to point it out here: a badly crafted regex is an OOM crash after a process lockup waiting to happen.

authentik/core/models.py
```diff
@@ -1392,3 +1394,61 @@ def from_request(request: HttpRequest, user: User) -> AuthenticatedSession | Non
...
+    group = models.TextField(blank=True)
+    key = models.TextField()
+
+    type = models.TextField(choices=AttributeType.choices)
+    is_unique = models.BooleanField(default=False)
+    is_required = models.BooleanField(default=False)
+    regex = models.TextField(blank=True)
+    is_array = models.BooleanField(default=False)
+
+    def run_validation(self, value: Any) -> None:
```
kensternberg-authentik, line 1471: You never check the type on validation.  Is that deliberate?  Maybe you should remind your future self (and anyone else who stumbles in here) why.

authentik/core/models.py
```diff
@@ -1392,3 +1394,61 @@ def from_request(request: HttpRequest, user: User) -> AuthenticatedSession | Non
...
+                if not all(re.fullmatch(self.regex, v) for v in value):
+                    raise ValidationError({err_key: _("Value does not match configured pattern.")})
+        else:
+            if self.is_unique:
+                model: type[models.Model] = self.object_type.model_class()
+                lookup_key = f"attributes__{self.key.replace(".", "__")}"
+                if model.objects.filter(**{lookup_key: value}).exists():
+                    raise ValidationError({err_key: _("Value is not unique.")})
+            if self.regex != "":
+                if not re.fullmatch(self.regex, value):
```
kensternberg-authentik: Okay, dumb question: do you prevent the user from supplying a regex for any type other than text?  Numeric attributes are passed back and forth as strings, but when they reach _here_, are they still strings?  (My Django is outdated, sorry). Because, uh, 
```
>>> re.fullmatch(r'\d+', "40")
<re.Match object; span=(0, 2), match='40'>
>>> re.fullmatch(r'\d+', 40)
  File "<python-input-2>", line 1, in <module>
    re.fullmatch(r'\d+', 40)
    ~~~~~~~~~~~~^^^^^^^^^^^^
  File "/Users/ken/.pyenv/versions/3.13.9/lib/python3.13/re/__init__.py", line 172, in fullmatch
    return _compile(pattern, flags).fullmatch(string)
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^
TypeError: expected string or bytes-like object, got 'int'
```

And it looks like the payload _to_ the server does encode booleans as booleans, so:
```
{
    "human_age": "0",
    "is_anthro": false,
    "settings": {
        "locale": ""
    }
}
```

A regex against `is_anthro` is gonna give you a bad time.

Since the regex doesn't grey out when you select "boolean," well... that's not good.  :-)

authentik/core/api/object_attributes.py
```diff
@@ -0,0 +1,94 @@
...
+from authentik.core.api.utils import ModelSerializer
+from authentik.core.models import AttributesMixin, ObjectAttribute
+from authentik.lib.utils.dict import get_path_from_dict
+
+
+class AttributesMixinSerializer(ModelSerializer):
+
+    def validate(self, data: dict[str, Any]) -> dict[str, Any]:
+        model = self.Meta.model
+        attrs = data.get("attributes", {})
```
kensternberg-authentik, line 18: From `SCIMSourceForm.ts`:
```
        update: (slug: string, patchedSCIMSourceRequest: SCIMSource) =>
            aki(SourcesApi).sourcesScimPartialUpdate({ slug, patchedSCIMSourceRequest }),
```

That's a PATCH.  What happens to the attributes when the patch doesn't include an attributes collection, but there's one in the database?  Does it get zeroed? Does this throw an erroneous error?

authentik/core/tests/test_object_attributes.py
```diff
@@ -0,0 +1,198 @@
...
+        )
+        res = self.client.patch(
+            reverse("authentik_api:user-detail", kwargs={"pk": self.user.pk}),
+            data={
+                "attributes": {attr.key: ["foo"]},
+            },
+        )
+        self.assertEqual(res.status_code, 400)
+        self.assertJSONEqual(
+            res.content, {f"attributes_{attr.key}": ["Value does not match configured pattern."]}
```
kensternberg-authentik, line 178: The front-end shows the keys as `name="attributes.settings.locale"` (taken from your mod to UserForm.ts); if this test shows the key names for errors as they'll be sent (underscored, rather than dotted) to the front-end, errors will not be propagated to the correct field; at best, they'll show up as "non-field errors" in the form's error handler.  (I listened to an entire Ayreon rock-opera trying to understand this, so tracking this down took me an hour and a half.)

web/src/admin/object-attributes/renderAttributes.ts
```diff
@@ -0,0 +1,166 @@
...
+ * Renders a single attribute based on its definition and the provided values.
+ *
+ * @param the current values of the attributes.
+ * @param def the definition of the attribute to render.
+ */
+function renderSingleAttribute(
+    values: Record<string, unknown>,
+    def: Pick<ObjectAttribute, "key" | "label" | "isRequired" | "type">,
+): SlottedTemplateResult {
+    const value = values[def.key] || "";
```
kensternberg-authentik: It bugs me, a little, that we only get away with `values[def.key] || ""` when the typeof resolves to "boolean" because "empty" is falsey with HTML. It feels artificial.

kensternberg-authentik (commented):
The "Group" field says it's optional, but without it the display isn't very pretty:

Image

authentik/core/models.py
BeryJu: values are converted to strings for regex compare now

authentik/core/tests/test_object_attributes.py
BeryJu: fixed, cheers

authentik/core/models.py
BeryJu: reworked the entire validation step

PR #21545, opened 2026-04-11
web/elements: P5 Drawer component with all capabilities

kensternberg-authentik:
## What

Extend ak-drawer to comply with the full specification; port ak-drawer to use Patternfly 5; vendor the Patternfly 5 subsystems directly responsible for the Drawer into the CSS.

## Why

To meet the requirements of the Drawer, of the LightDOM project, and of the Patternfly 5 vendoring port.

## Details

The Drawer’s internal CSS is now entirely within the Lit framework; the controlling CSS is namespaced to `ak-v2-c--drawer` and placed into the global CSS. Every bit of the drawer has a `part` name, so it can be customized to your heart’s content.

Added stylelint to make sure I’m doing this correctly.

-   [👽] The code has been formatted (`make web`)


web/stylelint.config.mjs
```diff
@@ -0,0 +1,21 @@
+/** @type { import("stylelint").Config } */
+export default {
+    extends: "stylelint-config-standard",
+    rules: {
+        "custom-property-pattern": [
```
kensternberg-authentik, line 5: Patternfly's names are much more varied than the standard.

GirlBossRush (approved):

PR #21548, opened 2026-04-11
fix(oauth2): replace Python 2 comma-exception syntax with Python 3 form in token endpoint

### LINKED ISSUE #21547: bug: Python 2 'except ValueError, Error:' syntax causes SyntaxError in OAuth2 token endpoint (client_credentials flow broken)
## Description

`authentik/providers/oauth2/views/token.py` line 336 contains Python 2 comma-exception syntax that is a **parse-time `SyntaxError`** in Python 3, making the entire module unimportable and breaking every OAuth2 token request.

## Root Cause

```python
# authentik/providers/oauth2/views/token.py  line 336
        try:
            user, _, password = b64decode(self.client_secret).decode("utf-8").partition(":")
            return self.__post_init_client_credentials_creds(request, user, password)
        except ValueError, Error:           # ← Python 2 syntax — SyntaxError in Python 3
            raise TokenError("invalid_grant") from None
```

`except ExcType, name:` is Python 2 syntax for binding the caught exception to a name.  Python 3 replaced it with `except ExcType as name:` and raises `SyntaxError` on the old form at parse time — before any code executes.

## Impact

The `SyntaxError` is raised when Python imports this module, so **all** OAuth2 token endpoint flows (authorization_code, refresh_token, client_credentials, device_code) fail with an import error until this is fixed.

## Reproduction

```bash
python3 -c "import authentik.providers.oauth2.views.token"
# SyntaxError: invalid syntax (token.py, line 336)
```

## Suggested Fix

Determine the intended semantics (catch `ValueError` only, or catch both `ValueError` and some `Error` class) and apply Python 3 syntax:

```python
# Option A — catch ValueError only (most likely intent, since 'Error' is not defined in the file):
except ValueError:
    raise TokenError("invalid_grant") from None

# Option B — catch ValueError and bind to a variable:
except ValueError as err:
    raise TokenError("invalid_grant") from None
```

---

**Note:** Once this SyntaxError is fixed, a follow-up hardening pass should also replace the plain `==` comparison on line 329 with `compare_digest` (already imported at line 7) to prevent timing side-channel attacks on the `client_credentials` secret, consistent with how lines 169-170 handle the other grant types.

CrepuscularIRIS:
## Summary

- `authentik/providers/oauth2/views/token.py` line 336 had `except ValueError, Error:` — Python 2 syntax that is a parse-time `SyntaxError` in Python 3, making the entire module unimportable
- Replace with `except ValueError:` (the `Error` name was not defined anywhere in the file, so `ValueError`-only is the correct intent)

## Test plan

- [ ] `python3 -c "import authentik.providers.oauth2.views.token"` — should import without error after this fix
- [ ] OAuth2 `client_credentials` grant with base64-encoded `username:password` as `client_secret` should raise `TokenError("invalid_grant")` on b64decode failure instead of crashing at import time
- [ ] All other grant types (authorization_code, refresh_token, device_code) unaffected

Fixes #21547

### DIFF
--- a/authentik/providers/oauth2/views/token.py
+++ b/authentik/providers/oauth2/views/token.py
@@ -333,7 +333,7 @@ class TokenParams:
         try:
             user, _, password = b64decode(self.client_secret).decode("utf-8").partition(":")
             return self.__post_init_client_credentials_creds(request, user, password)
-        except ValueError, Error:
+        except ValueError:
             raise TokenError("invalid_grant") from None
 
     def __post_init_client_credentials_creds(


CrepuscularIRIS:
Closing: automated submission needs further validation before resubmit.

PR #21557, opened 2026-04-13
fix: handle Attr objects in config path navigation

alfonsodg:
## Problem

When environment variables with `__` separators (e.g. `AUTHENTIK_POSTGRESQL__HOST`) are processed, they create nested config paths. If a YAML config file has already set a value at an intermediate path, that value is an `Attr` object. The path navigation functions then try to use dict operations on the `Attr` object, causing:

```
TypeError: 'Attr' object does not support item assignment
```

This affects Docker Swarm deployments where all config is passed via environment variables.

## Fix

- `dict.py`: Add `_unwrap_attr()` that extracts `.value` dict from `Attr` objects during path traversal
- `config.py`: Update `update()`, `get()`, `refresh()`, `get_keys()` to handle non-Attr results

Changes: +36 lines, -9 lines across 2 files.

## Testing

Tested on Docker Swarm 3-node cluster with env vars:
AUTHENTIK_POSTGRESQL__HOST, AUTHENTIK_POSTGRESQL__USER, AUTHENTIK_POSTGRESQL__NAME, AUTHENTIK_POSTGRESQL__PASSWORD, AUTHENTIK_REDIS__HOST, AUTHENTIK_SECRET_KEY

PostgreSQL 18 + Valkey 8.

### DIFF
--- a/authentik/lib/config.py
+++ b/authentik/lib/config.py
@@ -157,7 +157,10 @@ class ConfigLoader:
         """Recursively update dictionary"""
         for key, raw_value in updatee.items():
             if isinstance(raw_value, Mapping):
-                root[key] = self.update(root.get(key, {}), raw_value)
+                existing = root.get(key, {})
+                if isinstance(existing, Attr) and isinstance(existing.value, dict):
+                    existing = existing.value
+                root[key] = self.update(existing, raw_value)
             else:
                 if isinstance(raw_value, str):
                     value = self.parse_uri(raw_value)
@@ -172,7 +175,9 @@ class ConfigLoader:
 
     def refresh(self, key: str, default=None, sep=".") -> Any:
         """Update a single value"""
-        attr: Attr = get_path_from_dict(self.raw, key, sep=sep, default=Attr(default))
+        attr = get_path_from_dict(self.raw, key, sep=sep, default=Attr(default))
+        if not isinstance(attr, Attr):
+            return attr
         if attr.source_type != Attr.Source.URI:
             return attr.value
         attr.value = self.parse_uri(attr.source).value
@@ -257,8 +262,10 @@ class ConfigLoader:
         # Walk sub_dicts before parsing path
         root = self.raw
         # Walk each component of the path
-        attr: Attr = get_path_from_dict(root, path, sep=sep, default=Attr(default))
-        return attr.value
+        attr = get_path_from_dict(root, path, sep=sep, default=Attr(default))
+        if isinstance(attr, Attr):
+            return attr.value
+        return attr
 
     def get_int(self, path: str, default=0) -> int:
         """Wrapper for get that converts value into int"""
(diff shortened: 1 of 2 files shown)


rissson:
What's the scenario where this happens?

alfonsodg:
## Scenario

Docker Swarm passes all configuration via environment variables with `__` as path separator. For example:

```
AUTHENTIK_POSTGRESQL__HOST=authentik-db
AUTHENTIK_POSTGRESQL__USER=authentik
AUTHENTIK_POSTGRESQL__NAME=authentik
AUTHENTIK_POSTGRESQL__PASSWORD=secret
AUTHENTIK_REDIS__HOST=authentik-valkey
AUTHENTIK_SECRET_KEY=my-secret-key
```

Unlike Docker Compose or Kubernetes, Swarm does not support mounting config files into services easily — environment variables are the standard way to configure services.

### What happens

1. `authentik/lib/config.py` loads the default YAML config (`default.yml`), which creates nested `Attr` objects for paths like `postgresql.host`
2. Then `refresh()` processes environment variables and calls `set_path_in_dict()` to set `postgresql.host = authentik-db`
3. `set_path_in_dict()` walks the path: at `postgresql`, it finds an `Attr` object (from the YAML load), not a plain dict
4. It tries `root["host"] = value` on the `Attr` object → `TypeError: 'Attr' object does not support item assignment`

### Reproduction

```bash
# Docker Swarm (no compose file, pure env vars)
docker service create --name authentik-server \
  -e AUTHENTIK_POSTGRESQL__HOST=db \
  -e AUTHENTIK_POSTGRESQL__PASSWORD=secret \
  -e AUTHENTIK_SECRET_KEY=key \
  ghcr.io/goauthentik/server:2024.8.3
```

The server crashes on startup with the TypeError.

### Why this doesn't happen in Docker Compose

In Docker Compose, you typically mount `config.yml` which sets all values as plain dicts. The env vars only override leaf values, so the path traversal never encounters an `Attr` at an intermediate node. In Swarm, there's no config file — everything comes from env vars, and the default YAML creates `Attr` objects at intermediate paths.

rissson:
I feel like I'm talking with an LLM so why doesn't this happen with setting `AUTHENTIK_POSTGRESQL__PASSWORD=file:///secrets/postgres-password`?

alfonsodg:
Ha, fair enough on the LLM comment — I was trying to be thorough but I see how it reads that
way.

To your question: using file:/// doesn't help here because the crash happens before the 
value is even parsed. The TypeError is in set_path_in_dict when it walks the path 
postgresql.host — at the postgresql node it finds an Attr (from default.yml) and tries to do
root["host"] = value on it. It blows up there regardless of whether the value is "db" or 
"file:///some/path".

You can reproduce it with just AUTHENTIK_POSTGRESQL__HOST=db — no secrets involved. The 
issue is that Attr wraps the default dict but doesn't support item assignment, so any env 
var with __ that targets a path already defined in default.yml will crash.

rissson:
The thing is I have seen many multiple setup where configuration is done through both the default.yml file, another .yml file, and then overrides with environment variables on top, and I've never encountered this.

So without a minimal reproduction, this won't go anywhere.

PR #21572, opened 2026-04-13
High DB Load. Server and worker print loop "Found static bindings"

joaocfernandes:
## Details

During startup, the _reconcile() method in ManagedAppConfig accidentally evaluates @property descriptors by calling getattr() on every attribute. This triggers tenant_schedule_specs, which accesses schedule_specs on each outgoing sync provider (SCIM).

Each provider's schedule_specs calls get_sync_time_limit_ms(), which runs PolicyEngine for every user in the database just to estimate a timeout. With N providers and M users, this results in N × M × 2 database queries during startup.

Fix _reconcile() to check attributes on the class instead of the instance, avoiding accidental property evaluation. Fix get_object_sync_time_limit_ms() to use a simple COUNT(*) query instead of materializing the full policy-filtered queryset.

---

## Checklist

-   [] Local tests pass (`ak test authentik/`)

If applicable

### DIFF
--- a/authentik/blueprints/apps.py
+++ b/authentik/blueprints/apps.py
@@ -3,7 +3,6 @@
 import traceback
 from collections.abc import Callable
 from importlib import import_module
-from inspect import ismethod
 
 from django.apps import AppConfig
 from django.conf import settings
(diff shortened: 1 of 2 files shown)


rissson:
Closing in favour or https://github.com/goauthentik/authentik/pull/21575 and https://github.com/goauthentik/authentik/pull/21576

Thanks for the contribution!

PR #21577, opened 2026-04-13
packages/django-dramatiq-postgres: reset db connections in raise_connection_error

joaocfernandes:
## Details

When a task finishes running, the worker tries to mark it as done in the database (marking it as done or failed). If the database connection drops at that exact moment, the worker retries every 5 seconds. But the retry never works because the broken connection is never cleaned up between attempts. The worker keeps trying the same dead connection forever.
This leaves the worker process permanently stuck. The only way to recover is to restart the pod.

---

## Checklist

If applicable

### DIFF
--- a/packages/django-dramatiq-postgres/django_dramatiq_postgres/broker.py
+++ b/packages/django-dramatiq-postgres/django_dramatiq_postgres/broker.py
@@ -62,6 +62,10 @@ def raise_connection_error(func: Callable[P, R]) -> Callable[P, R]:  # noqa: UP0
             return func(*args, **kwargs)
         except DATABASE_ERRORS as exc:
             logger.warning("Database error encountered", exc=exc)
+            try:
+                connections.close_all()
+            except DATABASE_ERRORS:
+                pass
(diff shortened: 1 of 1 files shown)


rissson:
I've actually been thinking of removing this retry completely, which might cause tasks to be re-run but I think I can live with that. Thoughts?

joaocfernandes:
> I've actually been thinking of removing this retry completely, which might cause tasks to be re-run but I think I can live with that. Thoughts?

Agree. I never thought about this but if tasks are generally idempotent removing the retry is better. A task running twice is much less harmful than a worker stuck.

rissson:
We are agreed then, let's remove tenacity. I'll open a PR for that.

rissson:
Ah actually I didn't read any of this correctly.

The tenacity retry is done only for enqueuing tasks.

The retry for mark database connections as completed is done but dramatiq directly, and yeah I think this change will solve your issue. Do you mind targeting `main` for this commit?

rissson (approved):

PR #21586, opened 2026-04-13
web/flows: fix deadlock on continuous flow (2025.12)

BeryJu:
see https://github.com/goauthentik/authentik/pull/21482


GirlBossRush:
Closing in favor of #21583 21583

PR #21602, opened 2026-04-14
web: Tidy continuous login orchestrator

GirlBossRush:
## Details

Small fixes to the orchestrator:

- Replaced `setInterval` polling with `requestAnimationFrame` for tab-exit checks.
- Declare that `Broadcast` implements `Disposable`; switch private members to protected for subclass access. A gradual pattern we've adopted across the codebase
- Listener naming conventions


GirlBossRush:
Closing in favor of #21583

PR #21622, opened 2026-04-15
web: Close modal on route navigation

GirlBossRush:
## Details

This PR fixes a small issue where modals do not close after the router navigates to another page, such as when clicking an enterprise only link within a modal.


kensternberg-authentik (approved):
Elegant and excellent.

PR #21623, opened 2026-04-15
web: Fix table visibility checks, search params.

GirlBossRush:
## Details

This PR fixes a regression introduced in the modals clean up. Interactivity checks now handle tables outside of modals, allowing search params to synchronizing to the URL.


kensternberg-authentik (approved):
Nice re-org.

PR #21624, opened 2026-04-15
core: update refresh_groupancestry trigger for supporting logical replication

nalves599:
## Details

When we want to have logical replication on the PostgreSQL database, the trigger must run on the primary instance and on the replica. This PR updates the trigger
---

## Checklist

If applicable

### DIFF
--- a/authentik/core/models.py
+++ b/authentik/core/models.py
@@ -278,7 +278,7 @@ class GroupParentageNode(models.Model):
                 operation=pgtrigger.Insert | pgtrigger.Update | pgtrigger.Delete,
                 when=pgtrigger.After,
                 func="""
-                    REFRESH MATERIALIZED VIEW CONCURRENTLY authentik_core_groupancestry;
+                    REFRESH MATERIALIZED VIEW CONCURRENTLY public.authentik_core_groupancestry;
                     RETURN NULL;
                 """,
             ),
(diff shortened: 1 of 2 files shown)


authentik/core/models.py
```diff
@@ -278,7 +278,7 @@ class Meta:
                 operation=pgtrigger.Insert | pgtrigger.Update | pgtrigger.Delete,
                 when=pgtrigger.After,
                 func="""
-                    REFRESH MATERIALIZED VIEW CONCURRENTLY authentik_core_groupancestry;
+                    REFRESH MATERIALIZED VIEW CONCURRENTLY public.authentik_core_groupancestry;
```
rissson, line 281: This issue with not relying on the search path is that we don't know which schema is being used when creating the migration above, as that's determined at runtime with this option https://docs.goauthentik.io/install-config/configuration/#advanced-settings `AUTHENTIK_POSTGRESQL__DEFAULT_SCHEMA`

nalves599: Is there a way of getting the schema like we do for the `get_tenant_base_schema`?

rissson: We do have a helper function for that, but that won't work, since the trigger creates a PostgreSQL function on migration, that won't be updated with the schema changes.

PR #21625, opened 2026-04-15
packages/ak-common/tracing: get sentry config from API for outposts

rissson:
## Details

REPLACE ME

---

## Checklist

If applicable

### DIFF
--- a/packages/ak-common/src/api.rs
+++ b/packages/ak-common/src/api.rs
@@ -2,17 +2,18 @@
 
 use ak_client::apis::configuration::Configuration;
 use eyre::{Result, eyre};
+use url::Url;
 
 use crate::{config, user_agent_outpost};
 
-struct ServerConfig {
-    host: String,
-    token: String,
-    insecure: bool,
+pub struct ServerConfig {
+    pub host: Url,
+    pub token: String,
+    pub insecure: bool,
 }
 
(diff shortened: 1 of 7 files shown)


packages/ak-common/src/api.rs
```diff
@@ -0,0 +1,39 @@
...
+        .token
+        .clone()
+        .ok_or_else(|| eyre!("environment variable `AUTHENTIK_TOKEN` not set"))?;
+    let ak_insecure = config::get().insecure.unwrap_or(false);
+
+    let base_path = if ak_host.ends_with('/') {
+        format!("{ak_host}api/v3")
+    } else {
+        format!("{ak_host}/api/v3")
+    };
```
BeryJu: I'd prefer if we used a real URL parsing/manipulation instead of string-based

rissson: I actually do exactly that right after with host.parse(), which will automatically set the path to `/` if there are not trailing `/`, so yeah, fixed

rissson: actually I'll improve a bit more and do the same for the base url for the client, standby

rissson: and I've added tests :tada:

BeryJu (approved):

PR #21649, opened 2026-04-16
website/docs: clarify LDAP group attribute mappings

### LINKED ISSUE #5874: how to map ldap group's custom attribute to Authentik's group "attribute"
**Describe your question**
How to map a custom attribute set on ldap group can be mapped to Authentik group's attributes property.

Steps I performed
1> In JumpCloud ldap service, I added a custom property "acl" to the group to which users belong. This custom property has json data
2> Configured external LDAP service provider (JumpCloud) using Authentik's **Federation & Social login** option.
3> While configuring LDAP source in **Federation & Social login** "**LDAP Attribute mapping**" for group I selected option "authentik default OpenLDAP Mapping:cn"
4> When sync runs, I can see that LDAP group and users from JumpClound are copied into Authentik under Directory \ groups and Directory \ users.
5> When I check the **attributes** property of specific ldap group copied in Authentik, the value I get is
ldap_uniq: ldap
distinguishedName: cn=ldap,ou=Users,o=junkdata,dc=jumpcloud,dc=com

### What I want
the **attributes** property of Authentik group created by copying JumpCloud Ldap group, should contain values of custom property "acl" of Ldap group

### What I did
To achieve this I tried to create **Property Mapping** of type "**LDAP Property Mapping**" but,
 I don't understand what to put in field "**Object field** ", which has comment as "Field of the **user object** this value is written to", Does this mean that this property mapping only work for user not group ?
Second thing what expression should I use to extract the value of custom attribute "acl" of ldap groups and return with this mapping.

**Relevant infos**
 JumpCloud Ldap service

-   authentik version: [Version: 2023.5]
-   Deployment: [helm]

dominic-r:
Explain that LDAP source property mappings can be assigned to groups, add an example for copying a custom LDAP group attribute into authentik group attributes, and note how to decode JSON-encoded values.

Closes: https://github.com/goauthentik/authentik/issues/5874

### DIFF
--- a/website/docs/users-sources/sources/protocols/ldap/index.md
+++ b/website/docs/users-sources/sources/protocols/ldap/index.md
@@ -77,7 +77,7 @@ The same LDAP source property mapping type is used for both users and groups. A 
 
 ### Copy a custom LDAP group attribute
 
-To store a custom LDAP group field in authentik's group `attributes`, create an **LDAP Source Property Mapping** and assign it to **Group Property Mappings** on the source:
+To store a custom LDAP group attribute in authentik's group `attributes`, create an **LDAP Source Property Mapping** and assign it to **Group Property Mappings** on the source:
 
 ```python
 return {
--- a/website/docs/users-sources/sources/protocols/ldap/index.md
+++ b/website/docs/users-sources/sources/protocols/ldap/index.md
@@ -73,6 +73,33 @@ return {
 }
 ```
 
+The same LDAP source property mapping type is used for both users and groups. A mapping only applies to groups when you assign it under **Group Property Mappings** on the LDAP source. If you only use the built-in group naming mappings, synced groups will keep the automatically populated LDAP identifiers, such as `distinguishedName`, but custom LDAP attributes won't be copied unless you add your own group mapping.
+
+### Copy a custom LDAP group attribute
+
+To store a custom LDAP group field in authentik's group `attributes`, create an **LDAP Source Property Mapping** and assign it to **Group Property Mappings** on the source:
+
+```python
+return {
+    "attributes": {
+        "acl": list_flatten(ldap.get("acl")),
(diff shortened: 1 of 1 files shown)


website/docs/users-sources/sources/protocols/ldap/index.md
```diff
@@ -73,6 +73,33 @@ return {
 }
 ```
 
+The same LDAP source property mapping type is used for both users and groups. A mapping only applies to groups when you assign it under **Group Property Mappings** on the LDAP source. If you only use the built-in group naming mappings, synced groups will keep the automatically populated LDAP identifiers, such as `distinguishedName`, but custom LDAP attributes won't be copied unless you add your own group mapping.
```
dewi-tik, line 76:
```suggestion
The same LDAP source property mapping type is used for both users and groups. A mapping only applies to groups when you assign it under **Group Property Mappings** on the LDAP source. If you only use the built-in group property mappings, synced groups will keep the automatically populated LDAP attributes, such as `distinguishedName`, but custom LDAP attributes won't be copied unless you add your own group mapping.
```

website/docs/users-sources/sources/protocols/ldap/index.md
```diff
@@ -73,6 +73,33 @@ return {
...
+
+### Copy a custom LDAP group attribute
+
+To store a custom LDAP group field in authentik's group `attributes`, create an **LDAP Source Property Mapping** and assign it to **Group Property Mappings** on the source:
```
dewi-tik, line 80:
```suggestion
To store a custom LDAP group attribute in authentik's group `attributes`, create an **LDAP Source Property Mapping** and assign it to **Group Property Mappings** on the source:
```

dewi-tik (approved):
Small changes otherwise LGTM

PR #21690, opened 2026-04-18
core: actually omit users from the groups API response when include_users=false

SAY-5:
## What

`GroupSerializer` in `authentik/core/api/groups.py` gates the `users_obj` SerializerMethodField on `_should_include_users` correctly (returns `None` when `?include_users=false`). The sibling `users` field is **not** a SerializerMethodField, it's the implicit `PrimaryKeyRelatedField(many=True)` that DRF derives from the Meta.fields list, and nothing turned it off. So every list response still serialises the group-users M2M as an array of PKs.

Per #21278, on tenants with large groups this makes:

```
GET /api/v3/core/groups/?include_users=false
```

indistinguishable from `include_users=true` in wall-clock time, which is the opposite of what the caller asked for and what the OpenAPI parameter doc implies. The viewset already drops the `users` prefetch when `_should_include_users` is false (see the `prefetch_related("users", ...)` guard), so each group now takes a *cold* round-trip to walk the M2M. That makes the N+1 shape in this mode arguably worse than with `include_users=true`.

## Fix

Override `GroupSerializer.get_fields` to drop the `users` binding when the include flag is off:

```python
def get_fields(self):
    fields = super().get_fields()
    if not self._should_include_users:
        fields.pop("users", None)
    return fields
```

After this:

- `GET /core/groups/?include_users=false` returns neither `users` nor `users_obj`. The M2M is never walked.
- `GET /core/groups/` (default) still returns both.
- Write paths (POST `/core/groups/`, PATCH `/core/groups/{uuid}/`) do not set `include_users` in their query string, so `_should_include_users` stays at its `True` default and `users` remains a writable field. Create/update semantics for assigning users are unaffected.

## Test plan

- Baseline on main: `GET /core/groups/?include_users=false&page_size=20` on a tenant with large groups. Response body still contains `users: [pk, pk, ...]` per group; p95 latency unchanged vs `include_users=true`.
- Apply patch. Same request returns group objects with no `users` or `users_obj` keys. p95 drops back to the "no user walks" curve.
- `GET /core/groups/` (no query param) unchanged: both fields present.
- `POST /core/groups/ {"name": "...", "users": [1, 2]}` still assigns users correctly (test_groups_api covers this).

Fixes #21278


SAY-5:
Closing this, #23216 landed and covers the N+1 without changing the shape of the users/parents/children lists, which is the behaviour you wanted preserved.

PR #21691, opened 2026-04-18
stages/user_login: refuse to log in an unsaved pending user

SAY-5:
## What

`UserLoginStageView.do_login` reads `PLAN_CONTEXT_PENDING_USER` and hands the resulting `User` to `is_known_device()` and Django's `login()`:

```python
user: User = self.executor.plan.context[PLAN_CONTEXT_PENDING_USER]
if not user.is_active:
    ...
self.executor.plan.context[PLAN_CONTEXT_METHOD_ARGS].setdefault(
    PLAN_CONTEXT_METHOD_ARGS_KNOWN_DEVICE, self.is_known_device(user)
)
with audit_ignore():
    login(self.request, user, backend=backend)
```

The identification stage plants an *unsaved* `User` (`pk is None`) in `PLAN_CONTEXT_PENDING_USER` when the supplied identifier is unknown, it exists only so that `make_password()` runs against something and the timing side-channel that would otherwise leak "user exists vs doesn't" goes away (#18883). That dummy user was previously scoped to the identification stage, but the stage-state refactor (#18641) means it now propagates down the flow plan into `do_login`.

`is_known_device` immediately does:

```python
AuthenticatedSession.objects.filter(session__last_ip=client_ip, user=user).exists()
```

and `login()` fires the `user_logged_in` signal, which `user_logged_in_session` and `on_user_logged_in` handle with further ORM lookups. All of those are related-field filters, and Django rejects unsaved instances with:

```
ValueError: Model instances passed to related filters must be saved.
```

Observable effect: on a default authentication flow that has an enrollment flow linked, typing an unknown email into the identification stage returns HTTP 500 instead of redirecting to enrollment or showing the usual invalid-login message. Per #21036.

## Fix

Guard at the top of `do_login` before any user-dependent work runs:

```python
user: User = self.executor.plan.context[PLAN_CONTEXT_PENDING_USER]
if user.pk is None:
    self.logger.debug("pending user is not saved, refusing to log in")
    return self.executor.stage_invalid()
if not user.is_active:
    ...
```

Any code path that produced an unsaved `PENDING_USER` now ends in `stage_invalid`, never reaches `is_known_device`, `login()`, or the signal chain. The existing `is_active` check stays: an unsaved `User()` defaults to `is_active=True`, so the pk guard has to run first.

## Why here rather than in the signal handlers

The signal handlers (`user_logged_in_session`, `on_user_logged_in`) and `is_known_device` are three separate call sites that would each need a pk guard, and the bug would come back the next time someone added a fourth caller. Gating at `do_login` is the single chokepoint for "we are about to log a user in" and matches the existing `is_active` check placement.

Fixes #21036


authentik/stages/user_login/stage.py
```diff
@@ -151,6 +151,17 @@ def do_login(self, request: HttpRequest, remember: bool | None = None) -> HttpRe
...
         )
         user: User = self.executor.plan.context[PLAN_CONTEXT_PENDING_USER]
+        # The identification stage plants an unsaved User (pk is None) in
+        # PLAN_CONTEXT_PENDING_USER to normalise timing against make_password()
+        # when the supplied identifier is unknown. Previously that dummy user
+        # reached is_known_device (which does User-filtered ORM lookups) and
+        # the user_logged_in signal handlers, both of which rely on a saved
+        # instance and fail with:
+        #     ValueError: Model instances passed to related filters must be saved.
+        # Catch the unsaved user here and surface stage_invalid instead of a 500.
```
BeryJu:
```suggestion
        # This should realistically never happen, as even though the identification stage returns a fake user, any kind of authentication
        # would fail due to the user not existing
        # and as such this is very much just a fallback for a misconfiguration
        # to prevent an exception
```

BeryJu (approved):

PR #21697, opened 2026-04-19
web/admin: Allow binding users/groups in policy binding wizard and existing stage in stage binding wizard

BeryJu:
Currently we have two primary options on binding lists (policybinding and stagebinding)
The admin has to choose before hand if they want to bind an existing object, but when binding a new object gets the wizard anyways to make more choices.
This adds options to the wizard to allow binding existing objects.

Current:

https://github.com/user-attachments/assets/717d51db-1f0a-4d68-b2a2-45ec8d258ed4

With this PR:

https://github.com/user-attachments/assets/54282874-295f-4e97-887f-142bbc6853f7

Labelling probably needs to be revised


GirlBossRush (approved):
Looks great ✨👏🏻  

Added some small fixes to make wizards look and behave more consistently, navigation fixes, and clean up to make #18615 easier to merge.

PR #21702, opened 2026-04-20
web/flows: localize OR_LIST_FORMATTERS so the field-label connector follows the active locale

### LINKED ISSUE #20632: Identification stage: OR_LIST_FORMATTERS uses hardcoded "default" locale instead of active locale
## Describe the bug

On the identification stage, when `user_fields` contains multiple fields (e.g. `[email, username]`), the field label is generated by joining the translated field names with a disjunction word ("or", "ou", "oder", etc.).

The individual field names are correctly translated via `msg()` (e.g. "Courriel", "Nom d'utilisateur" in French), but the **"or" connector word stays in English** regardless of the selected locale.

### Root cause

In [`web/src/flow/stages/identification/IdentificationStage.ts`](https://github.com/goauthentik/authentik/blob/main/web/src/flow/stages/identification/IdentificationStage.ts), the `OR_LIST_FORMATTERS` is defined with a hardcoded `"default"` locale:

```ts
export const OR_LIST_FORMATTERS: Intl.ListFormat = new Intl.ListFormat(
    "default",
    { style: "short", type: "disjunction" }
);
```

This causes `Intl.ListFormat` to use the **browser's default locale** for the disjunction word, not authentik's currently active locale (set via the language switcher or `?locale=` URL parameter).

### Expected behavior

The `Intl.ListFormat` should use the currently active authentik locale, so the connector word is properly localized:
- French: "Courriel **ou** Nom d'utilisateur"
- German: "E-Mail **oder** Anmeldename"
- English: "Email **or** Username"

### Actual behavior

The connector word "or" remains in English (or whatever the browser's default locale is), while the field names themselves are correctly translated:
- French: "Courriel **or** Nom d'utilisateur"
- German: "E-Mail **or** Anmeldename"

### Suggested fix

Pass the active locale to `Intl.ListFormat` instead of `"default"`. For example:

```ts
const locale = getCurrentLocale(); // or however the active locale is accessed
const OR_LIST_FORMATTERS = new Intl.ListFormat(locale, { style: "short", type: "disjunction" });
```

Or make it a function that creates the formatter with the current locale at render time, since the locale can change dynamically via the language switcher.

## To Reproduce

1. Set up an identification stage with `user_fields: [email, username]`
2. Navigate to the login flow with `?locale=fr` (or `?locale=de`, or use the language switcher)
3. Observe the field label shows "Courriel **or** Nom d'utilisateur" — the "or" is not translated

## Version and Deployment

**authentik version:** 2026.2.0
**Deployment:** Docker Compose

## Related issues

- #16957 — reported the same symptom but root cause was not identified

Image
Image
Image

SAY-5:
## What

`IdentificationStage.ts` joins the translated `user_fields` with `Intl.ListFormat` when more than one field is configured. The formatter was hardcoded to the `"default"` locale:

```ts
export const OR_LIST_FORMATTERS: Intl.ListFormat = new Intl.ListFormat("default", {
    style: "short",
    type: "disjunction",
});
```

`"default"` binds `Intl.ListFormat` to whatever the *browser*'s default locale happens to be, which is not necessarily the authentik locale set via the language switcher or `?locale=` URL parameter. The individual field names pass through `msg()` and localize correctly, but the "or" connector stays in the browser locale. Per #20632 this produces labels like `Courriel or Nom d'utilisateur` in French or `E-Mail or Anmeldename` in German.

## Fix

Turn `OR_LIST_FORMATTERS` into a factory that reads `@lit/localize`'s `getLocale()` at call time:

```ts
import { getLocale, msg, str } from "@lit/localize";

export const orListFormatter = (): Intl.ListFormat =>
    new Intl.ListFormat(getLocale(), {
        style: "short",
        type: "disjunction",
    });

// usage
const label = orListFormatter().format(fields.map((f) => UI_FIELDS[f]));
```

Each render picks up the current locale. `Intl.ListFormat` construction is O(1) and the factory runs once per render path, so there's no meaningful perf cost. As a side benefit, this also means the label updates correctly when the user toggles languages without a full reload (the module-level constant never updated).

## Test plan

- Pre-patch, set locale to French (`?locale=fr`). Configure an identification stage with `user_fields: [email, username]`. Label reads `Courriel or Nom d'utilisateur`.
- Post-patch, same configuration: label reads `Courriel ou Nom d'utilisateur`.
- German (`?locale=de`): `E-Mail oder Anmeldename`.
- English: `Email or Username` (unchanged).
- Switching the locale dropdown without a reload now updates the connector on the next render.

Fixes #20632


web/src/flow/stages/identification/IdentificationStage.ts
```diff
@@ -52,10 +52,17 @@ export const PasswordManagerPrefill: {
...
-export const OR_LIST_FORMATTERS: Intl.ListFormat = new Intl.ListFormat("default", {
-    style: "short",
-    type: "disjunction",
-});
+// Build a disjunction-style list formatter bound to whatever locale is
+// active at call time, so `Email or Username` gets localized to
+// `Courriel ou Nom d'utilisateur` / `E-Mail oder Anmeldename` instead of
+// leaking the hardcoded English connector when the user-facing label
+// renders. Module-level `new Intl.ListFormat("default", ...)` would bind
+// once to the browser's default locale instead.
```
BeryJu:
```suggestion
```

web/src/flow/stages/identification/IdentificationStage.ts
```diff
@@ -52,10 +52,11 @@ export const PasswordManagerPrefill: {
...
-    type: "disjunction",
-});
+export const orListFormatter = (): Intl.ListFormat =>
+    new Intl.ListFormat(getLocale(), {
+        style: "short",
+        type: "disjunction",
+    });
 
 const UI_FIELDS: { [key: string]: string } = {
     [UserFieldsEnum.Username]: msg("Username"),
```
kensternberg-authentik, line 68: Did anyone else notice that these, too, are initialized at startup?  Are they initialized _before_ or _after_ we've figured out the locale?  (I feel bad about this; I wrote this particular code, but if we're gonna get serious about fixing our i18n problem, this may not be correct.)

web/src/flow/stages/identification/IdentificationStage.ts
```diff
@@ -298,7 +299,7 @@ export class IdentificationStage extends BaseStage<
 
         const offerRecovery = flowDesignation === FlowDesignationEnum.Recovery;
         const type = fields.length === 1 && fields[0] === UserFieldsEnum.Email ? "email" : "text";
-        const label = OR_LIST_FORMATTERS.format(fields.map((f) => UI_FIELDS[f]));
+        const label = orListFormatter().format(fields.map((f) => UI_FIELDS[f]));
```
kensternberg-authentik, line 308: Also, I feel like this is a function that's nicely isolated from the user experience.  It _contributes_, but it isn't rendering any HTML to the screen.  Where are the tests?

kensternberg-authentik (commented):
A request for tests, and a question about initialization order.

web/src/flow/stages/identification/IdentificationStage.ts
BeryJu: I would assume the `msg()` call is a lazy proxy which only gets resolved during rendering

PR #21729, opened 2026-04-20
stages/password: count failed attempts directly instead of deriving from reputation

### LINKED ISSUE #21600: Password Stage Failed Attempts Before Cancel Only Works once
### Describe the bug

In my auth flow, I have "Failed attempts before cancel" set to 5 in my Password Stage.  If I then fail the password 5 times, the flow successfully ends as expected.

However, if I go back to the login page, and try again with either the same or a different user, I can now try the password stage more than 5 times (seems to be indefinite).

Running Authentik 2026.2.1 with Docker.

### How to reproduce

Try logging in 5 times using bad password. 
After Flow fails, click "GO Back" or "Not You"
Try again - can attempt password more than 5 times.

### Expected behavior

Flow always ends after 5 password failures during stage.

### Screenshots

_No response_

### Additional context

_No response_

### Deployment Method

Docker

### Version

2026.2.1

### Relevant log output

```shell

```

SAY-5:
Fixes #21600.

The cancel-after-N-failures check in `PasswordStageView.challenge_invalid` compared reputation scores:

```python
if (initial_score - new_score) >= current_stage.failed_attempts_before_cancel:
```

That works on the first flow run but breaks on every subsequent one. `Reputation.update_score` clips the resulting score to `tenant.reputation_lower_limit`, so once a user has hit the floor on the first lockout, further `login_failed` signals leave the score unchanged. On the next flow plan `initial_score` captures the floor value, `new_score` stays at the floor on every failure, the diff is always 0, the guard never fires, and the user can retry indefinitely.

This tracks a per-flow-plan counter in the plan context instead:

```python
attempts = self.executor.plan.context.get(PLAN_CONTEXT_FAILED_ATTEMPTS, 0) + 1
self.executor.plan.context[PLAN_CONTEXT_FAILED_ATTEMPTS] = attempts
if attempts >= current_stage.failed_attempts_before_cancel:
    return self.executor.stage_invalid(_("Invalid password"))
```

Each new flow plan (including the one created after *Not you?* / *Go back*) starts from zero, counts exactly the attempts it served, and cancels at the configured threshold regardless of the user's reputation history. `PLAN_CONTEXT_INITIAL_SCORE` is left in place so `get_challenge` continues to capture it for observability (no other code paths rely on the diff).


SAY-5:
Good point, done. The plan now counts attempts directly but the allowance is min(initial score minus the tenant lower limit, failed_attempts_before_cancel), so someone already sitting at the reputation floor gets cancelled on their first try in a new flow instead of getting a fresh 5. Added a test for that case.

PR #20983, opened 2026-03-18
root: init rust workspace

rissson:
---

## Checklist

-   [ ] Local tests pass (`ak test authentik/`)
-   [ ] The code has been formatted (`make lint-fix`)

If an API change has been made

-   [ ] The API schema has been updated (`make gen-build`)

If changes to the frontend have been made

-   [ ] The code has been formatted (`make web`)

If applicable

-   [ ] The documentation has been updated
-   [ ] The documentation has been formatted (`make docs`)

PR #21732, opened 2026-04-21
root: add Rust tools to make install

GirlBossRush:
This PR adds a `rust-install` make target that runs `cargo install` for all four tools and hooks it into the existing `install` target alongside `node-install`, `docs-install`, and `core-install`. The prerequisite docs are updated to name the tools and point to `make install` so the path from "I just cloned the repo" to a passing `make` is unambiguous.

## Details

PR #20983 (\"root: init rust workspace\", March 2026) introduced four required Cargo tools — `cargo-deny`, `cargo-machete`, `cargo-llvm-cov`, and `cargo-nextest` — and wired them into `make lint`. CI installs them via `taiki-e/install-action`, but no local equivalent was added: the prerequisite bullet in `full-dev-environment.mdx` only mentioned Rust itself, leaving `make` silently broken for anyone on a fresh checkout (`cargo machete: no such command`, exit 2 on `ci-lint-cargo-machete`).

The one trade-off compared to CI: `cargo install` compiles from source rather than fetching pre-built binaries, so `make install` on a fresh machine will be slower. This is a one-time cost and avoids introducing a dependency on `cargo-binstall` or any other external installer beyond a standard Rust toolchain.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -125,7 +125,7 @@ core-i18n-extract:
 		--ignore website \
 		-l en
 
-rust-install:  ## Install required Cargo tools (cargo-deny, cargo-machete, cargo-llvm-cov, nextest)
+rust-install:  ## Install required Cargo tools
 	$(CARGO) install --locked cargo-deny cargo-machete cargo-llvm-cov cargo-nextest
 
 install: node-install docs-install core-install rust-install  ## Install all requires dependencies for `node`, `docs`, `core`, and `rust`
--- a/Makefile
+++ b/Makefile
@@ -125,7 +125,10 @@ core-i18n-extract:
 		--ignore website \
 		-l en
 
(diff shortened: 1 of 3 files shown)


gergosimonyi (commented):
`$TMPDIR` is part of POSIX (only recommended, not required). I'm interested what system you worked on where a fallback was needed.

rissson:
same question as gergo, I'm not too keen on having a hardcoded fallback for TMPDIR

Makefile
```diff
@@ -125,7 +125,10 @@ core-i18n-extract:
 		--ignore website \
 		-l en
 
-install: node-install docs-install core-install  ## Install all requires dependencies for `node`, `docs` and `core`
+rust-install:  ## Install required Cargo tools (cargo-deny, cargo-machete, cargo-llvm-cov, nextest)
```
rissson:
```suggestion
rust-install:  ## Install required Cargo tools
```

PR #21745, opened 2026-04-21
restructure `UserSerializer.create` and `UserSerializer.update`

gergosimonyi:


gergosimonyi:
Implemented in target branch

Issue #18208, opened 2025-11-18
Automatically redirect to initial-setup on first authentik boot


PeshekDotDev:
### Is your feature request related to a problem?

When setting up new authentik instances, you must manually take yourself to initial-setup

### Describe the solution you'd like

I want to automatically be redirected to initial-setup

### Describe alternatives that you've considered

Manually taking myself to initial-setup, which does work, but redirecting would be nicer

PR #21753, opened 2026-04-21
core: complete rework to oobe and setup experience

BeryJu:
The main ideas with this PR:

- single source of truth for whether an authentik instance has been setup or not
- don't make users have to enter the whole initial setup flow URL

Side goals:

- Users should be able to navigate to <authentik-ip>:9000 and be able to just let the tab sit until everything is ready

Notes:

- New setup flag using flag system (hidden)
- Determine whether authentik has been setup using version history
- Probably also worth checking if any non-akadmin users exist


authentik/core/tests/test_setup.py
```diff
@@ -0,0 +1,158 @@
...
+        self.assertEqual(res.status_code, HTTPStatus.FOUND)
+        self.assertRedirects(
+            res,
+            reverse("authentik_core:root-redirect"),
+            fetch_redirect_response=False,
+        )
+
+    @patch_flag(Setup, False)
+    def test_not_setup_no_flow(self):
+        """Test case on initial startup; setup flag is not set and oobe flow doesn
```
dominic-r:
```suggestion
        """Test case on initial startup; setup flag is not set and oobe flow does
```

dominic-r: ;)

dominic-r (approved):
lgtm docs

authentik/commands/management/commands/worker.py
```diff
@@ -0,0 +1,35 @@
+from os import getenv
```
rissson: this will clash with https://github.com/goauthentik/authentik/pull/21324

BeryJu: Fixed by moving logic to signal

authentik/core/migrations/0058_setup.py
```diff
@@ -0,0 +1,57 @@
...
+    ):
+        is_already_setup = True
+    # non-akadmin and non-guardian anonymous user exist
+    if (
+        User.objects.using(db_alias)
+        .exclude(username="akadmin")
+        .exclude(username="AnonymousUser")
+        .exists()
+    ):
+        is_already_setup = True
```
rissson: I would extract this to another function to avoid extra queries if they're not needed

authentik/core/migrations/0058_setup.py
```diff
@@ -0,0 +1,57 @@
...
+
+    if is_already_setup:
+        tenant = get_current_tenant()
```
rissson, line 48: Will this work in migrations?

BeryJu: seems to from the migration-from-stable tests

rissson:
Review and tests so far:
- migration from an already setup authentik instance correctly sets the setup flag to true
- on a new instance, I am indeed redirected to the initial-setup flow, and only after the blueprint has been applied

authentik/core/setup/views.py
```diff
@@ -0,0 +1,80 @@
...
+            return HttpResponse(status=HTTPStatus.SERVICE_UNAVAILABLE)
+        if not Flow.objects.filter(slug=self.setup_flow_slug).exists():
+            return HttpResponse(status=HTTPStatus.SERVICE_UNAVAILABLE)
+        return HttpResponse(status=HTTPStatus.OK)
+
+    def get(self, request: HttpRequest):
+        flow = Flow.objects.filter(slug=self.setup_flow_slug).first()
+        if not flow:
+            LOGGER.info("Setup flow does not exist yet, waiting for worker to finish")
+            return HttpResponse(
```
rissson, line 51: do we want the same logic here as https://github.com/goauthentik/authentik/blob/main/internal/web/proxy.go#L102 ?

authentik/core/apps.py
```diff
@@ -26,6 +32,10 @@ class AuthentikCoreConfig(ManagedAppConfig):
     mountpoint = ""
     default = True
 
+    def import_related(self):
+        super().import_related()
+        self.import_module("authentik.core.setup.signals")
```
rissson, line 37: This seems like a hack. Can we maybe turn `authentik.core.setup` into an app?

rissson (approved):

authentik/core/setup/views.py
BeryJu: I dont think we need all that tbh, especially since this is only returned on the root URL and the setup URL

authentik/core/apps.py
BeryJu: a little bit, it doesn't justify its own app for 2 files but I also dont want to just throw it in with core

authentik/core/setup/views.py
rissson: fair enough

authentik/core/apps.py
rissson: fair enough

rissson (approved):

PeshekDotDev:
closes #18208

PR #21787, opened 2026-04-23
web: Gracefully handle missing element construction.

GirlBossRush:
## Details

This PR introduces a fallback to component tag names which are provided from the server. If the component fails to load, the fallback will render an error message instead of breaking the entire page. 

This can be observed when a stage has been created on another branch:

Screenshot 2026-04-23 at 02 38 50

### DIFF
--- a/web/src/elements/errors/boundaries.ts
+++ b/web/src/elements/errors/boundaries.ts
@@ -1,5 +1,9 @@
+import { globalAK } from "#common/global";
+
 import { AKElement } from "#elements/Base";
 import { SlottedTemplateResult } from "#elements/types";
+
+import { CapabilitiesEnum } from "@goauthentik/api";
 
 import { msg } from "@lit/localize";
 import { customElement } from "@lit/reactive-element/decorators/custom-element.js";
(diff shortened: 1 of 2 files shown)


web/src/elements/errors/boundaries.ts
```diff
@@ -0,0 +1,31 @@
...
+    protected override render(): SlottedTemplateResult {
+        return html`<div class="pf-c-alert pf-m-danger" role="alert">
+            <div class="pf-c-alert__icon">
+                <i class="fas fa-exclamation-triangle" aria-hidden="true"></i>
+            </div>
+            <h4 class="pf-c-alert__title">${msg("Failed to load element")}</h4>
+            <div class="pf-c-alert__description">
+                ${msg(
+                    "The element could not be loaded. This may be due to a missing import or a version mismatch.",
+                )}
```
BeryJu: I'd have two different strings based on if capabilities include debug; for debug this is fine but for non-debug I'd phrase it more like "An element could not be loaded. Please try refreshing the page or clearing your cache" (or something along those lines)

BeryJu (commented):
nice QOL!

kensternberg-authentik (approved):
This looks fine, but it also feels like a lot of ceremony. I'm not sure we need it. Do other applications like Spectrum, Home Assistant, and so forth have anything like this?

PR #21848, opened 2026-04-27
feat(worker): Separate critical actors from the default queue

josegomezr:
## Details

Helps a lot to shape priorities in the queue, whenever long running tasks (like an LDAP Sync/SCIM Sync over a big dataset) span over multiples of the available worker threads.

Also helps alleviate unexpected behaviors like emails not sent and/or SCIM direct syncs [via the UI] not happening on due time as the job was not picked due to an overcrowded queue.

The default worker command will still process _all_ queues. The explicit queue names enable queue-specific workers and thus guarantee that, for example, emails will not be stalled waiting for synchronization tasks to finish.

There's a second side to this PR: Having the rust worker controller pass `argv` to the underlying `python -m worker ...` invocation, not there yet due to lack of rustiness on my side :melting_face: 

---

## Checklist

If applicable

### DIFF
--- a/authentik/events/tasks.py
+++ b/authentik/events/tasks.py
@@ -23,7 +23,10 @@ from authentik.tasks.middleware import CurrentTask
 LOGGER = get_logger()
 
 
-@actor(description=_("Dispatch new event notifications."))
+@actor(
+    description=_("Dispatch new event notifications."),
+    queue_name="event_dispatcher",
+)
 def event_trigger_dispatch(event_uuid: UUID):
     for trigger in NotificationRule.objects.all():
         event_trigger_handler.send_with_options(args=(event_uuid, trigger.name), rel_obj=trigger)
@@ -33,7 +36,8 @@ def event_trigger_dispatch(event_uuid: UUID):
     description=_(
         "Check if policies attached to NotificationRule match event "
         "and dispatch notification tasks."
-    )
+    ),
+    queue_name="event_dispatcher",
 )
 def event_trigger_handler(event_uuid: UUID, trigger_name: str):
     """Check if policies attached to NotificationRule match event"""
@@ -97,7 +101,10 @@ def event_trigger_handler(event_uuid: UUID, trigger_name: str):
     self.info(f"Created {count} notification tasks")
 
 
-@actor(description=_("Send notification."))
+@actor(
+    description=_("Send notification."),
+    queue_name="event_dispatcher",
+)
 def notification_transport(transport_pk: int, event_pk: str, user_pk: int, trigger_pk: str):
     """Send notification over specified transport"""
     event = Event.objects.filter(pk=event_pk).first()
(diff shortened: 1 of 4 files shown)


rissson:
Heya, thanks for the contribution. The idea for this was to have hierarchical queues, so for instance:

- sync
  - sources
    - ldap
    - kerberos
  - providers
    - scim
    - google
    - entra
- events
  - dispatch
  - mail

and so on. Selecting `sync` would selecting all the queues beneath that key.

As for passing which queue(s) to process for a given worker, we probably want to use a config key instead of a CLI argument, but @BeryJu might have a different opinion.

We also need to figure out how to ignore specific queues in a worker, so that if you have two worker pools, and the first one is only for `sync` tasks, then the second pool should ignore `sync` tasks and process all the other ones. We might also want to improve the Helm chart to allow multiple worker pool deployments.

And finally, this probably requires custom handling in the dramatiq consumer (`/packages/django-dramatiq-postgres/django_dramatiq_postgres/broker.py`) so that we don't spawn multiple consumers, which in turn results in added load on the database, but instead handle queue filtering in a single consumer.

josegomezr:
> Heya, thanks for the contribution. The idea for this was to have hierarchical queues.

Haven't checked so deep into dramatiq, but does it support filtering based on separator?

e.g.

```
ak worker --queue foo.*
```

would be able to prefix match all workers that start by foo?

---

> We also need to figure out how to ignore specific queues in a worker, so that if you have two worker pools, and the first one is only for sync tasks, then the second pool should ignore sync tasks and process all the other ones. We might also want to improve the Helm chart to allow multiple worker pool deployments.

I'm testing something for this case at the moment, the ability to spawn many worker deployments from the same template with custom args each (e.g. producing a deployment worker-defaut: `ak worker`, and a deployment worker-mailer: `ak worker --queue mailer`, ...)

rissson:
> Haven't checked so deep into dramatiq, but does it support filtering based on separator?
> 
> e.g.
> 
> ```
> ak worker --queue foo.*
> ```
> 
> would be able to prefix match all workers that start by foo?

I'm pretty sure it doesn't, but since we have our own consumer, we can do that ourselves.

We currently define the queues here https://github.com/goauthentik/authentik/blob/main/lifecycle/worker_process.py#L89, which are then passed to the worker here https://github.com/Bogdanp/dramatiq/blob/master/dramatiq/worker.py#L90 and used here https://github.com/Bogdanp/dramatiq/blob/master/dramatiq/worker.py#L216 to create new consumers for each queue (which we don't want).

What we'd need to do is override the `Worker` class there to only spawn on consumer always, and then do the filtering there. For instance in https://github.com/goauthentik/authentik/blob/main/packages/django-dramatiq-postgres/django_dramatiq_postgres/broker.py#L302

josegomezr:
Been giving it some rounds and I'm not _there_ yet fully, from one side I achieved the single consumer part by "hardcoding" the consumer in the Worker Class to be exclusively one. That solves the

> we don't spawn multiple consumers, which in turn results in added load on the database, but instead handle queue filtering in a single consumer.

Now, from there the issue is that the message selection (on the Broker side) can't happen before decoding, as changing `Task.queue_name` also alters the `pg_notify`, pretty much making the `LISTEN/NOTIFY` part of job fetching irrelevant.

I see two potential solutions:

- Add more data to the `pg_notify` call so `LISTEN`-ers can decide to ignore messages without having to unwrap the message.
- Bite the bullet and do the decoding.

On the latter, adding an option (say for the sake of the example: `ak_queue_name`) as a message option _does_ seem to be enough to "ignore" messages that the current consumer is not interested in. Addenda: This I did via a custom `actor_class` and allowing `ak_queue_name` as an accepted parameter in the broker.

Will continue experimenting...

PR #21856, opened 2026-04-27
outpost/proxy: rewrite in rust

rissson:
only merge this as a stack

### DIFF
--- a/.github/workflows/ci-main.yml
+++ b/.github/workflows/ci-main.yml
@@ -433,7 +433,10 @@ jobs:
       - name: Setup authentik env
         uses: ./.github/actions/setup
         with:
-          dependencies: rust,runtime
+          dependencies: system,python,rust,runtime
+      - name: Prepare database
+        run: |
+          uv run make migrate
       - name: run tests
         run: |
           cargo llvm-cov --no-report nextest --workspace
--- a/src/outpost/proxy/application/handlers/forward.rs
+++ b/src/outpost/proxy/application/handlers/forward.rs
@@ -1,0 +1,271 @@
+use std::{net::IpAddr, sync::Arc};
+
+use ak_axum::{error::Result, extract::client_ip::ClientIp};
+use axum::{
+    extract::{Request, State},
+    http::{HeaderMap, StatusCode, Uri, header},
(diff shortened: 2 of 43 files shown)


.github/workflows/ci-main.yml
```diff
@@ -465,7 +468,7 @@ jobs:
     needs:
       - build-compute-tags
       - build
-      - ci-core-mark
```
rissson, line 472: NOTE TO SELF: revert this

src/outpost/proxy/application/handlers/forward.rs
```diff
@@ -0,0 +1,271 @@
+use std::{net::IpAddr, sync::Arc};
```
BeryJu, line 1: I would've probably implemented the different proxies as implementations of traits but thats probably just my OOP brain speaking

src/outpost/proxy/application/mod.rs
```diff
@@ -0,0 +1,195 @@
...
+
+    /// Default session lifetime: the access token validity plus one second
+    /// (so the session never outlives indefinitely), or zero if unset.
+    pub(super) fn session_max_age(&self) -> Duration {
+        self.provider
+            .access_token_validity
+            .map_or(Duration::ZERO, |validity| {
+                Duration::from_secs_f64(validity + 1.0)
+            })
+    }
```
BeryJu, line 194: yeah I'd probably have a bunch of the other methods that take App as a function in here but same diff

BeryJu (approved):
Looks good

I'm surprised this didnt end up as packages/ak-outpost-proxy crate thats just pulled in by the server tbh

src/outpost/proxy/application/handlers/forward.rs
rissson: writing down for posterity what I said about this: "I think those are just implementation details. I tried staying as close as possible to the go code in terms of organisation"

src/outpost/proxy/application/mod.rs
rissson: writing down for posterity what I said about this: "I think those are just implementation details. I tried staying as close as possible to the go code in terms of organisation"

PR #21868, opened 2026-04-28
flows: preserve signed background URLs in CSS

### LINKED ISSUE #21867: SFE breaks s3 urls
ex:

most likely a side effect of my recent pr

dominic-r:
Flow background URLs can include signed S3 query parameters with & separators. These values are rendered inside <style> tags, where Django autoescaping changes & to &amp;; browsers then request the literal escaped query string from S3, causing 400 responses for presigned background images.

Mark the flow background URL values as safe in the CSS-only template contexts used by the standard flow interface, the SFE flow page, and the full-screen login background. Regression coverage asserts that signed URL query separators are preserved in the rendered CSS for both standard and SFE flows.

Rationale for iriencode and safe: https://authentiksecurity.slack.com/archives/C08ABMS9TRP/p1777428178724259?thread_ts=1777386224.358639&cid=C08ABMS9TRP

Closes: https://github.com/goauthentik/authentik/issues/21867


BeryJu (approved):

authentik/core/templates/login/base_full.html
```diff
@@ -12,7 +12,7 @@
 {% block head %}
 <style data-id="static-styles">
     :root {
-        --ak-global--background-image: url("{{ request.brand.branding_default_flow_background_url }}");
+        --ak-global--background-image: url("{{ request.brand.branding_default_flow_background_url|iriencode|safe }}");
```
kensternberg-authentik, line 15: Why is this one different from the other two?

kensternberg-authentik (approved):
I had a question, but I assume the answer is because the source inside `core` is different from inside `flow`.

authentik/core/templates/login/base_full.html
dominic-r: This one is rendered in raw css

Issue #21865, opened 2026-04-27
Could not determine id from oauth


slavb18:
### Describe the bug

I see id in profile from  trace logs 
```
server-1  | {"auth_via": "unauthenticated", "body": "{\"auth_type\":\"employer\",\"is_applicant\":false,\"is_employer\":true,\"is_admin\":false,\"is_hiring_manager\":f
alse,\"is_application\":false,\"is_employer_integration\":false,\"id\":\"155745500\",...
```

But next line `Could not determine id.`

```
server-1  | {"auth_via": "unauthenticated", "domain_url": "id.iconicompany.com", "event": "Authentication Failure", "host": "id.iconicompany.com", "level": "warning", "logger": "authentik.sources.oauth.views.callback", "pid": 112, "reason": "Could not determine id.", "request_id": "47d11360e3a742479acb2cc814d132b6", "schema_name": "public", "timestamp": "2026-04-28T04:55:38.745434"}
```

### How to reproduce

Configure custom OAuth provider (social login)

### Expected behavior

Should get id from profile

### Additional context

Related with https://github.com/goauthentik/authentik/issues/12491

### Deployment Method

Docker

### Version

2026.2

### Relevant log output

```shell

```

Issue #12491, opened 2024-12-26
get_user_id Yandex


utya:
# Issue: TypeError in get_user_id Method During OAuth Authentication

## Problem Description
The `get_user_id` method in https://github.com/goauthentik/authentik/blob/main/authentik/sources/oauth/views/callback.py
 used for processing OAuth callbacks returns `null` despite valid profile data containing an ID. The issue cannot be resolved by modifying the existing method and requires the creation of a separate method to handle the profile data correctly.

## Current Behavior
The application fails to authenticate users through OAuth, particularly with the Yandex OAuth provider. The `get_user_id` method returns `null` even though the profile data contains a valid ID field.

Log example showing profile data with ID present but the method returning null:
```json
"raw_info": {
    "id": "26728592",
    "default_email": "user@example.com",
    ...
},
"identifier": null
```

Root Cause

The issue arises because the original get_user_id method cannot properly handle the incoming data. Even modifying the method (e.g., changing the type hint to Any) does not solve the problem. The only working solution is to create a new method that fully handles the profile data extraction.

Example of the non-working method:
```
def get_user_id(self, info: dict[str, Any]) -> str | None:
    """Return a unique identifier from the profile data."""
    if "id" in info:
        return info["id"]
    return None
```

Example of the working method:

```
def custom_get_user_id(self, info: Any) -> str | None:
    """Return a unique identifier from the profile data."""
    if "id" in info:
        return str(info["id"])
    return None
```

The key difference lies not only in the type hints but in the necessity of creating a separate method instead of trying to fix the existing one.

Impact
	•	Users cannot authenticate through OAuth providers.
	•	The authentication flow fails with a “Could not determine id” error.
	•	All OAuth providers using this method for ID extraction are affected.

Solution
	1.	Create a separate custom_get_user_id method to handle profile data and return the unique identifier:
	2. 
```
def custom_get_user_id(self, info: Any) -> str | None:
    """Return a unique identifier from the profile data."""
    if "id" in info:
        return str(info["id"])
    return None
```

BeryJu:
Where are you putting this `custom_get_user_id` function/where are you calling it from?

Which OAuth source type are you using for this, I'm assuming the OpenID Connect source type?

utya:
I simply replaced the default get_user_id method with custom_get_user_id.
I am using OpenID Connect as the OAuth source.

PR #21880, opened 2026-04-28
sources/oauth: ensure user ID is returned as str

slavb18:
Fix from #12491 for number id

## Details

When oauth profile returns id as number, e.g. "id": "123456", function fails with `return self.handle_login_failure("Could not determine id.")`
#12491
#21865

---

## Checklist

-   [X] The API schema and clients have been updated (`make gen`)

If applicable


rissson (approved):

rissson:

Issue #21947, opened 2026-04-28
client_credentials JWT exchange silently drops group mappings from source property maps


albrodfer1:
### Describe the bug

When a JWT bearer token (e.g. a SPIRE SVID) is exchanged for an Authentik token via the client_credentials grant using an OAuth source, any groups key returned by a source property mapping is silently discarded. The user is created/updated correctly and attributes are persisted, but group membership is never applied.                        
The interactive browser-based OAuth source flow works correctly because it runs GroupUpdateStage as part of the flow pipeline. The JWT path in `TokenParams.__create_user_from_jwt` (authentik/providers/oauth2/views/token.py) bypasses all flow stages and has no equivalent group-sync step.    

### How to reproduce

1. Create an OAuth source with a property mapping that returns a groups key:

```                                                                                   
return {        
"username": token.get("sub"),                                                                                                                                          
"groups": ["My Group"],
}
```                                                                                                                            
2. Configure a provider to accept client_credentials with that source (e.g. a SPIRE/JWT source).
3. Exchange a valid JWT for an Authentik token:                                                                                                                            
POST /application/o/token/                                                                                                                                                 
grant_type=client_credentials&client_assertion_type=urn:ietf:params:oauth:client-assertion-type:jwt-bearer&client_assertion=<jwt>                                          
4. Inspect the created user's group membership, it is empty.                                                                                                              
5. Inspect the user's attributes, they are set correctly.                                                                                                                 

###### Root cause      

SourceMapper.build_object_properties correctly returns groups in its output, but __create_user_from_jwt only reads username, name, email, path, type, and attributes. The  
groups key is never consumed. No error is raised.

### Expected behavior

The user should be added to "My Group" (resolved according to the source's group_matching_mode) after a successful token exchange, consistent with the behavior of the     
interactive flow.

### Deployment Method

Kubernetes

### Version

805ff9f Based on 2026.5.0-rc1

### Relevant log output

```shell
{"auth_via": "oauth_client_secret", "domain_url": "idm.lab.control-plane.io", "event": "verifying JWT with source", "host": "idm.lab.control-plane.io", "level": "debug", "logger": "authentik.providers.oauth2.views.token", "pid": 5341, "request_id": "7809a8666157458f84b0e46758626b8b", "schema_name": "public", "source": "librechat-spire-oauth-source", "timestamp": "2026-04-28T12:35:39.522721"}
[...]

PR #21948, opened 2026-04-28
providers/oauth2: fix group mapping ignored in JWT client_credentials…

albrodfer1:
## Details

**Issue** https://github.com/goauthentik/authentik/issues/21947

When a JWT bearer token (e.g. a SPIRE SVID) is exchanged for an Authentik token via the client_credentials grant using an OAuth source, any groups key returned by a source property mapping is silently discarded. The user is created/updated correctly and attributes are persisted, but group membership is never applied.
The interactive browser-based OAuth source flow works correctly because it runs GroupUpdateStage as part of the flow pipeline. The JWT path in TokenParams.__create_user_from_jwt (authentik/providers/oauth2/views/token.py) bypasses all flow stages and has no equivalent group-sync step.

Added `__sync_groups_from_jwt` that mirrors `GroupUpdateStage.handle_groups()`:

The helper uses `SourceMatcher` + `GroupOAuthSourceConnection`. the same mechanism as the interactive flow, so it:
  - Respects the source's `group_matching_mode`
  - Only replaces groups assigned by this source
  - Creates `GroupOAuthSourceConnection` records

Three regression tests added
- test_successful_group_mapping  
- test_group_mapping_name_link_existing_group
- test_group_mapping_reruns_update_membership 
---

## Checklist

If applicable

### DIFF
--- a/authentik/providers/oauth2/tests/test_token_cc_jwt_source.py
+++ b/authentik/providers/oauth2/tests/test_token_cc_jwt_source.py
@@ -290,3 +290,135 @@ class TestTokenClientCredentialsJWTSource(OAuthTestCase):
         self.assertEqual(jwt["email"], "test-user@example.com")
         self.assertEqual(jwt["given_name"], "Mapped Test User")
         self.assertEqual(jwt["preferred_username"], test_username)
+
+    def test_successful_group_mapping(self):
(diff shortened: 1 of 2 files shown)


dewi-tik:
Thank you for the contribution. Are you happy to rebase this on main?

PR #21949, opened 2026-04-28
web: MDX without React

GirlBossRush:
Replaces the React-based `<ak-mdx>` (and the `@mdx-js/mdx` runtime evaluator behind it) with a build-time Markdown→HTML pipeline plus a pair of light-DOM custom-element wrappers. The rewrite removes the last runtime use of `new AsyncFunction(...)` from the bundle, so the page CSP no longer needs `'unsafe-eval'` for `<ak-mdx>` to function.

Build pipeline (`web/bundler/mdx-plugin/node.js`):
- Compiles each `.md` / `.mdx` to HTML via `unified` + remark + rehype.
- Resolves relative `./foo.md` hrefs against the docs site at build time and adds `target="_blank"` + `rel="noopener noreferrer"`.
- Wraps every `<a>` in `<ak-md-a>` and every `<pre><code class="language-mermaid">` in `<ak-md-mermaid>` (wrap-don't-replace, so the original semantic HTML survives for graceful degradation).
- Translates Docusaurus-style `:::name Title` admonitions into the spec form `:::name[Title]` before parsing, then promotes the title paragraph to `<strong>` inside `<ak-alert>`.
- Adds `caution` and `tip` to the admonition vocabulary, mapping them to the closest PFAlert level.
- Bakes frontmatter `title:` into a `<h1 part="title">` and wraps the whole document in `<div class="pf-c-content" part="content">`.

Runtime (`web/src/elements/ak-mdx/`):
- `<ak-mdx>` is a thin Lit element. URL mode fetches the JSON envelope and stamps the pre-rendered HTML through `CompiledMarkdownTrustPolicy` (a passthrough Trusted Types policy for our own build output). Content mode runs the same `unified` pipeline in-browser for admin-supplied prose, sanitised through `BrandedHTMLPolicy`.
- `<ak-md-a>` (light-DOM Lit element via `:host { display: contents }`) attaches a click interceptor on the slotted `<a>` so fragment links scroll within the host shadow tree instead of overwriting `location.hash` and yanking the hash-routed SPA off its current page.
- `<ak-md-mermaid>` delegates to the existing `<ak-diagram>` element, using the slotted `<pre><code>` as the source of mermaid markup.

Mermaid theming (`web/src/elements/Interface.ts`, `web/src/common/mermaid.ts`):
- `applyMermaidTheme` is now invoked once at startup with the resolved current theme so the first SVG render of a page loaded in dark mode comes out in the right colours, instead of waiting for the user's first explicit theme toggle to fix it.
- `<ak-diagram>` listens for `AKMermaidRefreshEvent` on `window` and calls `requestUpdate()` on each fire, picking up the latest global mermaid config without per-instance theme tracking.

Dependencies:
- Drop `@mdx-js/mdx`, `react`, `react-dom`, `rehype-mermaid`, `remark-mdx-frontmatter`.
- Promote `unified`, `remark-parse`, `remark-rehype`, `mdast-util-to-string` from transitive to direct.
- Convert the remaining `react` imports (`HTMLAttributes`, `ImgHTMLAttributes`, `LabelHTMLAttributes`) to `import type` so the bundler erases them entirely.

Tests:
- `web/test/browser/ak-mdx.test.ts` — Playwright e2e, exercises frontmatter, heading slugs, admonitions, link wrapping (external, relative, fragment), mermaid render, and theme-toggle re-rendering against the OAuth2 provider docs.


kensternberg-authentik:
Suggestion 1: Make the first change ([Extract Mermaid](https://github.com/goauthentik/authentik/pull/21949/changes/9914a935beb11464f8af856d327e3c9c348e7c30)) into its own PR.

web/src/admin/flows/FlowDiagram.ts
```diff
@@ -6,22 +6,29 @@ import { Diagram } from "#elements/Diagram";
...
-        this.diagram = undefined;
+    protected override updated(changedProperties: PropertyValues<this>): void {
+        super.updated(changedProperties);
+
+        if (changedProperties.has("flowSlug")) {
+            this.refresh();
+        }
+    }
+
+    protected refresh = (): void => {
```
kensternberg-authentik: `this.diagram` is already a property, so changing it should automatically trigger an update.  The reason you need to `requestUpdate()` is that you're doing it inside a method that prevents updates from being scheduled.

Suggested alternative: use pfe-core's `@observes('flowSlug')` to automagically trigger this cascade; it runs before `shouldUpdate` triggers, so you can use it to schedule this retrieval automagically.

web/src/admin/flows/FlowDiagram.ts
```diff
@@ -6,22 +6,29 @@ import { Diagram } from "#elements/Diagram";
...
         new FlowsApi(DEFAULT_CONFIG)
             .flowsInstancesDiagramRetrieve({
```
kensternberg-authentik, line 21: `aki(FlowsApi)` ?  Now that it's merged?

web/src/elements/mermaid/utils.ts
```diff
@@ -0,0 +1,49 @@
+import { DOM_PURIFY_STRICT } from "#common/purify";
```
kensternberg-authentik: I bet this whole commit works fine on its own; maybe consider pulling it out into its own PR?

GirlBossRush:
Split out into #22980, with both of the FlowDiagram suggestions applied over there.

web/src/common/purify.ts
```diff
@@ -50,6 +50,17 @@ export const SanitizedTrustPolicy = trustedTypes.createPolicy("authentik-sanitiz
...
+ * Trusted types policy for HTML produced by our own build-time markdown
+ * pipeline. The HTML is generated from source we own (the `mdx-plugin`),
+ * including custom elements like `<ak-md-a>` and `<ak-alert>` that
+ * DOMPurify's default tag list would strip. Treat the input as already
+ * trusted and pass it through unmodified.
+ */
+export const CompiledMarkdownTrustPolicy = trustedTypes.createPolicy("authentik-markdown", {
+    createHTML: (trustedHTML: string) => trustedHTML,
+});
+
```
kensternberg-authentik, line 79: Using my security audit prompt, Claude found at least two places where this is... worrisome.

> ProxyProviderViewPage.ts:419 -> ProxyProviderViewPage:160: A corrupted `provider.externalHost` value in the provider configuration, a detail entered by the admin, is generated before this is called, and it is passed to `ak-mdx`; this violates the security promise that "we own" the source code; that's not true here.

> Oauth2ProviderViewPage:404 : same problem; a provider configuration detail is passed to the markdown without any sanitization and checking.  The security promise that "we own" the source code is not true here.

Okay, yes, in practice, admins would have to be idiots to create a security hole through this mechanism.  Big companies can have problematic employees who love to mess with and push the limits.  I don't know if you want to address this, but I'm laying it out here because I can see how these violate the security promise.

web/src/elements/ak-mdx/markdown.ts
```diff
@@ -0,0 +1,46 @@
...
+ */
+export async function compileRuntimeMarkdown(source: string): Promise<string> {
+    if (!source.trim()) return "";
+
+    // Translate Docusaurus's `:::name Title` syntax to `:::name[Title]`
+    // before remark-directive parses it; otherwise it falls through as
+    // plain text.
+    const normalized = normalizeAdmonitionLabels(source);
+
+    const file = await unified()
```
kensternberg-authentik, line 44: Why is this _different_ from the process done in the Bundler, and is there a risk that we'll make changes to one without also changing the other?  

The left hoof wants to say "Centralize this process into a common function," and another that says," and on the other hoof I wonder if there's enough difference between the two to just put a comment here reminding people that if you change this, change the Bundler as well.

[‎web/bundler/mdx-plugin/compile.js](https://github.com/goauthentik/authentik/pull/21949/changes#diff-e87029400333c125cc60595e2ee35a5e92ca801ce7ecd895f78c54d18fcaec43)

kensternberg-authentik (commented):
What I see here is:

1. the `<ak-mdx>` and `<ak-md-a>` elements are _consumers_ of markdown to produce HTML.

2. The bundler is also a consumer of markdown to produce HTML.

3. The transform presets by remark are duplicated in two places, and could be unified and published as a separate package in the same way you've got packages for eslint or prettier.

4. The bundler plugin is probably also publishable as a separate package.  It takes the content as arguments and produces the corresponding output.

BeryJu (commented):
From a visual standpoint in the UI everything looks great, I'll leave the code review to @kensternberg-authentik

web/src/elements/ak-mdx/ak-mdx.ts
```diff
@@ -0,0 +1,149 @@
...
+     * reactive state. After this completes, Lit's render takes over.
+     */
+    protected async hydrate(): Promise<void> {
+        this.compiledTemplate = this.url
+            ? await this.#hydrateFromURL(this.url)
+            : await this.#hydrateFromContent(this.content ?? "");
+    }
+
+    public override render(): SlottedTemplateResult {
+        return this.compiledTemplate;
```
kensternberg-authentik, line 141:
```suggestion
        return this.compiledTemplate ?? nothing;
```

kensternberg-authentik: That's not a complete suggestion, since you have to import `nothing` as well to make it work.

kensternberg-authentik (approved):

PR #21976, opened 2026-04-30
website/docs: clarify blueprint identifiers

### LINKED ISSUE #15601: Specifics on using Identifiers in Blueprints
**Describe your question**
I'm using blueprints to setup our providers/applications etc and I'm a little unclear on how identifiers should be used.  The documentation states "Key:value filters to uniquely identify this object (required)".  My questions are:

- What values are acceptable for the keys/values? Are both the key and values arbitrary or do they have to match attrs? 
- In what cases do you need to provider multiple key/value pairs?
- In one case, I'm creating a authentik_policies.policybinding model and tried to set an identifier using 'name' as a key and the blueprint failed.

Any additional info/insight you could provide would be very helpful. Thank you!

**Relevant info**
Running Authntik v2025.6.3 on Docker

**Logs**
This is the error I receive when tyring to use name identifier in the policybinding model
```
Entry invalid: Invalid identifier field: Cannot resolve keyword 'name' into field. Choices are: enabled, failure_result, group, group_id, negate, order, policy, policy_binding_uuid, policy_id, target, target_id, timeout, user, user_id
```

dominic-r:
Closes: https://github.com/goauthentik/authentik/issues/15601

## Details

REPLACE ME

---

## Checklist

If applicable

### DIFF
--- a/website/docs/customize/blueprints/v1/models.mdx
+++ b/website/docs/customize/blueprints/v1/models.mdx
@@ -181,13 +181,15 @@ For example:
 
 ### Required fields
 
-Each PolicyBinding entry must have **exactly one** of the following set in `attrs`:
+Each PolicyBinding entry must provide **exactly one** of the following fields. These are commonly set in `attrs`, but can also be used in `identifiers` when they are part of the lookup for an existing binding:
 
 - `policy`: Reference to a Policy object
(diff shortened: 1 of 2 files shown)


tanberry (approved):
Thanks @dominic-r !

PR #21996, opened 2026-04-30
ci: run Playwright e2e suite on every PR

### LINKED ISSUE #21994: ci: run Playwright e2e tests on every PR
## Context

Today the only end-to-end coverage that runs on each PR in GitHub Actions is the Django/Selenium suite under `tests/e2e/`. The browser-side Playwright suite under `web/e2e/` (driven by `web/playwright.config.js`) is run **locally** by frontend developers — it never executes on CI. That gap means UI regressions only surface during manual review or after merge.

## Problem

- Frontend changes can land without any automated browser verification.
- Reviewers and release managers cannot trust that a green CI implies a working web UI.
- Local-only execution drifts: tests rot, fixtures break silently, and the bar to run them goes up over time.

## Proposal

Wire up the existing Playwright suite to run on every PR via GitHub Actions. Concretely:

- Add a `ci-web-e2e` job (or extend `ci-web.yml`) that:
  - Boots an authentik stack sufficient for the current Playwright fixtures (auth flows, providers, sessions).
  - Installs Playwright browsers via `corepack`-pinned npm.
  - Runs `npm run -w web playwright` against that stack.
  - Uploads the HTML report and any traces/videos as artifacts on failure.
- Decide on a sharding/parallelism story so the job stays under a sensible wall-clock budget.
- Make the job required for merge once it is stable.

## Dependencies

- Builds on top of the Corepack normalization in #20400 — Playwright install is sensitive to npm version drift, and the e2e job will use the same composite `setup-node` action.
- Branch `github-ci-playwright-e2e` already carries WIP scaffolding on top of `npm-corepack`.
- Related earlier WIP: #15693 (Playwright groundwork).

## Acceptance criteria

- [ ] Playwright e2e job runs on every PR touching `web/**` (and on `main`).
- [ ] Job is reproducible from a fresh clone — no hidden local state, no manual `playwright install` step in dev docs that CI skips.
- [ ] HTML report + traces are retrievable from the failed run's artifacts.
- [ ] Job is required for merge.
- [ ] Frontend developer docs are updated to point at the same npm scripts CI uses, so local and CI runs stay in lockstep.

## Out of scope

- Adding new Playwright tests beyond what's needed to validate the CI plumbing.
- Replacing the Django/Selenium suite — both stay until coverage parity is reached.

GirlBossRush:
## Details

This PR wires up the existing `web/test/browser/` Playwright suite as its own job in `ci-web.yml`, closing #21994.

Boots the full authentik stack from source (postgres + Go server + Rust worker) inside the workflow, applies migrations and the test-admin blueprint, then runs `corepack npm run --prefix web test:e2e` against `http://localhost:9000`. On failure, uploads the HTML report, traces/videos, and server + worker logs as artifacts so reviewers can debug without rerunning locally.

Also enables the HTML reporter and screenshot/video capture on CI in `playwright.config.js`, and updates the full dev-environment docs to point at the same npm scripts CI uses so local and CI runs stay in lockstep.

## Stacking

This PR targets `npm-corepack` (#20400) intentionally — the new job uses `corepack`-prefixed npm calls and the composite `setup-node` action that #20400 introduces. Will rebase onto `main` once #20400 lands.

## Notes for reviewer

- **Rust cold build is the long pole** (~5-10 min cold, ~1 min warm via the existing rust toolchain cache keyed on `Cargo.lock`). 60-min job timeout chosen with that in mind.
- **Not added to `ci-web-mark.needs`** — runs in parallel with lint/build to avoid cascading the heavy job into vitest. Branch protection should add `e2e (playwright)` directly as a required status check after the first green run.
- **No path filter** — consistent with the rest of `ci-web.yml`.
- Existing Django/Selenium suite under `tests/e2e/` is untouched; both suites coexist as the issue requests.

## Test plan

- [ ] First run on this PR completes within the 60-min timeout (validates cold rust build assumption)
- [ ] HTML report + traces + authentik logs are retrievable from a deliberately-failed run's artifacts
- [ ] After merge, branch protection updated to require `e2e (playwright)` status check

Closes #21994


.github/workflows/ci-web.yml
```diff
@@ -146,30 +155,137 @@ jobs:
...
+          } >> "$GITHUB_OUTPUT"
+      # The S3 publishing pair below is intentionally gated off until the
+      # `authentik-playwright-artifacts` bucket is provisioned by infra. Flip
+      # the repo variable `PLAYWRIGHT_S3_ENABLED=true` to turn it on; the URL
+      # baked into the PR comment below already points at the eventual key.
+      - name: Configure AWS credentials for HTML report upload
+        if: ${{ !cancelled() && github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name == github.repository && vars.PLAYWRIGHT_S3_ENABLED == 'true' }}
+        uses: aws-actions/configure-aws-credentials@ec61189d14ec14c8efccab744f656cffd0e33f37 # v6.1.0
+        with:
+          role-to-assume: "arn:aws:iam::016170277896:role/github_goauthentik_authentik"
```
dominic-r, line 237: this might open us to more exposure/risk, I'd keep this as an env var

GirlBossRush: Seems to be safe :)

https://github.com/goauthentik/authentik/blame/0b6d3a2850b5678bb189cf845e23cb1e4d5abfe5/.github/workflows/release-publish.yml#L196

https://github.com/goauthentik/authentik/commit/fda605428531aea00d160159dcec40c5292bf979#diff-c6d47dacf31662f6791d0cf218b1d34ee0ab48348c037819431cdd40e3fc23e4R187

dominic-r: In that case, fair :)

.github/workflows/api-ts-publish.yml
```diff
@@ -0,0 +1,65 @@
...
+          token: ${{ steps.generate_token.outputs.token }}
+      - uses: ./.github/actions/setup-node
+        with:
+          working-directory: web
+      - name: Generate API Client
+        run: make gen-client-ts
+      - name: Publish package
+        working-directory: gen-ts-api/
+        run: |
+          npm i
```
dominic-r: would there be a reason why we would need corepack'd npm in this file?

dominic-r: i'm thinking publish wise if anything changes and really making sure all our stuff uses the same version, no matter what it is

.github/workflows/ci-web.yml
```diff
@@ -12,51 +12,280 @@ on:
...
+      - name: Build web UI
+        run: corepack npm run --prefix web build
+      - name: Build authentik server (Go)
+        run: | # shell
+          go build -o ./bin/authentik-server ./cmd/server
+          sudo install -m 0755 ./bin/authentik-server /usr/local/bin/authentik-server
+      - name: Build authentik worker (Rust)
+        run: | # shell
+          cargo build --release --bin authentik
+          sudo install -m 0755 ./target/release/authentik /usr/local/bin/authentik
```
dominic-r, line 120:
```suggestion
          sudo install -m 0755 ./target/release/authentik /usr/local/bin/authentik-worker
```

.github/workflows/ci-web.yml
```diff
@@ -12,51 +12,280 @@ on:
...
+        id: playwright-cache
+        uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v4
+        with:
+          path: ~/.cache/ms-playwright
+          key: ${{ runner.os }}-playwright-${{ steps.playwright-version.outputs.version }}
+      - name: Install Playwright browsers
+        working-directory: web
+        run: | # shell
+          if [ "${{ steps.playwright-cache.outputs.cache-hit }}" = "true" ]; then
+            corepack npm exec -- playwright install-deps chromium
```
dominic-r: does this use the project's playwright version or the one of npm? Also, I don't see an install step, but it might be handled by the setup action? If that's the case, it would use the project version, right?

.github/workflows/ci-web.yml
```diff
@@ -12,51 +12,280 @@ on:
...
+          else
+            corepack npm exec -- playwright install --with-deps chromium
```
dominic-r: do we want to dare to test firefox based browsers as well? as we have encountered a few strange/unique bugs on there as well

.github/workflows/ci-web.yml
```diff
@@ -12,51 +12,280 @@ on:
...
+          # contains the <ak-flow-executor> custom element — both are absent from the
+          # worker's axum healthcheck router, so checking either rules out the
+          # port-collision failure mode.
+          timeout 240 bash -c '
+            until curl -fsS -o /dev/null http://localhost:9000/-/health/ready/; do
+              sleep 2
+            done'
+          timeout 300 bash -c '
+            until curl -fsS http://localhost:9000/if/flow/default-authentication-flow/ \
+              | grep -q "ak-flow-executor"; do
```
dominic-r: this feels a bit hacky? also, usually, would this be present even if no js is loaded client side via the bundles?  mabye just check for the inline presence of a brand config since that's a pretty permanent item?

.github/workflows/ci-web.yml
```diff
@@ -12,51 +12,280 @@ on:
...
+        run: | # shell
+          set -euo pipefail
+          if [ ! -d web/playwright-report ]; then
+            echo "No playwright-report/ produced; skipping S3 upload"
+            exit 0
+          fi
+          aws s3 cp \
+            --recursive \
+            --acl=public-read \
+            --cache-control "public, max-age=600" \
```
dominic-r: how likely are the reports to change once imported? i'd go for a longer cache control

.github/workflows/ci-web.yml
```diff
@@ -12,51 +12,280 @@ on:
...
+        if: ${{ !cancelled() && github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name == github.repository && vars.PLAYWRIGHT_S3_ENABLED == 'true' }}
+        uses: aws-actions/configure-aws-credentials@ec61189d14ec14c8efccab744f656cffd0e33f37 # v6.1.0
+        with:
+          role-to-assume: "arn:aws:iam::016170277896:role/github_goauthentik_authentik"
+          aws-region: eu-central-1
+      - name: Upload HTML report to S3
+        if: ${{ !cancelled() && github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name == github.repository && vars.PLAYWRIGHT_S3_ENABLED == 'true' }}
+        env:
+          S3_BUCKET: authentik-playwright-artifacts
+          S3_KEY_PREFIX: pr-${{ github.event.pull_request.number }}/run-${{ github.run_id }}/attempt-${{ github.run_attempt }}
```
dominic-r, line 243: ci-web also runs on main and version branches, wouldn't there be an issue with the key prefix to contain the pr number

.github/workflows/ci-web.yml
GirlBossRush: Second iteration :)

.github/workflows/ci-web.yml
GirlBossRush: Good call — bumped to `max-age=2592000, immutable`. Key includes `run_id`/`attempt-N`, so write-once. Thanks!

.github/workflows/ci-web.yml
GirlBossRush: Going to keep `authentik` — `lifecycle/ak:35` does `command -v authentik` and falls back to `cargo run --` otherwise, so renaming would silently slow CI down. I'll fix the misleading step label instead.

.github/workflows/ci-web.yml
GirlBossRush: Project version. The composite `setup` action runs `corepack npm ci --prefix web` first, so `@playwright/test` is pinned by `web/package-lock.json` and `npm exec` resolves to it.

.github/workflows/ci-web.yml
GirlBossRush: `ak-flow-executor` is in the SSR'd HTML from the Go proxy — no JS needed. That's what distinguishes a real Go response from the Rust worker's axum healthcheck on a port collision (the failure this probe exists to catch). Brand-config grep would work too but isn't materially less hacky.

.github/workflows/ci-web.yml
GirlBossRush: The upload step is gated on `github.event_name == 'pull_request'`, so `pr-<num>` only expands when there's a PR context. On main/version-* pushes the S3 upload is skipped entirely.

PR #22009, opened 2026-05-01
providers/scim: fix per-user PolicyEngine evaluation in get_object_qs

ryanpesek:
## Details
PR #13947 introduced this filter in `SCIMProvider.get_object_qs(User)`:
```python
if self.backchannel_application:
    base = base.filter(
        pk__in=[
            user.pk
            for user in base
            if PolicyEngine(self.backchannel_application, user, None).build().passing
        ]
    )
 ```
This iterates every user in the database and instantiates PolicyEngine once per user. get_object_qs(User) is called once per page-sync task (one per sync_page_size users), once per signal-driven sync, and once during _sync_cleanup. With N users and N/page_size pages, that's O(N²/page_size) PolicyEngine evaluations per full sync per provider; each evaluation runs a recursive group-membership CTE plus the static-binding aggregation. On a instance with thousands of users and multiple SCIM providers, this produced:
- several hundred thousand PolicyEngine evaluations per provider per full sync
- 15–20 minute per-page sync tasks (each task processes only 100 users but evaluates all users)
- Sustained high database load
- Millions of P_ENG: Found static bindings log lines

The fix:
Replace the Python loop with a SQL-level translation of PolicyEngine.compute_static_bindings:
- Fast path (purely static bindings, no Policy objects): translate each binding into a Q expression and combine with OR (MODE_ANY) or chained .filter() calls (MODE_ALL — since Q(groups__in=g1) & Q(groups__in=g2) collapses to a single M2M JOIN that can never match, chained filters force separate JOINs). Group bindings use with_descendants() to honor
  hierarchy, matching PolicyEngine's user.all_groups(). Negated bindings become ~Q(...). The application's policy_engine_mode is honored. Zero PolicyEngine instantiations.
- Slow path (real Policy objects bound): we can't translate user-defined Python policies to SQL, so fall back to per-user
  PolicyEngine. For MODE_ALL, pre-filter candidates via the static-binding SQL filter first (a static-failing user can never pass MODE_ALL overall). For MODE_ANY, still iterate every user since a static-failing user might pass via policy.
- AppAccessWithoutBindings tenant flag is honored at the SQL level (returning base or base.none() when no bindings exist) and on the slow path (setting engine.empty_result per instance). Related cleanups (in the same code paths)
- sync_signal_direct / sync_signal_m2m: drop not queryset (which calls __bool__ → _fetch_all() and materializes the entire queryset just to test non-emptiness) in favor of a single .exists() call. Also remove a no-op .filter() in sync_signal_m2m.
- _sync_cleanup: pass the in-scope queryset as a subquery rather than materializing it into a Python set, avoiding a NOT IN (...) clause with thousands of parameters.
- SCIMProvider.get_object_qs(Group): drop the prefetch_related on scimprovidergroup_set — never accessed during sync iteration (write() does its own per-object connection lookup; values_list() in cleanup bypasses prefetches).

---

## Checklist

If applicable

### DIFF
--- a/authentik/lib/sync/outgoing/tasks.py
+++ b/authentik/lib/sync/outgoing/tasks.py
@@ -119,9 +119,12 @@ class SyncTasks:
                 client = provider.client_for_model(object_type)
             except TransientSyncException:
                 continue
-            in_scope_pks = set(provider.get_object_qs(object_type).values_list("pk", flat=True))
+            # Pass the in-scope queryset as a subquery rather than materializing every
+            # PK into a Python set, which scales better and produces a single SQL
+            # statement rather than a NOT IN (...) clause with thousands of params.
+            in_scope_qs = provider.get_object_qs(object_type).values("pk")
             stale = client.connection_type.objects.filter(provider=provider).exclude(
-                **{f"{client.connection_type_query}__pk__in": in_scope_pks}
+                **{f"{client.connection_type_query}__pk__in": in_scope_qs}
             )
             for connection in stale:
                 try:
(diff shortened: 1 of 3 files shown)


BeryJu (commented):
@ryanpesek I like the general idea of this and the optimisation for checking policybindings on multiple objects, however I'd much prefer that optimisation to be more re-usable for other things such as listing applications or RAC endpoints. Now rewriting this optimisation to be usable for that will take a bit longer so I'll keep this open and up-to-date as a temporary bug fix

PR #22038, opened 2026-05-04
web: Consistent use of "User Dashboard"

GirlBossRush:
## Details

As requested in our 2026.5.0 release notes, this PR normalizes our use of "user library" and "user interface" labels with "User Dashboard".

### DIFF
--- a/web/src/admin/applications/ApplicationForm.ts
+++ b/web/src/admin/applications/ApplicationForm.ts
@@ -197,17 +197,15 @@ export class ApplicationForm extends WithCapabilitiesConfig(ModelForm<Applicatio
                         ?checked=${this.instance?.openInNewTab ?? false}
                         label=${msg("Open in new tab")}
                         help=${msg(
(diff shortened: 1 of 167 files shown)


dominic-r (approved):

dewi-tik (approved):

kensternberg-authentik (approved):
The front-end changes look excellent. The documentation changes look mechanical, unremarkable (in a good way), and correct.

fheisler:
channeling @tanberry:
- "if checked" seems clearer to me than "whether" (and is unrelated to this update)
- There's still inconsistency between "User Dashboard" and "the User Dashboard"
- It was referenced as "My Applications" because it shows up with a "My applications" header on the user's page. Looks like it now show up as "User Dashboard" on the page? As much as I'm not a fan of using "my" in the header, it seems even stranger to refer to the user as a user there. (Do we need a title there at all?)

PR #22043, opened 2026-05-04
stages/identification: add configurable passwordless button label

### LINKED ISSUE #22042: stages/identification: passwordless flow button label is hardcoded to "Use a security key"
**Describe the bug**

`IdentificationStage.passwordless_flow` is a generic mechanism — admins point it at any authentication flow. But the button text is hardcoded to "Use a security key", which does not match non-WebAuthn passwordless flows (e.g., email-based authentication).

**To reproduce**
1. Create an EmailStage flow for passwordless auth
2. Set `passwordless_flow` on an IdentificationStage to that flow
3. Visit the login page
4. Button says "Use a security key" regardless of what the flow does

**Expected behavior**
Admins should be able to configure the button label, or the label should reflect the flow's purpose.

**Proposed solution**
Add an optional `passwordless_label` field to IdentificationStage. When blank → existing behavior. When set → renders custom text.

dahux-net:
## Problem

`passwordless_flow` is a generic mechanism — admins point it at any
authentication flow. But the button text is hardcoded to "Use a
security key", which does not match non-WebAuthn flows.

Organizations already use `passwordless_flow` for email-based
authentication, but the misleading label forces them to either fork
or accept confusing UX.

Fixes #22042

## Solution

Add an optional `passwordless_label` TextField to IdentificationStage.
When blank → existing behavior. When set → renders custom text.

## Changes

- **Model**: `passwordless_label` TextField (blank=True, default="")
- **API**: `passwordlessLabel` in IdentificationChallenge, conditional in view
- **Frontend**: fallback to `msg("Use a security key")` when label is empty
- **Admin**: text input next to `passwordlessFlow` selector
- **Migration**: `0018_identificationstage_passwordless_label`
- **Tests**: label inclusion, label omission, passwordless flow integration

## Rationale

- One field, no new models, no new stages, no new APIs
- Follows the same pattern as `passwordless_flow` URL itself
- Zero behavior change for existing deployments
- Enables any passwordless method without custom code


dahux-net:
Hi, I've rebased this onto the latest main and squashed it into a single commit. All checks passed and tests included. Is there anything else needed to get this reviewed? This would let users use passwordless_flow for non-WebAuthn flows (email auth, etc.) without the misleading “Use a security key” label. Thanks!

dewi-tik:
Thank you for the contribution. However, we're going to resolve this in a different method that doesn't lead to the object having customization in this way.

PR #22103, opened 2026-05-06
Web/flow/expose flow in lightdom

kensternberg-authentik:
Put the Login input sequence into the lightDOM.

# What

1.  Separate the Flow and FlowExecutor into two different components; the Flow is concerned *only* with the layout of the page and the look and feel, and it hosts the FlowExecutor; the FlowExecutor, in turn, contains only the logic about managing the Flow and its stages.

    Because the FlowExecutor gets some information about the Flow’s look and feel as part of its payload, it communicates those parts (and only those parts) with the Flow via an event. This is a tighter coupling that I might have liked, but it serves our purposes.

2.  Introduces the `light()` directive, a powerful directive that takes a Lit TemplateResult in-line in a component’s overall template rendering process and replaces the TemplateResult with a `slot` element, then places the TemplateResult into the component’s lightDOM, thus allowing us to both grant easy traversal to those elements *and* dictate the location and behavior of those elements.

    One trade-off to this design is that the CSS for those elements that will be in the lightDOM must also be in the parent element’s adoptedStylesheets; since we’re trying to open up the “whole tree of accessible nodes,” that means the document as a whole. We could, in theory, inject those automatically, but the risks of duplication and the violation of the Web Component idiom doesn’t justify doing so at this time.

3.  Wrap the chain of render calls from Flow ⮕ FlowExecutor ⮕ IdentificationStage in the `light()` directive so that now all of those elements are in the lightDOM. Other stages have not been transformed this way. I am unaware of any that need updating at this time.

# Note

This commit is dependent upon the [P5 Drawer with all capabilities](https://github.com/goauthentik/authentik/pull/21545) review. By itself, this commit does not update the Patternfly styling, but we expect to in the near future.

-   [🐉] The code has been formatted (`make web`)


web/src/flow/Flow.ts
```diff
@@ -0,0 +1,307 @@
...
+    public background: ContextualFlowInfo["background"];
+
+    @state()
+    protected backgroundThemedUrls?: ContextualFlowInfo["backgroundThemedUrls"];
+
+    #abortController: AbortController | null = null;
+
+    readonly #wsController = new FlowWebsocketClientController(this);
+
+    readonly #logger = ConsoleLogger.prefix("flow");
```
GirlBossRush: Nit: switch to `protected`

web/src/flow/Flow.ts
```diff
@@ -0,0 +1,307 @@
...
+        PFBackgroundImage,
+        Styles,
+    ];
+
+    //#endregion
+
+    //#region Properties
+
+    @property()
+    public slug: string = window.location.pathname.split("/")[3];
```
GirlBossRush:
```suggestion
    @property({ type: String })
    public slug: string = window.location.pathname.split("/")[3];
```

Possibly union with null ?

web/src/flow/Flow.ts
```diff
@@ -0,0 +1,307 @@
...
+ *
+ */
+@customElement("ak-flow")
+export class Flow extends WithBrandConfig(Interface) {
+    //#region Static
+
+    public static readonly DefaultLayout: FlowLayoutEnum =
+        globalAK()?.flow?.layout || FlowLayoutEnum.Stacked;
+
+    static styles: CSSResult[] = [
```
GirlBossRush:
```suggestion
    public static styles: CSSResult[] = [
```

web/src/flow/Flow.ts
```diff
@@ -0,0 +1,307 @@
...
+
+    // Only used by the `sidebar_*_frame_backgrounds` to give customers a place to put their
+    // branding visuals, if they like.
+    //
+    protected renderFrameBackground() {
+        return guard([this.layout, this.background], () => {
+            if (!this.#layoutUsesSidebarFrames) return nothing;
+
+            const { background } = this;
+            if (!background) return nothing;
```
GirlBossRush:
```suggestion
        const { layout, background } = this
        return guard([layout, background], () => {
            if (!this.#layoutUsesSidebarFrames) return null;

            if (!background) return null;
```

web/src/flow/FlowExecutor.ts
```diff
@@ -76,70 +75,44 @@ import PFTitle from "@patternfly/patternfly/components/Title/title.css";
...
-
     //#endregion
 
     //#region Internal State
 
-    #logger = ConsoleLogger.prefix("flow-executor");
+    readonly #logger = ConsoleLogger.prefix("flow-executor");
 
-    #api: FlowsApi;
+    readonly #api: FlowsApi;
```
GirlBossRush:
```suggestion
    protected logger = ConsoleLogger.prefix("flow-executor");

    protected api: FlowsApi;
```

web/src/flow/FlowExecutor.ts
```diff
@@ -152,6 +125,10 @@ export class FlowExecutor extends WithBrandConfig(Interface) implements StageHos
         this.submit(payload, options);
     };
 
+    handleFlowUpdate() {
```
GirlBossRush:
```suggestion
    protected handleFlowUpdate() {
```

GirlBossRush (approved):

GirlBossRush (approved):

web/src/elements/directives/light.ts
```diff
@@ -0,0 +1,114 @@
...
+            ...renderOptions,
+        });
+
+        const rendered = this.sentinel.previousSibling;
+        if (rendered instanceof Element) {
+            rendered.slot = this.slotName;
+        }
+
+        if (!this.slot) {
+            this.slot = Object.assign(document.createElement("slot"), {
```
GirlBossRush, line 83:
```suggestion
            this.slot = Object.assign(this.ownerrDocument.createElement("slot"), {
```

web/src/flow/Flow.ts
kensternberg-authentik: The documentation for `window.location.pathname` states that this should _never_ be null.  If it is, this code is not being run in an environment where it should do anything, much less complain.

web/src/flow/Flow.ts
kensternberg-authentik: Accepted!

web/src/flow/Flow.ts
kensternberg-authentik: `nothing`, surely?  It's returned as a value to the template; it needs to be a template-aware value, something Lit recognizes.  The documentation for `guard` states: "valueFn is a function that returns a renderable value."  `nothing` has to be a better choice here.

(The other suggestions are great; in fact, I made `#layoutUsesSidebarFrames` into `private layoutUsesSidebarFrames` so that they can _all_ be destructured at start-up.

web/src/flow/FlowExecutor.ts
kensternberg-authentik: Your headcanon is acceptable to me.

web/src/elements/directives/light.ts
kensternberg-authentik: I don't think this wins us anything?  `document` is the canon source for creating new elements and fragments, but using it creates no obligations with either the lightDOM or a shadow, and if the generated reference goes completely out of scope it gets garbage collected regardless of originating source; if `document` isn't there, you've got bigger problems. https://developer.mozilla.org/en-US/docs/Web/API/Document/createElement

PR #22132, opened 2026-05-07
web: add grid/list view toggle to user application library

### LINKED ISSUE #22126: Add list view toggle to application library for long application names
## Problem

The current application library uses a card/grid layout that truncates long application names. Customers managing apps with naturally long identifiers (e.g. AWS accounts where the display name is `<account-name> (<account-id>)` for searchability) see names cut off, making the library look messy and harder to scan visually — even though the underlying search still works.

> Do you have a better display of long application names in the application library? we have some "apps" with pretty long names (aws accounts where we want to display a name string and the account id for easy searching) but they get cut off pretty quickly and it looks goofy (search still works fine). A list-style view would be amazing!

## Proposal

Add a **grid ↔ list view toggle** to the application library (My Applications page).

- **Toggle control**: a view-switcher button placed near the "My Applications" header.
- **Grid view**: unchanged — existing card layout for users who prefer it.
- **List view**: full application name displayed without truncation, with room for additional metadata (description, group, slug, etc.) on the same row to support search-by-substring use cases like "AWS account ID alongside account name".
- **Persistence**: remember the user's choice (localStorage or user setting) so they don't have to re-toggle on every visit.

## Why this helps

- Long names (AWS accounts, customer-prefixed app names, internal services with environment + region suffixes) become legible without hover-to-see-full-name.
- A denser layout helps users with many apps scan their library faster.
- Grid layout remains the default for users with shorter app names / smaller libraries.

## Notes

There was prior discussion of a list view that was never implemented; this issue tracks finishing that work.

GirlBossRush:
## Summary

Closes #22126.

Adds a grid ↔ list view toggle to the user-facing application library so customers with naturally long application names (e.g. AWS accounts displayed as `<account-name> (<account-id>)`) can scan their apps without name truncation.

- New icon button in the "My Applications" header swaps between **grid** (default) and **list** mode.
- Preference persisted per-browser via `StorageAccessor.local` (`ak-library-view-mode`), mirroring the existing `RememberMe` pattern.
- **List mode** renders each app as a wide row using a CSS subgrid so icon, name+meta, and action menu line up consistently across rows. Full application name displayed without clamp; description, publisher, and slug appear inline as supplementary metadata for substring search use cases (e.g. AWS account ID alongside account name).
- The kebab action menu (`CardMenu`) is shared with the grid card and re-anchored to the kebab's right edge in list mode so the popover stays inside the viewport.

## Screenshots

Screenshot 2026-05-07 at 21 21 59
Screenshot 2026-05-07 at 21 22 15

## Files

- `web/src/user/LibraryPage/ak-library-impl.ts` — toggle button, view-mode state, persistence
- `web/src/user/LibraryPage/ak-library-impl.css` — header layout for the toggle control
- `web/src/user/LibraryPage/ApplicationList.ts` / `.css` — list-mode container styles
- `web/src/user/LibraryPage/LibraryAppRow.ts` — new list-row component
- `web/src/user/LibraryPage/types.ts` — view-mode type
- `blueprints/example/long-named-aws-applications.yaml` — 25-app fixture blueprint


web/src/user/LibraryPage/ak-library-impl.css
```diff
@@ -14,20 +14,44 @@
...
+        flex: 0 0 auto;
+        font-size: var(--pf-global--FontSize--xl);
+        line-height: 1;
+
+        .ak-c-vector-icon {
+            width: 1.5rem;
+            height: 1.5rem;
+        }
+
+        @container (width < 800px) {
```
kensternberg-authentik, line 40: Can we invert this?  "When the width is greater than or equal to?" _Especially_ in the User app, the "mobile first" approach is really going to be valuable going forward.

web/src/user/LibraryPage/ak-library-impl.ts
```diff
@@ -41,6 +45,47 @@ import PFGrid from "@patternfly/patternfly/layouts/Grid/grid.css";
...
+                xmlns="http://www.w3.org/2000/svg"
+                class="ak-c-vector-icon"
+                fill="currentColor"
+                aria-hidden="true"
+                viewBox="0 0 32 32"
+            >
+                <path
+                    d="M10 6h18v2H10zM10 24h18v2H10zM10 15h18v2H10zM4 15h2v2H4zM4 6h2v2H4zM4 24h2v2H4z"
+                />
+            </svg>`,
```
kensternberg-authentik, line 67: [Strokes chin] Hmm, ze storage of ze icons in ze app precludes letting ze user choose zir own icons, which is a project we hope to integrate in ze near future, no?

web/src/user/LibraryPage/ApplicationList.ts
```diff
@@ -75,24 +106,11 @@ export const AKLibraryApplicationList: LitFC<AKLibraryApplicationListProps> = ({
...
-
-                            return AKLibraryApp({
-                                application,
-                                background,
-                                editURL,
-                                targetRef: selected ? targetRef : null,
-                            });
-                        },
-                    )}
+                    ${isList
```
kensternberg-authentik, line 109: Ah, I figured out what was bugging me about this.  You have _two different renderers_ here kinda intertwingled, which is why you have to ask `isList?` twice.  If you separate them out into different units chosen with a ternary, you would probably have an easier time reading this in the future.

kensternberg-authentik (approved):
I made three comments, all of which are about the maintainability of this beast.  None of them are _broken_, so I've approved, but do consider them.

PR #22161, opened 2026-05-08
events: fix `destination_group_obj` not being nullable

gergosimonyi:


authentik/events/api/notification_rules.py
```diff
@@ -11,7 +11,7 @@
 class NotificationRuleSerializer(ModelSerializer):
     """NotificationRule Serializer"""
 
-    destination_group_obj = GroupSerializer(read_only=True, source="destination_group")
+    destination_group_obj = GroupSerializer(read_only=True, source="destination_group", required=False, allow_null=True)
```
BeryJu: should probably also be `PartialGroupSerializer` instead

gergosimonyi: Generally agreed. This PR is specifically to backport though, so I'd keep it here.

BeryJu (approved):

PR #22162, opened 2026-05-08
tenants/settings: present unset flags as `False`

gergosimonyi:


authentik/tenants/api/settings.py
```diff
@@ -20,12 +20,15 @@
...
-        """Exclude any system flags that aren't modifiable"""
         new_value = value.copy()
         for flag in Flag.available(exclude_system=False):
             _flag = flag()
+            # Exclude any system flags that aren't modifiable
             if _flag.visibility == "system":
                 new_value.pop(_flag.key, None)
+            # Explicitly present unset flags as if they were set to False
+            if _flag.key not in value:
+                value[_flag.key] = False
```
BeryJu:
```suggestion
                value[_flag.key] = _flag.default
```

BeryJu: (Flags are currently all boolean but might not always be)

authentik/tenants/api/settings.py
```diff
@@ -20,12 +20,15 @@
...
 
     def to_representation(self, value: dict) -> dict:
-        """Exclude any system flags that aren't modifiable"""
         new_value = value.copy()
         for flag in Flag.available(exclude_system=False):
             _flag = flag()
+            # Exclude any system flags that aren't modifiable
             if _flag.visibility == "system":
                 new_value.pop(_flag.key, None)
+            # Explicitly present unset flags as if they were set to False
```
gergosimonyi:
```suggestion
            # Explicitly present unset flags as if they were set to default
```

BeryJu (approved):

Issue #3565, opened 2022-09-08
`userinfo` endpoint missing CORS headers


dcshoes23:
**Describe the bug**
Web application is calling the `/application/o/userinfo/` endpoint from the browser. 

It is failing with the message in the Chrom Debug Console 
```
Access to fetch at 'https://auth.mydomain.com/application/o/userinfo/' from origin 'https://myapp.mydomain.com' has been blocked by CORS policy: Response to preflight request doesn't pass access control check: No 'Access-Control-Allow-Origin' header is present on the requested resource. If an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled.
```
Analyzing the responce with CURL, I see that it is missing the following headers:
```
access-control-allow-credentials: true
access-control-allow-origin: https://myapp.mydomain.com
```

**To Reproduce**
1. Create an OAuth2 provider and set `Redirect URIs/Origins (RegEx)` to `https://myapp.mydomain.com/.*`
2. App logs in and gets token from `https://auth.mydomain.com/application/o/token/` (response has the correct CORS headers)
3. Make GET request to `https://auth.mydomain.com/application/o/userinfo/` with the bearer token 

**Expected behavior**
Response contains proper CORS headers with the origin specified on the provider
```
access-control-allow-credentials: true
access-control-allow-origin: https://myapp.mydomain.com
```

**Version and Deployment (please complete the following information):**
 - authentik version: 2022.8.2
 - Deployment: Helm

BeryJu:
There are already some CORS headers that are being set on the userinfo endpoint: https://github.com/goauthentik/authentik/blob/main/authentik/providers/oauth2/views/userinfo.py#L96
which allows these things https://github.com/goauthentik/authentik/blob/371f1878a8b3f3e6fad288a6315ed06e4418a0e2/authentik/providers/oauth2/utils.py#L61
so it should work correctly already

dcshoes23:
ya, it's confusing me as well when I look at the code but the request works through curl (copied curl request from Chrom developer tools for the request) I see the user info JSON response and you can see the headers missing:

`/application/o/userinfo/`
```
* Connection state changed (MAX_CONCURRENT_STREAMS == 250)!
< HTTP/2 200
< cache-control: no-store
< content-encoding: deflate
< content-type: application/json
< date: Fri, 09 Sep 2022 12:52:32 GMT
< pragma: no-cache
< referrer-policy: same-origin
[...]

PR #22179, opened 2026-05-09
providers/oauth2: honour RedirectURI matching_mode in CORS origin check

### LINKED ISSUE #3084: OAuth CORS does not take into account RegEx
**Describe the bug**
CORS rules are specified in the same field as redirect URLs. However, in contrast to redirect urls, CORS requests do not seem to use any form of RegEx.
 
https://github.com/goauthentik/authentik/blob/350f0d836580f4411524614f361a76c4f27b8a2d/authentik/providers/oauth2/utils.py#L44
In this line the URL is parsed and shortly afterwards compared directly to any of the passed rules. This function does not take any regex into account.

**To Reproduce**
1. Go to 'Providers'
2. Create an OAuth Provider
3. Specify ".*" in the Redirect URIs field
4. Perform a request to the `application/o/<slug>/.well-known/openid-configuration` endpoint.
5. The CORS headers in the response will be missing.

**Expected behavior**
CORS allowed origins respect the same RegEx rules as redirect URLs.

**Logs**
```
server_1       | {"allowed": [".+", "http://.*"], "event": "CORS: Origin is not an allowed origin", "host": "authentik.redacted.io", "level": "warning", "logger": "authentik.providers.oauth2.utils", "pid": 23, "request_id": "7197b556022548e1849a1814014b355f", "requested": ["http", "localhost:3000", "", "", "", ""], "timestamp": "2022-06-13T17:15:51.891612"}
```

 - authentik version: 2022.6.2
 - Deployment: docker-compose

**Additional context**
I would like to accept CORS requests from a range of subdomains. Possible workarounds are appreciated.

AndyCross:
# providers/oauth2: honour RedirectURI matching_mode in CORS origin check

Closes #3084
Refs #3565

## Details

Authentik's OAuth2 provider supports two `matching_mode` values for redirect URI entries: `STRICT` and `REGEX`. The authorize view honours both correctly when validating the `redirect_uri` query parameter. The CORS origin check in `authentik/providers/oauth2/utils.py::cors_allow`, however, ignored `matching_mode` entirely — it received a flat list of URL strings and compared scheme/host/port equality against the request's `Origin` header.

Symptom: a SPA hosted on a templated origin (Tailscale funnels, preview deployments, per-tenant subdomains, etc.) configured with a `REGEX` redirect URI entry can complete `/authorize/`, but its `fetch()` call to `/application/o/token/` is blocked by the browser because the response has no `Access-Control-Allow-Origin` header. Worker logs show `CORS: Origin is not an allowed origin` with the configured regex appearing verbatim in `allowed`, never matched.

The same issue affects the `/userinfo/` and provider-info endpoints (the latter via `views/provider.py`).

## Change

- `cors_allow` now accepts `RedirectURI` instances directly via `*allowed_origins: RedirectURI | str` and branches on `matching_mode`. Strict entries keep the existing scheme+host+port comparison (regression-safe). Regex entries are matched against the request `Origin` via `re.fullmatch`, mirroring the semantics already used by `views/authorize.py`.
- Malformed regexes raise `re.error`, which is caught, logged via the existing structlog logger (consistent with `views/authorize.py`), and skipped — a single bad entry will not 500 the request.
- The three call sites (`views/token.py`, `views/userinfo.py`, `views/provider.py`) now pass `RedirectURI` objects through to `cors_allow` instead of flattening them to bare URLs.
- Bare string entries are still accepted (treated as STRICT) for backwards compatibility with any external caller of `cors_allow`.
- No model or schema changes — `matching_mode` is already persisted on each `RedirectURI` entry.

## Origin vs. path semantics

The regex is matched against the bare request `Origin` (scheme + host[:port], no path). Existing regex redirect URI entries that include a path will not match a bare Origin out of the box; users have two clean options:

1. **Make the path group optional** (recommended — single entry covers both authorize and CORS):

   ```
   https://app-\w+\.example\.com(/callback)?
   ```

2. **Add a separate origin-only Regex entry** alongside the existing redirect URI entry:

   ```
   https://app-\w+\.example\.com/callback   # for /authorize redirect_uri
   https://app-\w+\.example\.com            # for CORS Origin
   ```

This avoids the brittle alternative of attempting to derive an origin pattern from a regex (regex syntax ≠ URL syntax — character classes, alternations, and escaped slashes make `urlparse`-on-regex unreliable).

## Tests

`authentik/providers/oauth2/tests/test_utils.py` adds focused unit tests on `cors_allow` covering:

- Strict regression (matches scheme+host+port, rejects mismatch).
- Regex match / no-match.
- Regex with optional `(/callback)?` path group matching a bare Origin.
- Regex with mandatory path *not* matching a bare Origin (documented behaviour).
- Malformed regex skipped, doesn't raise; later valid entries still match.
- Lone malformed regex returns response unchanged (no 500).
- OPTIONS preflight passthrough (existing behaviour).
- Bare-string backwards compatibility (treated as STRICT).
- Mixed strict + regex provider configurations.

Existing CORS integration tests in `tests/test_token.py` (`test_refresh_token_view_origin`, `test_refresh_token_view_invalid_origin`) continue to pass unchanged — the strict path is preserved byte-for-byte.

## Reproduction

1. Create an OAuth2 provider.
2. Add a Regex-mode redirect URI: `https://app-\w+\.example\.com(/callback)?`
3. SPA at `https://app-abc123.example.com` completes `/authorize/`, then `fetch()`es `/application/o/token/`.
4. Before this patch: browser blocks the response — no `Access-Control-Allow-Origin` header.
5. After this patch: `Access-Control-Allow-Origin: https://app-abc123.example.com` is set; browser allows the response.


PeshekDotDev:
@AndyCross It looks like there was an error updating this branch with main. Could you please fix your branch so we can review it? There isn't a realistic way for us to see your changes right now

AndyCross:
Hey @PeshekDotDev - you're right, something went wrong before. 

I rebased onto current main — branch is now a single commit (73d702286b) on top of main, no merge conflicts.   

I hope that makes it be reviewable now. Thanks for the patience.

AndyCross:
What should I do here maintainers? Is there any feedback on this PR?

BeryJu (approved):

PR #22194, opened 2026-05-10
Entra ID / OAuth Source Property Mapping - Add a checkbox/toggle to automatically default to UPN if Email is blank for Federation & Social Sources

### LINKED ISSUE #20448: Entra ID / OAuth Source Property Mapping - Add a checkbox/toggle to automatically default to UPN if Email is blank for Federation & Social Sources

I am using Entra ID to authenticate to various OIDC apps. I invited a user who was only an admin, with no mailbox license in entra ID to my apps. The email field was blank and it caused a whole bunch of issues with signing in that were hard to debug before realizing what happened. 

I'd like authentik to have a checkbox or toggle somewhere that could essentially enable the following OAuth source property mapping automatically for that source or specifically entra ID source.

```
return {
    "email": info.get("mail") or info.get("userPrincipalName"),
}
```

It is simple enough to just setup the Oauth source property mapping yourself, but arriving to the decision could be hard for slightly less technical users.

Another alternative would be mentioning this as an example in the documentation explaining that various apps will only work if you have an "email" field populated for your users, and to always populate the email field you can do the following steps. 

1. Create a custom OAuth source property mapping within customizations -> property mapping
2. Include an expression of the following:
```
return {
    "email": info.get("mail") or info.get("userPrincipalName"),
}
```
3. Assign the property mapping to your federation & social login source

### Additional context

Obviously, this is quite an edge case, and may not be aligned with the principles of the project, but I just wanted to submit the idea in case I could help someone out in the future with the same issue.

Cheers!

bkoragan:
## Details

When an Entra ID account has no Microsoft mailbox license (e.g. an admin user invited without a mailbox), Microsoft Graph returns `mail: null` and omits or empties `otherMails`. `EntraIDType.get_base_user_properties` then writes `email = None` onto the resulting authentik user. Downstream this surfaces as silent failures (notification emails fail to send, account-link-by-email cannot match, integrations that key on email return ambiguous results) and the root cause is hard to diagnose.

Extend the fallback chain from `mail` -> `otherMails[0]` to `mail` -> `otherMails[0]` -> `userPrincipalName`. UPN is always populated for Entra ID users, is the canonical sign-in identifier, and is already used as the default `username`, so no extra Graph calls are needed and the value is consistent with what authentik already trusts. Behavior is unchanged for users that have a populated `mail` or `otherMails` - the new branch only fires when both are empty.

While here, also handle `"otherMails": []` (an empty list, distinct from the field being absent) without raising IndexError, by replacing`info.get("otherMails", [None])[0]` with`(info.get("otherMails") or [None])[0]`.

Fixes #20448

Signed-off-by: Bapuji Koraganti <bapuk.2008@gmail.com>

---

## Checklist

-   [x ] Local tests pass (`ak test authentik/`)
-   [x ] The code has been formatted (`make lint-fix`)

If applicable

### DIFF
--- a/authentik/sources/oauth/tests/test_type_entra_id.py
+++ b/authentik/sources/oauth/tests/test_type_entra_id.py
@@ -46,6 +46,30 @@ class TestTypeAzureAD(TestCase):
         self.assertEqual(ak_context["email"], EID_USER["mail"])
         self.assertEqual(ak_context["name"], EID_USER["displayName"])
 
+    def test_enroll_context_other_mails(self):
+        """mail empty, otherMails populated -> use otherMails[0]"""
+        info = {**EID_USER, "mail": None, "otherMails": ["alt@goauthentik.io"]}
+        ak_context = EntraIDType().get_base_user_properties(source=self.source, info=info)
+        self.assertEqual(ak_context["email"], "alt@goauthentik.io")
+
+    def test_enroll_context_other_mails_empty_list(self):
+        """mail empty, otherMails is [] -> fall back to UPN (regression: would IndexError)"""
+        info = {**EID_USER, "mail": None, "otherMails": []}
+        ak_context = EntraIDType().get_base_user_properties(source=self.source, info=info)
+        self.assertEqual(ak_context["email"], EID_USER["userPrincipalName"])
+
+    def test_enroll_context_no_mailbox(self):
+        """No mail / otherMails (license-less account) -> fall back to UPN. Issue #20448."""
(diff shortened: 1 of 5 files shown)


dewi-tik:
The empty/null `otherMails `handling is a useful fix. However, a UPN is not necessarily a deliverable email address, particularly for Entra guests, so I’d prefer making that fallback opt in through a setting or property mapping rather than changing the default behavior.

Also the title of this PR doesn't align with the actual PR.

PR #22241, opened 2026-05-11
web/dependency: move the notifications components into the components folder

kensternberg-authentik:
web/maint: Move notifications into the components folder; adjust imports accordingly

# What

1.  Moves the notifications folder from elements to components: the API and Notifications drawers are API-aware. If we want to separate that out and do something unique, we can, but for now, let’s just get things where they should be.

2.  Adjusts all the imports correctly.

3.  (Minor): Mutating the array and then calling `requestUpdate()`, especially when the array is then sorted-and-reversed, doesn’t save anything over creating a new array with the new item shifted onto the head, sorted once, and then saved to the property, which triggers an update automatically.

-   [🦓] The code has been formatted (`make web`)


GirlBossRush (approved):
logical improvement 👍🏻

PR #22245, opened 2026-05-12
core: harden npm install against supply-chain attacks

GirlBossRush:
## Details

This PR adds a repo-root `.npmrc` that sets `ignore-scripts=true`, neutralizing the dominant npm supply-chain attack vector (malicious lifecycle scripts in transitive dependencies)

- Adds `engine-strict`, `save-exact`, `audit`, and `prefer-offline` to the same `.npmrc` for tighter install hygiene
- Refactors the Makefile to factor a `root-node-install` target out of `node-install`, makes `docs-install` depend on it, and explicitly rebuilds a small audited allowlist (`esbuild`, `chromedriver`, `tree-sitter`, `tree-sitter-json`) after the web install — these are the only dependencies whose install scripts are actually required for the toolchain to function
- Retires the `website/package.json` `preinstall` hook now that the same orchestration lives in `make`
- Documents the new install workflow in `web/README.md`, including the exact `npm rebuild` incantation for anyone who bypasses `make`

## Why

Two practical motivations:

1. **Defense in depth on every install.** Without `ignore-scripts=true`, a single compromised transitive dependency can run arbitrary code on every contributor and CI machine during `npm ci`. The lifecycle-script vector has been used in real attacks (`event-stream`, `ua-parser-js`, the recent `npm-debug` cluster). This change makes that vector a non-default behavior we opt back into for an explicit, audited list.

2. **Predictable orchestration.** The previous `website/preinstall` hook chained `npm ci --prefix ..` to make sure root deps existed before the website install. That worked but was invisible: it fired on every `npm install` in `website/`, didn't respect `--ignore-scripts`, and depended on developers starting from the right directory. Moving the dependency into `make` makes it explicit, composable, and auditable.


.npmrc
```diff
@@ -0,0 +1,22 @@
...
+engine-strict=true
+
+# Pin exact versions so `npm install <pkg>` writes "1.2.3" not "^1.2.3".
+save-exact=true
+
+# Surface CVE warnings during install; doesn't block.
+audit=true
+
+# Suppress funding banners.
+fund=false
```
dominic-r, line 20:
```suggestion
fund=true
```

Honestly... Might as well.

GirlBossRush: While most packages have good intentions, the attack surface is viable for external links and prompt injection.  

https://docs.npmjs.com/cli/v11/commands/npm-fund

https://docs.npmjs.com/cli/v11/configuring-npm/package-json#funding

.npmrc
```diff
@@ -0,0 +1,22 @@
+# Block lifecycle scripts (preinstall/install/postinstall/prepare) from
```
rissson: Let's add this to CODEOWNERS as well

rissson (approved):
a welcomed change!

.npmrc
dominic-r: Oh oops i completely misread it. figured it was disable funding and not enable. Yea better keep this off

Makefile
```diff
@@ -228,9 +228,19 @@ gen-dev-config:  ## Generate a local development config file
 ## Node.js
 #########################
 
-node-install:  ## Install the necessary libraries to build Node.js packages
+# Packages whose install/postinstall scripts are required for correct
+# operation (binary downloads, native bindings). The root .npmrc sets
+# `ignore-scripts=true` to block dependency lifecycle scripts by default;
+# this list is rebuilt explicitly with scripts re-enabled. Audit any
+# additions: each entry runs arbitrary code at install time.
+TRUSTED_INSTALL_SCRIPTS := esbuild chromedriver tree-sitter tree-sitter-json
```
dominic-r, line 236: just out of curiosity is there any risk these packages could be compromised? I assume it doesn't matter at all cause of the cooldown period?

dominic-r (approved):
!!!

.npmrc
kensternberg-authentik: Thanks for the explanation. I was just as confused as Dominic.

Makefile
kensternberg-authentik: The supply chain reliability for these is pretty damned strong, given how much is dependent on them.  Someone would have to compromise them internally, but that's true of any dependency.

web/README.md
```diff
@@ -3,6 +3,27 @@
 This is the default UI for the authentik server. The documentation is going to be a little sparse
 for awhile, but at least let's get started.
 
+# Setup
+
```
kensternberg-authentik, line 7: Should this also go into the docs for "front-end development" and "full development environments"?

kensternberg-authentik (approved):
These changes look good; I like Marc's suggestion, and maybe have a follow-up with docs about the development environment.

PR #22247, opened 2026-05-12
core: add opt-in containerized npm install via Apple container

GirlBossRush:
## Summary

Adds an opt-in `make node-install-containerized` target and a small wrapper script (`scripts/container-sandbox`) that runs the npm install steps inside an ephemeral Linux microVM via Apple's [`container` runtime](https://github.com/apple/container). The container mounts only the repo (at `/work`) and sees nothing else — no `~/.ssh`, `~/.aws`, Keychain, or other host process state.

This is defense in depth *on top of* the `.npmrc` `ignore-scripts=true` baseline introduced in #22245. The default `make node-install` is unchanged.

**Depends on #22245** — once that lands and is rebased, this PR rebases onto `main`.

## Why container, why not sandbox-exec / bubblewrap

- **sandbox-exec** (macOS): Apple has officially deprecated it since 2013. Still functional, but the profile language is sparsely documented and we'd be writing userspace allow/deny rules on a shared kernel. Real but fragile.
- **bubblewrap / systemd-run** (Linux): both are good, but they don't help our macOS contributors — who are also the cohort most likely to have plaintext credentials in `~/.aws` and `~/.ssh` for daily use, since macOS doesn't sandbox CLI tools out of the box.
- **Apple `container`**: real kernel namespacing via `Virtualization.framework`. Per-container Linux microVM. No deprecation risk. The opt-in cost is `brew install container` and accepting that `node_modules` ends up with Linux binaries.

We're shipping container only — no sandbox-exec, no Linux-side wrappers in this PR. If Linux contributors later want a similar opt-in, that's a follow-up.

## Workflow tradeoffs (documented in `web/README.md`)

- **Linux node_modules.** The container is Linux; native binaries (esbuild, chromedriver, tree-sitter, …) install as `linux-arm64`. Contributors who also want to run the dev server natively on macOS need a separate native install. Pick one workflow per session.
- **`brew install container` required.** macOS 15+ on Apple Silicon. Intel Macs and older macOS get the `.npmrc` baseline only.
- **Image is tag-pinned by default.** `node:22-bookworm`. Override with `CONTAINER_IMAGE=node:22-bookworm@sha256:...` for stronger supply-chain guarantees.
- **Network is configurable.** `CONTAINER_NETWORK=none` for hermetic operations (tests, builds). Default leaves network on because install needs the registry.

This is opt-in; nothing in CI changes by default. The wrapper is intentionally minimal — a Bash script around `container run`, not a framework. If we later want to deepen this (per-phase containers, digest-pinned images by default, podman parity for Linux), those are follow-up PRs.

### DIFF
--- a/scripts/container-sandbox
+++ b/scripts/container-sandbox
@@ -1,0 +1,77 @@
+#!/usr/bin/env bash
+#
+# Run a command inside an ephemeral Linux container via Apple's
+# `container` runtime (macOS 15+, Apple Silicon). Each invocation is a
+# fresh container with the repo mounted read-write at /work; nothing
+# outside the repo, the npm cache, or /tmp is visible.
+#
+# Examples:
+#   scripts/container-sandbox npm ci
+#   CONTAINER_NETWORK=none scripts/container-sandbox npm test
+#   CONTAINER_IMAGE=node:22-bookworm scripts/container-sandbox sh -c '...'
+#
+# This is an opt-in defense-in-depth wrapper for macOS contributors. The
+# .npmrc at the repo root is the primary defense; the container adds
+# real kernel-level isolation on top.
+#
+# Install requirements:
+#   brew install container        (macOS 15+, Apple Silicon only)
+#
+# Caveats:
+#   - The container is Linux, not Darwin. node_modules installed inside
+#     will contain linux-arm64 binaries (esbuild, chromedriver, etc.).
+#     If you also want to run the dev server on the host, you'll need a
+#     second native install. Pick one workflow per session.
+#   - Image is pinned to a tag, not a digest, for ergonomics. For
+#     stronger supply-chain guarantees, set CONTAINER_IMAGE to a digest.
+
+set -euo pipefail
+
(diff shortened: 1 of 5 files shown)


scripts/container-sandbox
```diff
@@ -0,0 +1,77 @@
...
+#
+# Run a command inside an ephemeral Linux container via Apple's
+# `container` runtime (macOS 15+, Apple Silicon). Each invocation is a
+# fresh container with the repo mounted read-write at /work; nothing
+# outside the repo, the npm cache, or /tmp is visible.
+#
+# Examples:
+#   scripts/container-sandbox npm ci
+#   CONTAINER_NETWORK=none scripts/container-sandbox npm test
+#   CONTAINER_IMAGE=node:22-bookworm scripts/container-sandbox sh -c '...'
```
dominic-r, line 11:
```suggestion
#   CONTAINER_IMAGE=node:25-trixie scripts/container-sandbox sh -c '...'
```

PR #22336, opened 2026-05-13
website: migrate brand assets to pkg

BeryJu:
analogue to https://github.com/goauthentik/website/pull/588


website/docs/docusaurus.config.esm.mjs
```diff
@@ -33,16 +33,22 @@ const releaseEnvironment = prepareReleaseEnvironment();
 
 //#region Copy static files
 
-const files = [
-    // ---
-    resolve(authentikModulePath, "lifecycle/container/compose.yml"),
-];
+const brandFiles = new Map([
```
dominic-r, line 38: perhaps turn the package name into a constant?

website/static/img/.gitignore
```diff
@@ -0,0 +1,7 @@
+icon.png
```
dominic-r, line 1: perhaps gitignore * and exclude LICENSE instead of hardcoding files, if more icon variants get added or stuff gets renamed

dominic-r (approved):

PR #22340, opened 2026-05-13
core: Defer status posts until GitHub Actions finishes.

GirlBossRush:
## Details

This PR defers status posts until GitHub Actions finishes, otherwise the status flips red while partial coverage uploads trickle in from jobs.


dominic-r (approved):
Wow 🤩

BeryJu (approved):
We used to have this and it never really worked, so if this now fixes issues i'm all for it

PR #22342, opened 2026-05-13
website/integrations: add grommunio

snxRCS:
## Summary

Adds a community integration guide for [grommunio](https://grommunio.com/), an open-source groupware and collaboration platform with Outlook/MAPI compatibility.

grommunio-web ships with built-in Keycloak/OIDC support (`keycloak.json`), but since authentik uses different endpoint paths, a small nginx bridge config and a one-line patch to `class.keycloak.php` are needed to make the integration work. This guide documents the full setup based on a working production deployment.

**What the guide covers:**
- Creating the OAuth2/OIDC provider in authentik (including the `offline_access` scope mapping, which is required for token refresh and causes a redirect loop if missing)
- Exporting the signing key for gromox JWT verification (`bearer_pubkey`)
- The nginx location config that bridges Keycloak-style paths to authentik endpoints
- The `keycloak.json` config file
- The `offline_access` scope fix in `class.keycloak.php`
- Verification steps

**Why `offline_access` matters:**  
grommunio-web's token flow calls `validate_grant()`, which requires a `refresh_token` in the response. Without the `offline_access` scope in both the authentik provider config and the PHP scope string, no `refresh_token` is issued and the user ends up in an endless login redirect loop. This is the most common pitfall and is documented clearly in the guide.

Follows the same format and structure as the existing `mailcow` integration guide.


dewi-tik:
Thanks for the contribution @snxRCS! We really appreciate it.

We'll make some changes and get this merged ASAP.

dominic-r (approved):
Thank you for contributing this integration. I pushed a cleanup pass to align it with our integration docs standards and clarify a few grommunio-specific details.

GirlBossRush (approved):

PR #22343, opened 2026-05-13
core: Ensure npm global bin is available for Corepack

dominic-r:
## Details

`make install` fails during corepack setup on systems where
`npm config get prefix` resolves to a directory whose `bin/` is not
on `PATH` (nix, asdf, volta, manual prefix override). The freshly
installed `corepack` shim is unreachable on the next line.

Fix:

- `Makefile`: prepend `$(npm config get prefix)/bin` to `PATH` for
  every recipe, gated on an `NPM_EXISTS` probe.
- `scripts/node/setup-corepack.mjs`: same adjustment inside the Node
  process so the script also works when invoked directly.
- `scripts/node/utils/commands.mjs`: surface `stdout` / `stderr` from
  failed `exec` calls so the next subprocess failure isn't opaque.

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,6 +21,11 @@ BREW_PKG_CONFIG_PATH :=
 
 CARGO := cargo
 UV := uv
+NPM_PREFIX := $(shell npm config get prefix 2>/dev/null || true)
+
+ifneq ($(NPM_PREFIX),)
+	export PATH := $(NPM_PREFIX)/bin:$(PATH)
+endif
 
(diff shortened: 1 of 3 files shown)


GirlBossRush (approved):

GirlBossRush:
@rissson Seems this bug was found on a nix setup. Please let me know if this PR helps 🙏🏻

dominic-r:
superceeded by pnpm (migration) pr

PR #22373, opened 2026-05-15
stages/captcha: add Cap and JSON verification support

### LINKED ISSUE #22314: Feaure Request: Add tiagozip/cap and other JSON-based CAPTCHA support

I have been trying to use https://github.com/tiagozip/cap with Authentik (Community Edition), but it is not currently supported. While you can provide a custom `jsUrl = https://cdn.jsdelivr.net/npm/@cap.js/widget` and provide your own `apiUrl`, Authentik does not understand how to load or verify the CAPTCHA widget. I get this error:

```txt
The CAPTCHA challenge failed to load.

Could not find a suitable CAPTCHA provider.
```

No error codes seem to be logged when running on a production instance.

I believe that this is an issue due to limited support of newer CAPTCHA protocols that are not based on the currently supported CAPTCHA providers.

For example, `application/json` is used by cap. Currently, Authentik only supports the older default choice of `application/x-www-form-urlencoded`.

https://github.com/goauthentik/authentik/blob/00f0cfe6e40b6f0986099d2b1f119af69b61ab2a/authentik/stages/captcha/stage.py#L36-L83

I propose to add support for `application/json` based CAPTCHA's, while also adding `cap` as a dropdown option for a CAPTCHA stage.

We should be able to do this by:

1. Adding a new dropdown to CaptchaStage, where we can set the Content-Type to either use `application/x-www-form-urlencoded` or `application/json`.
- If custom is selected, a user should be able to select any of these.
- If a specific pre-defined CAPTCHA is selected, then the default Content-Type header should be auto-populated.
2. Use the selected Content-Type option in `verify_captcha_token` in [/authentik/stages/captcha/stage.py](https://github.com/goauthentik/authentik/blob/00f0cfe6e40b6f0986099d2b1f119af69b61ab2a/authentik/stages/captcha/stage.py#L36-L83) instead of defaulting to `application/x-www-form-urlencoded`.
3. (Optional) Create a new tiagozip/cap CAPTCHA controller with default `jsUrl = 'https://cdn.jsdelivr.net/npm/cap-widget'`.
- It is important that we also provide an alert on selection that links to the docs at [trycap.dev](https://trycap.dev/guide/) to help users set up the CAPTCHA server.

An alternative approach to supporting different Content-Type servers can also include creating an adapter that can convert from a supported API to cap (and other CAPTCHA servers) that use `application/json`.

While this could work in theory, the solution is more involved on the user's end and would require an adapter for every custom CAPTCHA server. It would be much simpler to simply support any of the two most common `Content-Type`'s to cover most cases.

### Additional context

Related Links

Authentik:
- CAPTCHA Stage: [/web/src/flow/stages/captcha/CaptchaStage.ts](https://github.com/goauthentik/authentik/blob/71025a83ad6c468a44bb1f8152a92ba778d3409b/web/src/flow/stages/captcha/CaptchaStage.ts#L59)
- CAPTCHA Controller Base Class: [/web/src/flow/stages/captcha/controllers/CaptchaController.ts](https://github.com/goauthentik/authentik/blob/71025a83ad6c468a44bb1f8152a92ba778d3409b/web/src/flow/stages/captcha/controllers/CaptchaController.ts)
- Verify CAPTCHA Logic: [/authentik/stages/captcha/stage.py](https://github.com/goauthentik/authentik/blob/00f0cfe6e40b6f0986099d2b1f119af69b61ab2a/authentik/stages/captcha/stage.py#L36-L83)

cap:
- tiagozip/cap documentation: https://trycap.dev/guide/

dominic-r:
Add a configurable verification request content type so CAPTCHA providers can use either form-encoded or JSON token verification.

Add Cap as a preset and flow controller, including module-script loading, interactive widget handling, generated API/client types, tests, and docs.

image

Closes https://github.com/goauthentik/authentik/issues/22314


web/src/flow/stages/captcha/CaptchaStage.ts
```diff
@@ -12,6 +12,7 @@ import { AKFormErrors, ErrorProp } from "#components/ak-field-errors";
 import { FlowUserDetails } from "#flow/FormStatic";
 import { BaseStage } from "#flow/stages/base";
 import Styles from "#flow/stages/captcha/CaptchaStage.css";
+import { CapController } from "#flow/stages/captcha/controllers/cap";
```
kensternberg-authentik: First observation: I do not know what "Cap" is (I assume it's some kind of Captcha) and nowhere in this, not even in the documentation, do we describe Cap.  I assume most people building websites know what Google Captcha, and Cloudflare's Captcha are, but when we start supporting non-commercial, semi-obscure variants, it would be helpful to be more explicit about what they are.

web/src/admin/stages/captcha/CaptchaStageForm.ts
```diff
@@ -140,20 +151,34 @@ export class CaptchaStageForm extends BaseStageForm<CaptchaStage> {
...
+                      id: "captcha.public-key.description",
+                      desc: "Description for CAPTCHA public key field.",
+                  });
+
         return html`
             <ak-text-input
-                label=${msg("Public Key")}
+                label=${publicKeyLabel}
                 required
                 name="publicKey"
```
kensternberg-authentik: Claude (rightly) points out that this change will result in a URL being the value where the key `publicKey` appears in any blueprints based on Cap.  Maybe we need to change the name of this field?

web/src/admin/stages/captcha/shared.ts
```diff
@@ -108,19 +137,42 @@ export const CAPTCHA_PROVIDERS = {
...
         interactive: false,
         supportsScore: true,
         score: { min: 0.5, max: 1.0 },
     },
 } as const satisfies Record<CaptchaProviderKey, CaptchaProviderPreset>;
 
+export function deriveCapSiteVerifyURL(endpoint: string): string {
+    const normalizedEndpoint = endpoint.endsWith("/") ? endpoint : `${endpoint}/`;
+
+    return new URL("siteverify", normalizedEndpoint).toString();
```
kensternberg-authentik, line 187: If you give this a bad URL, it's [going to throw an exception](https://developer.mozilla.org/en-US/docs/Web/API/URL#constructor).  What happens?  Is it caught and handled?

web/src/admin/stages/captcha/shared.ts
```diff
@@ -108,19 +137,42 @@ export const CAPTCHA_PROVIDERS = {
             }),
         keyURL: "https://dash.cloudflare.com",
     },
+    cap: {
+        formatDisplayName: () =>
+            msg("Cap", {
+                id: "captcha.providers.cap",
+            }),
+        jsUrl: "https://cdn.jsdelivr.net/npm/cap-widget",
```
kensternberg-authentik: Consider adding a resource integrity marker and CAP-widget versioning to this URL. While it's unlikely that, should this endpoint be compromised, anything worse than "Sorry, you can't log in" will happen, it's still worth our while to ensure our customers don't get bitten by silly security neglecfulness.

https://developer.mozilla.org/en-US/docs/Web/Security/Defenses/Subresource_Integrity

kensternberg-authentik (approved):
I made comments for improvment; nothing here looks _broken_, but the documentation could be better and the semantic overload of `publicKey` where you (sometimes) put a URL is one of those things that can make looking at the raw database or at a blueprint confusing.  There is one place where the user could type in a bad URL and I don't see that handled.  That shouldn't stop deployment, but it should be addressed.

web/src/flow/stages/captcha/CaptchaStage.ts
grivera64: I can give some insight since I opened an issue to request for this feature in #22314.

Cap is a small, self-hostable CAPTCHA server. It relies on using background Proof-of-Work challenges to slow down bot traffic. I don't believe they have their own hosted instance, but you can self-host it easily using docker.

They have a GitHub repo at https://github.com/tiagozip/cap and host their documentation at https://trycap.dev/guide/. We can probably link to these to help users interested in using Cap set up a server and connect it to Authentik.

website/docs/add-secure-apps/flows-stages/stages/captcha/index.md
```diff
@@ -56,6 +56,23 @@ Recommended values:
...
+Cap is a self-hostable CAPTCHA server that uses proof-of-work challenges.
+
+See https://trycap.dev/guide/.
+
+Recommended values:
+
+- **Public key**: public Cap endpoint for the site key path, for example `https://cap.example.com/site-key/`
+- **Private key**: Cap secret key
+- **Interactive**: enabled
+- **JS URL**: `https://cdn.jsdelivr.net/npm/cap-widget`
```
grivera64: We should add the recommendation to use a versioned JS URL, or a self-hosted version that cap's server exposes, so that people are aware about potential supply chain attacks that could result from always using the latest version of npm packages (since this has been common over the last few months).

dominic-r: Yep! Mentioned by Ken, but i've been thinking about a good way. https://github.com/goauthentik/authentik/pull/22373#discussion_r3250375779 Any ideas on a way to keep the versioned widget URL up to date without having to manually update it each time in the docs/code?

grivera64: I think it would be a good idea to either:
- Use a placeholder url like `https://cdn.jsdelivr.net/npm/cap-widget@version` and provide a link to find the latest on GitHub
- Use a self hosted link instead `https://captcha.example/assets/widget.js`


We could also just link to the GitHub released for cap. This way we can avoid updating the docs to point to the newest widget URL in the case that it was the target of a supply chain attack.

https://github.com/tiagozip/cap/releases

The alternative would be to create a release script that updates the tag to the latest capjs version, but that still risks linking a version that has been compromised. A local version probably is the safest, since you have to host the API locally anyway to use cap.

dominic-r: hmmyea I'll do that. Thanks for bringing it up again

grivera64 (approved):
LGTM 🚀 I think we could also include a comment in the docs about only supporting the default widget and not the floating version of the widget (https://trycap.dev/guide/floating.html).

To keep consistent with other captcha providers, I think we can skip supporting the floating widget.

gergosimonyi (approved):

PR #22376, opened 2026-05-15
web/table: fetch on first render when already visible

GirlBossRush:
## Summary

Tables inside `<ak-modal>` rendered empty until the user clicked the refresh button — surfaced by a beta tester on the 2026.5 RC against the RAC endpoint launcher in the user library.

## Root cause

The native-`<dialog>` modal migration (#21336) taught `AKModal.updated()` to force `visible = true` on its slotted child. `Table.firstUpdated()` was delegating to `#synchronizeRefreshSchedule()`, which only flushes a *previously deferred* refresh. With visibility forced on before the first update cycle, no deferred refresh was ever queued — so the synchronizer no-op'd and nothing fetched.

## Fix

Call `fetch()` directly from `Table.firstUpdated()`. `fetch()` already handles both states correctly:

- table visible → issue the request immediately.
- table not yet visible → queue the deferred refresh that the synchronizer flushes on visibility flip.

The change therefore also covers any future caller that mounts a `Table` already-visible, not just the modal path.

```diff
 protected override firstUpdated(changedProperties: PropertyValues<this>): void {
     super.firstUpdated(changedProperties);
-    this.#synchronizeRefreshSchedule();
+    this.fetch();
 }
```

## Test plan

- [x] Manually reproduced against the user-library RAC endpoint launcher in the playpen dev stack: modal opens empty on `main`, populates without a manual refresh with this change.
- [x] Playwright e2e regression test added at `web/test/browser/rac-launch-modal.test.ts` — seeds a RAC provider + two endpoints via the API, opens the launcher modal, asserts the endpoint rows appear without any user interaction. Fails on `main`, passes with this change.
- [x] `make lint-fix` clean.

## Disclosure

Triaged and authored end-to-end by an agent running in a sandbox against `goauthentik/authentik@main`. Reviewed and shipped by @GirlBossRush.


web/src/admin/providers/rac/EndpointForm.ts
```diff
@@ -73,38 +71,28 @@ export class EndpointForm extends ModelForm<Endpoint, string> {
                 ?autofocus=${!this.instance}
             >
             </ak-text-input>
-            <ak-form-element-horizontal required name="protocol">
-                ${AKLabel(
+
+            <ak-radio-input
```
GirlBossRush, line 75: ARIA fixes for the e2e test.

web/src/user/LibraryPage/ak-library-impl.ts
```diff
@@ -174,7 +174,6 @@ export class LibraryPage extends WithSession(AKElement) {
         includeScore: true,
         shouldSort: true,
         ignoreFieldNorm: true,
-        useExtendedSearch: true,
```
GirlBossRush, line 177: AFAICT this is a vestigial setting and not something we've intended to support in this context. Disabling it here because programmatically generated application names can include search control characters that prevent results from matching.

web/e2e/fixtures/FormFixture.ts
```diff
@@ -22,8 +22,14 @@ export class FormFixture extends PageFixture {
...
         context: LocatorContext = this.page,
     ) => {
+        // The `ak-text-input` / `ak-number-input` wrappers set role="presentation"
+        // *and* aria-labelledby on themselves, so getByLabel matches the wrapper
+        // in addition to the inner <input>.
+        // .filter({ hasNot: ... }) only checks descendants, not self.
+        // So we exclude self-presentation explicitly.
         const control = context
             .getByLabel(fieldName, { exact: true })
+            .and(context.locator(':not([role="presentation"])'))
```
kensternberg-authentik: This feels like a _hack_.

web/src/elements/table/Table.ts
```diff
@@ -405,7 +405,12 @@ export abstract class Table<T extends object, D = T>
 
     protected override firstUpdated(changedProperties: PropertyValues<this>): void {
         super.firstUpdated(changedProperties);
-        this.#synchronizeRefreshSchedule();
+
+        // Use `fetch()` rather than `#synchronizeRefreshSchedule()` here: the
+        // latter only flushes a *previously deferred* refresh and would no-op
+        // when a parent (e.g. `AKModal`) has already forced `visible = true`
+        // before the first update cycle, leaving the table empty on open.
+        this.fetch();
```
kensternberg-authentik: Not `connectedCallback()` ?

I'm surprised Lit doesn't whine about a render being scheduled from an `updated()` command. It usually does. You can work around it by microtasking the fetch, but this still doesn't feel like the place to do that.

kensternberg-authentik (approved):

PR #22410, opened 2026-05-17
website/integrations: update nextcloud username mapping

JATippit:
Map `nextcloud_user_id` to Authentik username.

Mapping to UUID in the current state creates a new user where the Nextcloud Display name is the Authentik UUID.  The proposed change correctly links accounts if the Authentik username matches an existing Nextcloud username.


website/integrations/chat-communication-collaboration/nextcloud/index.mdx
```diff
@@ -98,7 +98,7 @@ If you want to control user storage and designate Nextcloud administrators, you
                 # Set a quota by using the "nextcloud_quota" property in the user's attributes
                 "quota": user.group_attributes().get("nextcloud_quota", None),
                 # To connect an existing Nextcloud user, set "nextcloud_user_id" to the Nextcloud username.
-                "user_id": user.attributes.get("nextcloud_user_id", str(user.uuid)),
+                "user_id": user.attributes.get("nextcloud_user_id", str(user.username.lower())),
```
tanberry: Thanks so much for this contribution @JATippit . How does this work if the username attribute is not set? Can we use an `or` to cover both situations?

JATippit: I'm unsure if the username is unset.  However, the info section further down states that this must match the Nextcloud Display Name, which UUID will most likely not do for existing users.

My setup had a user account in Authentik and a local Nextcloud account I wanted to link.  Using UUID, it created a new user with the UUID as the Display Name/account name instead of linking to the existing user in Nextcloud.

Username correctly linked the accounts.

I then tried with a different Authentik account (with no matching Nextcloud user) using username, and that also correctly created the new account.

JATippit: I double checked and could not make a user without a username in Authentik.

dominic-r (approved):
Thanks!

dewi-tik (approved):

PR #22455, opened 2026-05-19
providers/saml: handle XML declarations in unified endpoint

dominic-r:
## Summary
- Normalize decoded redirect-binding SAML requests to bytes before the unified endpoint parses them with lxml.
- Add regression coverage for redirect-binding AuthnRequest and LogoutRequest payloads that include XML declarations.

## RCA
The unified SAML endpoint decodes redirect-binding `SAMLRequest` payloads with `decode_base64_and_inflate()`, which returns a Unicode string. Some service providers include an XML declaration with an encoding, for example `<?xml version="1.0" encoding="UTF-8"?>`. Passing that Unicode string directly to `defusedxml.lxml.fromstring()` raises before authentik can identify the message type. The broad exception handler then reports `Unsupported SAML message type: unknown` instead of routing the request to SSO or SLO handling.

Encoding the decoded XML back to bytes before parsing lets lxml consume the declaration correctly and keeps the existing SSO/SLO delegation behavior intact.

## Context
Slack: https://authentiksecurity.slack.com/archives/D09DHA650QH/p1779152665243569

## Validation
- `.venv/bin/python -m pytest authentik/providers/saml/tests/test_views_unified.py`
- `uv run ruff check authentik/providers/saml/views/unified.py authentik/providers/saml/tests/test_views_unified.py`


authentik/providers/saml/tests/test_views_unified.py
```diff
@@ -0,0 +1,49 @@
...
+)
+
+
+class TestDetectSAMLMessageType(SimpleTestCase):
+    """Test SAML request type detection."""
+
+    def test_redirect_authn_request_with_xml_declaration(self):
+        """Detect redirect-binding AuthnRequest with an XML declaration."""
+        request = deflate_and_base64_encode(
+            '<?xml version="1.0" encoding="UTF-8"?>'
```
PeshekDotDev: For the saml requests, can you make them via the processor or make these ones fixtures?

PeshekDotDev (approved):

rissson (approved):

PR #22462, opened 2026-05-19
root: bind-mount .npmrc into Dockerfile npm ci stages

GirlBossRush:
## Summary

Bind-mounts the repo `.npmrc` into the two Dockerfiles that run `npm ci` during image builds:

- `lifecycle/container/Dockerfile` (the web build stage)
- `website/Dockerfile` (the docs build stage)

## Why

`npm` walks up from cwd looking for `.npmrc`. Both Dockerfiles bind-mount `package.json` / `package-lock.json` for the install step, but neither mounts `.npmrc`. As a result the project-level settings are not present inside the build — most importantly:

- `ignore-scripts=true` — without this, a malicious dep's `preinstall`/`postinstall` payload would execute *inside the image build*, which is far worse than executing in CI (it would land in the published artifact)
- `engine-strict=true`
- `audit=true`

Local dev and our existing CI workflows already honor the project `.npmrc` because they invoke `npm ci` from inside the working tree. This PR closes the same gap for the image-build path.

The mount is `type=bind` (read-only by default) and only present during the `npm ci` step, so it adds nothing to the resulting image layers.

## Test plan

- [ ] `make docker` (or the equivalent buildx command for `lifecycle/container/Dockerfile`) succeeds and produces a working web bundle.
- [ ] The website Docker build produces a working docs image (`make website-build` / equivalent).
- [ ] Image size unchanged — no layer impact from the new mount.

### DIFF
--- a/lifecycle/container/Dockerfile
+++ b/lifecycle/container/Dockerfile
@@ -24,7 +24,8 @@ WORKDIR /work/web
 COPY ./packages /work/packages
 COPY ./web/packages /work/web/packages
 
-RUN --mount=type=bind,target=/work/web/package.json,src=./web/package.json \
+RUN --mount=type=bind,target=/work/.npmrc,src=./.npmrc \
+    --mount=type=bind,target=/work/web/package.json,src=./web/package.json \
     --mount=type=bind,target=/work/web/package-lock.json,src=./web/package-lock.json \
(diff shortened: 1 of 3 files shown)


rissson (changes requested):
This should also be done in `lifecycle/container/proxy.Dockerfile`

rissson (approved):

PR #22523, opened 2026-05-21
web: adopt @goauthentik/theme tokens in the app styles

GirlBossRush:
Adopts the `@goauthentik/theme` tokens in the application styles.

Stacked on #23341 — the base is `slice/theme-package` so the diff here shows only this change. GitHub will retarget to `main` when #23341 merges.

## What this changes

- `web/src/styles/global/theme/token-bridge.css` bridges the package's semantic tokens onto the existing PatternFly 4 variables. Each PF variable falls back to its own prior value where a token is absent.
- `variables.css` imports `@goauthentik/theme/index.css` and the bridge, and drops the hand-rolled `--ak-v2-global--*` block.
- `shadow/patternfly-base.css` imports the bridge too, so per-shadow-root sheets stay in sync with the document cascade.
- `ak-drawer` moves onto the semantic tokens. It was the only consumer of `--ak-v2-global--*`, which is why that block goes in the same commit.

## Rebase notes

The branch predated the styles restructure, so a few things moved rather than merged:

- `styles/authentik/base/` → `styles/global/theme/`, and `styles/patternfly/base.css` → `styles/shadow/patternfly-base.css`.
- The cascade-layer reservation the branch added to the global entrypoints (`@layer reset, base, tokens, components, utilities, overrides`) is **dropped**. Layers shipped for real in #23366 as `@layer reset, vendor, components, theme, mode, brand`; re-declaring a second, different order in the same stylesheet would merge the two and disturb the shipped cascade.

## Value drift

Of the 49 bridged PatternFly variables, 38 are byte-identical and 8 more are the same color expressed differently — main stores the palette as `oklab()`, the theme emits `oklch()` with a hex comment. Converting them agrees exactly (`oklab(0.522 -0.0434 -0.1717)` = `#0066cc` = `--ak-color-primary`, and so on for text, muted text, surface, info, success, warning, danger).

Three are genuine changes:

| Variable | Was | Now |
| --- | --- | --- |
| `--pf-global--BorderRadius--lg` | `30em` | `30rem` |
| `--pf-global--BoxShadow--xl` | `0 1rem 2rem 0 rgba(3,3,3,.16), …` | `0 .75rem 1.5rem 0 rgba(3,3,3,.2), …` |
| dark surface (drawer) | `#18191a` | `#121212` |

## Worth a careful look in review

In dark mode the bridge is layered differently in the two cascade contexts, and the results diverge:

- **Document**: `variables.css` is imported into `layer(theme)`, and `mode.css` into `layer(mode)`, which sorts after it. So `mode.css` still wins and `--pf-global--BackgroundColor--100` stays `#1b1d21`.
- **Shadow roots**: the bridge is adopted unlayered on `:host`, so it wins there and the same variable resolves to `--ak-color-surface`, i.e. `#121212`.

That is a visible dark-mode shift inside every PF-backed component, and it is worth confirming in a browser. If the intent is parity, the theme's dark `color.surface` wants to be `#1b1d21`. Note the token's own source comment says surfaces are "pinned near PatternFly 4's `BackgroundColor--100` (#151515) and the legacy drawer surface (#18191a)" — `#121212` is darker than both, which suggests drift rather than intent.

Co-Authored-By: Ken Sternberg <ken@goauthentik.io>

### DIFF
--- a/web/src/elements/ak-drawer/ak-drawer.root.css
+++ b/web/src/elements/ak-drawer/ak-drawer.root.css
@@ -1,14 +1,26 @@
 /* ----------- CSS Custom Properties for DRAWER  --------------------------- */
 
 :root {
+    /* Component-local directional shadows for the panel's expanded state.
+     * Defined here so the drawer carries its own shadow values rather than
+     * exposing four directional shadow tokens on the public --ak-* surface. */
+    --_drawer-shadow-top: 0 -0.75rem 0.75rem -0.5rem rgba(3, 3, 3, 0.18);
+    --_drawer-shadow-right: 0.75rem 0 0.75rem -0.5rem rgba(3, 3, 3, 0.18);
+    --_drawer-shadow-bottom: 0 0.75rem 0.75rem -0.5rem rgba(3, 3, 3, 0.18);
+    --_drawer-shadow-left: -0.75rem 0 0.75rem -0.5rem rgba(3, 3, 3, 0.18);
+
+    /* Flips translateX() offsets in the RTL-scoped rules of
+     * ak-drawer.styles.ts; only those rules consume it. */
+    --_drawer-inverse-multiplier: -1;
+
     --ak-v2-c-drawer__content--FlexBasis: 100%;
-    --ak-v2-c-drawer__content--BackgroundColor: var(--ak-v2-global--ContentSurface);
-    --ak-v2-c-drawer__content--ZIndex: var(--ak-v2-global--ZIndex--xs, auto);
+    --ak-v2-c-drawer__content--BackgroundColor: var(--ak-color-surface);
+    --ak-v2-c-drawer__content--ZIndex: 100;
     --ak-v2-c-drawer__panel--MinWidth: 50%;
     --ak-v2-c-drawer__panel--MaxHeight: auto;
-    --ak-v2-c-drawer__panel--ZIndex: var(--ak-v2-global--ZIndex--sm);
-    --ak-v2-c-drawer__panel--BackgroundColor: var(--ak-v2-global--ContentSurface);
-    --ak-v2-c-drawer__panel--TransitionDuration: var(--ak-v2-global--TransitionDuration);
+    --ak-v2-c-drawer__panel--ZIndex: 200;
+    --ak-v2-c-drawer__panel--BackgroundColor: var(--ak-color-surface);
+    --ak-v2-c-drawer__panel--TransitionDuration: var(--ak-duration-normal);
     --ak-v2-c-drawer__panel--TransitionProperty: margin, transform, box-shadow, flex-basis;
     --ak-v2-c-drawer__panel--FlexBasis: 100%;
     --ak-v2-c-drawer__panel--md--FlexBasis--min: 1.5rem;
@@ -29,14 +41,14 @@
     --ak-v2-c-drawer--m-panel-bottom__panel--m-resizable--MinHeight: 1.5rem;
     --ak-v2-c-drawer__splitter--Height: 0.5625rem;
     --ak-v2-c-drawer__splitter--Width: 100%;
-    --ak-v2-c-drawer__splitter--BackgroundColor: var(--ak-v2-global--ContentSurface);
+    --ak-v2-c-drawer__splitter--BackgroundColor: var(--ak-color-surface);
     --ak-v2-c-drawer__splitter--Cursor: row-resize;
     --ak-v2-c-drawer__splitter--m-vertical--Height: 100%;
     --ak-v2-c-drawer__splitter--m-vertical--Width: 0.5625rem;
     --ak-v2-c-drawer__splitter--m-vertical--Cursor: col-resize;
     --ak-v2-c-drawer--m-inline__splitter--focus--OutlineOffset: -0.0625rem;
-    --ak-v2-c-drawer__splitter--after--BorderColor: var(--ak-v2-global--BorderColor--100);
-    --ak-v2-c-drawer__splitter--after--border-width--base: var(--ak-v2-global--BorderWidth--sm);
+    --ak-v2-c-drawer__splitter--after--BorderColor: var(--ak-color-border);
+    --ak-v2-c-drawer__splitter--after--border-width--base: var(--ak-border-width-sm);
     --ak-v2-c-drawer__splitter--after--BorderTopWidth: 0;
     --ak-v2-c-drawer__splitter--after--BorderRightWidth: var(
         --ak-v2-c-drawer__splitter--after--border-width--base
@@ -70,26 +82,20 @@
     --ak-v2-c-drawer--m-panel-bottom__splitter-handle--Top: calc(
         50% - var(--ak-v2-c-drawer__splitter--after--border-width--base)
     );
-    --ak-v2-c-drawer__splitter-handle--after--BorderColor: var(--ak-v2-global--Color--200);
-    --ak-v2-c-drawer__splitter-handle--after--BorderTopWidth: var(--ak-v2-global--BorderWidth--sm);
+    --ak-v2-c-drawer__splitter-handle--after--BorderColor: var(--ak-color-text-muted);
+    --ak-v2-c-drawer__splitter-handle--after--BorderTopWidth: var(--ak-border-width-sm);
     --ak-v2-c-drawer__splitter-handle--after--BorderRightWidth: 0;
-    --ak-v2-c-drawer__splitter-handle--after--BorderBottomWidth: var(
-        --ak-v2-global--BorderWidth--sm
-    );
+    --ak-v2-c-drawer__splitter-handle--after--BorderBottomWidth: var(--ak-border-width-sm);
     --ak-v2-c-drawer__splitter-handle--after--BorderLeftWidth: 0;
-    --ak-v2-c-drawer__splitter--hover__splitter-handle--after--BorderColor: var(
-        --ak-v2-global--Color--100
-    );
-    --ak-v2-c-drawer__splitter--focus__splitter-handle--after--BorderColor: var(
-        --ak-v2-global--Color--100
-    );
+    --ak-v2-c-drawer__splitter--hover__splitter-handle--after--BorderColor: var(--ak-color-text);
+    --ak-v2-c-drawer__splitter--focus__splitter-handle--after--BorderColor: var(--ak-color-text);
     --ak-v2-c-drawer__splitter--m-vertical__splitter-handle--after--BorderTopWidth: 0;
     --ak-v2-c-drawer__splitter--m-vertical__splitter-handle--after--BorderRightWidth: var(
-        --ak-v2-global--BorderWidth--sm
+        --ak-border-width-sm
     );
     --ak-v2-c-drawer__splitter--m-vertical__splitter-handle--after--BorderBottomWidth: 0;
     --ak-v2-c-drawer__splitter--m-vertical__splitter-handle--after--BorderLeftWidth: var(
-        --ak-v2-global--BorderWidth--sm
+        --ak-border-width-sm
     );
     --ak-v2-c-drawer__splitter-handle--after--Width: 0.75rem;
     --ak-v2-c-drawer__splitter-handle--after--Height: 0.25rem;
@@ -105,25 +111,19 @@
 
 :root {
     --ak-v2-c-drawer__panel--BoxShadow: none;
-    --ak-v2-c-drawer--m-expanded--m-panel-bottom__panel--BoxShadow: var(
-        --ak-v2-global--BoxShadow--lg-top
-    );
-    --ak-v2-c-drawer--m-expanded__panel--BoxShadow: var(--ak-v2-global--BoxShadow--lg-left);
+    --ak-v2-c-drawer--m-expanded--m-panel-bottom__panel--BoxShadow: var(--_drawer-shadow-top);
+    --ak-v2-c-drawer--m-expanded__panel--BoxShadow: var(--_drawer-shadow-left);
 }
 
 :root {
-    --ak-v2-c-drawer--m-expanded--m-panel-left__panel--BoxShadow: var(
-        --ak-v2-global--BoxShadow--lg-right
-    );
+    --ak-v2-c-drawer--m-expanded--m-panel-left__panel--BoxShadow: var(--_drawer-shadow-right);
 }
 
 :root {
-    --ak-v2-c-drawer__panel--after--Width: var(--ak-v2-global--BorderWidth--sm);
-    --ak-v2-c-drawer--m-panel-bottom__panel--after--Height: var(--ak-v2-global--BorderWidth--sm);
+    --ak-v2-c-drawer__panel--after--Width: var(--ak-border-width-sm);
+    --ak-v2-c-drawer--m-panel-bottom__panel--after--Height: var(--ak-border-width-sm);
     --ak-v2-c-drawer__panel--after--BackgroundColor: transparent;
-    --ak-v2-c-drawer--m-inline--m-expanded__panel--after--BackgroundColor: var(
-        --ak-v2-global--BorderColor--100
-    );
+    --ak-v2-c-drawer--m-inline--m-expanded__panel--after--BackgroundColor: var(--ak-color-border);
     --ak-v2-c-drawer--m-inline__panel--PaddingLeft: var(--ak-v2-c-drawer__panel--after--Width);
     --ak-v2-c-drawer--m-panel-left--m-inline__panel--PaddingRight: var(
         --ak-v2-c-drawer__panel--after--Width
@@ -136,6 +136,12 @@
 html[data-theme="dark"],
 .ak-t-dark,
 .pf-t-dark {
-    --ak-v2-c-drawer__panel--BackgroundColor: var(--ak-v2-global--ContentSurface);
+    --ak-v2-c-drawer__panel--BackgroundColor: var(--ak-color-surface);
     --ak-v2-c-drawer__splitter--BackgroundColor: transparent;
+
+    /* Dark theme: shadows need stronger opacity to read against dark surfaces. */
+    --_drawer-shadow-top: 0 -0.75rem 0.75rem -0.5rem rgba(3, 3, 3, 0.72);
+    --_drawer-shadow-right: 0.75rem 0 0.75rem -0.5rem rgba(3, 3, 3, 0.72);
+    --_drawer-shadow-bottom: 0 0.75rem 0.75rem -0.5rem rgba(3, 3, 3, 0.72);
+    --_drawer-shadow-left: -0.75rem 0 0.75rem -0.5rem rgba(3, 3, 3, 0.72);
 }
--- a/web/src/styles/global/theme/token-bridge.css
+++ b/web/src/styles/global/theme/token-bridge.css
@@ -1,0 +1,114 @@
+/*
+ * Runtime bridge from authentik's public semantic tokens to existing
+ * PatternFly 4 variables. This keeps old component CSS working while new
+ * custom CSS can target the shorter --ak-* surface.
+ *
+ * Each PF variable falls back to a literal copy of the light-theme default
+ * it overrides when the matching --ak-* token is absent, so the bridge
+ * overrides PatternFly only where a token is defined and otherwise preserves
+ * PF's defaults. The literals are spelled out because a var() self-reference
+ * (`--x: var(--token, var(--x))`) is a dependency cycle and computes to
+ * invalid rather than falling through. The font families are the exception:
+ * their defaults are themselves var() references into ./fonts.css, so they
+ * fall back to the same references. (The theme is always loaded in
+ * authentik, so the fallback is a safety net, not the common path.)
+ */
+
+:root,
+:host {
+    /* Legacy authentik bridge */
+    --ak-accent: var(--ak-color-accent, #fd4b2d);
+
+    /* PatternFly 4 compatibility bridge */
+    --pf-global--primary-color--100: var(--ak-color-primary, oklab(0.522 -0.0434 -0.1717));
+    --pf-global--primary-color--200: var(--ak-color-primary-hover, oklab(0.3763 -0.0324 -0.1182));
+    --pf-global--link--Color: var(--ak-color-link, oklab(0.522 -0.0434 -0.1717));
+    --pf-global--link--Color--hover: var(--ak-color-link-hover, oklab(0.3763 -0.0324 -0.1182));
+    --pf-global--link--Color--visited: var(--ak-color-link-visited, oklab(0.3679 0.0535 -0.1797));
+    --pf-global--Color--100: var(--ak-color-text, oklab(0.1957 -0 0));
+    --pf-global--Color--200: var(--ak-color-text-muted, oklab(0.5364 -0.0026 -0.0089));
+    --pf-global--BackgroundColor--100: var(--ak-color-surface, oklab(1 -0 0));
+    --pf-global--BackgroundColor--150: var(--ak-color-surface-raised, oklab(0.9851 -0 0));
+    --pf-global--BackgroundColor--200: var(--ak-color-surface-muted, oklab(0.9551 -0 0));
+    --pf-global--BorderColor--100: var(--ak-color-border, #d2d2d2);
+    --pf-global--BorderColor--200: var(--ak-color-border-strong, #8a8d90);
+    --pf-global--info-color--100: var(--ak-color-info, oklab(0.6689 -0.0608 -0.1513));
+    --pf-global--success-color--100: var(--ak-color-success, oklab(0.5549 -0.1071 0.0852));
+    --pf-global--warning-color--100: var(--ak-color-warning, oklab(0.7864 0.0298 0.1604));
+    --pf-global--danger-color--100: var(--ak-color-danger, oklab(0.5331 0.1798 0.1034));
+    --pf-global--FontFamily--sans-serif: var(
+        --ak-font-family-body,
+        var(--ak-font-family-sans-serif)
+    );
+    --pf-global--FontFamily--heading--sans-serif: var(
+        --ak-font-family-heading,
+        var(--ak-generic-display)
+    );
+    --pf-global--FontFamily--monospace: var(--ak-font-family-code, var(--ak-font-family-monospace));
+    --pf-global--FontSize--xs: var(--ak-font-size-xs, 0.75rem);
+    --pf-global--FontSize--sm: var(--ak-font-size-sm, 0.875rem);
+    --pf-global--FontSize--md: var(--ak-font-size-md, 1rem);
+    --pf-global--FontSize--lg: var(--ak-font-size-lg, 1.125rem);
+    --pf-global--FontSize--xl: var(--ak-font-size-xl, 1.25rem);
+    --pf-global--FontSize--2xl: var(--ak-font-size-2xl, 1.5rem);
+    --pf-global--FontSize--3xl: var(--ak-font-size-3xl, 1.75rem);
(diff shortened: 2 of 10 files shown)


website/docs/developer-docs/frontend/cascade-layers.md
```diff
@@ -0,0 +1,115 @@
...
+1. **Document cascade.** Driven by `interface.global.css`, `flows.global.css`, `static.global.css`. Loads PatternFly, fonts, tokens, globals, locale CSS.
+2. **Per-shadow-root cascade.** Each Lit element adopts `PFBase`, `AKBase`, its own component CSS, and brand custom CSS via `Base.ts`.
+
+These cascades are independent. Layers in one have no effect in the other. We can adopt layers in one context without disturbing the other.
+
+## Proposed migration
+
+### Step 0 — Reserve the layer order
+
+Add a single `@layer` declaration at the top of the document entrypoints and of the per-shadow-root base sheets. **No rule moves into a layer yet.** This is a zero-behavior change that lets downstream PRs reference a stable order.
```
dewi-tik:
```suggestion
Add a single `@layer` declaration at the top of the document entrypoints and at the top of the per-shadow-root base sheets. **No rule moves into a layer yet.** This is a zero-behavior change that lets downstream PRs reference a stable order.
```

website/docs/sys-mgmt/brands/design-language.mdx
```diff
@@ -0,0 +1,115 @@
...
+
+This page describes the **what** and **why** of the design language. For practical "how do I change X" recipes, see [Custom CSS](./custom-css.mdx).
+
+:::warning Advanced topic
+
+Custom CSS is an advanced feature. Improper use may lead to unexpected behavior or visual issues. Test changes in a safe environment before applying them to production.
+
+:::
+
+## How the design system is organised
```
dewi-tik:
```suggestion
## How the design system is organized
```

website/docs/sys-mgmt/brands/design-language.mdx
```diff
@@ -0,0 +1,115 @@
...
+| -------------------------- | ------------------------------------------------------------- | ------------------------------------------------------------------------------- | ----------------------------------------- |
+| **Semantic tokens**        | `--ak-color-primary`, `--ak-space-md`, `--ak-shadow-lg`       | The shared design decisions every component reads. Affects the whole interface. | Brand custom CSS, themes.                 |
+| **Component-local tokens** | `--ak-c-login__main--BackgroundColor`, drawer width variables | One component's tunable knobs.                                                  | When a single component needs adjustment. |
+| **CSS parts**              | `ak-flow-executor::part(locale-select)`                       | Structural hooks a component intentionally exposes.                             | Targeted layout or visibility changes.    |
+
+Most brand customisation stays at the first layer. The lower layers are escape hatches.
```
dewi-tik:
```suggestion
Most brand customization stays at the first layer. The lower layers are escape hatches.
```

website/docs/sys-mgmt/brands/design-language.mdx
```diff
@@ -0,0 +1,115 @@
...
+}
+```
+
+A third theme — `reduced` — is reserved for users who explicitly opt into reduced motion (in addition to the browser-level `prefers-reduced-motion` media query, which fires automatically). authentik respects both, so most operators don't need to do anything to support reduced motion users.
+
+When you override semantic tokens in your custom CSS, follow the same pattern: provide light values under `:root`, then dark values under `html[data-theme="dark"]`. See [Custom CSS · Theme-based styling](./custom-css.mdx) for a worked example.
+
+## Consuming tokens outside authentik
+
+If your organisation has its own design tooling, you can pull authentik's tokens in directly. The token system is published as the [`@goauthentik/theme`](https://www.npmjs.com/package/@goauthentik/theme) npm package, with three output formats:
```
dewi-tik:
```suggestion
If your organization has its own design tooling, you can pull authentik's tokens in directly. The token system is published as the [`@goauthentik/theme`](https://www.npmjs.com/package/@goauthentik/theme) npm package, with three output formats:
```

website/docs/sys-mgmt/brands/design-language.mdx
```diff
@@ -0,0 +1,115 @@
...
+
+The design language ships with two themes out of the box. Both are toggled by the `data-theme` attribute on `html`:
+
+```css title="Switching themes"
+html[data-theme="dark"] {
+    /* dark theme overrides apply here */
+}
+```
+
+A third theme — `reduced` — is reserved for users who explicitly opt into reduced motion (in addition to the browser-level `prefers-reduced-motion` media query, which fires automatically). authentik respects both, so most operators don't need to do anything to support reduced motion users.
```
dewi-tik:
```suggestion
A third theme — `reduced` — is reserved for users who explicitly opt into reduced motion (in addition to the browser-level `prefers-reduced-motion` media query, which fires automatically). authentik respects both, so most operators do not need to do anything to support reduced motion users.
```

website/docs/sys-mgmt/brands/design-language.mdx
```diff
@@ -0,0 +1,115 @@
...
+## Stability
+
+The `--ak-*` semantic tokens are the public API. authentik treats them as a stable contract within a major version:
+
+- Names won't be renamed without a deprecation cycle.
+- Values may shift to follow design improvements, but the _role_ a token plays (primary, surface, text-muted, etc.) is stable.
+- New tokens may be added.
+- Tokens may be removed only after a deprecation notice in the release notes.
+
+Anything _not_ in the `--ak-*` namespace — `--ak-c-*`, `--pf-*`, `--_*`, internal class names — is implementation detail. Custom CSS targeting those names may break between releases. Treat them as escape hatches for situations the semantic surface doesn't cover yet, and report any cases where you find yourself reaching for them so we can promote those to public tokens.
```
dewi-tik:
```suggestion
Anything _not_ in the `--ak-*` namespace — `--ak-c-*`, `--pf-*`, `--_*`, internal class names — is implementation detail. Custom CSS targeting those names may break between releases. Treat them as escape hatches for situations the semantic surface does not cover yet, and report any cases where you find yourself reaching for them so we can promote those to public tokens.
```

website/docs/sys-mgmt/brands/design-language.mdx
```diff
@@ -0,0 +1,115 @@
...
+| -------------------------- | ------------------------------------------------------------- | ------------------------------------------------------------------------------- | ----------------------------------------- |
+| **Semantic tokens**        | `--ak-color-primary`, `--ak-space-md`, `--ak-shadow-lg`       | The shared design decisions every component reads. Affects the whole interface. | Brand custom CSS, themes.                 |
+| **Component-local tokens** | `--ak-c-login__main--BackgroundColor`, drawer width variables | One component's tunable knobs.                                                  | When a single component needs adjustment. |
+| **CSS parts**              | `ak-flow-executor::part(locale-select)`                       | Structural hooks a component intentionally exposes.                             | Targeted layout or visibility changes.    |
+
+Most brand customisation stays at the first layer. The lower layers are escape hatches.
+
+## Token groups
+
+The semantic surface is around 60 tokens, organised by concern. Every token name follows the pattern `--ak-{group}-{member}`.
```
dewi-tik:
```suggestion
The semantic surface consists of around 60 tokens, organised by concern. Every token name follows the pattern `--ak-{group}-{member}`.
```

website/docs/developer-docs/frontend/css-architecture.md
```diff
@@ -0,0 +1,158 @@
...
+  -> generated CSS variables
+  -> imported runtime token CSS
+  -> Lit component styles and brand custom CSS
+```
+
+Start with handwritten runtime CSS tokens. Add generated files only when token ownership, review workflow, and Figma sync are defined.
+
+## Migration phases
+
+1. Establish public `--ak-*` semantic tokens mapped to current PF values.
```
dewi-tik:
```suggestion
1. Establish public `--ak-*` semantic tokens mapped to current PatternFly values.
```

website/docs/developer-docs/frontend/css-architecture.md
```diff
@@ -0,0 +1,158 @@
...
+2. Update custom CSS docs to prefer `--ak-*` over `--pf-*`.
+3. Migrate new or touched authentik components to consume semantic tokens first, PF globals second.
```
dewi-tik:
```suggestion
3. Migrate new or touched authentik components to consume semantic tokens first, PatternFly globals second.
```

website/docs/developer-docs/frontend/css-architecture.md
```diff
@@ -0,0 +1,158 @@
...
+4. Add token metadata docs generated from the public token list.
+5. Introduce DTCG source files and a compiler when the public token list is stable.
+6. Add cascade layers after global, shadow, and custom CSS order are audited.
+7. De-PF component markup incrementally, starting with components whose styling is already mostly custom.
```
dewi-tik:
```suggestion
7. De-PatternFly component markup incrementally, starting with components whose styling is already mostly custom.
```

website/docs/developer-docs/frontend/css-architecture.md
```diff
@@ -0,0 +1,158 @@
...
+authentik's UI is currently a customized PatternFly 4 system wrapped by Lit web components. Treat PatternFly as the compatibility substrate, not the future public API.
+
+## Current state
+
+- `web/package.json` depends on `@patternfly/patternfly` 4.x and `@patternfly/elements`.
+- `web/src/elements/Base.ts` injects PatternFly base CSS, component CSS, and authentik base CSS into each Lit element.
+- `web/src/common/stylesheets.ts` centralizes `CSSStyleSheet` creation and adopted stylesheet updates.
+- `web/src/styles/authentik/base/*.css` overrides PatternFly global variables and already contains light and dark theme branches.
+- Brand custom CSS is appended to component style roots, including nested shadow roots.
+- Components mix PF class markup, PF per-component imports, global CSS, inline styles, `part` attributes, and local `--ak-*` variables.
```
dewi-tik:
```suggestion
- Components mix PatternFly class markup, PatternFly per-component imports, global CSS, inline styles, `part` attributes, and local `--ak-*` variables.
```

website/docs/developer-docs/frontend/cascade-layers.md
```diff
@@ -0,0 +1,115 @@
...
+- `PFBase` and `AKBase` get wrapped in `@layer components` at their CSS sources.
+- Component styles get wrapped in `@layer components` by their authoring template.
+- Brand custom CSS gets wrapped in `@layer overrides` by `createStyleSheetUnsafe` (or a new variant) before adoption.
+- Each shadow root has a layer-order declaration at the top of the first adopted sheet.
+
+After this step, brand custom CSS wins inside shadow roots without specificity duels. The document cascade is unchanged.
+
+### Step 3 — Layer the document cascade
+
+Wrap the per-file imports in `interface.global.css`, `flows.global.css`, and `static.global.css` into their target layers via `@import ... layer(...)`. Audit ordering against current behavior using visual regression on the storybook before merging.
```
dewi-tik:
```suggestion
Wrap the per-file imports in `interface.global.css`, `flows.global.css`, and `static.global.css` into their target layers via `@import ... layer(...)`. Audit ordering against current behavior using visual regression on the Storybook before merging.
```

dewi-tik (approved):
Some minor suggestions otherwise LGTM for docs

packages/theme/LICENSE.txt
```diff
@@ -0,0 +1,18 @@
+The MIT License (MIT)
+
+Copyright (c) 2025 Authentik Security, Inc.
```
dominic-r:
```suggestion
Copyright (c) 2026 Authentik Security, Inc.
```

packages/theme/lib/tokens/color.js
```diff
@@ -0,0 +1,56 @@
...
+export const colorInfo = variable("color.info", "oklab(0.6689 -0.0608 -0.1513)");
+export const colorSuccess = variable("color.success", "oklab(0.5549 -0.1071 0.0852)");
+export const colorWarning = variable("color.warning", "oklab(0.7864 0.0298 0.1604)");
+export const colorDanger = variable("color.danger", "oklab(0.5331 0.1798 0.1034)");
+
+// Dark theme overrides. Surface values are pinned near PatternFly 4's
+// BackgroundColor--100 (#151515) and the legacy drawer surface (#18191a) — the
+// earlier oklab(0.23) value visibly brightened every PF-backed dark panel.
+theme("dark", (ctx) => {
+    ctx.variable(colorText, "oklab(0.9067 -0 0)");
```
kensternberg-authentik: Styleframe converts everything to OKLCH under the covers and uses that for its styling guide. I can't help but wonder if there's a way to create a design system where we lean heavily on that capability and use the tint/shade/brightness controls instead.

packages/theme/lib/tokens/color.js
```diff
@@ -0,0 +1,56 @@
...
+ * re-points links to their own oklab values rather than chaining through
+ * primary because dark mode links need higher luminance than primary buttons.
+ *
+ * `warning` and `danger` deliberately stay on light values in dark mode — state
+ * colors keep consistent intensity across themes so warnings read as urgent.
+ */
+
+import { ref, theme, variable } from "../shared.js";
+
+export const colorAccent = variable("color.accent", "#fd4b2d");
```
kensternberg-authentik: Suggestion: make this a table.  (I know, I know, "You want everything in tables."  Yes!  So much of what we do is just maintaining spreadsheets of "Oh, that looks nice.").  You can then export the table as the default, and it'll still work just fine.

packages/theme/scripts/build.mjs
```diff
@@ -0,0 +1,212 @@
+/**
+ * @file Build script for `@goauthentik/theme`.
+ *
+ * Runs the styleframe transpiler against the configured token tree, then
+ * fans out the single emitted CSS string into one file per category
+ * (color, typography, spacing, shape, shadow, motion, z-index) and a top-
+ * level `index.css` that `@import`s them in deterministic order.
+ *
```
kensternberg-authentik: I'd prefer the final product being a single CSS file, our top-level "design" file, the way Patternfly does it.  Otherwise, we're making some other part of the system understand the `@import` syntax. It's pretty common, but it's not necessary. We have our division-of-cognitive-labor at the top level, in the Typescript modules; don't put more work on future devs or future machinery, to "undo the assembly only to re-assemble it in the next step."

A theme is a single semantic unit.  Put it in a single syntactical file.

packages/theme/scripts/format-dist.mjs
```diff
@@ -0,0 +1,56 @@
...
+ * not-particularly-pretty output (tab indentation, long lines that aren't
+ * wrapped, mixed spacing in shadow values). Running prettier afterwards
+ * gives anyone inspecting the published artefacts — `dist/index.css`,
+ * `dist/color.css`, `dist/tokens.dtcg.json`, etc. — a consistent and
+ * readable shape.
+ *
+ * Run as the last step of `npm run build` so every file consumers see is
+ * formatted with the repository's own prettier config.
+ */
+
```
kensternberg-authentik: This, this I can get behind.  Ship pretty!

website/docs/developer-docs/frontend/css-architecture.md
```diff
@@ -0,0 +1,158 @@
...
+@layer reset, base, tokens, components, utilities, overrides;
+```
+
+Use layers only after the active CSS entrypoints can establish a single layer order. PatternFly CSS and existing unlayered rules should stay outside layers until their order is audited. In CSS, unlayered normal declarations override layered normal declarations, so partially layering token files can silently lose to legacy declarations.
+
+Recommended meaning:
+
+- `reset`: minimal browser normalization.
+- `base`: document, typography, default color scheme, accessibility defaults.
+- `tokens`: semantic `--ak-*` values and theme variants.
```
kensternberg-authentik: Recommendation: Break `tokens` into _usage_: `globals` and `theme`.

website/docs/developer-docs/frontend/css-architecture.md
```diff
@@ -0,0 +1,158 @@
...
+- `components`: reusable component class and shadow styles.
```
kensternberg-authentik: Recommendation: Have a separate layer, `elements`, for component classes that we expect to find in the lightDOM, i.e. slotted things.

website/docs/developer-docs/frontend/css-architecture.md
```diff
@@ -0,0 +1,158 @@
...
+
+Use `::part()` only for exposed structure:
+
+```css
+ak-flow-executor::part(locale-select) {
+    display: none;
+}
+```
+
+Use slots for composition, not styling. A component should expose stable parts only when a user or brand can reasonably style that substructure without coupling to internal DOM.
```
kensternberg-authentik: Any recommendations for `::slotted()`?

website/docs/developer-docs/frontend/css-architecture.md
```diff
@@ -0,0 +1,158 @@
...
+
```
kensternberg-authentik, line 94: I don't know if we should put recommendations for component styling here for new components. I want to get it down _somewhere_, this is battle-hardened knowledge: How we use `:host([attribute])` to control some styling and layout, how we prefer changing custom properties over providing alternative concrete stylings, and how much we want in slots vs in the component. This doesn't seem like the right place, but I'm not sure where to put it.

website/docs/sys-mgmt/brands/design-language.mdx
```diff
@@ -0,0 +1,115 @@
...
+| `duration`     | `normal`                                                                                                                                                                                                          | The default transition duration. Adjust to make the whole interface feel snappier or calmer.                   |
+| `easing`       | `standard`                                                                                                                                                                                                        | The default cubic-bezier curve.                                                                                |
+| `z-index`      | `xs` through `2xl`                                                                                                                                                                                                | The stacking scale matches PatternFly's tiers (100..600).                                                      |
+
```
kensternberg-authentik, line 47: Breakpoints?
Opacities?
Grids?
Any discussion of container queries, now that they're considered baseline widely available?
Gradients? 

... just some things to think about.

packages/theme/lib/tokens/color.js
kensternberg-authentik:
```
const v = variable;

const colorPrimary = v("color.primary", "oklab(0.522 -0.0434 -0.1717)");
colorPrimaryHover = v("color.primary-hover", "oklab(0.3763 -0.0324 -0.1182)");

// prettier-ignore
const colors = {
    colorPrimary,
    colorPrimaryHover,
    colorAccent:        v("color.accent",         "#fd4b2d"),
    colorText:          v("color.text",           "oklab(0.1957 -0 0)"),
    colorTextMuted:     v("color.text-muted",     "oklab(0.5364 -0.0026 -0.0089)"),

    colorLink:          v("color.link",           ref(colorPrimary)),
    colorLinkHover:     v("color.link-hover",     ref(ColorPrimaryHover)),
    colorLinkVisited:   v("color.link-visited",   "oklab(0.3679 0.0535 -0.1797)"),

    colorSurface:       v("color.surface",        "oklab(1 -0 0)"),
    colorSurfaceMuted:  v("color.surface-muted",  "oklab(0.9551 -0 0)"),
    colorSurfaceRaised: v("color.surface-raised", "oklab(0.9851 -0 0)"),

    colorBorder:        v("color.border",         "#d2d2d2"),
    colorBorderStrong:  v("color.border-strong",  "#8a8d90"),

    colorInfo:          v("color.info",           "oklab(0.6689 -0.0608 -0.1513)"),
    colorSuccess:       v("color.success",        "oklab(0.5549 -0.1071 0.0852)"),
    colorWarning:       v("color.warning",        "oklab(0.7864 0.0298 0.1604)"),
    colorDanger:        v("color.danger",         "oklab(0.5331 0.1798 0.1034)"),
};

// Dark theme overrides. Surface values are pinned near PatternFly 4's
// BackgroundColor--100 (#151515) and the legacy drawer surface (#18191a) — the
// earlier oklab(0.23) value visibly brightened every PF-backed dark panel.
//
// prettier-ignore
theme("dark", (ctx) => {
    ctx.variable(colors.colorText,          "oklab(0.9067 -0 0)");
    ctx.variable(colors.colorTextMuted,     "oklab(0.7407 -0.0007 -0.0017)");
    ctx.variable(colors.colorLink,          "oklab(70.367% -0.07498 -0.139)");
    ctx.variable(colors.colorLinkHover,     "oklab(0.7706 -0.0485 -0.1012)");
    ctx.variable(colors.colorLinkVisited,   "oklab(0.7137 0.0522 -0.151)");
    ctx.variable(colors.colorSurface,       "oklab(0.183 -0 0)");
    ctx.variable(colors.colorSurfaceMuted,  "oklab(0.14 -0 0)");
    ctx.variable(colors.colorSurfaceRaised, "oklab(0.225 -0 0)");
    ctx.variable(colors.colorBorder,        "oklab(0.3906 0.0001 -0.0052)");
    ctx.variable(colors.colorBorderStrong,  "oklab(0.4602 -0.0003 -0.0034)");
    ctx.variable(colors.colorInfo,          "oklab(0.7706 -0.0485 -0.1012)");
    ctx.variable(colors.colorSuccess,       "oklab(0.6488 -0.1066 0.0846)");
});

export default colors;
```

I just think this is easier to read.  Not necessarily maintain (spacing by hand is a pain), but it will help future users find stuff.

web/src/styles/global/theme/variables.css
```diff
@@ -5,6 +5,8 @@
 @import "./shadows.css";
 @import "./z-indexes.css";
 @import "./borders.css";
+@import "@goauthentik/theme/index.css";
+@import "./token-bridge.css";
```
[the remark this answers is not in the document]
GirlBossRush: Wired up: `web`'s `build` and `build-proxy` wireit scripts now depend on `../packages/theme:build`, so the theme package compiles and emits `dist/index.css` before ESBuild resolves the import. Verified with a clean `packages/theme/dist`.

web/src/elements/ak-drawer/ak-drawer.root.css
```diff
@@ -1,14 +1,22 @@
 /* ----------- CSS Custom Properties for DRAWER  --------------------------- */
 
 :root {
+    /* Component-local directional shadows for the panel's expanded state.
+     * Defined here so the drawer carries its own shadow values rather than
+     * exposing four directional shadow tokens on the public --ak-* surface. */
+    --_drawer-shadow-top: 0 -0.75rem 0.75rem -0.5rem rgba(3, 3, 3, 0.18);
+    --_drawer-shadow-right: 0.75rem 0 0.75rem -0.5rem rgba(3, 3, 3, 0.18);
+    --_drawer-shadow-bottom: 0 0.75rem 0.75rem -0.5rem rgba(3, 3, 3, 0.18);
+    --_drawer-shadow-left: -0.75rem 0 0.75rem -0.5rem rgba(3, 3, 3, 0.18);
```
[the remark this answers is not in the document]
GirlBossRush: Fixed — `--_drawer-inverse-multiplier: -1` is now defined in `ak-drawer.root.css` next to the other component-local properties.

web/src/styles/global/theme/token-bridge.css
```diff
@@ -0,0 +1,127 @@
...
+ * token is defined and otherwise leaves PF's defaults untouched. (The theme
+ * is always loaded in authentik, so the fallback is a safety net, not the
+ * common path.)
+ */
+
+:root,
+:host {
+    /* Legacy authentik bridge */
+    --ak-accent: var(--ak-color-accent, #fd4b2d);
+
```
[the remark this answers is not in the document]
GirlBossRush: Fixed — the hardcoded `--ak-accent: #fd4b2d` is removed from `variables.css`; the bridge is now the single definition in both cascade contexts.

web/src/styles/global/theme/token-bridge.css
```diff
@@ -0,0 +1,127 @@
...
+    /* PatternFly 4 compatibility bridge */
+    --pf-global--primary-color--100: var(--ak-color-primary, var(--pf-global--primary-color--100));
+    --pf-global--primary-color--200: var(
+        --ak-color-primary-hover,
+        var(--pf-global--primary-color--200)
+    );
```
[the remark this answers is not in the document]
GirlBossRush: Correct — a self-reference in the fallback is a dependency cycle and computes to invalid. The bridge now spells out each PF variable's prior light-theme value as a literal fallback instead.

packages/theme/lib/tokens/color.js
GirlBossRush: Done — `color.ts` is now the spreadsheet form you sketched: aligned value columns under `// prettier-ignore`, grouped by concern (brand / text / link / surface / border / state), for both the light map and the dark override pairs. Values are authored as hex now (the build transforms to `oklch()` with a hex swatch comment), and the emitted CSS is byte-identical to before the reformat.

packages/theme/lib/tokens/color.js
GirlBossRush: The plumbing for this exists — every color already emits as `oklch()`, so lightness/chroma are directly adjustable — but each variant is still an explicit stop. Recorded as a direction in `css-architecture.md` → Still to do: derive relational colors (hover/tint/shade) algorithmically in OKLCH space. Felt like a follow-up rather than something to fold into this PR.

packages/theme/scripts/build.mjs
GirlBossRush: This is how it works now: `dist/index.css` is fully self-contained — every token in one file, zero `@import` — and it is the only file the app consumes. The per-category files (`color.css`, `typography.css`, …) are additive slices for external consumers that want just one concern (e.g. the docs site); happy to drop them if you would rather not carry the surface.

website/docs/developer-docs/frontend/css-architecture.md
GirlBossRush: Wrote this down in `developer-docs/frontend/css-architecture.md` → "Styling components": `:host([attribute])` for variants, reassign-custom-properties-over-alternative-concrete-stylings (with the drawer as the worked example), and slots-are-for-composition with an ownership rule for what gets slotted. Worth a read to check it matches the knowledge you meant to capture.

website/docs/developer-docs/frontend/css-architecture.md
GirlBossRush: Added to the same "Styling components" section: `::slotted()` is a boundary tool — it can only select the slotted elements themselves, so keep it to what a container owns about its children (layout, spacing, alignment); typography/color inside slotted content belongs to the document cascade, and needing deep control is a sign the content should be component-owned template instead.

website/docs/developer-docs/frontend/css-architecture.md
GirlBossRush: Noted in `cascade-layers.md` → Adding CSS: light-DOM/slotted-content classes are document-level component rules and live in `layer(components)` today, with the dedicated-layer split (between `components` and `theme`) recorded as the path if that category grows. Did not extend the shipped six-layer order in this PR since re-ordering layers is a cascade-visible change.

website/docs/sys-mgmt/brands/design-language.mdx
GirlBossRush: Added a "What is deliberately not a token" section to `design-language.mdx` covering these: breakpoints (components use container queries — baseline now — and the remaining media queries follow PF grid tiers), opacities (baked into composed values; a standalone scale layers unpredictably against themed surfaces), grids/layout (component behavior, not brand identity), and gradients (the existing ones — skeleton shimmer, scroll shadows — are component implementation details composed from semantic colors). Container queries got their mention; nothing tokenized until usage demands it.

kensternberg-authentik (approved):

PR #22533, opened 2026-05-21
ci: Fix Node script hang. Add timeouts to docker build, lint, and setup-node install steps

GirlBossRush:
## Summary

- Add explicit `process.exit(0)` calls to scripts to drain logger.
- Add `timeout-minutes: 30` to `_reusable-docker-build-single.yml` (build job) and the `lint` matrix in `ci-main.yml` — both were relying on GitHub's 6-hour default.
- Add `timeout-minutes: 5` to the two `corepack npm ci` steps in `.github/actions/setup-node` so a hung npm install surfaces as a real failure.

### DIFF
--- a/.github/actions/setup-node/action.yml
+++ b/.github/actions/setup-node/action.yml
@@ -54,10 +54,9 @@ runs:
     - name: Install monorepo dependencies
       if: ${{ contains(inputs.dependencies, 'monorepo') }}
       shell: bash
-      timeout-minutes: 5
       run: | #shell
         node ./scripts/node/lint-lockfile.mjs
-        corepack npm ci
+        timeout 5m corepack npm ci
     - name: Setup Node.js (Working Directory)
       if: ${{ contains(inputs.dependencies, 'working-directory') }}
       uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v4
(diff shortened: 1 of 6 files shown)


rissson:
if we're doing this, let's add timeouts to everything

.github/actions/setup-node/action.yml
```diff
@@ -56,7 +56,7 @@ runs:
       shell: bash
       run: | #shell
         node ./scripts/node/lint-lockfile.mjs
-        corepack npm ci
+        timeout 5m corepack npm ci
```
rissson, line 59: Should this be a timeout on that step instead? If that's supported, I'm not 100% on that

.github/workflows/_reusable-docker-build-single.yml
```diff
@@ -31,6 +31,7 @@ jobs:
   build:
     name: Build ${{ inputs.image_arch }}
     runs-on: ${{ inputs.runs-on }}
+    timeout-minutes: 30
```
rissson, line 34: I'd rather have one PR that fixes the hang, and then another big one that adds timeouts to everything

GirlBossRush:
Splitting this into two focused PRs per @rissson's review:

- #22794 — the load-bearing hang fix (explicit `process.exit(0)` in the three `setup-node` lint scripts).
- #22795 — the CI timeouts, with a reply to the `setup-node` step-timeout question in the PR description (short version: `timeout-minutes` isn't supported on composite-action steps — actions/runner#1979 — so the `timeout 5m` shell wrapper is the correct workaround).

Closing in favor of those two. Thanks @rissson!

PR #22548, opened 2026-05-21
root: fix make gen-diff command

PeshekDotDev:
## Details

I have not been able to make `make gen-diff` work without these changes. I am not an expert on this, so please review

---

## Checklist

If applicable

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -201,8 +201,8 @@ gen-diff:  ## (Release) generate the changelog diff between the current schema a
 		/local/schema-old.yml \
 		/local/schema.yml
 	rm schema-old.yml
-	$(SED_INPLACE) 's/{/&#123;/g' diff.md
-	$(SED_INPLACE) 's/}/&#125;/g' diff.md
+	$(SED_INPLACE) 's/{/\&#123;/g' diff.md
+	$(SED_INPLACE) 's/}/\&#125;/g' diff.md
 	npx prettier --write diff.md
 
 gen-client-go:  ## Build and install the authentik API for Golang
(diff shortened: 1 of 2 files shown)


scripts/compose.yml
```diff
@@ -32,11 +32,13 @@ services:
     restart: always
   diff:
     image: docker.io/openapitools/openapi-diff:2.1.7
-    entrypoint: "/bin/true"
+    # Only run when explicitly invoked (e.g. `make gen-diff`), never on `docker compose up`.
+    profiles:
```
rissson, line 36: TIL!

Makefile
rissson, line 198: Do we need a `--profile diff` here or not?

Makefile
```diff
@@ -201,8 +201,8 @@ gen-diff:  ## (Release) generate the changelog diff between the current schema a
 		/local/schema-old.yml \
 		/local/schema.yml
 	rm schema-old.yml
-	$(SED_INPLACE) 's/{/&#123;/g' diff.md
-	$(SED_INPLACE) 's/}/&#125;/g' diff.md
+	$(SED_INPLACE) 's/{/\&#123;/g' diff.md
+	$(SED_INPLACE) 's/}/\&#125;/g' diff.md
```
rissson, line 205: cc. @gergosimonyi

Makefile
PeshekDotDev: Since we are explicitly running diff, we don't have to pass the profile. A good example of this is the `docker compose run db-migrations` example in the docker docs -- https://docs.docker.com/compose/how-tos/profiles/#auto-starting-profiles-and-dependency-resolution

rissson: perfect

rissson (approved):

PR #22554, opened 2026-05-22
website/docs: Fix release notes cards

dominic-r:
Render release note version labels without Docusaurus' leading digit icon split, and make the generated release index description more useful.

The release index cards now bypass Docusaurus' fallback icon extraction for release docs, which was splitting the leading `2` out of labels like `2026.8`. The generated `/releases/` page also has a fuller description that calls out upgrade guidance, breaking changes, features, and fixes.

The lint follow-up removes effect-driven derived state in the glossary cards and 404 redirect route handling, so the website lint target passes without suppressions. A final prettier-only update fixes the import ordering reported by CI.

Before:

Release notes cards before fix

After:

Release notes cards after fix

Validation:
- corepack npm run --prefix website prettier-check
- corepack npm run --prefix website lint-check
- corepack npm run --prefix website check-types


PeshekDotDev (approved):
I approve but don't merge until someone from front-end reviews this

GirlBossRush (approved):

PR #22590, opened 2026-05-22
providers/oauth2: OpenID key binding support

EthanHeilman:
## Details

This PR adds OpenID Key Binding to authentik. OpenID Key Binding binds a public key to an ID Token using mechanisms defined in [RFC9449](https://openid.github.io/connect-key-binding/main.html#RFC9449), OAuth 2.0 Demonstrating Proof of Possession (DPoP). For more details see the draft OpenID spec for Key Binding https://openid.github.io/connect-key-binding/main.html

We add key binding support for the authorization code, device code and refresh flows fully supporting all flows specified in the draft standard. The draft standard does not key bind the access token and we follow that here.

The PR contains:
* The code implementing the key binding feature
  * Adds `key_bound` scope to list of scopes
  * Adds new ID Token typ `dpop+id_token`
  * Mitigation to add new field `dpop_jkt` to authorizationcode, devicetoken and refreshtoken models. This is needed to remember the public key bound to token for each of those flows.
   * Stores `jti` claims in the authentik cache to protect against replay attacks
* Unit tests

Currently this doesn't make any changes to docs since it doesn't seem there is a doc per flow type in the docs directory, but if docs for this feature is desired, I do have some I could can add.

---

## Checklist

If applicable

Appendix
==

These are largely for my own notes to document my design decisions. Feel free to read, but not required to understand to review the PR.

Replay protection  and JTI
====

Following the recommendation given in [RFC 9449 Token Replay](https://datatracker.ietf.org/doc/html/rfc9449#Token_Replay), we use the `jti` claim in the DPOP proof to detect replayed DPoP proofs:

> In the context of the target URI, servers can store the jti value of each DPoP proof for the time window in which the respective DPoP proof JWT would be accepted to prevent multiple uses of the same DPoP proof. HTTP requests to the same URI for which the jti value has been seen before would be declined.

To do this we employ authentiks caching mechanism to store these proofs for 3 minutes. Any proof older or newer than 60 seconds is considered expired and rejected. We enforce a max JTI string size of 256 characters

An assumption we make here is that the cache is shared between all instances of authentik or all instances of authentik at an URI. If this was not the case, would be possible a DPoP proof can be replayed. Preventing replay is not a requirement of DPoP with the standard stating:

> When strictly enforced, such a single-use check provides a very strong protection against DPoP proof replay, but it may not always be feasible in practice, e.g., when multiple servers behind a single endpoint have no shared state.

JTI or (JTI AND HTU)
========

> When strictly enforced, such a single-use check provides a very strong protection against DPoP proof replay, but it may not always be feasible in practice, e.g., when multiple servers behind a single endpoint have no shared state.

We check only uniqueness of the `jti` in the cache. It would be allowed by the standard to store the `jti` and the `htu` (URI of the request)  in the cache and then only reject if the pair (`jti`, `htu`) was not unique.  We choose to check uniqueness of `jti` and not the pair (`jti`, `htu`) to "fail faster" and make finding bugs easier. If someone is using a non-unique `jti` we want them to get errors early so they can fix it in testing. Using the `htu` might hide a non-random `jti` bug.

Is JTI strictly needed here?
========

We support three flows: authorization code, device code and refresh flow. The authorization code and the device code flows use the hash of the auth_code or the device code as the claim `c_s256` in the DPoP proof functioning like a nonce. This should prevent DPoP replays and makes the `jti` check in these flows unnecessary. Only the refresh flow has a security benefit from the `jti` check. So why not skip the `jti` check for the authorization code and device code?

* The `jti` check is recommended by standard and expected. If a developer noticed that the `jti` check is not happening, they might assume this is a security bug and report it. Better to do the expected thing to reduce false positives. 
* Skipping the check would not simplify the code and would require special casing for different flows.
* Checking the `jti` allows the dpop code added in this PR to be used for dpop as authentik adopts dpop more generally.


gergosimonyi (approved):
Seems clean, thank you for the contribution!

I'll leave it to @BeryJu to merge

PR #22642, opened 2026-05-26
sources/oauth: add discord roles sync as a native feature

### LINKED ISSUE #22367: Discord source improvement - native guild and avatar sync
Update discord source so that users don't need to manually create mappings.

Benricheson101:
## Details

Adds Discord avatar and role syncing as a native feature with the Discord OAuth source. This replaces the complexity of adding a property mapping for assigning authentik groups according to Discord roles.

A new Discord guild ID field is added to the OAuth Source form. When this is set, the `guilds.members.read` scope is automatically appended to the Discord authorization URL. Any roles the user has in that guild are synced to authentik as groups named `<role-id>@discord`. These roles are automatically created, assigned, and revoked when a user enrolls/authenticates with a Discord Source. If the guild ID is not set, no automatic role sync is performed.

The Discord user's avatar URL is added to the authentik user as an attribute called `discord_avatar`, which can be used as an authentik avatar by adding `attributes.discord_avatar` to the avatar source list.

image

Closes #22367 

---

## Checklist

If applicable


authentik/sources/oauth/models.py
```diff
@@ -84,6 +84,14 @@ class OAuthSource(NonCreatableType, Source):
         ),
     )
 
+    discord_guild_id = models.CharField(
+        max_length=20,
```
rissson: I think let's just have a `TextField` in DB here. We don't want to prevent users from changing this if discord ever decides to break their API.

authentik/sources/oauth/types/discord.py
```diff
@@ -34,9 +61,34 @@ class DiscordType(SourceType):
...
-            "name": info.get("username"),
+            "name": info.get("global_name", info.get("username")),
+            "groups": role_ids,
+            "attributes": {
+                "discord_avatar": avatar_url,
+            },
         }
+
+    def get_base_group_properties(self, source: OAuthSource, group_id: str, **kwargs):
+        return {"name": f"{group_id}@discord"}
```
rissson, line 94: I'm almost 100% sure we'll get feature requests about customising group names from roles. A better solution to the UX problem with long docs might be to provide the property mapping we currently have in our docs as a built-in property mapping that can be selected instead? So that then, users can select it if they just want the default, or create a custom one changing what they need.

Actually disregard all of that, you can create a property mapping that overrides this, nevermind. But we should probably document that as well.

authentik/sources/oauth/types/discord.py
```diff
@@ -34,9 +61,34 @@ class DiscordType(SourceType):
...
+        )
+
         return {
             "username": info.get("username"),
             "email": info.get("email", None),
-            "name": info.get("username"),
+            "name": info.get("global_name", info.get("username")),
+            "groups": role_ids,
+            "attributes": {
+                "discord_avatar": avatar_url,
```
rissson, line 89: Does that attribute get removed from the end attributes set on the user if it's None?

authentik/sources/oauth/types/discord.py
Benricheson101: Maybe the current code snippet on the docs (that uses the `discord_role_id` attribute) could be added as an example blueprint with `ACCEPTED_GUILD_ID` templated in. I thought of this as a potential alternative to it being a proper first-party feature, but it's a bit less intuitive than just having roles sync automatically. 

Unfortunately Discord's guild member endpoint doesn't return role names without involving a bot account, so naming them this way was the best way I could think of.

authentik/sources/oauth/types/discord.py
Benricheson101: It gets added as null, which is skipped when the avatar source is chosen (since they're or'd).

BeryJu (commented):
This is pretty much what I imagined 🎉 

One thing we need to figure out is if we're ok with storing provider_type specific settings on the root OAuthSource model (as this PR currently does), or if we bite the bullet and have OAuthSource subclasses that aren't just proxies to store these fields. There's also a Microsoft Entra source issue that would require a similar field on the source.

Benricheson101:
I had the same thought. I tried implementing it one other way, which used a separate table for the DiscordOAuthSource model (`abstract=False`), but it ended up being needing too many special cases across too many places for me to feel comfortable with it.

BeryJu:
after some discussion I think its fine to keep this in the base class for now and if more oauth source types need specific configuration we can migrate to separate concrete classes

PR #22659, opened 2026-05-26
core: add user switching

### LINKED ISSUE #15095: Simultaneous multi user login
I have two accounts setup with Authentik, one for myself, and one as myself but with administrative privilages. I find myself having to swap between these two accounts often while setting up a new application.

It would be great to support an account selector (similar to google and others) that lets me have 2 logged in users and a selector (optional only if there are multiple users) that shows up when an authentication request is coming from an application (in my case using oidc)

Currently I have to open my authentik page directly and logout before going back to the app to login with the other account. I have to do this as i have auto redirect in most cases and it just automatically logins with the current account otherwise.

**Additional context**
I am really looking for something like the google account selector for when there are multiple users logged in to the authentik instance from a current browser session. This way it allows account switching with less hassle.

### LINKED ISSUE #18375: Multi-account authentication

In business deployments, it's very common to have multiple administrator accounts, or multiple accounts with different levels of permissions. As a sysadmin, I often have to conduct changes in authentik's configuration, such as group assignments, use deactivation, etc. Though, I cannot use this same privileged account to login to internal apps.

Having to logout, log back in again to a standard user account makes performing day-to-day tasks in authentik tedious.

In the same manner as services like Google, being able to switch between connected accounts. We already have impersonation.

There are no acceptable alternatives as of today.

### Additional context

As mentioned, it could function in the same as all the big cloud providers.

When being presented with an authentication flow, a client with more than one connected account would be prompted to choose the account to continue the action with. When, inevitably, one account gets disconnected (whether be it a timeout or anything else), the account selection prompt would display so.

dominic-r:
Adds user switching so users can keep multiple signed-in users available in the same browser and switch between them from the User interface header.

Each login keeps its own authenticated session. When a user selects another signed-in user, authentik starts the configured user switch authentication flow instead of silently reusing another session. The flow receives user-switch context so policies can decide which stages apply, while normal authentication and MFA policy behavior remain in control. Sessions from the same browser remain available as switch targets, but superseded sessions are not accepted for normal request handling.

Brands can set a dedicated User switch flow. If unset, user switching is disabled; selecting the default authentication flow explicitly reuses it for user switching.

Closes #18375
Closes https://github.com/goauthentik/authentik/issues/15095


website/docs/add-secure-apps/flows-stages/stages/account_selection/index.md
```diff
@@ -0,0 +1,68 @@
+---
+title: Account selection stages
```
dewi-tik:
```suggestion
title: Account selection stage
```

website/docs/add-secure-apps/flows-stages/stages/account_selection/index.md
```diff
@@ -0,0 +1,68 @@
...
+---
+
+Account selection stages let users choose which remembered authentik account should continue a flow.
```
dewi-tik:
```suggestion
The account selection stage allows users choose which remembered authentik account should continue a flow.
```

website/docs/add-secure-apps/flows-stages/stages/account_selection/index.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Overview
+
+authentik can remember multiple accounts in the same browser after successful login. The remembered account list is stored in a signed browser cookie, but an account is only shown when its authentik session is still active.
```
dewi-tik:
```suggestion
authentik can remember multiple accounts within the same browser. The list of remembered accounts is stored in a signed browser cookie, and an account only appears if its authentik session is still active.
```

website/docs/add-secure-apps/flows-stages/stages/account_selection/index.md
```diff
@@ -0,0 +1,68 @@
...
+
+Two stages work together:
```
dewi-tik: I think this would need to be split in to 2 docs. Maybe within a directory in the sidebar with an overview page?

dewi-tik (commented):
Will leave further doc review until the underlying changes go through review

authentik/core/api/users.py
```diff
@@ -451,13 +452,25 @@ class Meta:
         }
 
 
+class AccountSelectionUserSerializer(PassiveSerializer):
+    """Browser-local account that can be selected from the UI."""
```
gergosimonyi: Why isn't this frontend-only?

dominic-r: https://authentiksecurity.slack.com/archives/D09CQHGBS7K/p1780601476828949

web/src/admin/flows/utils.ts
```diff
@@ -10,6 +10,8 @@ export function DesignationToLabel(designation: FlowDesignationEnum): string {
     switch (designation) {
         case FlowDesignationEnum.Authentication:
             return msg("Authentication");
+        case FlowDesignationEnum.AccountSelection:
```
kensternberg-authentik: (suggestion) switch/case is of the devil:

```
const designationToLabel = {
    FlowDesignationEnum.Authentication: msg("Authentication"),
    FlowDesignationEnum.AccountSelection: msg("Account selection"),
    FlowDesignationEnum.Authorization: msg("Authorization"),
    FlowDesignationEnum.Enrollment: msg("Enrollment"),
    FlowDesignationEnum.Invalidation: msg("Invalidation"),
    FlowDesignationEnum.Recovery: msg("Recovery"),
    FlowDesignationEnum.StageConfiguration: msg("Stage Configuration"),
    FlowDesignationEnum.Unenrollment: msg("Unenrollment"),
    FlowDesignationEnum.UnknownDefaultOpenApi: msg("Unknown designation"),
}


export function DesignationToLabel(designation: FlowDesignationEnum): string {
    return designationToLabel[designation] ?? designationToLabel[FlowDesignationEnum.UnknownDefaultOpenApi];
}
```

kensternberg-authentik (approved):
I had only one suggestion, and it's a minor one.  I'm not recommending you implement it, just something to keep in mind.  Its like camping: leave the grounds better than how you found them.  I'll keep whining about switch/case everywhere until it's nowhere.

web/src/admin/stages/user_selection/UserSelectionStageForm.ts
```diff
@@ -0,0 +1,51 @@
...
+        return html`<span> ${msg("Prompt the user to select a browser-local account.")} </span>
+            <ak-form-element-horizontal label=${msg("Name")} required name="name">
+                <input
+                    type="text"
+                    value="${ifDefined(this.instance?.name || "")}"
+                    class="pf-c-form-control"
+                    required
+                />
+            </ak-form-element-horizontal>`;
+    }
```
GirlBossRush: May need to replace this <ak-text-input> to fix the ARIA tree

web/src/admin/stages/account_selection/AccountSelectionStageForm.ts
```diff
@@ -0,0 +1,51 @@
...
+import { AccountSelectionStage, StagesApi } from "@goauthentik/api";
+
+import { msg } from "@lit/localize";
+import { html, TemplateResult } from "lit";
+import { customElement } from "lit/decorators.js";
+import { ifDefined } from "lit/directives/if-defined.js";
+
+@customElement("ak-stage-account-selection-form")
+export class AccountSelectionStageForm extends BaseStageForm<AccountSelectionStage> {
+    loadInstance(pk: string): Promise<AccountSelectionStage> {
```
GirlBossRush:
```suggestion
    protected loadInstance(pk: string): Promise<AccountSelectionStage> {
```

web/src/admin/stages/account_selection/AccountSelectionStageForm.ts
```diff
@@ -0,0 +1,51 @@
...
+        return new StagesApi(DEFAULT_CONFIG).stagesAccountSelectionSelectionRetrieve({
+            stageUuid: pk,
+        });
+    }
+
+    async send(data: AccountSelectionStage): Promise<AccountSelectionStage> {
```
GirlBossRush:
```suggestion
    protected async send(data: AccountSelectionStage): Promise<AccountSelectionStage> {
```

web/src/admin/stages/account_selection/AccountSwitchStageForm.ts
```diff
@@ -0,0 +1,53 @@
...
+                accountSwitchStageRequest: data,
+            });
+        }
+        return new StagesApi(DEFAULT_CONFIG).stagesAccountSelectionSwitchCreate({
+            accountSwitchStageRequest: data,
+        });
+    }
+
+    protected override renderForm(): TemplateResult {
+        return html`<span>
```
GirlBossRush: Similar ARIA fixes.

web/src/components/ak-account-switcher.ts
```diff
@@ -0,0 +1,161 @@
...
+            return `${globalAK().api.base}if/flow/${accountSelectionFlow}/?${query.toString()}`;
+        }
+        if (!account?.uid) {
+            query.set("add_account", "true");
+        }
+        if (loginHint) {
+            query.set("login_hint", loginHint);
+        }
+        return `${globalAK().api.base}flows/-/default/authentication/?${query.toString()}`;
+    }
```
GirlBossRush: Break this out into a function if you can. Makes for easier unit testing

web/src/components/ak-account-switcher.ts
```diff
@@ -0,0 +1,161 @@
...
+
+    protected renderAvatar(account?: Pick<AccountSelectionUser, "avatar">) {
```
GirlBossRush:
```suggestion
    protected renderAvatar(account?: Pick<AccountSelectionUser, "avatar">): SlottedTemplateResult {
```

Among a few places

web/src/components/ak-account-switcher.ts
```diff
@@ -0,0 +1,161 @@
...
+        const content = html`
+            <span class="pf-c-dropdown__menu-item-main account-switcher-item">
+                ${this.renderAvatar(account)}
+                <span class="account-switcher-labels">
+                    <span class="account-switcher-name">${label}</span>
+                    <span class="account-switcher-description">${description}</span>
+                </span>
+                ${account.isCurrent
+                    ? html`<i class="fas fa-check account-switcher-current" aria-hidden="true"></i>`
+                    : nothing}
```
GirlBossRush: Unless you've got some complex optional attribute assignment, Prefer `null` over `nothing` when adding new code. There's a mega PR coming that'll do it everywhere soon.

```suggestion
                    : null}
```

web/src/flow/stages/account_selection/AccountSelectionStage.ts
```diff
@@ -0,0 +1,164 @@
...
+@customElement("ak-stage-account-selection")
+export class AccountSelectionStage extends BaseStage<
+    AccountSelectionChallenge,
+    AccountSelectionChallengeResponseRequest
+> {
+    static styles: CSSResult[] = [
+        PFLogin,
+        PFForm,
+        PFButton,
+        css`
```
GirlBossRush: Separate CSS file.

web/src/flow/stages/account_selection/AccountSelectionStage.ts
```diff
@@ -0,0 +1,164 @@
...
+
+                <fieldset class="ak-c-fieldset pf-c-form__group pf-m-action">
+                    <legend class="sr-only">${msg("Form actions")}</legend>
+                    <button
+                        type="button"
+                        class="pf-c-button pf-m-link pf-m-block"
+                        @click=${() =>
+                            this.host.submit({
+                                action: AccountSelectionChallengeResponseActionEnum.Login,
+                            })}
```
GirlBossRush: Move into protected method:

```suggestion
                        @click=${this.dispatchAccountSelectionSubmit}
```

web/src/components/ak-account-switcher.ts
```diff
@@ -0,0 +1,161 @@
...
+                        </a>
+                    </li>
+                    <li role="presentation">
+                        <a
+                            class="pf-c-dropdown__menu-item"
+                            role="menuitem"
+                            href="${globalAK().api.base}flows/-/default/invalidation/"
+                        >
+                            <i class="fas fa-sign-out-alt" aria-hidden="true"></i>
+                            ${msg("Sign out current account")}
```
GirlBossRush: When adding messages, especially outside the admin, provide an message ID and optional description.

https://github.com/goauthentik/authentik/blob/main/web/AGENTS.md#i18n

web/src/components/ak-account-switcher.css
```diff
@@ -0,0 +1,115 @@
+:host {
+    display: contents;
+}
+
+.account-switcher {
```
GirlBossRush: Prefer the [part] attribute selectors. A user may be less likely to fully customize the component, but it helps them have a uniform understanding of how to adjust the UI in a pattern that matches other page elements.

For example:

- https://github.com/goauthentik/authentik/blob/main/web/src/elements/commands/ak-command-palette-modal.css

web/src/admin/stages/account_selection/AccountSwitchStageForm.ts
```diff
@@ -0,0 +1,53 @@
...
+
+@customElement("ak-stage-account-switch-form")
+export class AccountSwitchStageForm extends BaseStageForm<AccountSwitchStage> {
+    loadInstance(pk: string): Promise<AccountSwitchStage> {
+        return new StagesApi(DEFAULT_CONFIG).stagesAccountSelectionSwitchRetrieve({
+            stageUuid: pk,
+        });
+    }
+
+    async send(data: AccountSwitchStage): Promise<AccountSwitchStage> {
```
GirlBossRush:
```suggestion
    protected async send(data: AccountSwitchStage): Promise<AccountSwitchStage> {
```

web/src/admin/stages/account_selection/AccountSwitchStageForm.ts
```diff
@@ -0,0 +1,53 @@
...
+import { AccountSwitchStage, StagesApi } from "@goauthentik/api";
+
+import { msg } from "@lit/localize";
+import { html, TemplateResult } from "lit";
+import { customElement } from "lit/decorators.js";
+import { ifDefined } from "lit/directives/if-defined.js";
+
+@customElement("ak-stage-account-switch-form")
+export class AccountSwitchStageForm extends BaseStageForm<AccountSwitchStage> {
+    loadInstance(pk: string): Promise<AccountSwitchStage> {
```
GirlBossRush:
```suggestion
    protected loadInstance(pk: string): Promise<AccountSwitchStage> {
```

GirlBossRush (dismissed):
Over all a very exciting feature 👏🏻 

With some requested changed it should be approved ASAP.

dominic-r:
Thanks @GirlBossRush ! Suggestions have been applied ^^

authentik/stages/account_selection/apps.py
```diff
@@ -0,0 +1,12 @@
+"""authentik account selection stage config."""
+
+from authentik.blueprints.apps import ManagedAppConfig
+
+
+class AuthentikStageAccountSelectionConfig(ManagedAppConfig):
+    """authentik account selection stage config."""
+
+    name = "authentik.stages.account_selection"
```
BeryJu: if anything this should be `user_selection` to match with `user_write`/`user_login`/etc

authentik/providers/oauth2/views/authorize.py
```diff
@@ -24,13 +24,19 @@
     PROMPT_CONSENT,
     PROMPT_LOGIN,
     PROMPT_NONE,
+    PROMPT_SELECT_ACCOUNT,
```
BeryJu: For the stake of reviewing this, imo the OAuth2 stuff should be in a follow-up PR

BeryJu (changes requested):
Chatted about this with @gergosimonyi, some thoughts

- I'd like this to function similarly to how google does it with multiple accounts
- It would probably make sense to have something above or below a session since we don't just want to swap the user in/out of the session
- Some metric for knowing "how authenticated" a user is (do we just remember the username? did they just login? somewhere in between?)

web/src/components/ak-nav-buttons.ts
```diff
@@ -154,7 +142,7 @@ export class NavigationButtons extends WithNotifications(WithSession(AKElement))
         </div>`;
     }
 
-    renderImpersonation() {
+    protected renderImpersonation(): SlottedTemplateResult {
```
kensternberg-authentik: `protected` is good; you really don't need the return type, though. Consumers will tell you if Yuir Doin' It Wrong.

authentik/brands/models.py
```diff
@@ -57,6 +57,11 @@ class Brand(SerializerModel):
     flow_authentication = models.ForeignKey(
         Flow, null=True, on_delete=models.SET_NULL, related_name="brand_authentication"
     )
+    # Optional authentication flow that account switches go through.
+    # Account switching is disabled when unset.
```
gergosimonyi:
```suggestion
```

authentik/core/models.py
```diff
@@ -1368,6 +1368,13 @@ class AuthenticatedSession(SerializerModel):
 
     user = models.ForeignKey(User, on_delete=models.CASCADE)
 
+    # Browser cookie value grouping logins from the same browser.
+    # Unbound sessions are never offered for account switching.
```
gergosimonyi:
```suggestion
    # This is matched against the value of the `authentik_browser` cookie.
```

authentik/core/models.py
```diff
@@ -1368,6 +1368,13 @@ class AuthenticatedSession(SerializerModel):
...
+    browser_key = models.CharField(max_length=64, null=True, db_index=True)
+    # Whether this is the login the browser most recently switched to.
+    # Stored server-side to reject replayed cookies for superseded logins.
```
gergosimonyi:
```suggestion
```

authentik/core/models.py
```diff
@@ -1368,6 +1368,13 @@ class AuthenticatedSession(SerializerModel):
...
+    is_current = models.BooleanField(default=False)
```
gergosimonyi: Only one `AuthenticatedSession` per `browser_key` is allowed to be `is_current`, right? If so, add a unique constraint so it's enforced.

authentik/core/models.py
```diff
@@ -1384,11 +1391,20 @@ def __str__(self) -> str:
     @staticmethod
     def from_request(request: HttpRequest, user: User) -> AuthenticatedSession | None:
         """Create a new session from a http request"""
+        from authentik.root.middleware import SessionMiddleware
```
gergosimonyi: Importing a middleware in models smells. Please relocate the function to a more suitable place.

authentik/core/models.py
```diff
@@ -1384,11 +1391,20 @@ def __str__(self) -> str:
...
+
         if not hasattr(request, "session") or not request.session.exists(
             request.session.session_key
         ):
             return None
+        browser_key = SessionMiddleware.ensure_browser_key(request)
+        if browser_key:
+            # The new login takes over the browser; older logins become switch targets.
+            # Concurrent logins may briefly leave two current sessions; cookies pick the winner.
```
gergosimonyi:
```suggestion
```

authentik/core/sessions.py
```diff
@@ -32,29 +32,44 @@ def get_model_class(cls):
...
 
+    @staticmethod
+    def _check_superseded(session) -> None:
+        """Reject browser logins superseded by a newer login."""
+        authenticated_session = getattr(session, "authenticatedsession", None)
+        if (
+            authenticated_session is not None
+            and authenticated_session.browser_key
+            and not authenticated_session.is_current
+        ):
```
gergosimonyi: Is there any reason this isn't on `Session`? Perhaps a `@property` called `active`/`inactive`?

authentik/core/urls.py
```diff
@@ -44,6 +45,12 @@
         RedirectToAppLaunch.as_view(),
         name="application-launch",
     ),
+    path(
+        # Turns the frontend-selected UID into backend-verified flow context.
+        "account/switch/<str:user_uid>/",
```
gergosimonyi: Why did you pick `uid` of all the identifiers?

authentik/flows/planner.py
```diff
@@ -42,6 +42,8 @@
 PLAN_CONTEXT_DEVICE = "device"
 PLAN_CONTEXT_SOURCE = "source"
 PLAN_CONTEXT_OUTPOST = "outpost"
+PLAN_CONTEXT_IS_ACCOUNT_SWITCH = "is_account_switch"
+PLAN_CONTEXT_ACCOUNT_SWITCH_FROM_USER = "account_switch_from_user"
```
gergosimonyi: Why two values? Can't you get `PLAN_CONTEXT_IS_ACCOUNT_SWITCH` by `bool(PLAN_CONTEXT_ACCOUNT_SWITCH_FROM_USER)`?

authentik/core/views/account_switch.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def get(self, request: HttpRequest, user_uid: str) -> HttpResponse:
+        flow = request.brand.flow_account_switch
+        if not flow:
+            return TemplateResponse(
+                request,
+                "if/error.html",
+                {
+                    "title": _("Account switching disabled"),
+                    "message": _("Account switching is disabled for this brand."),
```
gergosimonyi:
```suggestion
                    "message": _("Account switching is disabled."),
```

authentik/core/views/account_switch.py
```diff
@@ -0,0 +1,80 @@
...
+        sessions = (
+            AuthenticatedSession.objects.filter(
+                browser_key=browser_key,
+                session__expires__gt=timezone.now(),
+                user__is_active=True,
+            )
+            .select_related("session", "user")
+            .order_by("-session__last_used")
+        )
+        return next((session for session in sessions if session.user.uid == user_uid), None)
```
gergosimonyi: Why is this done on the python side?

authentik/core/views/account_switch.py
```diff
@@ -0,0 +1,80 @@
...
+        session = self.get_browser_session(request, user_uid)
+        stale_user_uid = None
+        if session:
+            context.update(
+                {
+                    PLAN_CONTEXT_PENDING_USER: session.user,
+                    PLAN_CONTEXT_IS_ACCOUNT_SWITCH: True,
+                }
+            )
+            if request.user.is_authenticated:
```
gergosimonyi: Feels like this should be a guard clause above. "Switch" doesn't seem to make sense when there's nothing to switch from.

authentik/core/views/account_switch.py
```diff
@@ -0,0 +1,80 @@
...
+                context[PLAN_CONTEXT_ACCOUNT_SWITCH_FROM_USER] = request.user
+        else:
+            stale_user_uid = user_uid
+        planner = FlowPlanner(flow)
+        # The context decides which stages policies skip, so cached plans don't apply
```
gergosimonyi: It's not clear to me what this means. Please elaborate.

authentik/core/views/account_switch.py
```diff
@@ -0,0 +1,80 @@
...
+        planner.use_cache = False
+        try:
+            plan = planner.plan(request, context)
+        except FlowNonApplicableException:
+            raise Http404 from None
+        response = plan.to_redirect(request, flow)
+        if stale_user_uid:
```
gergosimonyi: I believe this should also be a guard clause. Again, as far as the backend is concerned, "stale user" means "no user".

authentik/root/middleware.py
```diff
@@ -28,6 +30,13 @@
 ACR_AUTHENTIK_SESSION = "goauthentik.io/core/default"
 SIGNING_HASH = sha512(settings.SECRET_KEY.encode()).hexdigest()
 
+# Opaque browser identifier used to group logins for account switching.
+# Unlike the session cookie it survives logins and logouts.
+COOKIE_NAME_BROWSER = "authentik_browser"
+COOKIE_NAME_ACCOUNTS_LEGACY = "authentik_accounts"
+BROWSER_KEY_LENGTH = 32
+BROWSER_COOKIE_AGE = int(timedelta(days=365).total_seconds())
```
gergosimonyi: Use a common prefix for these please

authentik/root/middleware.py
```diff
@@ -28,6 +30,13 @@
 ACR_AUTHENTIK_SESSION = "goauthentik.io/core/default"
 SIGNING_HASH = sha512(settings.SECRET_KEY.encode()).hexdigest()
 
+# Opaque browser identifier used to group logins for account switching.
+# Unlike the session cookie it survives logins and logouts.
+COOKIE_NAME_BROWSER = "authentik_browser"
+COOKIE_NAME_ACCOUNTS_LEGACY = "authentik_accounts"
```
gergosimonyi: Is this not just a previous name within this PR? If so, please delete.

authentik/root/middleware.py
```diff
@@ -145,6 +197,24 @@ def process_response(self, request: HttpRequest, response: HttpResponse) -> Http
                         httponly=settings.SESSION_COOKIE_HTTPONLY or None,
                         samesite=same_site,
                     )
+                    # Keep the browser cookie longer-lived than the sessions it groups.
+                    # It is intentionally not deleted with the session cookie.
+                    if getattr(request, "browser_key", None):
```
gergosimonyi: Why `getattr`? When can `request.browser_key` not exist? It was set just above.

```suggestion
                    if request.browser_key:
```

authentik/stages/user_login/stage.py
```diff
@@ -138,6 +140,22 @@ def is_known_device(self, user: User):
             self.logger.info("eh", exc=exc)
             return False
 
+    def detach_session(self):
+        """Continue on a new session so the current login stays usable as a switch target."""
+        old_session = self.request.session
+        flow_get = old_session.get(SESSION_KEY_GET)
```
gergosimonyi: Rename please (or alternatively: can be inlined)

authentik/stages/user_login/stage.py
```diff
@@ -138,6 +140,22 @@ def is_known_device(self, user: User):
...
+        # Move the active plan so audit events keep context and old logins cannot resume it.
+        if SESSION_KEY_PLAN in old_session:
+            del old_session[SESSION_KEY_PLAN]
+        old_session.save()
+        self.request.session = SessionStore(
+            last_ip=ClientIPMiddleware.get_client_ip(self.request),
+            last_user_agent=self.request.META.get("HTTP_USER_AGENT", ""),
```
gergosimonyi: These are already computed on the old session, no?

authentik/brands/models.py
dominic-r: 6345da0551

authentik/core/models.py
dominic-r: e3165c8dcb

authentik/core/models.py
dominic-r: 3b7ac61240

authentik/core/models.py
dominic-r: c4e144af05

authentik/core/models.py
dominic-r: 0ee28a3ec0

authentik/core/models.py
dominic-r: 0ee28a3ec0 + 1a67de998a

authentik/core/sessions.py
dominic-r: 11f498b31c + 8dbdba5987

authentik/core/urls.py
dominic-r: switching to something which seems more sensible while also being exposed to the frontend without needing an extra mgiration (also backend now filters by user_id directly instead of iterating sessions and comparing uid which is a nice little simplification)

dominic-r: eafe2e73fe

authentik/flows/planner.py
dominic-r: 3d5f2d733a

authentik/core/views/account_switch.py
dominic-r: 01b23af5e6

authentik/core/views/account_switch.py
dominic-r: eafe2e73fe

authentik/core/views/account_switch.py
dominic-r: c52cab6476.

authentik/core/views/account_switch.py
dominic-r: 4b69fc2bfa

authentik/core/views/account_switch.py
dominic-r: 4574bd4b6c

authentik/root/middleware.py
dominic-r: 1101f40e6d

authentik/root/middleware.py
dominic-r: 1101f40e6d

authentik/root/middleware.py
dominic-r: fd92d63bf4

authentik/stages/user_login/stage.py
dominic-r: 526af5358e

authentik/stages/user_login/stage.py
dominic-r: ec15a98693

authentik/core/sessions.py
```diff
@@ -1,19 +1,64 @@
...
+
+    browser_key = SessionMiddleware.ensure_browser_key(request)
+    if browser_key:
+        AuthenticatedSession.objects.filter(browser_key=browser_key).exclude(
+            session=authenticated_session.session
+        ).update(is_current=False)
+        if authenticated_session.browser_key != browser_key or not authenticated_session.is_current:
+            authenticated_session.browser_key = browser_key
+            authenticated_session.is_current = True
+            authenticated_session.save(update_fields=["browser_key", "is_current"])
```
BeryJu: should probably be an atomic transaction

authentik/core/sessions.py
```diff
@@ -1,19 +1,64 @@
...
+    from authentik.core.models import AuthenticatedSession, User
+
 LOGGER = get_logger()
 
 
+def authenticated_session_from_request(
+    request: HttpRequest, user: User
+) -> AuthenticatedSession | None:
+    """Create a new authenticated session from an HTTP request."""
+    from authentik.core.models import AuthenticatedSession, Session
```
BeryJu: I think I'd prefer to keep this on the AuthenticatedSession object...?

authentik/core/views/account_switch.py
```diff
@@ -0,0 +1,99 @@
...
+        if not flow:
+            return TemplateResponse(
+                request,
+                "if/error.html",
+                {
+                    "title": _("Account switching disabled"),
+                    "message": _("Account switching is disabled."),
+                },
+                status=400,
+            )
```
BeryJu: `bad_request_message()` from authentik.lib.views

authentik/core/views/account_switch.py
```diff
@@ -0,0 +1,99 @@
...
+        if not request.user.is_authenticated:
```
BeryJu: I think we can just have the LoginRequired Mixin on this maybe?

authentik/core/views/account_switch.py
```diff
@@ -0,0 +1,99 @@
...
+        # Reusing a cached plan would skip that planning pass.
+        planner.use_cache = False
+        try:
+            plan = planner.plan(request, context)
+        except FlowNonApplicableException:
+            raise Http404 from None
+        response = plan.to_redirect(request, flow)
+        if stale_user_pk:
+            separator = "&" if "?" in response["Location"] else "?"
+            response["Location"] += separator + urlencode({QS_ACCOUNT_SWITCH_STALE: stale_user_pk})
```
BeryJu: this is valid but just not nice

BeryJu: does this `QS_ACCOUNT_SWITCH_STALE` need to be a query string? since we're planning the flow right above, can't we just have it in the flow context?

web/src/components/ak-account-switcher.ts
```diff
@@ -0,0 +1,179 @@
...
+        return new URLSearchParams({ next }).toString();
+    }
+
+    protected accountSwitchURL(account: BrowserLocalAccount): string {
+        return `${globalAK().api.base}account/switch/${account.pk}/?${this.nextQuery}`;
+    }
+
+    protected addAccountURL(): string {
+        return `${globalAK().api.base}flows/-/default/authentication/?${this.nextQuery}`;
+    }
```
BeryJu: I really dislike this, at the least it should be URL-joined correctly, but better yet those URLs come from the server

BeryJu (commented):
- not a fan of the "browser_key" term
- do we want to configure a default switch flow? even if its the authentication flow?
- still think this should maybe be called user switching?

dominic-r:
pffff

authentik/core/views/user_switch.py
```diff
@@ -0,0 +1,82 @@
...
+)
+from authentik.flows.stage import PLAN_CONTEXT_PENDING_USER_IDENTIFIER
+from authentik.lib.views import bad_request_message
+
+
+class UserSwitchView(LoginRequiredMixin, View):
+    """Authenticate another login held by this browser."""
+
+    def get(self, request: HttpRequest, user_pk: int) -> HttpResponse:
+        flow = request.brand.flow_user_switch
```
BeryJu: Probably check if flow is applicable to current user..?

authentik/core/models.py
```diff
@@ -1315,6 +1316,13 @@ class Meta:
         verbose_name_plural = _("Property Mappings")
 
 
+class SessionSuperseded(SuspiciousOperation):
```
BeryJu: does this need to be in models? minor nit but the file is already too long

authentik/core/models.py
```diff
@@ -1394,14 +1402,84 @@ def serializer(self) -> type[Serializer]:
...
+        )
+        token = user_switching.ensure_request_token(request)
+        if token:
+            authenticated_session.bind_to_user_switching_token(token)
+        return authenticated_session
+
+    #: Attempts to win the current-session slot before giving up on a concurrent bind.
+    USER_SWITCHING_BIND_ATTEMPTS = 3
+
+    def bind_to_user_switching_token(self, user_switching_token: str) -> Self:
```
BeryJu: not sure how I feel about this, I'd prefer to just show an error instead of silently retrying tbh

authentik/core/sessions.py
```diff
@@ -77,6 +33,15 @@ def get_model_class(cls):
     def model_fields(self):
         return [k.value for k in self.model.Keys]
 
+    def _discard_session_on_load_error(self, exc: Exception):
+        """Log why a stored session was rejected and forget its key."""
+        if isinstance(exc, SessionSuperseded):
+            # An expected, routine outcome of user switching.
```
BeryJu: not a fan of having a "routine expected" exception tbh

authentik/flows/challenge.py
```diff
@@ -50,6 +50,7 @@ class ContextualFlowInfo(PassiveSerializer):
     background_themed_urls = ThemedUrlsSerializer(required=False, allow_null=True)
     cancel_url = CharField()
     layout = ChoiceField(choices=[(x.value, x.name) for x in FlowLayout])
+    user_switch_stale_user = CharField(required=False)
```
BeryJu: not a fan of stale but dont have a better suggestion

authentik/core/user_switching.py
```diff
@@ -0,0 +1,72 @@
...
+    try:
+        payload = decode(raw, _SIGNING_HASH, algorithms=["HS256"])
+    except PyJWTError:
+        return None
+    return validate_token(payload.get("user_switching"))
+
+
+def ensure_request_token(request: HttpRequest) -> str | None:
+    """Return the request's user switching token, creating one if the session
+    middleware initialised the request but no token is present yet."""
```
dominic-r:
```suggestion
    middleware initialized the request but no token is present yet."""
```

authentik/core/models.py
dominic-r: Moved it to authentik/core/exceptions.py

Any other ideas on where? Tho this might be fine as other places seem to do this as well.

authentik/core/sessions.py could be better, but sessions.py imports authentik.root.middleware, and root.middleware imports authentik.core.models at module import time. If models.py then imports SessionSuperseded from sessions.py, that seems quite complicated

website/docs/users-sources/user/user-switching.md
```diff
@@ -0,0 +1,88 @@
+---
+title: User switching
+---
+
+authentik can keep multiple users signed in within the same browser and let users switch between them from the User interface header.
```
dewi-tik:
```suggestion
authentik can keep multiple user accounts signed in within the same browser. Users can switch between their user accounts by clicking... <insert short explanation>
```

website/docs/users-sources/user/user-switching.md
```diff
@@ -0,0 +1,88 @@
+---
+title: User switching
```
dewi-tik:
```suggestion
title: User account switching
```

Not strictly essential but useful I think

website/docs/users-sources/user/user-switching.md
```diff
@@ -0,0 +1,88 @@
...
+---
+
+authentik can keep multiple users signed in within the same browser and let users switch between them from the User interface header.
+
+## How it works
+
+Every login creates its own session. authentik groups sessions created by the same browser, so a user can only switch to users that are already signed in from that browser.
```
dewi-tik:
```suggestion
Each login creates its own session. authentik groups sessions created by the same browser, so a user can only switch to user accounts that are already signed in from that browser.
```

website/docs/users-sources/user/user-switching.md
```diff
@@ -0,0 +1,88 @@
...
+Only the most recent login is active at a time. When a user selects another user from the switcher, authentik verifies that the browser has a live session for that user, then starts the configured user switch flow. authentik verifies the target session again immediately before completing the switch. The user must complete the authentication stages required by that flow, such as password or MFA validation.
+
+The previous user's session stays signed in as a switch target, but its session cookie can no longer be used to act as that user. This prevents an old recorded session cookie from being replayed after the browser has switched to another user.
+
+Logins performed through a user switch are recorded in the event log.
+
+## Configuration
+
+User switches authenticate through the brand's **User switch flow**. If no user switch
+flow is set, the switcher is disabled and the switch endpoint cannot be used.
```
dewi-tik, line 26:
```suggestion
User switches authenticate through the brand's **User switch flow**. If no user switch flow is set, the switcher is disabled and the switch endpoint cannot be used.
```

website/docs/users-sources/user/user-switching.md
```diff
@@ -0,0 +1,88 @@
...
+title: User switching
+---
+
+authentik can keep multiple users signed in within the same browser and let users switch between them from the User interface header.
+
+## How it works
+
+Every login creates its own session. authentik groups sessions created by the same browser, so a user can only switch to users that are already signed in from that browser.
+
+Only the most recent login is active at a time. When a user selects another user from the switcher, authentik verifies that the browser has a live session for that user, then starts the configured user switch flow. authentik verifies the target session again immediately before completing the switch. The user must complete the authentication stages required by that flow, such as password or MFA validation.
```
dewi-tik:
```suggestion
Only the most recent login is active at a time. When a user selects another user account from the switcher, authentik verifies that the browser has a live session for that user account, then starts the configured user switch flow. authentik verifies the target session again immediately before completing the switch. The user must complete the authentication stages required by that flow, such as password or MFA validation.
```

website/docs/users-sources/user/user-switching.md
```diff
@@ -0,0 +1,88 @@
...
+
+The previous user's session stays signed in as a switch target, but its session cookie can no longer be used to act as that user. This prevents an old recorded session cookie from being replayed after the browser has switched to another user.
```
dewi-tik:
```suggestion
The previous user's session stays signed in as a switch target, but its session cookie can no longer be used to act as that user account. This prevents an old recorded session cookie from being replayed after the browser has switched to another user account.
```

website/docs/users-sources/user/user-switching.md
```diff
@@ -0,0 +1,88 @@
...
+
+## Configuration
+
+User switches authenticate through the brand's **User switch flow**. If no user switch
+flow is set, the switcher is disabled and the switch endpoint cannot be used.
+
+Users start switches from the User interface header. To add another switch target,
+users select **Add user** from the header menu and complete a normal login in the same
+browser. After the login completes, that user appears in the same menu and can be
+selected as long as the browser still has a live session for that user.
```
dewi-tik:
```suggestion
Users can start an account switch from the User interface header. To add another user account, users select **Add user** from the header menu and complete a normal login. 

After the login completes, that user account appears in the same menu and can be selected as long as the browser still has a live session for that user account.
```

website/docs/users-sources/user/user-switching.md
```diff
@@ -0,0 +1,88 @@
...
+The previous user's session stays signed in as a switch target, but its session cookie can no longer be used to act as that user. This prevents an old recorded session cookie from being replayed after the browser has switched to another user.
+
+Logins performed through a user switch are recorded in the event log.
+
+## Configuration
+
+User switches authenticate through the brand's **User switch flow**. If no user switch
+flow is set, the switcher is disabled and the switch endpoint cannot be used.
+
+Users start switches from the User interface header. To add another switch target,
```
dewi-tik: This section shouldn't be under the Configuration header. Needs its own header.

website/docs/users-sources/user/user-switching.md
```diff
@@ -0,0 +1,88 @@
...
+users select **Add user** from the header menu and complete a normal login in the same
+browser. After the login completes, that user appears in the same menu and can be
+selected as long as the browser still has a live session for that user.
+
+1. Log in to authentik as an administrator and open the Admin interface.
+2. Navigate to **System** > **Brands** and edit the brand you want to configure.
+3. Under **Default flows**, set **User switch flow** to the authentication flow users should complete when switching users.
```
dewi-tik:
```suggestion
3. Under **Default flows**, set **User switch flow** to the authentication flow users should complete when switching user accounts.
```

website/docs/users-sources/user/user-switching.md
```diff
@@ -0,0 +1,88 @@
...
+4. Click **Update**.
+
+Any authentication flow can be used. To reuse the default authentication flow for switching, set it
```
dewi-tik, line 33: Is the default authentication flow that's already set on the brand automatically get set for the user switch flow? If not, maybe it should?

website/docs/users-sources/user/user-switching.md
```diff
@@ -0,0 +1,88 @@
...
+explicitly as the **User switch flow**.
```
dewi-tik, line 34:
```suggestion
Any authentication flow can be used. To reuse the default authentication flow for switching, set it explicitly as the **User switch flow**.
```

website/docs/users-sources/user/user-switching.md
```diff
@@ -0,0 +1,88 @@
...
+
+During a user switch, the selected user is already known and is available to the flow as
+`pending_user`. To skip a stage only during user switches, bind an expression policy to that
+stage binding:
```
dewi-tik:
```suggestion
During a user switch, the selected user account is already known and is available to the flow as `pending_user`. To skip a stage only during user switches, bind an expression policy to that stage binding:
```

website/docs/users-sources/user/user-switching.md
```diff
@@ -0,0 +1,88 @@
...
+```python
+flow_plan = request.context.get("flow_plan")
+is_user_switch = bool(flow_plan and flow_plan.context.get("user_switch_from_user"))
+
+return not is_user_switch
+```
+
+Bind the policy to the identification, password, authenticator validation, or other stage binding
+depending on what users should confirm when they switch back to a signed-in user. Without these
+policies, the selected user switch flow runs in full.
```
dewi-tik:
```suggestion
Bind the policy to the identification, password, authenticator validation, or other stage bindings to control which stages are shown to users when they switch to another user account. 

Without these policies, the selected user switch flow runs in full.
```

website/docs/users-sources/user/user-switching.md
```diff
@@ -0,0 +1,88 @@
...
+
+## Require only MFA for recently used users
+
+To require only MFA when the selected user was used in the last 24 hours, create a dedicated
+user switch flow with a Password stage, an Authenticator Validation stage, and a User Login stage.
+Then bind the following expression policy to the Password stage binding:
```
dewi-tik:
```suggestion
To require only MFA when the selected user account was used in the last 24 hours, create a dedicated user switch flow with a Password stage, an Authenticator Validation stage, and a User Login stage.

Then bind the following expression policy to the Password stage binding:
```

website/docs/users-sources/user/user-switching.md
```diff
@@ -0,0 +1,88 @@
...
+    session__expires__gt=now,
+    session__last_used__gte=now - timedelta(hours=24),
+).exists()
+
+return not recently_used
+```
+
+When the policy returns `False`, authentik skips the Password stage and continues to the
+Authenticator Validation stage. When no live session for that user has been used in the last
+24 hours, the Password stage runs and the user completes the full password plus MFA flow.
```
dewi-tik:
```suggestion
When the policy returns `False`, authentik skips the Password stage and continues to the Authenticator Validation stage. 

When the user account has not had an active session in the past 24 hours, the Password stage is triggered and they must complete the full flow.
```

dewi-tik (dismissed):

authentik/core/migrations/0061_userswitchingsession_and_more.py
```diff
@@ -0,0 +1,57 @@
+# Generated by Django 5.2.15 on 2026-07-09 20:22
```
gergosimonyi: Please merge this migration with the previous one if at all possible.

authentik/core/models.py
```diff
@@ -1368,27 +1369,66 @@ class AuthenticatedSession(SerializerModel):
...
 
+    user_switching_session = models.ForeignKey(
+        user_switching.UserSwitchingSession,
+        null=True,
+        on_delete=models.SET_NULL,
+        related_name="authenticated_sessions",
+    )
+
+    @property
+    def is_superseded(self) -> bool:
```
gergosimonyi: Not in love with naming this `superseded`. Perhaps use the inverse and name it `current`?

gergosimonyi (commented):
Backend-wise I'm okay with this, would like @BeryJu to do a second pass. (Also please resolve the conflicts.)

authentik/core/api/users.py
```diff
@@ -451,13 +463,43 @@ class Meta:
...
 
+class UserSwitchTargetSerializer(PassiveSerializer):
+    """Live user session available to the current browser."""
+
+    pk = IntegerField(read_only=True)
+    username = CharField(read_only=True)
+    name = CharField(read_only=True, allow_blank=True)
+    email = CharField(read_only=True, allow_blank=True)
+    avatar = CharField(read_only=True)
+    is_current = BooleanField(read_only=True)
```
BeryJu: maybe merge this with UserSelfSerializer and add an is_current flag there, instead of having duplicate users in the SessionUserSerializer

authentik/core/api/users.py
```diff
@@ -451,13 +463,43 @@ class Meta:
...
+
+
+class UserSwitchSerializer(PassiveSerializer):
+    """Request to add or switch users in the current browser."""
+
+    action = ChoiceField(choices=["add", "switch"], default="switch")
```
BeryJu: should probably be models.TextChoices for schema sanity

authentik/core/api/users.py
```diff
@@ -624,6 +666,30 @@ def get_ql_fields(self):
...
+    ) -> dict[str, str | int | bool]:
+        """Serialize the small user shape needed by the browser switcher."""
+        return {
+            "pk": user.pk,
+            "username": user.username,
+            "name": user.name,
+            "email": user.email,
+            "avatar": get_avatar(user, self.request),
+            "is_current": is_current,
+        }
```
BeryJu: This shouldn't be needed, can pass the user instance into the serializer and it'll do this, and pass is_current via context

authentik/core/api/users.py
```diff
@@ -661,6 +727,33 @@ def get_queryset(self):
...
+        request=UserSwitchSerializer,
+        responses={200: UserSwitchResponseSerializer},
+    )
+    @action(
+        detail=False,
+        methods=["POST"],
+        permission_classes=[IsAuthenticated],
+        pagination_class=None,
+        filter_backends=[],
+        url_path="switch",
```
BeryJu: I think this is redundant if it matches the function name

authentik/core/api/users.py
```diff
@@ -661,6 +727,33 @@ def get_queryset(self):
...
 
+    @extend_schema(
+        parameters=[OpenApiParameter("next", str, required=False)],
+        request=UserSwitchSerializer,
+        responses={200: UserSwitchResponseSerializer},
+    )
+    @action(
+        detail=False,
+        methods=["POST"],
+        permission_classes=[IsAuthenticated],
```
BeryJu, line 702: Please test with low-permission user

authentik/core/api/users.py
```diff
@@ -661,6 +727,33 @@ def get_queryset(self):
...
+        pagination_class=None,
+        filter_backends=[],
+        url_path="switch",
+    )
+    @validate(UserSwitchSerializer)
+    def switch(self, request: Request, body: UserSwitchSerializer) -> HttpResponse | Response:
+        """Start browser user switching."""
+        if body.validated_data["action"] == "add":
```
BeryJu: remark: not sure how I feel about this (returning responses and then checking the code) but technically fine

authentik/core/tests/test_sessions.py
```diff
@@ -1,205 +1,82 @@
...
+        for invalid in (None, "", valid, "too-short", "!" * user_switching.TOKEN_LENGTH):
+            self.assertIsNone(user_switching.decode_cookie(invalid))
 
-        response = SessionMiddleware(view)(RequestFactory().get("/"))
 
-        cookie = response.cookies.get(user_switching.COOKIE_NAME)
-        self.assertIsNotNone(cookie)
-        self.assertIsNotNone(user_switching.decode_cookie(cookie.value))
-        self.assertFalse(cookie["httponly"])
+class TestUserSwitchingConcurrency(TransactionTestCase):
```
BeryJu: remark: I generally prefer there to be a single test case class per file

authentik/core/tests/test_token_api.py
```diff
@@ -57,6 +57,7 @@ def test_token_set_key(self):
             reverse("authentik_api:token-set-key", kwargs={"identifier": token.identifier}),
             {"key": new_key},
         )
+        print(f"token owner={token.user_id}, acting user={self.admin.pk}")
```
BeryJu:
```suggestion
```

authentik/core/tests/test_user_switch.py
```diff
@@ -43,15 +47,29 @@ def setUp(self):
...
-        """Log the test client in through the flow executor, returning the session key"""
+    def _post_switch(self, user):
+        return self._post({"user_pk": user.pk})
+
+    def _assert_redirect(self, response):
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(
+            response.json()["redirect"],
+            reverse("authentik_core:if-flow", kwargs={"flow_slug": self.flow.slug}),
+        )
```
BeryJu: remark: I personally prefer no helpers in test classes,

authentik/core/views/user_switch.py
```diff
@@ -1,82 +1,114 @@
...
-                session__expires__gt=timezone.now(),
-                user__is_active=True,
-                user_id=user_pk,
-            )
-            .select_related("session", "user")
-            .order_by("-session__last_used")
-            .first()
-        )
+
+def user_add_response(request: HttpRequest) -> HttpResponse:
```
BeryJu: just for consistency sake, please make these CBV

authentik/core/user_switching.py
```diff
@@ -1,41 +1,75 @@
...
+    )
 
+    class Meta:
+        verbose_name = _("User Switching Session")
+        verbose_name_plural = _("User Switching Sessions")
+        default_permissions = []
 
-def generate_token() -> str:
+    def __str__(self) -> str:
+        return f"User Switching Session {self.token[:10]}"
```
BeryJu: For consistency this should be in core/models.py (or a completely separate django app but thats a larger refactor)

authentik/core/user_switching.py
```diff
@@ -1,41 +1,75 @@
...
-    if raw and len(raw) == TOKEN_LENGTH and raw.isalnum():
-        return raw
-    return None
+    if not token:
+        return None
+    if len(token) != TOKEN_LENGTH:
+        return None
+    if not token.isalnum():
+        return None
+    return token
```
BeryJu: I think it's fine to just look it up and fail if it doesn't exist...?

authentik/core/user_switching.py
```diff
@@ -67,6 +101,73 @@ def ensure_request_token(request: HttpRequest) -> str | None:
...
+    return AuthenticatedSession.objects.filter(
+        user_switching_session_id=token,
+        session__expires__gt=timezone.now(),
+        user__is_active=True,
+    )
+
+
+def target_sessions(request: HttpRequest, user_id: int, session_key: str) -> QuerySet:
+    """Filter to an exact live switch target held by this browser."""
+    from authentik.core.models import AuthenticatedSession
```
BeryJu: ideally these functions wouldn't need imports within them

authentik/stages/user_login/stage.py
```diff
@@ -175,11 +185,38 @@ def do_login(self, request: HttpRequest, remember: bool | None = None) -> HttpRe
...
+        # would fail because the user doesn't exist. This is just a fallback
+        # for a misconfiguration to prevent an exception.
+        if user.pk is None:
+            self.logger.debug("pending user is not saved, refusing to log in")
+            return self.executor.stage_invalid()
+        if PLAN_CONTEXT_USER_SWITCH_TARGET_SESSION in self.executor.plan.context:
+            target_session = self._get_valid_user_switch_target(user)
+            if not target_session:
+                self.logger.warning("User switch target session is no longer valid.")
+                return self.executor.stage_invalid()
```
BeryJu: maybe include a message for the user in stage_invalid()

dominic-r:
didn't mean to click the button anywys. gotta find a way to hide copilot

authentik/core/user_switching.py
```diff
@@ -0,0 +1,149 @@
...
+from django.utils import timezone
+from jwt import PyJWTError, decode, encode
+
+from authentik.core.models import AuthenticatedSession, UserSwitchingSession
+from authentik.lib.generators import generate_id
+from authentik.lib.utils.crypto import get_cookie_signing_key
+from authentik.policies.types import PolicyRequest
+
+TOKEN_LENGTH = 32
+COOKIE_NAME = "authentik_user_switching"
```
BeryJu: should be in settings.py

authentik/core/user_switching.py
```diff
@@ -0,0 +1,149 @@
...
+COOKIE_AGE = int(timedelta(days=365).total_seconds())
```
BeryJu: should probably be a config option

BeryJu (approved):

PR #22670, opened 2026-05-26
providers/oauth2: evaluate regex redirect URIs for CORS

ggilley:
The problem is to that the regex in the oauth provider don't work because it fails soon after because of CORS. CORS doesn't match the regex to approve the origin.

## Summary
- Fix CORS on OAuth2 token, userinfo, and OpenID configuration endpoints when redirect URIs use regex matching mode
- Browser-based OIDC clients (e.g. oidc-client-js) can now complete token exchange from origins matched by regex redirect URI patterns

## Problem
Redirect URI regex validation uses `re.fullmatch` during authorize/token body validation, but CORS used literal `urlparse()` hostname comparison against the raw pattern string. Regex patterns never matched real `Origin` headers, causing `Access-Control-Allow-Origin` to be omitted even after a successful authorize redirect.

## Changes
- Add `origin_matches_redirect_uri()` helper that evaluates strict and regex redirect URI entries for CORS
- Pass `RedirectURI` objects to `cors_allow()` instead of raw URL strings
- Add unit tests covering strict mode, regex match, regex mismatch, and OPTIONS preflight

## Test plan
- [ ] `make test-authentik authentik.providers.oauth2.tests.test_cors`
- [ ] Configure OAuth2 provider with regex redirect URI, complete browser OIDC login from matching preview origin
- [ ] Verify strict redirect URIs still work (regression)

Made with [Cursor](https://cursor.com)


ggilley:
Any chance someone can look at this? I have dynamic PR environments and I need the wildcards to work :-)

dewi-tik:
Hi @ggilley. Thank for the contribution. Can you please rebase this on .8?

ggilley:
@dewi-tik done. Thanks for looking at it!

BeryJu:
closing in favour of https://github.com/goauthentik/authentik/pull/22179 since that came first

PR #22671, opened 2026-05-26
tasks: preserve task history on related object delete

dominic-r:
## Details

This fixes an intermittent outpost deletion failure seen from the Terraform provider acceptance tests, where `DELETE /api/v3/outposts/instances/<uuid>/` could return `405 Method Not Allowed` during teardown. The HTTP error was a symptom: authentik was failing the delete transaction at commit time because PostgreSQL rejected a foreign-key violation from `authentik_tasks_tasklog.task_id` to `authentik_tasks_task.id`.

The root cause is that task history was modeled as a `GenericRelation` on `TasksModel` and `Schedule`. Django treats that relation as cascade-owned by the related object, but task rows are history/audit records, not child state owned by an outpost or schedule. Deleting an outpost could therefore delete related `Task` rows while a worker was still inserting or updating `TaskLog` rows for the same task. That race left `TaskLog` pointing at a deleted `Task`, causing the database transaction to fail and surfacing to the API client as the teardown 405.

This PR makes task history query-only from related objects instead of cascade-owned, so deleting an outpost, schedule, or other `TasksModel` object no longer deletes its task history. It also defers outpost teardown side effects until the delete transaction commits and avoids recreating a missing outpost service account during delete cleanup.

Validation run:

- `uv run ruff check authentik/tasks/models.py authentik/tasks/schedules/models.py authentik/tasks/tests/test_worker_middleware.py authentik/outposts/tests/test_api.py`
- `uv run manage.py makemigrations --check --dry-run authentik_tasks authentik_tasks_schedules`
- `uv run coverage run manage.py test authentik.tasks.tests.test_worker_middleware.TestWorkerMiddleware --keepdb`
- `uv run coverage run manage.py test authentik.outposts.tests.test_api.TestOutpostServiceConnectionsAPI --keepdb`
- Terraform provider e2e stress run against this source branch: https://github.com/goauthentik/terraform-provider-authentik/actions/runs/26458925287

---

## Checklist

-   [x] Local tests pass (`ak test authentik/`) - targeted tests listed above
-   [x] The code has been formatted (`make lint-fix`) - ruff check passed for touched files

If applicable


rissson:
So initially this was intended, because once the related objects have been deleted, there is no point in keeping the task history. But I understand the issue, so yeah, I don't really know.

authentik/outposts/signals.py
```diff
@@ -151,13 +152,21 @@ def outpost_reverse_related_post_save(sender, instance: CertificateKeyPair | Bra
...
+
+    def send_teardown():
+        cache.set(CACHE_KEY_OUTPOST_DOWN % outpost_pk, instance)
+        outpost_controller.send_with_options(
+            args=(outpost_pk,),
+            kwargs={"action": "down", "from_cache": True},
+            uid=outpost_name,
+        )
+
+    transaction.on_commit(send_teardown)
```
rissson, line 169: Are we sure we're always in a transaction here?

authentik/tasks/models.py
```diff
@@ -216,13 +216,17 @@ def to_log_event(self) -> LogEvent:
...
 class TasksModel(models.Model):
-    tasks = GenericRelation(
-        Task, content_type_field="rel_obj_content_type", object_id_field="rel_obj_id"
-    )
-
     class Meta:
         abstract = True
 
+    @property
+    def tasks(self):
```
rissson:
```suggestion
    def tasks(self) -> QuerySet[Task]:
```

authentik/outposts/signals.py
dominic-r: Yea. django handles both cases here. inside `atomic()` it queues the callback until the outer transaction commits, and outside an active transaction in normal autocommit mode it executes the callback immediately. So we do not have to always be in a transaction for this to work; the only unsupported case is manual transaction management with autocommit disabled, which this delete path should not be using.

And that keeps the old immediate behavior for non-transactional deletes (while avoiding enqueueing the teardown if a transactional delete rolls back).

rissson:
thoughts from discussion with @BeryJu, we actually prefer leaving this as is, and instead fail silently when creating TaskLog entries where the task does not exist.

PR #22672, opened 2026-05-26
sources/oauth: Fallback to `id` field when `sub` is missing in OIDC callback

v-kamerdinerov:
# Fix: Fallback to `id` field when `sub` is missing in OIDC callback

## Problem

Some OAuth2 providers that expose an OIDC-compatible endpoint do not return the standard `sub` claim in their profile response. Instead, they use a provider-specific `id` field.

Example is **Yandex OAuth** — its profile endpoint returns `id` rather than `sub`, which caused `OpenIDConnectOAuth2Callback.get_user_id()` to always return `None` and authentication to fail with:

```
Authentication Failure: Could not determine id.
```

Also can help with issue - https://github.com/goauthentik/authentik/issues/16115

## Solution

Updated `get_user_id` in `OpenIDConnectOAuth2Callback` to fall back to the `id` field when `sub` is not present:

```python
def get_user_id(self, info: dict[str, str]) -> str | None:
    return str(info.get("sub") or info.get("id"))
```

This keeps full compatibility with standard OIDC providers that return `sub`, while also supporting non-standard providers like Yandex that use `id`.

## Affected providers

- **Yandex OAuth** — returns `id` instead of `sub`
- Any other OAuth2/OIDC provider that uses `id` as the unique user identifier

## Step for reproducing

- Create an OAuth app on the Yandex OAuth provider — https://oauth.yandex.ru/
- Create a Source with type **OpenID Connect OAuth Source**
    - Fill in all required parameters: authorization URL, token URL, etc.
    - Choose your default authentication and enrollment flows
- Try to authenticate with the new application
 
 
Verified against Yandex OAuth source — authentication now completes successfully and the correct user identifier is extracted from the profile response.


BeryJu (approved):

mrekin:
Hi!
May be stupid question - when this fix will be included in any release?

v-kamerdinerov:
> Hi! May be stupid question - when this fix will be included in any release?

Fixed at 2026.8.0!

PR #22691, opened 2026-05-27
Update polish translation

phatterit:
Polish translation updates and missing strings


phatterit (commented):
Update pl translation

gergosimonyi:
[authentik uses transifex for translations](https://docs.goauthentik.io/developer-docs/translation/#online-translation); therefore, the contents of this PR would get overwritten the next time a sync runs from transifex to this repo.

Please add these translation strings over on transifex. Thank you in advance!

PR #22758, opened 2026-05-29
stages/email: fix ungrammatical expiry time in password reset templates

### LINKED ISSUE #22757: fix(email): password_reset templates use naturaltime filter causing ungrammatical output
Discovered while testing a recovery flow - following the [flow docs example](https://docs.goauthentik.io/add-secure-apps/flows-stages/flow/examples/flows/#recovery-with-email-and-mfa-verification) - and briefly discussed in Discord `#general` which allowed to pinpoint the issue.

The default email templates used by the recovery flow (`password_reset.txt` and `password_reset.html`) render the expiry time using Django’s `|naturaltime` filter inside a prepositional phrase. Because Django’s `naturaltime` unconditionally injects its own preposition for future datetimes (e.g. **"in 23 hours"**), the resulting sentence is grammatically incorrect **even in English**:

- **English:** `The above link is valid for **in** 23 hours.`
- **German:** `Der obige Link ist gültig für **in** 23 Stunden.`

Turned out this is a template-level bug, not a translation issue as initially assumed. The translation files (e.g., [de_DE/django.po](https://github.com/goauthentik/authentik/blob/main/locale/de_DE/LC_MESSAGES/django.po#L4758) are correct; the problem is that the variable being interpolated already contains a preposition that clashes with the surrounding sentence.

**To Reproduce**

1. Configure a recovery flow using the documented example / default email stage.
2. Trigger the recovery flow for any user.
3. Inspect the received email.
4. Observe the ungrammatical sentence in the email body.

**Expected behavior**

The sentence should read correctly without clashing prepositions:

- **English:** `The above link is valid for 23 hours.`
- **German:** `Der obige Link ist gültig für 23 Stunden.`

**Root cause**

Django’s `django.contrib.humanize.naturaltime`. For future datetimes it always outputs a phrase like `"in <time>"`. When the template embeds this inside the existing phrase *"valid for …"*, the prepositions collide.

### Link
**Affected files**

- [`authentik/stages/email/templates/email/password_reset.txt`](https://github.com/goauthentik/authentik/blob/main/authentik/stages/email/templates/email/password_reset.txt#L7)
- [`authentik/stages/email/templates/email/password_reset.html`](https://github.com/goauthentik/authentik/blob/main/authentik/stages/email/templates/email/password_reset.html#L39)

### Solution

**2 proposed solutions**

**Option A: Replace `naturaltime` with `timeuntil` (recommended)**

Use Django’s `timeuntil` filter instead, which returns a bare duration string (e.g., `"23 hours"` / `"23 Stunden"`) without any injected prepositions.

```django
{# Before (both templates) #}
{% blocktrans with expires=expires|naturaltime %}
The above link is valid for {{ expires }}.
{% endblocktrans %}

{# After (both templates) #}
{% blocktrans with expires=expires|timeuntil %}
The above link is valid for {{ expires }}.
{% endblocktrans %}
```

This produces correct output in English and all localized languages without needing to restructure the sentence.

**Option B: Restructure the surrounding sentence**

Keep `naturaltime` but change the template text so the preposition becomes semantically correct (e.g., `"This link will expire {{ expires }}"`). However, this is less robust across locales and places the burden on every translation to match the filter's behavior.

### Additional context

- This affects **any language** where Django’s `naturaltime` includes prepositions.
- The bug is reproducible in the default English output; translations only make it more obvious.

---

I believe **Option A** is the cleanest fix and I am happy to open a PR implementing it if the maintainers agree. This would be my first contribution to authentik.

lwillek:
## Details

Django's `naturaltime` filter outputs a prepositional phrase (e.g., **"in 23 hours"**). When embedded in the existing sentence "The above link is valid for {{ expires }}", the prepositions clash, producing ungrammatical output in English ("valid for in 23 hours") as well as in translations (e.g., German: "gültig für in 23 Stunden").

This PR replaces `naturaltime` with `timeuntil` in the default password reset email templates. `timeuntil` returns a bare duration string (e.g., **"23 hours"**), which fits correctly into the existing sentence structure across all locales.

## Changes

- `authentik/stages/email/templates/email/password_reset.txt`
- `authentik/stages/email/templates/email/password_reset.html`

## Related issues

- Closes #22757

---

*This is my first contribution to authentik — please let me know if anything needs adjustment.*


BeryJu (approved):

dominic-r (approved):

lwillek:
Thanks @BeryJu for your review, contribution to this PR and approval! 

However, I think that the "invite" email template does not need a fix, as there the phrasing is different, so there naturaltime is correct and timeuntil would produce awkward output.

The fix only applies to the password reset templates I think the invitation templates should remain unchanged - as they are fine already. Hence I reverted https://github.com/goauthentik/authentik/pull/22758/commits/d069784ff08a70b5728d60da4aadad22181c46ef in https://github.com/goauthentik/authentik/pull/22758/commits/991faf68483f52397cde71a1e101f9010995b66d - I hope this is OK. 

Looking forward for a new review!

BeryJu (approved):

PR #22762, opened 2026-05-31
sources/telegram: restore next= redirect after pre_authentication_flow

### LINKED ISSUE #18450: Authentik redirects to interface instead of app
### Describe the bug

I set up several applications to use authentik as oauth2 provider. Now, after i authorize in authentik social login(telegram) and click continue i'm being redirected not in the app, but to dashboard with all applications. But If i'm already authorized via telegram everything works fine

### How to reproduce

1. Set up oauth in app
2. Click proceed with authentik
3. Log in into authentik
4. Click continue
5. See authentik interface instead of app

### Expected behavior

After clicking continue, open callback url of application

### Screenshots

_No response_

### Additional context

_No response_

### Deployment Method

Docker

### Version

2025.10.1

### Relevant log output

```shell

```

alexeyqu:
## Details
  
  Telegram is the only source that runs authentication through a second flow executor invocation -- the `pre_authentication_flow` hosts the Telegram Login Widget as a stage (`TelegramLoginView`). When FlowExecutorView.dispatch() is called for that second flow, it unconditionally overwrites SESSION_KEY_GET with an empty dict (the redirect URL from TelegramStartView carries no query string), losing the original next= parameter. By the time TelegramLoginView.challenge_valid() runs, the flow manager falls back to "authentik_core:if-user" instead of the original application callback URL.

Fix: restore SESSION_KEY_GET from PLAN_CONTEXT_REDIRECT in the current plan context before constructing TelegramSourceFlowManager. TelegramStartView already saved the correct redirect there before the overwrite -- the same pattern used in `authentik/sources/saml/views.py:174-178`.

A regression test (`test_challenge_view_preserves_next_redirect`) reproduces the exact failure sequence and asserts the enrollment plan carries the correct PLAN_CONTEXT_REDIRECT.

Tested on a live 2026.5.2 deployment: first-time Telegram login inside an OIDC authorization flow now correctly redirects back to the application callback URL instead of /if/user/.

Closes #18450

  ---
  Checklist

  - [x] Local tests pass (ak test authentik/)
  - [x] The code has been formatted (make lint-fix)


authentik/sources/telegram/views.py
```diff
@@ -86,6 +86,14 @@ def challenge_valid(self, response: TelegramChallengeResponse) -> HttpResponse:
         raw_info.pop("hash")
         raw_info.pop("auth_date")
         source = self.source
+        # The pre_authentication_flow executor overwrites SESSION_KEY_GET with an empty
+        # dict (its redirect URL carries no query string), losing the original next= param.
+        # Restore it from PLAN_CONTEXT_REDIRECT, which was saved before that overwrite.
+        # Mirrors the same fix in authentik/sources/saml/views.py:174-178.
+        if SESSION_KEY_PLAN in self.request.session:
```
atereshkin: This check seems redundant.

authentik/sources/telegram/views.py
```diff
@@ -86,6 +86,14 @@ def challenge_valid(self, response: TelegramChallengeResponse) -> HttpResponse:
         raw_info.pop("hash")
         raw_info.pop("auth_date")
         source = self.source
+        # The pre_authentication_flow executor overwrites SESSION_KEY_GET with an empty
+        # dict (its redirect URL carries no query string), losing the original next= param.
+        # Restore it from PLAN_CONTEXT_REDIRECT, which was saved before that overwrite.
+        # Mirrors the same fix in authentik/sources/saml/views.py:174-178.
```
atereshkin: Let's not refer to line numbers in comments since those shift all the time.

atereshkin (commented):
Lgtm, apart from a couple of small nits.

alexeyqu:
Thank you @atereshkin! -- both comments are fixed with [8125d2a](https://github.com/goauthentik/authentik/pull/22762/commits/8125d2a6634a8ccd87bb378358d9437657cb569c). Let me know if anything else attracts your attention.

atereshkin (commented):
Good to go, as far as I'm concerned.

BeryJu (approved):

PR #22765, opened 2026-05-31
providers/scim: add flatten_nested_groups option

leoleoasd:
## Details

When a provider's `flatten_nested_groups` is enabled, group sync expands each group's members to include the direct members of all descendant groups (via the `parents` M2M, with cycle protection). This applies to:

- the initial group resource sent on POST/PUT (`to_schema`)
- the `_patch_add_users` call right after a group is first created
- the diff computed in `patch_compare_users` for ongoing PATCH updates

SCIM endpoints such as AWS IAM Identity Center, Slack and several others do not support nested groups in any form — the parent group on the remote side shows up empty even when authentik's local hierarchy has the relationships set up correctly. This option makes authentik flatten that hierarchy at sync time so the remote sees the same effective membership.

Default is `False` (no behavior change for existing providers). Surfaces in the API serializer and as a switch on the provider form.

Includes tests covering both the flatten=True (nested user reaches parent PATCH) and flatten=False (no leakage) paths.

---

## Checklist

If applicable


leoleoasd:
Hi @BeryJu Can you help take a look?

kensternberg-authentik (approved):
Very straightforward for the client-side code.

dewi-tik:
@leoleoasd Thanks for the contribution. Are you able to look into the conflicts? We can then give you a review if that's alright.

PR #22770, opened 2026-06-01
tests: avoid fixed proxy e2e ports

dominic-r:
## Summary
Fix proxy e2e failures caused by tests binding containers to fixed host ports. The tests now allocate per-test host ports and still route container-to-container traffic over Docker DNS where possible.

The shared Docker cleanup also removes stopped labelled containers, so a failed container start does not leave names behind and cascade into later test failures.

## Related
The separate endpoint e2e failure from blank agent OS versions is handled in goauthentik/platform#960.

## Validation
- `uv run ruff check tests/docker.py tests/e2e/test_provider_proxy.py tests/e2e/test_provider_proxy_forward.py`
- `uv run ruff format --check tests/docker.py tests/e2e/test_provider_proxy.py tests/e2e/test_provider_proxy_forward.py`
- `git diff --check`

### DIFF
--- a/blueprints/schema.json
+++ b/blueprints/schema.json
@@ -12110,11 +12110,6 @@
         "model_authentik_sources_kerberos.groupkerberossourceconnection": {
             "type": "object",
             "properties": {
-                "group": {
-                    "type": "string",
-                    "format": "uuid",
-                    "title": "Group"
-                },
                 "source": {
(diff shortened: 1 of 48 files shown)


BeryJu:
fixed by https://github.com/goauthentik/authentik/pull/22785

PR #22782, opened 2026-06-01
web/admin/stages/captcha: fix float values in score tresholds fix provider parsing

### LINKED ISSUE #22409: Captcha stage creation
### Describe the bug

I can't create any other captcha stage than the google ReCaptcha v2. When I create it with another type of captcha like v3 or Cloudflare Turnstile and hit the save button it always jumps back to Google ReCaptcha v2. I did everything as in the documentation but it doesn't work. 

In the docs it says "Score minimum threshold: 0.5" but in Authentik it is not possible to set it to 0.5 nor 0,5

Image

Image

When using the captcha on my identification stage i get an error as expected, keytype is invalid
(I used v3 credentials)

### How to reproduce

1. Go to Flows and Stages
2. Create new stage.
3. captcha stage
4. Use for example Recaptcha v3
5. Enter values
6. hit save
7. Open edit menu again
8. It should say recaptcha v2 again

### Expected behavior

The Provider type of my captcha should stay the same after hitting save button.

### Screenshots

_No response_

### Additional context

_No response_

### Deployment Method

Docker

### Version

2026.2.3

### Relevant log output

```shell

```

joshua-siw:
## Details

resolves: #22409
Added Property to ak-number-input to allow float values
made detectProviderFromInstance method more distinctive to correctly fill in the previously choosen captcha provider because it is not stored in the database

---

## Checklist

If applicable


kensternberg-authentik (approved):
Not broken. I even checked; `parseFloat("-1.0") === parseInt("-1", 10)`, which is both surprising and delightful.

The number of comparisons in `shared.ts` bothers me; that's a lot of configuration that has to match for confirmation, which is a code smell meaning that there's a missing name or missing matcher for these two objects.

But it's not broken.

joshua-siw:
> The number of comparisons in `shared.ts` bothers me; that's a lot of configuration that has to match for confirmation, which is a code smell meaning that there's a missing name or missing matcher for these two objects.
>

Yes I don't like it as well, I guess an approach would be to store the provider within the database, so the whole function detectProviderFromInstance could be removed (also getCapWidgetUrl), but I didn't wanted to make changes to the scheme, because I didn't knew if the decision to not include the provider in it is intentional.

PR #22828, opened 2026-06-03
ci: harden cherry-pick action against template injection

emilburzo:
As the cherry-pick action uses the potentially dangerous `pull_request_target`, move all interpolated fields to the `env` section, which prevents expanding directly inside the `run` script where command injection can happen.

There is currently no field that can be abused by a fork. The only potentially dangerous one is the label field, e.g. if someone is able to create custom labels, which requires repository permissions. But we should do it for all fields anyway as good practice.

Reference: https://docs.zizmor.sh/audits/#template-injection

---

## Checklist

If applicable

### DIFF
--- a/.github/actions/cherry-pick/action.yml
+++ b/.github/actions/cherry-pick/action.yml
@@ -20,12 +20,22 @@ runs:
       shell: bash
       env:
         GITHUB_TOKEN: ${{ inputs.token }}
+        # Untrusted/event-derived values are passed via the environment (never
+        # interpolated into the script body) to avoid template injection.
+        EVENT_NAME: ${{ github.event_name }}
+        REPOSITORY: ${{ github.repository }}
+        ISSUE_NUMBER: ${{ github.event.issue.number }}
+        LABEL_NAME_CTX: ${{ github.event.label.name }}
+        PR_NUMBER_CTX: ${{ github.event.pull_request.number }}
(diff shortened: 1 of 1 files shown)


.github/actions/cherry-pick/action.yml
```diff
@@ -117,16 +127,18 @@ runs:
...
         if [ "${REASON}" = "label_added_to_merged_pr" ]; then
           # Only process the specific label that was just added
-          if [ "${{ github.event_name }}" = "issues" ]; then
-            LABEL_NAME="${{ github.event.label.name }}"
+          if [ "$EVENT_NAME" = "issues" ]; then
+            LABEL_NAME="$LABEL_NAME_CTX"
           else
-            LABEL_NAME="${{ github.event.label.name }}"
+            LABEL_NAME="$LABEL_NAME_CTX"
           fi
```
emilburzo, line 142: I'm not sure what's going on here, it seems like a mistake?

BeryJu: I think its a leftover from something...?

BeryJu (approved):

.github/actions/cherry-pick/action.yml
emilburzo: handled in a [separate PR](https://github.com/goauthentik/authentik/pull/22859)

PR #22454, opened 2026-05-19
core, docs: Podman Quadlet Support

GirlBossRush:
## Details

This PR adds support for [Quadlet-based](https://podman-desktop.io/blog/podman-quadlet) Podman deployments of authentik.

## Code changes 

— authentik/outposts/tasks.py:192-214: the local-socket auto-detect now also looks for Podman's rootful (`/run/podman/podman.sock`) and rootless (`$XDG_RUNTIME_DIR/podman/podman.sock`) sockets, not just Docker's. A fresh authentik install on a Podman host automatically gets a working local service connection for outpost auto-deploy. The Docker SDK already handles the Podman socket fine; only the discovery hook was Docker-only.

## Documentation changes

- New install guide with full walkthrough: requirements, rootful vs. rootless table, Quadlet .container / .network / .volume units for postgresql/server/worker, upgrade flow, SELinux notes, and a pre-2026.5 fallback section.
- Docker Compose admonition pointing Podman users at the Quadlet guide.
- New "Podman" section in the Docker doc explaining the rootful/rootless socket paths.

PR #22831, opened 2026-06-03
lifecycle/quadlet: unit files, wire into dependabot

### LINKED ISSUE #22829: docs/quadlet: ship unit files under lifecycle/quadlet/ + wire into dependabot and release workflows
Follow-up to #22454, which landed the Podman Quadlet install guide as docs-only. Two reviewer asks from that PR are deferred here because they go beyond documentation:

### Ship the unit files under `lifecycle/quadlet/`

The guide currently inlines the `.pod` / `.container` / `.volume` unit definitions. Extract them into `lifecycle/quadlet/` (mirroring how `lifecycle/container/` holds the compose file) and have the guide reference the committed files instead of inlining them.

### Wire the new directory into dependabot and the release workflows

- Add a dependabot entry for `lifecycle/quadlet/` so the pinned image tags get bumped like the rest.
- Hook the image tags into the release workflows so they bump alongside the compose file on each release.

Originally raised by @dominic-r on #22454 (the `lifecycle/quadlet/` placement and the dependabot/release-workflow comments).

GirlBossRush:
Follow-up to #22454. Closes #22829.

The Podman Quadlet install guide inlined the .pod / .container / .volume definitions. Extract them into committed unit files under lifecycle/quadlet/ (rootful) and lifecycle/quadlet/rootless/ (rootless), the same way lifecycle/container/compose.yml lives in the repo.

- scripts/generate_quadlet.py mirrors generate_compose.py: reads authentik_version() with the same rc-to-previous-patch fallback and writes both unit sets with a concrete pinned image tag.
- Makefile gains gen-quadlet, called from bump alongside gen-compose, so make bump rewrites the tag in the same place the release workflow already runs it (release-tag.yml -> bump-authentik).
- .github/dependabot.yml adds /lifecycle/quadlet and /lifecycle/quadlet/rootless to the docker ecosystem block. An inline comment notes that dependabot does not parse .container files yet (dependabot/dependabot-core#13066), so the entries are a no-op until upstream lands quadlet support; the authentik tag itself bumps via make bump.
- website/Dockerfile and docs/docusaurus.config.esm.mjs copy the directory into the docs build so units are served at docs.goauthentik.io/quadlet/{name} and /quadlet/rootless/{name}, matching the compose.yml pattern.
- podman-quadlet.mdx replaces the inline blocks with a curl loop per install mode and keeps the brief notes on the worker port shift and User=0:0.

Drive-by from the previous round: the .pod unit's Description was "authentik server", which collided with the server container's own Description. Fixed to "authentik" on authentik.pod.

The rootless set ships full units rather than only diffing the worker, which also corrects the EnvironmentFile path the inlined doc had under the Rootless tab (the postgres and server containers used the rootful /etc/authentik/authentik.env path).


dominic-r (approved):
will push directly any updates to main pr

PR #22892, opened 2026-06-05
web/maintenance: eliminate the need for DEFAULT_CONFIG boilerplate everywhere

kensternberg-authentik:
web/maintenance: eliminate the need for DEFAULT_CONFIG boilerplate everywhere

# Promise:

There is only one file change in the entire PR. I promise. It is `./web/src/common/api/client.ts`.
Every other change is mechanical.  See the guideposts I provided below; they'll take you straight to that file.

# What

Replace the repetitive boilerplate of importing DEFAULT_CONFIG everywhere by replacing it with an enclosing function that does both instantiation and configuration for you.

# How !?!?!?! (I hear you cry)

Read the `client.ts` file. It shows you the new mechanism.

# Mechanical update:

Every other change to the source code was performed with the following seven-line script: use `prettier --print-width 400` to ensure all the construction expressions would be single-line (they weren’t necessarily always so); identify the files that used `DEFAULT_CONFIG`, replace their `import` statements, replace their instantiation expressions, then clean up the debris.

    $ ./node_modules/.bin/prettier --cache --write -u . --print-width 400
    $ rg -l 'import.*\{ DEFAULT_CONFIG \}' ./src > client-only.txt
    $ perl -pi.bak -e 's{import \{ DEFAULT_CONFIG \} from "\#common/api/config"}{import \{ aki \} from "#common/api/client"}' $(cat client-only.txt )
    $ perl -pi.bak -e 's{new ([A-Z][A-Za-z]+Api)\(DEFAULT_CONFIG\)}{aki(\1)}g' $(cat client-only.txt )
    $ rm client-only.txt
    $ find . -name '*.bak' -exec rm {} \;
    $ npm run prettier

---

## Checklist

-   [🌺] The project has been linted, built, and tested (`make all`)
-   [🌺] The documentation has been updated and formatted (`make docs`)


web/src/common/api/client.ts
```diff
@@ -0,0 +1,47 @@
...
+    LoggingMiddleware,
+} from "#common/api/middleware";
+import { globalAK } from "#common/global";
+import { SentryMiddleware } from "#common/sentry/middleware";
+
+import { CapabilitiesEnum, Configuration } from "@goauthentik/api";
+
+type APIConstructor<T> = new (config: Configuration) => T;
+
+let configuration: Configuration | null = null;
```
kensternberg-authentik, line 15: Cache the configuration formerly known as `DEFAULT_CONFIG`.

web/src/common/api/client.ts
```diff
@@ -0,0 +1,47 @@
...
+
+const endpoints = new Map<APIConstructor<unknown>, unknown>();
```
kensternberg-authentik, line 17: Once instantiated, the API endpoints are static objects maintaining only their own configuration details.  Cache them to make them re-usable.

web/src/common/api/client.ts
```diff
@@ -0,0 +1,47 @@
...
+
+function apiConfiguration(): Configuration {
+    if (!configuration) {
```
kensternberg-authentik, line 20: Build and cache the `DEFAULT_CONFIG` if it hasn't been built yet, otherwise return the existing one.

web/src/common/api/client.ts
```diff
@@ -0,0 +1,47 @@
...
+                    : []),
+            ],
+        });
+        Object.freeze(configuration);
+    }
+    return configuration;
+}
+
+export function aki<T>(APIClass: APIConstructor<T>): T {
+    let endpoint = endpoints.get(APIClass) as T | undefined;
```
kensternberg-authentik, line 41: For a given API class-level constructor passed in, see if there's an instance already. If not, build one, cache it. Return the instance.

web/src/common/api/client.ts
```diff
@@ -0,0 +1,47 @@
+import {
```
kensternberg-authentik, line 1: THIS IS THE NEW API INSTANTIATION API.

BeryJu (approved):

PR #22897, opened 2026-06-05
Web/maintenance/explicit model endpoints

kensternberg-authentik:
web/maintenance: eliminate boilerplate methods from ModelForm

## What

Provide an alternative syntax for defining the load/create/update methods in ModelForm.

Provide a default message for `getSuccessMessage`: `Successfully created` or `Successfully updated` plus the verboseName, if defined.

The old syntax still exists, and is required for a number of complicated and crusty forms with multiple fetches or unusual return values.

`getSuccessMessage()` can still be overridden, if you feel it’s necessary.

### Retrieve / Create / Update

This builds on the insights from the `aki()` authentik API wrapper: find a way to wrap boilerplate in a function. Every ModelForm supplies three methods: `load()`, `send()`, and `getSuccessMessage()`. `send()` is secretly two different methods in a trenchcoat: it’s `create()` and `update()`, depending on whether or not the object has an instancePk.

This commit adds a new field to ModelForm: `endpoints?`, which contains fields for `load`, `create`, and `update`; these contain *only* the minimal code needed to perform these duties. For example, the FlowForm’s endpoints look like this:

       protected endpoints = {
           load: (slug: string) => aki(FlowsApi).flowsInstancesRetrieve({ slug }),
           create: (flowRequest: Flow) => aki(FlowsApi).flowsInstancesCreate({ flowRequest }),
           update: (slug: string, flowRequest: Flow) =>
               aki(FlowsApi).flowsInstancesUpdate({ slug, flowRequest }),
       };

New concrete implementations of `load` and `send` in FlowForm, if not overridden (as they typically are now), look for `endpoints.load`, `endpoints.create`, and (if `instancePk` is set) `endpoints.update`, and call the functions defined in `endpoints`, providing the same functionality as before. Both calls have exception handlers for “Neither overridden nor properly parameterized”, preserving the existing error condition semantics.

### Success… messages.

`getSuccessMessage()` was just a funny fix: every form defines a `verboseName`; it exists to populate this message: `An error occurred while loading ${this.verboseName}.` That’s it. That’s all it does.

And then every form implements this stanza:

        public override getSuccessMessage(): string {
            return this.instance
                ? msg("Successfully updated flow.")
                : msg("Successfully created flow.");
        }

So I provided a better default in ModelForm.

        public override getSuccessMessage() {
            if (!this.verboseName) return super.getSuccessMessage();
            return this.instancePk === null
                ? msg(str`Successfully created ${this.verboseName}`)
                : msg(str`Successfully updated ${this.verboseName}`);
        }

Define the verboseName and don’t worry, be happy:

## Checklist

-   [🌤️] The project has been linted, built, and tested (`make all`)
-   [🌤️] The documentation has been updated and formatted (`make docs`)


web/src/elements/forms/ModelForm.ts
```diff
@@ -96,7 +107,30 @@ export abstract class ModelForm<
...
+            ? this.endpoints.create(data)
+            : this.endpoints.update(this.instancePk, data);
+    }
+
+    public override getSuccessMessage() {
+        if (!this.verboseName) return super.getSuccessMessage();
+        return this.instancePk === null
+            ? msg(str`Successfully created ${this.verboseName}`)
+            : msg(str`Successfully updated ${this.verboseName}`);
+    }
```
GirlBossRush, line 133: I want to say that we some machinery for this sort of thing,


https://github.com/goauthentik/authentik/blob/main/web/src/elements/forms/Form.ts#L392-L428


Aaaand, it lacks a past tense `submittedVerb` and `formatSubmittedLabel`.

web/src/admin/flows/FlowForm.ts
```diff
@@ -41,32 +41,12 @@ export class FlowForm extends WithCapabilitiesConfig(ModelForm<Flow, string>) {
...
-        return this.#api.flowsInstancesCreate({
-            flowRequest: data,
-        });
-    }
+    protected endpoints = {
+        load: (slug: string) => aki(FlowsApi).flowsInstancesRetrieve({ slug }),
+        create: (flowRequest: Flow) => aki(FlowsApi).flowsInstancesCreate({ flowRequest }),
+        update: (slug: string, flowRequest: Flow) =>
+            aki(FlowsApi).flowsInstancesUpdate({ slug, flowRequest }),
+    };
```
GirlBossRush, line 49: Something about this has seems like a lateral move vs a relief from the underlying problem, but I'm having trouble verbalizing my thoughts.

GirlBossRush (commented):
There's a definitive improvement here, but I'm not quite sure what gives me apprehension. Let's collab on it :)

GirlBossRush (approved):

PR #22903, opened 2026-06-07
lib/sync/outgoing: make outgoing sync cleanup provider-agnostic

czxtm:
### What does this PR change?

`SyncTasks._sync_cleanup` (in `authentik/lib/sync/outgoing/tasks.py`) is shared across all outgoing-sync providers (SCIM, Google Workspace, Microsoft Entra), but it referenced `connection.scim_id` directly.

This PR adds `BaseOutgoingSyncClient.get_identifier()`, which resolves the remote identifier generically (the single non-relational `TextField` on the connection model), and uses it in `_sync_cleanup` instead of the hard-coded `scim_id`.

Only the SCIM connection models (`SCIMProviderUser` / `SCIMProviderGroup`) have a `scim_id` field. Google Workspace uses `google_id` and Microsoft Entra uses `microsoft_id`. As a result, the cleanup step crashed for those providers:

```
AttributeError: 'GoogleWorkspaceProviderUser' object has no attribute 'scim_id'
```

Stack:
```
authentik/enterprise/providers/google_workspace/tasks.py:20  google_workspace_sync
authentik/lib/sync/outgoing/tasks.py:106                     sync
authentik/lib/sync/outgoing/tasks.py:128                     _sync_cleanup
```

This meant out-of-scope users/groups were never cleaned up for Google Workspace or Microsoft Entra providers.

### How was this tested?

- Added a regression test `GoogleWorkspaceUserTests.test_sync_cleanup` that creates an out-of-scope connection and asserts the sync cleanup deletes it without raising a `SYSTEM_EXCEPTION`.
- Verified the fix against a live Google Workspace provider: `_sync_cleanup` now completes and correctly resolves `google_id`, where it previously raised the `AttributeError`.
- `ruff check` and `ruff format --check` pass on all changed files.

### Linked issues

---

## Checklist


authentik/lib/sync/outgoing/base.py
```diff
@@ -53,6 +53,21 @@ def __init__(self, provider: TProvider):
...
+        TextField (scim_id / google_id / microsoft_id depending on provider).
+        Resolve it generically so cleanup works across all providers."""
+        for field in connection._meta.get_fields():
+            if (
+                field.get_internal_type() == "TextField"
+                and not field.is_relation
+                and field.name != "attributes"
+            ):
+                return getattr(connection, field.name)
+        raise AttributeError(f"No remote identifier field found on {type(connection).__name__}")
```
rissson: I'd rather override this in each of the dependent classes, instead of doing this

rissson (changes requested):

authentik/lib/sync/outgoing/base.py
czxtm: good call thats way better, updated.

rissson (approved):

rissson (approved):

PR #22906, opened 2026-06-07
providers/scim: Add GitLab compatibility mode

dominic-r:
Add a GitLab SCIM compatibility mode that skips ServiceProviderConfig probing, which GitLab does not return as SCIM JSON.

This also wraps non-JSON SCIM responses as SCIM request failures so the existing ServiceProviderConfig fallback path can handle HTML redirects/defaults consistently.

Validation:
- make gen
- uv run ak makemigrations --check --dry-run authentik_providers_scim
- uv run coverage run manage.py test --keepdb authentik/providers/scim/tests/test_client.py
- uv run ruff check authentik/providers/scim/clients/base.py authentik/providers/scim/models.py authentik/providers/scim/tests/test_client.py authentik/providers/scim/migrations/0020_alter_scimprovider_compatibility_mode.py
- git diff --check


BeryJu (changes requested):
Looks good, aside from the migration; since the choices are not actually saved in the database this is the one case where editing a previous migration to update the choices is fine (and adding an extra migration just slows down first start even more)

dominic-r:
Fixed!

BeryJu (approved):

PR #22938, opened 2026-06-08
website/docs: add Splunk event forwarding docs

### LINKED ISSUE #22223: Splunk documentation
1. How to configure logging into splunk with authentik as idp
2. How to do log pushing

dominic-r:
## Details

Adds Splunk HEC forwarding guidance under system event docs and links it from the Splunk Enterprise SAML integration guide. The SAML guide also now uses browser-facing Splunk URLs instead of assuming port 8000.

Closes #22223

## Validation

- `make docs`
- `make integrations`


PeshekDotDev:
The content is good. My only thought is I pictured this more as part of the *integration* section of the docs as opposed to being in our main docs. I'm not necessarily a big fan of having products promoted like that in our core documentation. I'd see this as more of an additional way to integration with Splunk. I also like people googling "authentik Splunk" and getting to one page. What are your thoughts @dewi-tik

website/integrations/log-forwarding/splunk-enterprise/index.mdx
```diff
@@ -0,0 +1,118 @@
...
+1. Log in to Splunk Enterprise with an administrator account.
+2. Navigate to **Settings** > **Data inputs**.
+3. Click **HTTP Event Collector**.
+4. Click **New Token** and create a token for authentik events.
+5. Enable the token and allow it to write to the target index.
+
+The HEC URL for JSON events is:
+
+```text
+https://splunk.company:8088/services/collector/event
```
PeshekDotDev: https://help.splunk.com/en/splunk-enterprise/get-started/get-data-in/9.2/get-data-with-http-event-collector/set-up-and-use-http-event-collector-in-splunk-web

<port> is the HEC port number
8088 on Splunk Cloud Platform free trials
443 by default on Splunk Cloud Platform instances

It looks like according to the docs, the port number for HEC is different depending on different factors. So instead of putting the port number, let's maybe explain to double check your port based on your version and link to the splunk docs or add a info/note

dominic-r:
> The content is good. My only thought is I pictured this more as part of the _integration_ section of the docs as opposed to being in our main docs. I'm not necessarily a big fan of having products promoted like that in our core documentation. I'd see this as more of an additional way to integration with Splunk. I also like people googling "authentik Splunk" and getting to one page. What are your thoughts @dewi-tik

Usually, we reserve integration docs for integrating sign in with SSO, so I figured it would be best to keep this doc seperate, cause its entirely event forwarding. But, looking forward to see what dewi has to say as well

dewi-tik:
We have guides for using nginx, caddy etc in our core docs. 

However, this is an enterprise only product (so it seems different to freely available reverse proxies), it's also much more involved than those guides, and it clutters up the events section a little. This will only get worse when we inevitably add more guides for other log forwarding options. Plus this document already takes the form of an integration doc anyway. 

So my thinking is:

1. Add a "Log forwarding" page in the events section of the docs that covers the concept, why you'd want to do this, and the different options.
2. Move the splunk guide to integrations under a new "Log forwarding" category.
3. Link from the log forwarding doc to the integration doc.
4. Update the integrations landing page to be less specific about authentication. This is a good idea anyway because a lot of the docs cover SCIM which isn't authentication.

website/docs/sys-mgmt/events/log-forwarding.mdx
```diff
@@ -0,0 +1,21 @@
...
+
+## Forward selected events
+
+To send selected events to another system, create an event matcher policy, a notification transport, and a notification rule. This forwards only events that match the notification rule, which is useful for security alerts, high-value audit events, or integrations that should receive a narrower event stream.
+
+Notification transports can send events locally, by email, or to a webhook. Webhook transports can be adapted to systems that accept HTTP event ingestion.
+
+## Log forwarding integrations
+
+For a worked example, see [Forward events to Splunk Enterprise](/integrations/log-forwarding/splunk-enterprise/). That guide uses Splunk HTTP Event Collector (HEC), a generic webhook notification transport, and a notification rule to forward matching authentik events.
```
dewi-tik:
```suggestion
For an example of integrating with a log forwarder, see [Forward events to Splunk Enterprise](/integrations/log-forwarding/splunk-enterprise/). That guide uses Splunk HTTP Event Collector (HEC), a generic webhook notification transport, and a notification rule to forward matching authentik events.
```

dewi-tik (approved):

PR #22940, opened 2026-06-08
web/polyfill: polyfill customElements.getName for Safari < 17.4

### LINKED ISSUE #22611: window.customElements.getName is not a function.
### Describe the bug

Hi,

After the update to 2026.5 i noticed my flows stopped working on some older IOS devices with error:

window.customElements.getName is not a function. (In 'window.customElements.getName(o)', 'window.customElements.getName' is undefined)

They had worked this morning but I haven’t had time yet to check whether it’s an issue on my end or possibly fixable via a legacy setting, but checking MDN leads me to believe that all older Safari and WebView browsers before version 17 will be unable to render it.

If this was an explicit choice, feel free to close this. Gracefully falling back or implementing check for customElements.getName() might be more work then the small amount of environments in the wild

### How to reproduce

1. Grab a browser in IOS16 (p. Emulated)
2. Open a flow
3. Greeted with error

### Expected behavior

Flow to continue

### Screenshots

_No response_

### Additional context

_No response_

### Deployment Method

Docker

### Version

2026.5

### Relevant log output

```shell
window.customElements.getName is not a function. (In 'window.customElements.getName(o)', 'window.customElements.getName' is undefined)
```

GirlBossRush:
## What

`window.customElements.getName()` is undefined on iOS 16 Safari and on any WebKit WebView older than 17.4. The flow renderer began calling it in 2026.5, so those clients throw `window.customElements.getName is not a function` at the first stage render, and the flow never recovers. Two reporters hit this on iOS 16 (#22611).

`getName()` is the reverse of `customElements.get()`: given a constructor, it returns the tag name that constructor was registered under. It shipped in Chrome/Edge 117, Firefox 119, and Safari 17.4, so all of iOS 16 and the 17.0 through 17.3 releases reach the renderer without it.

## How

A polyfill in `web/src/polyfill/custom-elements-get-name.ts`:

- Installs only when `getName` is absent, so browsers that already have the native method are left untouched.
- Wraps `define()` to record each tag/constructor pair in a `WeakMap`, then answers `getName(ctor)` from that map.
- Loads from the polyfill entry point ahead of every interface bundle, so the wrapped `define()` is in place before any `customElements.define()` call runs. Order matters here: an install that ran later would miss every element defined before it.

Unit tests cover the install-when-absent path, round-tripping a defined element, forwarding the full `define()` argument list (including `options`), the unregistered-constructor case, and the native no-op path.

## Disclosure

Triaged and authored end-to-end by an automated agent operating under my account and supervision. I reviewed the full diff and confirmed the local build, lint, and type-checks before opening.

Fixes #22611


BeryJu (approved):

sweetpeteza:
Confirming this fix also resolves a crash on Android — not just Safari/WebKit <17.4 as described above.

We hit the identical `TypeError: window.customElements.getName is not a function` in `FlowInterface` on 2026.5.2, reproducible in both Android's system WebView and Chrome for Android (desktop Chromium at any viewport width was unaffected). Upgrading to 2026.5.3 resolved it there as well.

Worth broadening the polyfill's documented scope beyond Safari, since the underlying API gap apparently exists on some Chromium-based mobile builds too.

PR #22980, opened 2026-06-10
web/elements: extract mermaid runtime, modernize `<ak-diagram>`

GirlBossRush:
Splits the first commit of #21949 into its own PR per @kensternberg-authentik's review. Same commit, rebased onto current `main`, with the two `FlowDiagram` follow-ups from the inline comments applied. #21949 gets rebased on top once this lands.

## What this changes

- Mermaid configuration moves from `Diagram` into a new `#common/mermaid` module. `loadMermaid(uiTheme)` lazy-loads the library on first call and reinitialises it when the active UI theme changes, so the whole app shares one mermaid instance instead of each `<ak-diagram>` constructor calling `mermaid.initialize` for itself.
- `<ak-diagram>` switches to `@listen(AKRefreshEvent, { target: window })` for its sync hook and renders through `guard([diagram, activeTheme], () => until(this.renderMermaid(), loadingPlaceholder))`. Inline styles move out to `Diagram.css`, and the SVG host selector is renamed from `mermaid-svg-` to `diagram` to match the new mermaid render id.
- `FlowDiagram` and `OAuthSourceDiagram` adopt the renamed `syncDiagramContent` hook from the parent.

## Review feedback addressed

Two inline comments on `FlowDiagram.ts` from the #21949 review:

- The `requestUpdate()`-inside-a-guarded-method dance is replaced with pfe-core's `@observes("flowSlug")`. The decorator wires up a `PropertyObserverController` that fires the callback during `hostUpdate`, ahead of `shouldUpdate`, so setting `flowSlug` schedules the API fetch without any manual update plumbing.
- API client construction switches from `new FlowsApi(DEFAULT_CONFIG)` to the `aki(FlowsApi)` helper that landed on `main` after #21949 was cut.

## Test plan

- `npm run lint-check`: clean.
- `npm run lint:types`: only error is `OAuthSourceViewPage.ts(34,63)`, which is also present on plain `upstream/main` and unrelated to this change.
- `npm run build`: production bundle builds cleanly.
- No diagram-specific unit tests exist in the tree. The mermaid e2e coverage in `web/test/browser/ak-mdx.test.ts` belongs to the MDX commits of #21949 and is deliberately not part of this PR.

## Disclosure

Triaged and authored end-to-end by an agent running in a sandbox against `goauthentik/authentik@main`. Reviewed and shipped by @GirlBossRush.


web/src/admin/flows/FlowDiagram.ts
```diff
@@ -6,24 +6,25 @@ import { Diagram } from "#elements/Diagram";
...
 export class FlowDiagram extends Diagram {
-    @property()
-    flowSlug?: string;
+    @property({ type: String, useDefault: true })
+    flowSlug: string | null = null;
 
-    refreshHandler = (): void => {
-        this.diagram = undefined;
+    @observes("flowSlug")
+    protected refresh(): void {
```
kensternberg-authentik, line 19: Pretty!  :-)

kensternberg-authentik (approved):

PR #23007, opened 2026-06-11
web/i18n: Fix stale flow locale, unsynchronized locale selector options

### LINKED ISSUE #22954: Login-screen locale selector (added in 2025.12) does not translate server-rendered flow strings (e.g. identification "Log in" button) — switching the locale never re-fetches the challenge
### Describe the bug

The locale selector shown on the flow/login screen (introduced in 2025.12) only updates **client-side** (lit-localize) strings when the user picks a language. It does **not** update **server-rendered** flow strings — most visibly the identification stage's primary action button (`primary_action`, e.g. "Log in" / "Anmelden"), but also the flow title and any other server-translated text in the challenge.

As a result, after switching the language the page is left in a **mixed-language state**: the client-side strings switch immediately, but the submit button (and other server-provided strings) keep the language that was active when the challenge was first fetched. To the end user the button appears to change "only sometimes" — in practice only when a fresh challenge happens to be fetched (e.g. a full page reload, or advancing to the next stage).

The server side itself is **not** broken: requesting the flow with `?locale=<code>` makes the server render everything (including the button) in that language correctly. The problem is purely that the selector never triggers a server round-trip, never persists the choice, and never re-fetches the current challenge.

### How to reproduce

1. Have a browser whose `Accept-Language` is, say, German (`de`). Open any flow that uses the **Identification** stage, e.g. the default authentication flow: `https://<authentik>/if/flow/<flow-slug>/`.
2. Observe the submit button is rendered in German ("Anmelden") — correct, because the server rendered the challenge in `de`.
3. Use the locale selector in the top-right and switch to **English**.
4. Observe: client-side strings (e.g. "Passwort" → "Password") update, **but the submit button stays "Anmelden"** and the flow title stays unchanged. The page is now a German/English mix.
5. Reload the page → the selector resets to the `Accept-Language` value (German); the manual choice was not persisted.

### Expected behavior

Selecting a language in the login-screen locale selector should result in the **entire** flow being rendered in that language, including server-provided strings such as the identification primary action button. The choice should also persist for the session/reload.

### Screenshots

_No response_

### Additional context

**Evidence / root-cause analysis** (observed live in the browser):

- **The button text is server-provided.** The identification component's challenge object already has the translated string baked in: `challenge.primaryAction === "Anmelden"` — the server translated it (gettext) at challenge-fetch time.
- **Switching the locale fires no server request.** Watching the network while changing the selector: **zero** `GET /api/v3/flows/executor/<slug>/` calls. The only executor request is the initial page load, so the existing challenge (with its already-translated button) is reused unchanged.
- **Nothing is persisted.** After switching, `document.cookie` is empty, `localStorage` is empty, and `sessionStorage` contains only `authentik_tab_id`. A reload therefore falls back to `Accept-Language`.
- **The server resolves locale correctly.** Loading `…/if/flow/<slug>/?locale=en` makes the server return `challenge.primaryAction === "Log in"`. So locale resolution and server-side translation work fine — the selector simply never exercises that path.

In short: the selector calls client-side `setLocale()` and stops there. It neither persists the locale (session/cookie) nor re-fetches/re-renders the current challenge.

**Suggested fix (either):**
1. On locale change, persist the chosen locale (e.g. to the session) **and** re-fetch the current challenge from `/api/v3/flows/executor/<slug>/` so server-rendered strings are re-translated; or
2. Reload the flow with `?locale=<code>` appended (preserving existing query parameters) so the middleware re-renders the whole challenge.

**Scope note:** the flow *title* is a statically configured field and is expected to stay as authored — this report concerns strings the server *does* translate per-locale (the identification `primary_action` button and other server-provided labels), not the static title.

**Related:** #9433 (general "locale not respected", closed as not planned / legacy) and #12088 (server-side locale leak, fixed) are related but distinct. This report is specifically about the **2025.12 login-screen selector** not triggering a server round-trip.

### Deployment Method

Kubernetes 

### Version

2026.5.2

### Relevant log output

```shell

```

GirlBossRush:
## Details

This PR fixes two intertwined issues:

### Unlocalized locale selector options

The locale selector is supposed to render each entry in the target language, and in the language you currently are viewing the page in. 

The labels are now fixed and additional comments are added to clarify how this works.

### Server-side localized labels

Server-side localized UI elements previously did not re-request the flow with the updated locale. This behavior was delayed to give us more confidence that the change wouldn't cause any issues in more complex flows. 

After a few releases with the locale selector in production, we're ready to test this behavior and await any user feedback.

Closes #22954

### DIFF
--- a/web/src/common/ui/locale/format.ts
+++ b/web/src/common/ui/locale/format.ts
@@ -196,6 +196,13 @@ export function formatLocaleDisplayNames(
     return entries.sort(createIntlCollator(activeLocaleTag, collatorOptions));
 }
 
+/**
+ * Format the display name for a single locale, using the same logic as the options list.
+ *
+ * @param languageTag The locale to format.
+ * @param localizedDisplayName The localized display name for the locale
+ * @param relativeDisplayName The relative display name for the locale.
+ */
 export function formatRelativeLocaleDisplayName(
     languageTag: TargetLanguageTag,
     localizedDisplayName: string,
--- a/web/src/elements/locale/ak-locale-select.ts
+++ b/web/src/elements/locale/ak-locale-select.ts
@@ -3,6 +3,7 @@ import { formatLocaleDisplayNames } from "#common/ui/locale/format";
 import { setSessionLocale } from "#common/ui/locale/utils";
 
 import { AKElement } from "#elements/Base";
+import { listen } from "#elements/decorators/listen";
 import Styles from "#elements/locale/ak-locale-select.css";
 import { LocaleOptions } from "#elements/locale/utils";
 import { WithCapabilitiesConfig } from "#elements/mixins/capabilities";
@@ -12,8 +13,8 @@ import { CapabilitiesEnum } from "@goauthentik/api";
 
 import { LOCALE_STATUS_EVENT, LocaleStatusEventDetail, msg } from "@lit/localize";
 import { html, PropertyValues } from "lit";
+import { guard } from "lit-html/directives/guard.js";
 import { customElement, state } from "lit/decorators.js";
-import { guard } from "lit/directives/guard.js";
 import { createRef, ref } from "lit/directives/ref.js";
 
 @customElement("ak-locale-select")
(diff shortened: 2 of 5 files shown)


kensternberg-authentik (approved):
Solid.

PR #23010, opened 2026-06-11
core: cache deterministic anonymous authentication entry points

ryanpesek:
## Why

Anonymous authentication entry points can receive substantial unauthenticated traffic. Their deterministic responses can be served by a shared HTTP cache, reducing requests to authentik and PostgreSQL.

Some responses from these endpoints are request-dependent, however. This revision only marks responses as public when authentik can determine that they are safe to share.

## Cache behavior

Cacheable responses receive:

`Cache-Control: public, s-maxage=60, max-age=0`

The 60-second shared-cache lifetime limits how long configuration changes can remain stale. `max-age=0` prevents browsers from reusing an anonymous response after login.

All cacheable responses retain `Vary: Cookie`. This prevents a shared cache from serving a cookieless anonymous response to a request containing authentication or other cookies.

Only completely cookieless `GET` and `HEAD` requests are marked public.

### Root redirect

Cookieless requests to `GET /` are cacheable. The response redirects to the fixed default authentication entry point.

### Default-flow redirect

`GET /flows/-/default/<designation>/` is cacheable only when the selected flow is explicitly configured on the current brand.

Policy-selected fallback flows are not cacheable. This is important because expression, GeoIP, reputation, and other policies can produce request-dependent results.

### Flow interface

`GET /if/flow/<slug>/` is cacheable only for the normal cookieless flow shell.

The response varies on `Cookie`, `Accept-Language`, and `User-Agent`.

Caching is disabled for:

- Requests containing any cookie
- Inspector mode
- Simplified flow executor mode
- Legacy user agents requiring the simplified executor
- Managed media that produces expiring signed URLs
- Non-GET/HEAD requests

Request-specific Sentry trace metadata is omitted from cacheable flow shells so unrelated clients are not attached to the trace that populated the cache.

Brand variation is naturally separated by the request host, which is part of the standard cache key. Query-string variants remain separate URLs.

## Deployment behavior

This change only supplies safe HTTP caching metadata. It does not force a reverse proxy or CDN to cache HTML.

## Tests

Sixteen tests cover:

- Public caching headers and shared-cache TTL
- Browser-cache prevention
- `Vary: Cookie`
- Requests with session and unrelated cookies
- Explicit brand defaults versus policy-selected fallback
- Language and user-agent variance
- Inspector and simplified executor exclusions
- Legacy browser exclusions
- Generated signed-media exclusions
- Request-specific tracing metadata
- Non-read request exclusions

---

## Checklist

### DIFF
--- a/authentik/core/tests/test_interface_views.py
+++ b/authentik/core/tests/test_interface_views.py
@@ -1,5 +1,8 @@
 """Test interface view redirect behavior by user type"""
 
+from http import HTTPStatus
+
+from django.conf import settings
 from django.test import TestCase
 from django.urls import reverse
 
@@ -7,6 +10,10 @@ from authentik.brands.models import Brand
 from authentik.core.apps import Setup
 from authentik.core.models import Application, UserTypes
 from authentik.core.tests.utils import create_test_brand, create_test_user
+from authentik.core.views.interface import (
+    ANONYMOUS_ROOT_REDIRECT_CACHE_SECONDS,
+    anonymous_redirect_cache_control,
+)
 
 
 class TestInterfaceRedirects(TestCase):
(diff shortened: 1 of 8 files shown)


gergosimonyi:
Perhaps we should not care about this, but please note: response to `GET /if/flow/<slug>/` can technically vary, e.g. by binding a non-deterministic expression policy to the flow or the first stage.

gergosimonyi:
Yeah I'm almost sure this can't be merged in its current form. It fails not only due to the previous comment, but also other dynamic parts like locale (e.g. `AuthorizationFlowInitView.dispatch_with_language`) or brand (e.g. `ToDefaultFlow.get_flow`), and possibly others.

Something smaller in scope is required.

ryanpesek:
Ok lets give this a try maybe? I rebased the branch onto current `main` and reworked the caching behavior so it no longer assumes that all anonymous flow responses are deterministic.

PR #23013, opened 2026-06-11
root: PostgreSQL direct-endpoint support for LISTEN/NOTIFY and advisory locks

ryanpesek:
## Why

authentik has three subsystems that hold state on a single PostgreSQL
backend across multiple statements: the Channels Postgres layer (LISTEN
on a long-lived connection), the dramatiq worker (LISTEN + session-scoped
`pg_advisory_lock`), and the startup migration's `pg_advisory_lock(1000)`.
Behind a transaction-pooling pooler (PgBouncer `pool_mode=transaction`,
Cloud SQL managed pooling) all three break silently — the pooler returns
the backend to the pool at every COMMIT, severing the LISTEN registration
and releasing the lock without anyone telling the app.

This blocks deployments from using a transaction-mode pooler in front of
authentik at all, even though such a pooler is the standard way to deal
with PG connection exhaustion in production. Today operators run direct
connections from every app thread, hitting the `max_connections` ceiling
under load.

## What

Adds an optional `postgresql.direct.*` config block. When set, a Django
database alias named `direct` is added; the three session-stable
subsystems route through it while ORM traffic continues through the
`default` alias (which can now safely point at a transaction-pooling
pooler).

## Safety

Opt-in. When `postgresql.direct.*` is unset, `postgresql_direct_db_enabled()`
returns False, no Django alias is added, and all subsystems use `default`
as before. Identical behavior to the current codebase.

---

## Checklist


rissson:
Yes! In principle I agree and want this change to make it in, but I have a bit of thinking to do about how this will affect future plans of moving this stuff out of Python.

Also, I don't see this affecting advisory locks everywhere, like where we use `pglock` instead of raw SQL.

ryanpesek:
> Yes! In principle I agree and want this change to make it in, but I have a bit of thinking to do about how this will affect future plans of moving this stuff out of Python.
> 
> Also, I don't see this affecting advisory locks everywhere, like where we use `pglock` instead of raw SQL.

@rissson Good point about pglock. I just pushed a commit that I think covers all of that now.

rissson:
Thanks! I'll take care of implementing this in `packages/ak-common/src/db.rs` as well

rissson (approved):

rissson:
@ryanpesek do you think it makes sense to force migrations to go through the direct connection if configured?

ryanpesek:
> @ryanpesek do you think it makes sense to force migrations to go through the direct connection if configured?

That seems like the safest option to me as that would avoid any `quirks` that a pooler could introduce.

PR #23014, opened 2026-06-11
Per-request DB query reductions

ryanpesek:
## Why

Two small, independent optimizations that compound under load. Each
removes queries from the request hot path without changing behavior.

### `TENANT_LIMIT_SET_CALLS = True`

`django_tenants` issues `SET search_path` before every cursor by default.
Setting this flag makes it issue the `SET` only when the schema actually
changes for a given connection. Roughly halves the statement count on
request hot paths. Safe for both single-tenant and multi-tenant setups
as long as the PG session is stable across statements.

### `select_related` for Brand FKs in `get_brand_for_request`

`get_brand_for_request` is called from `BrandMiddleware` on every
request. `CurrentBrandSerializer` then reads 7 Flow FK fields on the
brand (`flow_authentication`, `flow_invalidation`, `flow_recovery`,
`flow_unenrollment`, `flow_user_settings`, `flow_device_code`,
`flow_lockdown`) plus `default_application`. Without `select_related`,
each access is a lazy load — up to 8 extra queries per HTML render of
`/if/flow/<slug>/`. With it, all 8 hydrate in the same SELECT.

A regression test guards against future additions to the FK set being
missed (and against the symmetric "this list contains a field that no
longer exists, causing FieldError" failure mode).

### Memoize `get_current_tenant` per Django connection

`get_current_tenant()` is called multiple times per request (once per
public flag from `Flag.get()` in `CurrentBrandSerializer.get_flags()`,
plus a handful of other call sites). Each call without memoization is a
separate `SELECT ... FROM authentik_tenants_tenant WHERE schema_name = $1`.
Memoization keyed on `(schema_name, frozenset(only))` turns N=5+ queries
into 1.

The `only` argument is part of the key for migration safety — different
callers pass different `only` sets (notably `Flag.get` uses
`only=["flags"]`), and during migrations the Tenant table may not yet
have all columns. A `request_started` signal handler clears the cache at
request boundaries as defense-in-depth for persistent-connection
deployments.

## Tests

- `test_brand_fks_are_loaded_in_single_query` — `assertNumQueries(1)`
  while accessing every brand FK.
- `test_brand_related_fk_list_complete` — fails CI if a new FK is added
  to the Brand model without being included in `_BRAND_RELATED_FK_FIELDS`.
- `test_brand_related_fks_all_exist_on_model` — fails CI if a stale FK
  is in the list (would cause `FieldError` on every request).
- 5 tests for `get_current_tenant` memoization, including signal-handler
  registration and `schema_name`-based invalidation.

---

## Checklist

### DIFF
--- a/authentik/brands/tests.py
+++ b/authentik/brands/tests.py
@@ -2,13 +2,17 @@
 
 from json import loads
 
+from django.http import HttpRequest
+from django.test import TestCase
 from django.urls import reverse
 from rest_framework.test import APITestCase
 
 from authentik.blueprints.tests import apply_blueprint
 from authentik.brands.models import Brand
+from authentik.brands.utils import _BRAND_RELATED_FK_FIELDS, get_brand_for_request
 from authentik.core.models import Application
-from authentik.core.tests.utils import create_test_admin_user, create_test_brand
+from authentik.core.tests.utils import create_test_admin_user, create_test_brand, create_test_flow
+from authentik.flows.models import FlowDesignation
 from authentik.lib.generators import generate_id
 from authentik.providers.oauth2.models import OAuth2Provider
 from authentik.providers.saml.models import SAMLProvider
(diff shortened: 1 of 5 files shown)


rissson:
So for this PR specifically, I'll extract the brand changes as well, but leave the tenant changes here. We are planning on removing tenants in 1 or 2 releases, and as such am not sure if we're planning on making those changes or not.

PR #23015, opened 2026-06-11
root: in-process per-IP rate throttle

ryanpesek:
## Why

DRF's stock `AnonRateThrottle` uses Django's default cache to track
per-IP request counters. With authentik's default cache backed by
PostgreSQL (`django_postgres_cache`), every API request that hits the
throttle machinery issues a `cache.get` + `cache.set` against PG just
to make the throttle decision. Under flood conditions this _adds_ DB
load instead of reducing it — the protective layer becomes an amplifier.

## What

- New `LocalAnonRateThrottle(AnonRateThrottle)` subclass with
  `cache = caches["throttle"]`.
- New `throttle` cache alias in `CACHES` backed by `LocMemCache` with
  `MAX_ENTRIES=10000` 
- `DEFAULT_THROTTLE_CLASSES` points at the new subclass.

## Tests

5 tests: throttle uses the `throttle` cache alias, the alias is backed
by LocMemCache, `MAX_ENTRIES` is sufficiently above the default 300,
throttle writes are isolated from the default PG cache, and
`DEFAULT_THROTTLE_CLASSES` references the new subclass.

---

## Checklist


gergosimonyi (approved):
Seems reasonable to me. @rissson can you think of a reason not to?

rissson:
> Seems reasonable to me. @rissson can you think of a reason not to?

I cannot. We probably want to call this out in release notes however. @BeryJu thoughts?

rissson (approved):

melizeche (approved):

PR #23017, opened 2026-06-11
Eliminate hot DB queries on metrics scrape and login paths

ryanpesek:
## Why

Three full-table scans against the task and cache tables were the top
CPU consumers under sustained load:

1. `SELECT DISTINCT queue_name, actor_name FROM authentik_tasks_task`
   issued on every Prometheus scrape per server pod.
   Scans the entire task table to recover a tiny set of strings that are statically known
   from the dramatiq actor registry.

2. `SELECT cache_key FROM django_postgres_cache_cacheentry WHERE
   cache_key::text ~ $1` issued on every successful login. Django's
   `update_last_login` fires `post_save(User)`; `invalidate_policy_cache`
   responds with a regex full-table scan of the cache table — even
   though `last_login` doesn't affect policy evaluation or application
   access at all.

3. The cache backend's `keys()` method translates every glob pattern to
   a regex, forcing PostgreSQL into per-row regex evaluation. Even after
   2) is fixed, the remaining `cache.keys(...)` callers (admin actions,
   monitoring scrapes, the other signal handlers in `policies/signals.py`
   and `flows/signals.py`) all benefit from a cheaper path.

### `authentik/tasks/signals.py` — `monitoring_set_queued_tasks`

Replaces `Task.objects.values("queue_name", "actor_name").distinct()`
with iteration over `dramatiq.broker.get_broker().actors.values()`. The
set of `(queue_name, actor_name)` combinations is defined at process
startup when actors register with the broker; the database is the wrong
source. The remaining query (the `WHERE state = QUEUED` count) uses the
existing `(queue_name, state)` index and only touches active rows.

### `authentik/policies/signals.py` — `invalidate_policy_cache`

Adds an early return when a User save touches only the `last_login`
field. Django's `update_last_login` calls
`user.save(update_fields=["last_login"])` on every successful login;
detecting this case skips the broad cache invalidation entirely.
Profile updates, password changes, group membership changes, and
Policy/PolicyBinding/Group saves continue to invalidate as before.

### `packages/django-postgres-cache` — `DatabaseCache.keys`

Detects simple `prefix*` glob patterns and uses Django's `__startswith`
lookup (translates to `LIKE 'prefix%'`) instead of `__regex`. Patterns
without any wildcard reduce to primary-key equality. Complex patterns
(wildcards in non-suffix position, multiple wildcards) fall back to the
original regex path — preserving the public method contract.

All authentik `cache.keys()` call sites use the simple `prefix*`
pattern, so they all take the fast path.

## Performance characteristics

- For databases with a B-tree index using `text_pattern_ops` on
  `cache_key`: `LIKE 'prefix%'` becomes an index-only scan.
- Without such an index (default Cloud SQL with non-C collation): `LIKE`
  still requires a sequential scan, but the per-row cost is dramatically
  lower than regex automaton evaluation.
- Exact-equality reduces to a primary-key lookup.

## Tests

14 new tests across three files. Of particular note:

- `test_does_not_call_values_distinct_on_task_objects` — regression
  guard that fails if anything reintroduces a direct
  `Task.objects.values(...).distinct()` on the table.
- `test_user_save_with_only_last_login_does_not_invalidate` — regression
  guard for the per-login hot path. Confirmed to fail against the
  pre-fix code path with the exact symptom (`cache.keys` called).
- `test_simple_prefix_glob_uses_startswith` — verifies the SQL lookup
  chosen for each pattern shape.

---

## Checklist

### DIFF
--- a/authentik/policies/signals.py
+++ b/authentik/policies/signals.py
@@ -29,8 +29,18 @@ def monitoring_set_policies(sender, **kwargs):
 @receiver(post_save, sender=PolicyBindingModel)
 @receiver(post_save, sender=Group)
 @receiver(post_save, sender=User)
-def invalidate_policy_cache(sender, instance, **_):
-    """Invalidate Policy cache when policy is updated"""
+def invalidate_policy_cache(sender, instance, update_fields=None, **_):
+    """Invalidate Policy cache when policy is updated.
+
+    Skips when the save touched only ``last_login`` — Django's auth flow runs
+    ``user.save(update_fields=["last_login"])`` on every successful login, and
+    the broad invalidation below would otherwise issue a full-table cache
+    scan on every login. ``last_login`` doesn't affect policy evaluation or
+    application access, so there's nothing to invalidate.
+    """
+    if sender == User and update_fields and set(update_fields) <= {"last_login"}:
+        return
+
     if sender == Policy:
         total = 0
         for binding in PolicyBinding.objects.filter(policy=instance):
--- a/authentik/policies/tests/test_signals.py
+++ b/authentik/policies/tests/test_signals.py
@@ -1,0 +1,105 @@
+"""Tests for ``authentik.policies.signals.invalidate_policy_cache``.
+
+Regression guards for the per-login cache-invalidation skip. ``last_login``-
+only User saves must not trigger the broad ``cache.keys(...)`` invalidation.
+"""
+
+from unittest import TestCase, mock
+
+
+class _FakeUser:
+    pk = 1
+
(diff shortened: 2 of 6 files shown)


rissson:
closing in favour of the 3 linked PRs above. Thanks for the contribution!

ryanpesek:
Great, thank you for taking a look. Any chance the fixes will be cherry picked into 2026.5?

rissson:
> Any chance the fixes will be cherry picked into 2026.5?

@BeryJu

PR #23037, opened 2026-06-12
providers/oauth: Properly return error via post and for request objects

PeshekDotDev:
### What does this PR change?

Previously all authorization errors were returned as a fragment or redirect with the error in the query params, now we return some errors via form post. This is a requirement for OIDC certification.

We also parse request objects just enough to properly reject them, which is also a requirement for OIDC certification.

Withouth this, we do not meet certain oidc requirements.

### How was this tested?

Via test suite and openid.net cert testing suite

### Linked issues

---

## Checklist


BeryJu (approved):

authentik/providers/oauth2/views/authorize.py
```diff
@@ -151,45 +152,89 @@ def __post_init__(self, github_compat=False):
             LOGGER.warning("Invalid client identifier", client_id=self.client_id)
             raise ClientIdError(client_id=self.client_id)
         self.check_redirect_uri()
-        self.check_grant()
-        self.check_scope(github_compat)
         if self.request:
+            # We don't support request objects, so we buttThe error must still be returned to the
```
PeshekDotDev:
```suggestion
            # We don't support request objects, but the error must still be returned to the
```

PR #23062, opened 2026-06-15
web: Fix user list default paths.

GirlBossRush:
## Details

This PR fixes an issue where the user list default paths does not match our documented usage behavior.

Previously, `UserListPage` resolved a single `activePath` that fell back to `"" when` no URL param or brand override was set, showing all users. This conflicted with expected behavior where the compiled-in default path should still seed the tree view and wizard. The fix splits resolution into two properties: `activePath` (which still clears to "" for "show all") and `defaultActivePath` (which always resolves through the full fallback chain).

### DIFF
--- a/web/src/elements/TreeView.ts
+++ b/web/src/elements/TreeView.ts
@@ -2,16 +2,19 @@ import { EVENT_REFRESH } from "#common/constants";
 
 import { AKElement } from "#elements/Base";
 import { setURLParams } from "#elements/router/RouteMatch";
+import { SlottedTemplateResult } from "#elements/types";
 import { ifPresent } from "#elements/utils/attributes";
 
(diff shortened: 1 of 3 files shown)


web/src/elements/TreeView.ts
```diff
@@ -84,85 +89,87 @@ export class TreeViewNode extends AKElement {
         );
     };
 
-    render(): TemplateResult {
+    protected override render(): SlottedTemplateResult {
         const shouldRenderChildren = (this.item?.childItems || []).length > 0 && this.open;
         const itemLabel = this.item?.label || msg("Unnamed");
         const current = this.host?.activeNode === this;
 
-        return html`
```
GirlBossRush, line 92: whitespace change

BeryJu (approved):

PR #23064, opened 2026-06-15
packages/django-postgres-dramatiq: add support for task dependencies

rissson:
### What does this PR change?

Introduce a `dependencies` field on tasks, and a new task state `waiting_for_dependencies`, so that tasks may depend on other tasks finishing before running.

When enqueuing a task, a new option called `dependencies` has been added, which can be set to a list of tasks (or pks, or whatever django accepts). If that list is not empty, the task state is set to `waiting_for_dependencies` instead of `queued`.

When a task is done, it now looks for dependent tasks, and for each dependent task, checks if all its dependencies are done. If so, the dependent is marked as `queued`. When rejecting a task, all its dependents are marked as rejected.

Also, when the broker is grabbing new tasks, there is a new subroutine that runs when no tasks are found (to avoid blocking the broker when there is work to do) to find `waiting_for_dependencies` tasks that might have been missed when (n)ack-ing a task. The same logic as above applies here.

To get rid of top-level tasks for synchronization processes, and to all for new processes for tracking the status of syncs. See https://github.com/goauthentik/authentik/pull/23126 for an example use case.

### How was this tested?

With https://github.com/goauthentik/authentik/pull/23126

### Linked issues

---

## Checklist


dewi-tik (approved):
Docs approval

BeryJu (commented):
LGTM but I think @gergosimonyi would prefer a concrete m2m through model

GirlBossRush (approved):

authentik/tasks/models.py
```diff
@@ -149,6 +159,20 @@ def error(self, message: str | Exception, **attributes) -> None:
         self.log(self.uid, TaskStatus.ERROR, message, **attributes)
 
 
+class TaskDependency(models.Model):
```
rissson:
```suggestion
class TaskDependency(InternallyManagedMixin, models.Model):
```

I think

GirlBossRush (approved):

BeryJu (approved):

PR #23151, opened 2026-06-17
web, core: Fix server-side message race condition, type mismatch.

GirlBossRush:
## Details

This PR fixes two issues:

- Server-side messages can be dispatched before the message container is prepared to listen for the event.
- #19141 included a change to message events, switching from `CustomEvent` to `AKMessageEvent` which has a different runtime type. This prevented initial server-side events from dispatching. Messages are now serialized as JSON and drained after the message container component connects to the DOM.

### DIFF
--- a/authentik/core/templates/base/header_js.html
+++ b/authentik/core/templates/base/header_js.html
@@ -16,18 +16,14 @@
             relBase: "{{ base_url_rel }}",
         },
     };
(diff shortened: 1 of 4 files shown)


BeryJu (approved):

kensternberg-authentik (approved):
Well, that's a bit more simple than the version I proposed.  :-).

PR #23168, opened 2026-06-18
core: cache S3 file storage clients

dominic-r:
## Summary
- Reuses the S3 boto client while the configured endpoint, region, TLS, addressing style, signature version, and credentials remain unchanged.
- Invalidates the cached client when relevant S3 session or client config changes.
- Adds unit coverage for client reuse and config-driven refresh without requiring the S3 test service.

## Why
The media file listing path can call into S3 URL generation repeatedly. Creating a fresh boto client for every access adds avoidable overhead, which is especially visible on slow S3-compatible backends.

## Validation
- `uv run ruff check authentik/admin/files/backends/s3.py authentik/admin/files/backends/tests/test_s3_backend.py`
- `uv run ruff format --check authentik/admin/files/backends/s3.py authentik/admin/files/backends/tests/test_s3_backend.py`
- `uv run coverage run manage.py test --keepdb authentik/admin/files/backends/tests/test_s3_backend.py` (2 passed, 25 skipped; live S3 tests skipped because the local S3 test service is not running)


authentik/admin/files/backends/s3.py
```diff
@@ -77,41 +82,38 @@ def session(self) -> boto3.Session:
...
-        addressing_style = CONFIG.get(
-            f"storage.{self.usage.value}.{self.name}.addressing_style",
-            CONFIG.get(f"storage.{self.name}.addressing_style", "auto"),
-        )
-        signature_version = CONFIG.get(
-            f"storage.{self.usage.value}.{self.name}.signature_version",
-            CONFIG.get(f"storage.{self.name}.signature_version", "s3v4"),
-        )
+        endpoint_url, endpoint_url_r = self._get_config("endpoint", None)
+        use_ssl, use_ssl_r = self._get_config("use_ssl", True)
```
rissson: This will not return the expected `True` as a boolean I think.

Besides, why would we reload those? I understand reloading `endpoint`, but for the other ones, let's not bother and say that the server must be restarted for changes to take place.

authentik/admin/files/backends/tests/test_s3_backend.py
```diff
@@ -1,14 +1,84 @@
...
 from django.test import TestCase
 
+from authentik.admin.files.backends.s3 import S3Backend
 from authentik.admin.files.tests.utils import FileTestS3BackendMixin, s3_test_server_available
 from authentik.admin.files.usage import FileUsage
 from authentik.lib.config import CONFIG
 
 
+class TestS3BackendClientCache(TestCase):
+    """Test S3 client caching without requiring an S3 test server."""
```
dominic-r:
```suggestion
    """Test S3 client caching."""
```

rissson (approved):

PR #23178, opened 2026-06-18
tasks: ignore integrity error on tasklog insertion

rissson:
successor for https://github.com/goauthentik/authentik/pull/22671

### Linked issues

---

## Checklist


rissson:
@dominic-r would you be able to try this out?

BeryJu (commented):
should probably write the logs to the console if we can't save them in the db, no?

rissson:
> should probably write the logs to the console if we can't save them in the db, no?

yeah that makes sense

BeryJu (approved):

PR #23126, opened 2026-06-16
sources/ldap: use task dependencies instead of main task

rissson:
needs https://github.com/goauthentik/authentik/pull/23064

### Linked issues

---

## Checklist

-   [ ] The project has been linted, built, and tested (`make all`)
-   [ ] The documentation has been updated and formatted (`make docs`)

PR #23184, opened 2026-06-18
lib/sync: add Sync base model for storing past sync history

rissson:
### What does this PR change?

Extracted from https://github.com/goauthentik/authentik/pull/23126

Add a `Sync` model (looking for feedback on the name) that can store past sync statuses and related tasks. See the above PR for use case.

Currently 20 syncs are kept, but I am also looking for feedback regarding that, especially since we might create Sync objects for partial syncs (in SCIM providers for instance).

Relying on past tasks is flaky and hard to correctly track. A persistent model to store that state is better suited for that task.

### Linked issues

---

## Checklist

### DIFF
--- a/authentik/core/tasks.py
+++ b/authentik/core/tasks.py
@@ -15,6 +15,7 @@ from authentik.core.models import (
     ExpiringModel,
     User,
 )
+from authentik.lib.sync.models import Sync
 from authentik.lib.utils.db import chunked_queryset
 from authentik.tasks.middleware import CurrentTask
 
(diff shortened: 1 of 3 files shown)


authentik/core/tasks.py
```diff
@@ -45,6 +45,9 @@ def clean_expired_models():
             obj.delete()
         LOGGER.debug("Expired models", model=cls, amount=amount)
         self.info(f"Expired {amount} {cls._meta.verbose_name_plural}")
+    for cls in Sync.__subclasses__():
+        amount = cls.cleanup()
+        LOGGER.debug("Expired models", model=cls, amount=amount)
```
BeryJu, line 50: if we're doing this in here why not inherit from expiringmodel

rissson: How do you handle the count-based expiration with ExpiringModel?

BeryJu: good point

BeryJu (approved):

PR #23185, opened 2026-06-18
packages/django-dramatiq-postgres/broker: purge at start of loop to ensure it runs

### LINKED ISSUE #22713: authentik_tasks_task records with expired result_expiry are never deleted under load, causing unbounded database growth
### Describe the bug

Authentik 2026.2.3 / 2026.5.0

If you have an authentik worker instance that never ends up idle, the clean up job will never fire off.

I had over 6.5 million rows and many days worth of task tables that were never purged.

Symptom in my environment was OOMkilled since the task tables had built up to the point where once I did have an idle period the job tried to run and it ended up ballooning memory and it exceeded its limits rapidly and killed the pod.

From what I can see - auto_purge() in packages/django-dramatiq-postgres/django_dramatiq_postgres/broker.py:514 is the only cleanup path for authentik_tasks_task/_tasklog. Its single call site (broker.py:437) is in the # No message to process branch of consume() — it only runs when the broker is idle. Under sustained load (event-trigger traffic) the broker is never idle, so purge never runs. Confirmed: 0 occurrences of Running garbage collector / Purged messages in all queues in the worker logs over the full 5d18h+ uptime. Identical call site in version/2026.5.0 and main. (Previously discussed in closed issue [#21269](https://github.com/goauthentik/authentik/issues/21269), which was resolved as "the code exists" — which is true, but it's unreachable under load.)

Second issue, if we have an environment with very busy tasks, once we do get some idle time the ORM delete ends up OOM'ing the worker. Essentially the table became too large to manage.

Table sizes
authentik_tasks_tasklog   1809 MB   6,547,483 rows
authentik_tasks_task      1552 MB   2,113,658 rows

99.98% of done tasks were already past result_expiry but never deleted
SELECT count(*) FROM authentik_tasks_task
WHERE state='done' AND result_expiry < now();          -- 2,122,563

Queue was NOT backlogged — proves it's the purge, not a processing backlog
SELECT state, count(*) FROM authentik_tasks_task GROUP BY state;
--  done     2,123,424
--  rejected         6
--  queued           1

Growth rate: ~60k–90k task rows/day, accumulating since first run
SELECT min(mtime)::date, max(mtime)::date FROM authentik_tasks_task;
--  2026-04-19 | 2026-05-27   (38 days, no cleanup ever)

Top producer: 75%+ of rows:
authentik.events.tasks.event_trigger_handler    1,588,644
authentik.events.tasks.event_trigger_dispatch     397,156

Suggest that fixes are:

1) Make the purge reachable under load. invoke _auto_purge() from _scheduler() on its interval, not only the idle branch.

2) Make the delete memory-safe — batch the purge with a LIMIT-bounded loop / raw SQL DELETE instead of a single unbounded ORM queryset.delete(), so it can't load millions of rows into memory regardless of how large the table has grown.

I've had temporarily create a clean up job that runs directly on Postgres to periodically clean up these tasks.

### How to reproduce

An environment that is busy with tasks but has no idle time

### Expected behavior

I expected that the cleanup job would run regardless of load on the authentik worker, and appropriately batch in cases where the table is excessively large.

### Screenshots

_No response_

### Additional context

_No response_

### Deployment Method

Kubernetes

### Version

Authentik 2026.2.3 / 2026.5.0

### Relevant log output

```shell

```

rissson:
## Details

see https://github.com/goauthentik/authentik/issues/22713

### Linked issues

closes https://github.com/goauthentik/authentik/issues/22713

---

## Checklist

### DIFF
--- a/packages/django-dramatiq-postgres/django_dramatiq_postgres/broker.py
+++ b/packages/django-dramatiq-postgres/django_dramatiq_postgres/broker.py
@@ -442,6 +442,7 @@ class _PostgresConsumer(Consumer):
         # Run required processes first
         self._scheduler()
         self._purge_locks()
+        self._auto_purge()
 
         # If we don't have a connection yet, fetch missed notifications from the table directly
         if self._listen_connection is None and not self.pending:
@@ -487,9 +488,6 @@ class _PostgresConsumer(Consumer):
             else:
                 self.logger.debug("Message already consumed. Skipping.", message_id=message_id)
                 continue
-
-        # No message to process, we can do some cleaning
-        self._auto_purge()
 
         self.misses = 0
         return None


BeryJu (approved):

melizeche (approved):
LGTM, but let's wait for the tests to pass

PR #23195, opened 2026-06-18
web/css: de-dupe and simplify CSS customizations

kensternberg-authentik:
## What

1.  Moves the duplicated six lines pointing to Patternfly (base + Spinner) in both `flows.global.css` and `interface.global.css` into their own file (`#styles/vendor/patternfly-base.css`) and references them instead.

2.  Moves all the custom login properties (about 200 lines) that were duplicated in `flows.global.css` and `static.global.css`, which exist mostly to handle/avoid FOUC when loading a new template, into their own file, `#styles/authentik/flow-loading.css`, and references them instead.

3.  Deletes duplicate imports of `#elements/locale/ak-locale-select.css`. That looks it started out as a fat finger on one of them, and was duplicated by cut/paste.

4.  Removes the `pf-c-login` class from `<ak-flow-executor>` in the template. The only thing this did help the template avoid some FOUC on start-up; this has been replaced by much simpler code (found in `#styles/components/Login/flow-loading.css`) that control the look and feel of the template *before* `CustomElementRegistry.define()` is called and Lit takes over rendering the templates.

5.  Moves the responsibility for dictating the layouts into the FlowExecutor directly, by wrapping the whole template in a `<div>`. This both controls the layouts and activates the `--pf-c-login-*` CSS variables for anything contained inside the Flow DOM hierarchy, as well as including the customizations, which have also been moved to `#styles/components/Login/login-tokens.css` which is also used by static, and `#styles/components/Login/login-layout.css`, which is not.

6.  Removed how the global sheets had to reference FlowExecutor.css for some imports. Now that responsibility flows only one-way.

I have tested this in Chrome and Safari, and under “degraded network” and “slow CPU” conditions, and I’m actually happy with how good it looks. Other than the background loading very slowly, it acts just the way we expect, and the performance isn’t bad at all.

## Why

As I was working on the "enabling layers in our CSS," I realized that the changes I'd proposed on a per-phase basis were getting larger than I'd expected.  This seemed like a good place to call it, after the de-duplication.  There will be more CSS updates, but this was a biggie.


web/src/flow/FlowExecutor.css
```diff
@@ -1,10 +1,9 @@
-@import "../styles/authentik/components/Login/login.css";
+@import "../styles/authentik/components/Login/login-layout.css";
 
-:host,
-ak-flow-executor.style-scope {
+:host {
     display: flex;
+    min-width: 100%;
     min-height: 100dvh;
-    flex-flow: column nowrap;
```
kensternberg-authentik, line 7: Not needed; didn't do anything.

authentik/flows/templates/if/flow.html
```diff
@@ -50,7 +50,6 @@
 <ak-drawer id="flow-drawer">
     <ak-flow-executor
         slug="{{ flow.slug }}"
-        class="pf-c-login"
```
kensternberg-authentik, line 53: Not only not needed, it shows genuine confusion over what we're using pf-c-login for.

web/src/flow/FlowExecutor.ts
```diff
@@ -431,7 +431,7 @@ export class FlowExecutor extends WithBrandConfig(Interface) implements StageHos
     protected override render(): SlottedTemplateResult {
         const { challenge, loading } = this;
 
-        return html`<ak-locale-select
+        return html`<div class="pf-c-login" data-layout=${this.layout} part="login"><ak-locale-select
```
kensternberg-authentik: Putting this here activates it for everything for which FlowExecutor is responsible.

web/src/styles/authentik/components/Login/flow-loading.css
```diff
@@ -0,0 +1,15 @@
+/* `:defined` applies after `customElementRegistry.define()` has been called on the tag. So this
+   instructs the flow layout on what to do *before* that happens, to avoid FOUC.
+ */
+ak-flow-executor:not(:defined) {
+      display: grid;
+      place-items: center;
+      width: 100%;
+      min-height: 100dvh;
+}
```
kensternberg-authentik, line 9: This is all we really want for "What do we show inside this custom tag *before* it's registered as a custom element?"  The answer is: show whatever isn't controlled by the template dead center of the screen.  That "whatever" is **Spinner**, provided by the template.

web/src/styles/authentik/components/Login/login-tokens.css
```diff
@@ -0,0 +1,88 @@
+:root {
```
kensternberg-authentik, line 1: This code was copy/pasted into two places, both of which you can see below.

Now, it's not.

GirlBossRush (approved):

BeryJu (approved):

PR #23215, opened 2026-06-19
providers/saml: support wildcard * in ACS URL

boussaffawalid:
### What does this PR change?
Allows SAML providers to configure a wildcard ACS URL (e.g. https://*.example.com/saml/acs) to handle multiple subdomains without creating one provider per host.

- Adds `WildcardDomainlessURLValidator` to `authentik/lib/models.py`. accepts * in URLs by substituting it before URL format validation
- Updates `SAMLProvider.acs_url` to use the new validator
- Replaces the exact-match ACS URL check in `authn_request_parser.py` with `fnmatch` glob matching when the configured URL contains `*`; stores the resolved concrete URL on `AuthNRequest`
- `assertion.py` and `flows.py` now use the resolved concrete URL (not the pattern) in Recipient, Destination, and POST/Redirect binding targets

Support handling multiple subdomains without creating one provider per host.

### How was this tested?
I tested and deployed it in my homelab.

## Checklist

-   [] The project has been linted, built, and tested (`make all`)

### DIFF
--- a/authentik/lib/models.py
+++ b/authentik/lib/models.py
@@ -94,6 +94,13 @@ class DomainlessURLValidator(URLValidator):
         super().__call__(value)
 
 
+class WildcardDomainlessURLValidator(DomainlessURLValidator):
+    """Like DomainlessURLValidator but allows '*' wildcards in the URL."""
+
+    def __call__(self, value: str):
+        super().__call__(value.replace("*", "wildcard"))
+
+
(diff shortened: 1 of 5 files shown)


dewi-tik:
We're going to implement this in a different way by allowing an array of ACS values. You can track the feature here: https://github.com/goauthentik/authentik/issues/24843

But thank you very much for the contribution.

PR #23216, opened 2026-06-19
core/groups: fix n+1 queries for list groups when include_users=false

ryanpesek:
## Summary

Fix the `/api/v3/core/groups/?include_users=false` N+1 query shape without regressing large-group performance.

`include_users=false` still returns the `users` PK list, but no longer runs one membership query per group. Instead, the list view batches raw group/user PK pairs for the current page using the M2M through table and attaches those PK lists to each `Group` instance. The serializer then returns those raw PKs directly.

This gives us the “best of both worlds”:

- Keeps the existing API behavior where `users: [pk, ...]` is included.
- Avoids N+1 queries for group list responses.
- Avoids Django `Prefetch("users", queryset=User.objects.only("pk"))`, which would still materialize `User` model instances.
- Preserves the prior optimization for large groups and `members_by_pk + include_users=false`.

## Background

The previous optimization (https://github.com/goauthentik/authentik/pull/20431) intentionally removed the `users` prefetch for `include_users=false` so `BulkManyRelatedField.to_representation()` could use: `values_list("pk", flat=True)`
That avoided loading full or partial User model instances just to serialize PKs. This mattered for large groups, especially the Admin UI path that uses members_by_pk + include_users=false.
However, in a paginated group list this still ran once per group, producing an N+1 query shape. Discussed in https://github.com/goauthentik/authentik/pull/21690#issuecomment-4749094713

---

## Checklist


authentik/core/api/groups.py
```diff
@@ -392,7 +417,21 @@ def get_queryset(self):
...
+        if page is not None:
+            groups = list(page)
+            self._attach_user_pk_lists(groups)
+            serializer = self.get_serializer(groups, many=True)
+            return self.get_paginated_response(serializer.data)
+
+        groups = list(queryset)
+        self._attach_user_pk_lists(groups)
+        serializer = self.get_serializer(groups, many=True)
+        return Response(serializer.data)
```
gergosimonyi, line 435: Can you please include a test for this path?

gergosimonyi (commented):
Thank you for yet another contribution!

From a technical point this seems to work well, but I reaaaaaly wouldn't like doing this. It's a patch on a patch, an exceptional case within an exceptional case.

However, if a "prettier" solution is not found, I'll merge this in time for the next major release.

authentik/core/api/groups.py
ryanpesek: added in 627b5ae7cd94a8a344ed539af5f5383c3ac8e181

gergosimonyi (approved):
:( :) :( :)

PR #23315, opened 2026-06-23
core: make all m2m models explicit

gergosimonyi:
This PR adds explicit models for all many-to-many relationships. The immediate benefit is to RBAC: this will result in Django creating `add_`/`view_*`/`change_*`/`delete_*` permissions for all many-to-many models. (Some more distant benefits are having no exceptional cases, finer control, easier extension.)

The method chosen results in no database changes whatsoever. This was deemed to be easy to upgrade and also to result in relatively well contained "non-standard django code".

Implementation-wise, we went with `SeparateDatabaseAndState`, [as recommended](https://docs.djangoproject.com/en/5.2/howto/writing-migrations/#changing-a-manytomanyfield-to-use-a-through-model) in the Django documentation.


authentik/brands/models.py
```diff
@@ -166,6 +167,23 @@ class Meta:
         ]
 
 
+class BrandClientCertificate(models.Model):
```
BeryJu: uuid pk pls (for all of these)

rissson: CompositePk goes brrr :eyes:

gergosimonyi: Offline resolution: we'll do our very best to keep database changes to a minimum (at least in this PR):

- no table renames
- no column renames
- no changes to primary key

This is because it's not worth it to write code that would work with no uptime loss during an upgrade. It's not the prettiest, but the affected areas are relatively self-contained.

Subsequent m2m tables should not follow this pattern.

rissson:
Shouldn't most of those inherit from `InternallyManagedMixin`?

authentik/core/api/utils.py
```diff
@@ -21,6 +21,40 @@
     raise_errors_on_nested_writes,
 )
 
+# This is in label form to avoid an acute case of circular dependencies
+SIMPLE_THROUGH_MODELS = [
```
gergosimonyi: Make this a mixin to inherit from

BeryJu (approved):

BeryJu (changes requested):
All models need verbose_name and verbose_name_plural

authentik/lib/models.py
```diff
@@ -68,6 +68,13 @@ class InternallyManagedMixin:
     """Mixin for models that should _not_ be manageable via blueprint."""
 
 
+class SimpleThroughModel(models.Model, InternallyManagedMixin):
+    """Mixin for explicit many-to-many through models"""
+
+    class Meta:
+        abstract = True
```
BeryJu, line 83: Any reason for making this a separate class instead of just having InternallManagedMixin in the models?

gergosimonyi: Yes, `InternallyManagedMixin` is very specifically a part of `SimpleThroughModel`, since the aim of `SimpleThroughModel` is to be as close to a django-created implicit through model as possible.

((DISREGARD THE EDIT HISTORY OF THIS COMMENT))

authentik/lib/models.py
```diff
@@ -73,6 +73,13 @@ class InternallyManagedMixin:
     """Mixin for models that should _not_ be manageable via blueprint."""
 
 
+class SimpleThroughModel(models.Model, InternallyManagedMixin):
+    """Base class for explicit many-to-many through models"""
```
BeryJu:
```suggestion
    """Base class for explicit many-to-many through models, mainly used for adjusting API behaviour as if there were no explicit through models"""
```

BeryJu (approved):

PR #23354, opened 2026-06-24
website/docs: document proxy JWT identity verification

horner:
## Details

Documents the proxy provider's signed-JWT identity headers, which were previously undocumented, and explains the recommended verification pattern for upstream applications.

### Changes

- **`index.md`**
    - Add the `X-authentik-jwt` row to the user headers table and `X-authentik-meta-jwks` to the application headers table (both are sent by the outpost but were not listed).
    - Add a new **Verifying identity with the signed JWT** section describing how an upstream app should verify the JWT (read `X-authentik-jwt`, verify signature against a pinned JWKS URL, validate `iss`/`aud`/`exp`/`sub`, treat plain headers as convenience only), plus a security warning about pairing verification with network isolation.
    - Add a comparison table mapping the pattern to **Cloudflare Access** (signed token, JWKS/certs endpoint, `iss`, `aud`/AUD tag, and network isolation via Cloudflare Tunnel).
- **`header_authentication.mdx`**
    - Cross-link to the new section from an info admonition.

The header names were confirmed against `internal/outpost/proxyv2/application/mode_common.go` (`X-authentik-jwt` = `c.RawToken`, `X-authentik-meta-jwks` = `a.endpoint.JwksUri`). The Cloudflare details were confirmed against Cloudflare's official "Validate JWTs" documentation.

This is a documentation-only change.

### DIFF
--- a/website/docs/add-secure-apps/providers/proxy/header_authentication.mdx
+++ b/website/docs/add-secure-apps/providers/proxy/header_authentication.mdx
@@ -2,6 +2,10 @@
 title: Header authentication
 ---
 
+:::info
+To let upstream applications cryptographically verify the user's identity instead of trusting plain identity headers, the proxy also forwards a signed JWT. See [Verifying identity with the signed JWT](./index.md#verifying-identity-with-the-signed-jwt).
+:::
+
 ## Sending authentication
 
 ### Send HTTP Basic authentication
--- a/website/docs/add-secure-apps/providers/proxy/header_authentication.mdx
+++ b/website/docs/add-secure-apps/providers/proxy/header_authentication.mdx
@@ -2,10 +2,6 @@
 title: Header authentication
 ---
 
-:::info
-To let upstream applications cryptographically verify the user's identity instead of trusting plain identity headers, the proxy also forwards a signed JWT. See [Verifying identity with the signed JWT](./index.md#verifying-identity-with-the-signed-jwt).
-:::
-
 ## Sending authentication
 
 ### Send HTTP Basic authentication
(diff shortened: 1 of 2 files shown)


dewi-tik:
Thank you for your contribution @horner. We'll review and merge this ASAP.

dewi-tik:
authentik isn't using asymteric keys, therefore this method doesn't currently work. There is an open issue to switch to asymetric keys, however this is being held up at the moment because we don't have a good automated way to handle key rotation. But it is on our internal roadmap.

horner:
Thanks, that helps. I think I was mixing two different models:

```mermaid
sequenceDiagram
    participant B as Browser
    participant N as NGINX / reverse proxy
    participant O as authentik outpost
    participant A as authentik
    participant U as Upstream app

    B->>N: Request /app
    N->>O: Forward-auth / auth check
    O->>A: Validate session / policy
    A-->>O: User context
    O-->>N: Headers, including X-authentik-jwt
    N->>U: Request + identity headers
```

My proposed docs change assumed `X-authentik-jwt` was an app-verifiable signed JWT, similar to the Cloudflare Access pattern:

```mermaid
flowchart LR
    P[Auth proxy] -->|Signed JWT header| U[Upstream app]
    U -->|Fetch pinned JWKS/certs| J[JWKS endpoint]
    U -->|Validate iss, aud, exp, sub| OK[Accept identity]
```

But based on your comment, it sounds like the proxy provider/outpost token is not currently signed with an asymmetric key pair that an upstream app can verify against the advertised `X-authentik-meta-jwks` URL.

So the correct interpretation is probably:

1. `X-authentik-jwt` is emitted by the outpost, but it should **not** currently be documented as independently verifiable by the upstream app using JWKS.
2. The plain `X-authentik-*` headers remain trusted only when the upstream is network-isolated behind the outpost/reverse proxy.
3. A Cloudflare Access-style “verify signed JWT via JWKS/certs endpoint” pattern should wait until authentik supports asymmetric signing and key rotation for this use case.

Is that accurate?


For context, I was comparing this to our current OAuth2 Proxy setup (https://oauth2-proxy.github.io/oauth2-proxy/features/endpoints) , where NGINX uses the auth endpoint and the upstream receives identity only after the proxy has authenticated the request. I wanted to document a stronger app-side verification option for authentik, but if the outpost JWT cannot currently be verified that way, I’m happy to revise the PR to avoid implying that security property.
 
 
Is there a ticket for the "open issue to switch to asymetric keys"?  I am willing to contribute to working on it.

dewi-tik:
Yes, your interpretation is accurate. `X-authentik-jwt` cannot currently be independently verified by the upstream app via JWKS. Your Cloudflare Access-style pattern is held back until asymmetric signing is in place.

I can't find a relevant issue for the asymmetric key change. You're welcome to create a feature request. We always appreciate contributors.

I'm going to to close this PR for now - but feel free to re-open once the required changes are in place.

PR #23363, opened 2026-06-24
core: Agent instructions.

GirlBossRush:
## Details

This PR adds an AGENTS.md files to the repo root, giving agents some clarity on how to go about traversing the codebase.

This iteration is an attempt at a lean set of instructions rather than a deep dive into the codebase, but we can expand on it in future iterations if we find that agents are still struggling to get up to speed.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,170 @@
+## Project Overview
+
+This is the **authentik** monorepo — an open-source Identity Provider (IdP) for modern SSO. It speaks SAML, OAuth2/OIDC, LDAP, RADIUS, and SCIM, and is built to be self-hosted from a homelab to a large production cluster. The company is **Authentik Security, Inc.**; the product name is **always lowercase `authentik`**, even at the start of a sentence.
+
+It is a **polyglot monorepo**. Most work lands in one of the subtrees below; where a subtree has its own deeper guide, read it before working there:
+
+| Language       | Where                      | What it is                                                                              | Deeper guide                                |
+| -------------- | -------------------------- | --------------------------------------------------------------------------------------- | ------------------------------------------- |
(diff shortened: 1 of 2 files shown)


kensternberg-authentik (approved):
Looks like a solid start to me.  Nothing particularly controversial or wacky in there.

AGENTS.md
```diff
@@ -0,0 +1,170 @@
...
+| Outposts        | Go 1.26 (`goauthentik.io` module) — LDAP, proxy, RAC, RADIUS             |
+| Native services | Rust (2024 edition, `axum`) — server/worker components + shared crates   |
+| Web UI          | TypeScript, Lit 3, PatternFly 4 (see `web/`)                             |
+| Docs            | Docusaurus 3 (see `website/`)                                            |
+| API             | OpenAPI (`drf-spectacular`) → generated Go/Rust/TS clients               |
+| Python tooling  | `uv`, black, ruff, mypy (`--strict`), bandit                             |
+| Build hub       | GNU Make + per-language toolchains                                       |
+| CI / hosting    | GitHub Actions; distributed as Docker images and a Helm chart            |
+
+## Issue and PR Guidelines
```
dominic-r, line 166: how does this impact both our workflows?

AGENTS.md
```diff
@@ -0,0 +1,170 @@
...
+
+### Run
+
+```bash
+make run               # Run the authentik server + worker (uv run ak allinone)
+make run-watch         # Same, auto-reloading on .py/.rs/.go changes (needs watchexec)
+make migrate           # Apply Django migrations
+```
+
+### Test
```
dominic-r, line 95: would be nice to give it the command to test one specific file, or one specific file rather than the whole app like using the django test runner

dominic-r (approved):
Approving minus the last bit, internal discussion regarding it

PR #23385, opened 2026-06-25
website/docs: fix 2026.5 release notes mentioning docker-compose.yml

PeshekDotDev:
### Linked issues

---

## Checklist

### DIFF
--- a/website/docs/releases/2026/v2026.5.md
+++ b/website/docs/releases/2026/v2026.5.md
@@ -161,7 +161,7 @@ An integration is how authentik connects to third-party applications, directorie
 This release does not introduce any new requirements. You can follow the upgrade instructions below; for more detailed information about upgrading authentik, refer to our [Upgrade documentation](../../install-config/upgrade.mdx).
 
 :::warning Upgrade in order and upgrade all components
-Upgrades MUST happen in order, meaning that if you are 2 major releases behind, you need to upgrade to the previous major release before upgrading to this release. Refer to our [Upgrade documentation](../../install-config/upgrade.mdx) for more information on upgrade sequence.
+Upgrades MUST be performed sequentially by major version. If you are two or more major releases behind, you must first upgrade to each intermediate major release before upgrading to this one. Refer to our [Upgrade documentation](../../install-config/upgrade.mdx) for more information on upgrade sequence.
 
 The version of the authentik instance and of any outposts must be the same. We recommend that you always upgrade any outposts at the same time you upgrade your authentik instance.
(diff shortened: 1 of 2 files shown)


website/docs/releases/2026/v2026.5.md
```diff
@@ -166,15 +166,17 @@ When you upgrade, be aware that the version of the authentik instance and of any
...
 
 ```shell
-wget -O docker-compose.yml https://goauthentik.io/version/2026.5/lifecycle/container/compose.yml
+wget -O compose.yml https://goauthentik.io/version/2026.5/lifecycle/container/compose.yml
 docker compose up -d
 ```
 
 The `-O` flag retains the downloaded file's name, overwriting any existing local file with the same name.
 
+Upgrades MUST happen in order, meaning that if you are 2 releases behind, you need to upgrade to the last release before upgrading to this release.
```
dominic-r: I don't think this should be added on a release-specific basis and instead we should just link or mention the upgrade docs which already cover this point

PeshekDotDev: I definitely think we should mention it here, because a lot of people may look at this and want to quickly upgrade, not being aware of this. We have had a few issues be created because people jumped to the latest release and didn't follow this process, which is a sign that they don't know and aren't reading that section of the docs beforehand. Many will not click the links to the docs and scroll through and read enough to learn this fact.

However, I am very open to the wording of how we say this and also think this section should link to the docs section on upgrading, even having the warning link to the docs section could be a good idea

dewi-tik: Most people don't look at the upgrade documentation. Every release comes with issues caused by people not upgrading in sequence. So I'm happy having an extra warning in the release notes.

I've pushed some changes to change the wording slightly, put the warning in the upgrade section rather than under installation, and added an extra link to the upgrade documentation from the warning itself.

Happy to change the wording further or move the warning.

I merged the 2026.2 PR before seeing this. I've created a PR to revert the change and closed the 2026.2 cherry pick. Once we merge this PR, I'll update the other PR to match this one but for 2026.2.

PeshekDotDev: I think this is a better option. I added a warning to 2026.2, so we will have 2 in that section, but I think that's fine for one release

PeshekDotDev: A part of me is a little nervous that people may skim past the component wording, but I don't have a better answer

dewi-tik (approved):

PR #23396, opened 2026-06-25
stages/user_login: remember devices after login prompt

### LINKED ISSUE #22466: `authentik_device` cookie not set when "Stay signed in" is configured, causing `known_device` to always be `False`
### Describe the bug

When the "Stay signed in" duration is configured on a User Login Stage, the `authentik_device` cookie is never set. As a result, `known_device` is always `False` for every login, making the "Remember device" feature completely non-functional when used together with "Stay signed in".

This makes it impossible to use the `known_device` flag in notification rule expression policies, as it will always be `False` regardless of whether the device has logged in before.

### How to reproduce

1. Configure a User Login Stage with both `remember_device` (e.g. `days=30`) and `stay_signed_in` (e.g. `days=7`) set to a non-zero value
2. Log in — the "Stay signed in" prompt is shown to the user
3. Complete the login (regardless of whether "Stay signed in" is accepted or declined)
4. Check browser cookies — `authentik_device` is **not** present
5. Log out and log in again
6. Check the login event context in Admin UI → Events → `auth_method_args.known_device` is `False`

### Expected behavior

The `authentik_device` cookie should be set independently of the "Stay signed in" choice. "Remember device" and "Stay signed in" are two separate features and should not be mutually exclusive.

### Screenshots

_No response_

### Additional context

Root cause is in `/authentik/stages/user_login/stage.py`. The `do_login` method only calls `set_known_device_cookie` when `remember is None`:

```python
if remember is None:
    return self.set_known_device_cookie(user)
return self.executor.stage_ok()  # ← cookie is never set when remember is True/False
```

When "Stay signed in" is configured, the user is shown a prompt and `remember` is always `True` or `False` — never `None`. Therefore `set_known_device_cookie` is never reached and the `authentik_device` cookie is never set.

Suggested fix:

```python
response = self.set_known_device_cookie(user)
if remember is not None:
    return self.executor.stage_ok()
return response
```

### Deployment Method

Docker

### Version

2026.2.3

### Relevant log output

```shell

```

dominic-r:
- Route successful user-login completion through the known-device cookie helper even after the stay-signed-in prompt.
- Add a regression test that first creates the `authentik_device` cookie through the prompt path, clears server-side session evidence, and verifies the next login reports `known_device=True`.

## Root cause

`UserLoginStageView.do_login()` only set the known-device cookie when `remember is None`, which is the no-prompt path. When `remember_me_offset` is configured, the challenge response passes `remember` as `True` or `False`, so login returned a fresh `stage_ok()` response before the `authentik_device` cookie could be set.

Closes #22466

### DIFF
--- a/authentik/stages/user_login/stage.py
+++ b/authentik/stages/user_login/stage.py
@@ -180,6 +180,4 @@ class UserLoginStageView(ChallengeStageView):
             Session.objects.filter(
                 authenticatedsession__user=user,
             ).exclude(session_key=self.request.session.session_key).delete()
-        if remember is None:
-            return self.set_known_device_cookie(user)
-        return self.executor.stage_ok()
+        return self.set_known_device_cookie(user)
--- a/authentik/stages/user_login/tests.py
+++ b/authentik/stages/user_login/tests.py
@@ -21,6 +21,7 @@ from authentik.flows.views.executor import NEXT_ARG_NAME, SESSION_KEY_PLAN
 from authentik.lib.generators import generate_id
 from authentik.lib.utils.time import timedelta_from_string
 from authentik.root.middleware import ClientIPMiddleware
+from authentik.stages.password.stage import PLAN_CONTEXT_METHOD_ARGS
 from authentik.stages.user_login.middleware import (
     SESSION_KEY_BINDING_NET,
     BoundSessionMiddleware,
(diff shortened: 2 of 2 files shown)


dominic-r:
to turn in a more bug-fix PR for the entire stage

PR #23399, opened 2026-06-25
events: group application usage by ID

### LINKED ISSUE #21782: "Apps with most usage" should depend on application id and not name
### Describe the bug

When renaming an application, its usage starts from scratch:
Image

I'd not expected this and led to some confusion. It should be application-id specific, and not specific on the name.

### How to reproduce

1. Rename application
2. Login to this application
3. New entry will be there

### Expected behavior

The entry being renamed too.

### Screenshots

_No response_

### Additional context

_No response_

### Deployment Method

Docker

### Version

2026.2.2

### Relevant log output

```shell

```

### LINKED ISSUE #25278: Apps-with-most-usage aggregates by display name instead of slug
## What

Admin “Apps with most usage” aggregates by application name (breaks on rename) instead of slug/id.

## Want

- Admin “Apps with most usage” aggregates by application name (breaks on rename) instead of slug/id

## Ideas this came from

- [#11990 `Apps with most usage` to use `slug` instead of `name` to group the stats](https://github.com/goauthentik/authentik/issues/11990) — When checking the dashboard Apps with most usage in admin interface, the stats are not group correctly and in case the name changes, the stats are wrong. In Adm
- [#21782 "Apps with most usage" should depend on application id and not name](https://github.com/goauthentik/authentik/issues/21782) — When renaming an application, its usage starts from scratch: <img width="372" height="384" alt="Image" src="https://github.com/user-attachments/assets/6ec6f782-

dominic-r:
This fixes “Apps with most usage” aggregation so applications are grouped by the stable serialized application `pk` instead of the full serialized application payload.

Before, renaming an application could cause the same application to appear as multiple rows in the usage response. The stored event context includes serialized application details, including `authorized_application.name`, so a rename changed the grouped value even though the underlying application was the same.

Closes #21782
Closes #25278

### DIFF
--- a/authentik/events/api/events.py
+++ b/authentik/events/api/events.py
@@ -257,7 +257,7 @@ class EventViewSet(
             .annotate(application=KeyTransform("authorized_application", "context"))
             .annotate(user_pk=KeyTextTransform("pk", "user"))
             .values("application_pk")
-            .annotate(counted_events=Count("event_uuid"))
+            .annotate(counted_events=Count("pk"))
             .annotate(unique_users=Count("user_pk", distinct=True))
             .annotate(applications=ArrayAgg("application", order_by=("-created", "-event_uuid")))
             .values("unique_users", "applications", "counted_events")
--- a/authentik/events/api/events.py
+++ b/authentik/events/api/events.py
@@ -15,7 +15,7 @@ from django.utils.timezone import now
 from djangoql.schema import DateTimeField as QLDateTimeFIeld
 from djangoql.schema import IntField, StrField
 from drf_spectacular.types import OpenApiTypes
-from drf_spectacular.utils import OpenApiParameter, extend_schema, extend_schema_field
+from drf_spectacular.utils import OpenApiParameter, extend_schema
 from guardian.shortcuts import get_objects_for_user
 from rest_framework.decorators import action
 from rest_framework.fields import (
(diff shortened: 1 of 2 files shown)


authentik/events/api/events.py
```diff
@@ -242,18 +243,23 @@ def top_per_user(self, request: Request):
...
+            .values("application_pk")
+            .annotate(counted_events=Count("event_uuid"))
             .annotate(unique_users=Count("user_pk", distinct=True))
-            .values("unique_users", "application", "counted_events")
+            .annotate(applications=ArrayAgg("application", order_by=("-created", "-event_uuid")))
+            .values("unique_users", "applications", "counted_events")
             .order_by("-counted_events")[:top_n]
         )
+        for event in events:
+            event["application"] = event.pop("applications")[0]
```
BeryJu: Can't we do this on the database level? Or I'd rather adjust the serializer than doing this

authentik/events/api/events.py
```diff
@@ -242,16 +248,19 @@ def top_per_user(self, request: Request):
...
             get_objects_for_user(request.user, "authentik_events.view_event")
             .filter(action=filtered_action)
             .exclude(context__authorized_application=None)
+            .annotate(application_pk=application_pk)
             .annotate(application=KeyTransform("authorized_application", "context"))
             .annotate(user_pk=KeyTextTransform("pk", "user"))
-            .values("application")
-            .annotate(counted_events=Count("application"))
+            .values("application_pk")
+            .annotate(counted_events=Count("event_uuid"))
```
dominic-r:
```suggestion
            .annotate(counted_events=Count("pk"))
```

PR #23429, opened 2026-06-26
website/docs: clarify S3 storage provider configuration

dominic-r:
The existing S3 guide does not show how authentik settings map to common S3-compatible services, while maintaining full third-party setup tutorials would create untested and fast-changing documentation.

This keeps one shared setup and migration guide plus a compact provider reference limited to endpoint and authentication differences. It links to each service's own documentation, removes the non-actionable Azure gateway example, and makes clear that authentik does not test every listed service. Validated with `make docs`.


dominic-r (commented):
not sure of all these partials, but do let me know dewi

website/docs/sys-mgmt/ops/storage-s3/azure-blob-s3-gateway.mdx
```diff
@@ -0,0 +1,41 @@
+---
+title: Azure Blob Storage through an S3-compatible gateway
+---
+
+Azure Blob Storage does not natively expose an S3-compatible API. To use Azure Blob Storage with authentik's S3 backend, place an S3-compatible gateway in front of Azure Blob Storage and configure authentik to use the gateway endpoint.
```
dominic-r: yea this should probbaly show one provider as example or just remove it

website/docs/releases/2024/v2024.2.md
```diff
@@ -116,7 +116,7 @@ slug: /releases/2024.2
 
 - **S3 file storage**
 
-    Media files can now be stored on S3. Follow the [setup guide](../../sys-mgmt/ops/storage-s3.md) to get started.
+    Media files can now be stored on S3. Follow the [setup guide](../../sys-mgmt/ops/storage-s3/index.mdx) to get started.
```
dewi-tik:
```suggestion
    Media files can be stored on S3. Follow the [setup guide](../../sys-mgmt/ops/storage-s3/index.mdx) to get started.
```

website/docs/sys-mgmt/ops/storage-s3/wasabi.mdx
```diff
@@ -0,0 +1,46 @@
+---
+title: Wasabi
+---
+
+Use Wasabi with authentik through Wasabi's S3-compatible API.
```
dewi-tik:
```suggestion
Use Wasabi as an S3 storage backend for authentik by connecting to Wasabi's S3-compatible API.
```

I would change the opening sentences for all the different providers to something similar to this. Rather than "Use <provider> with authentik".

The change doesn't have to match mine, just needs to be a bit clearer.

dewi-tik (changes requested):

website/docs/sys-mgmt/ops/storage-s3/wasabi.mdx
antoniojrod: Accepted. Added a chart-context-specific `CHART_CONTEXT_ERROR_RESPONSES` whose 400 description covers all request-shape errors (malformed service date, unknown/empty context, missing patient id) and notes that collection-time open/verify/read failures come back as `status="failed"` in the body rather than a 400. The endpoint now uses it instead of `DATE_ERROR_RESPONSES`. (main.py:1023)

rissson:
maybe we want to call out that we are not testing with each of those providers

dominic-r:
good point, marc

dewi-tik (approved):

PR #23518, opened 2026-06-29
web/flows: fix untranslated user field labels in identification stage

### LINKED ISSUE #23517: Identification Stage user-field label/placeholder is not translated (stuck in English) for all non-English locales
### Describe the bug

In the Identification Stage (e.g. the bundled default-authentication-flow), the identifier input's label and placeholder — "Username", "Email", "UPN", or a combination like "Username or Email" — always render in English, regardless of the UI locale. Other strings on the same screen (e.g. "Log in to continue to…", "Password", "Need an account?") translate correctly, so this is specific to the user-field labels.

The strings themselves exist and are translated in the locale catalogs; they are simply never applied at runtime.

### How to reproduce

Steps to reproduce:

1. Open the login page served by the default-authentication-flow (or any flow whose first stage is an Identification Stage).
2. Set the locale to any non-English locale. German is a good example.
3. Look at the username/email input field's label and placeholder.

Observed: the label/placeholder is in English ("Username") even though the rest of the page is translated.

### Expected behavior

The identifier field's label and placeholder should be shown in the selected locale (e.g. "Anmeldename" in German), and should update when the locale changes, consistent with every other translated string on the page.

### Screenshots

_No response_

### Additional context

_No response_

### Deployment Method

Docker

### Version

dev

### Relevant log output

```shell

```

mohamed-karam:
### What does this PR change?

Fixes the Identification Stage so that the identifier field's label and placeholder ("Username", "Email", "UPN", or a combined variant like "Username or Email") are translated into the active UI locale.

The labels were built from a module-level `UI_FIELDS` constant whose `msg()` calls were evaluated once at import time. This constant is replaced with a `uiFieldLabels()` function that is invoked inside `renderInput()`, so the strings are resolved and rendered with the active locale.

Authentik's web UI uses `@lit/localize` in runtime mode, where only `msg()` calls made during render pick up the active locale and re-render when it changes. Because `UI_FIELDS` ran its `msg()` calls at module-import time - before the locale is applied — the field labels were frozen at the source locale (English) for every non-English UI. The translations already existed in the catalogs; they were simply never applied at runtime. As a result, the Identification Stage on the default authentication flow always showed an English field label even when the rest of the page was translated.

### How was this tested?

- Manually: Open `default-authentication-flow`  and change the locale to a non-English locale (German). Before the change the field label/placeholder rendered in English; after the change, it's renders localized (e.g.  "Anmeldename") and updates when the locale changes.
- Verified no other references to the removed `UI_FIELDS` constant remain.
- `make all` (lint, build, test).
### Linked issues

closes #23517

---

## Checklist

### DIFF
--- a/web/src/flow/components/ak-flow-card.ts
+++ b/web/src/flow/components/ak-flow-card.ts
@@ -27,10 +27,10 @@ export class FlowCard extends AKElement {
     role = "presentation";
 
     @property({ type: Object })
-    challenge?: Pick<FormStaticChallenge, "flowInfo">;
+    public challenge: Pick<FormStaticChallenge, "flowInfo"> | null = null;
 
     @property({ type: Boolean })
-    loading = false;
+    public loading = false;
 
     static styles: CSSResult[] = [PFLogin, PFTitle, Styles];
 
--- a/web/src/flow/stages/identification/IdentificationStage.ts
+++ b/web/src/flow/stages/identification/IdentificationStage.ts
@@ -15,6 +15,11 @@ import CaptchaDisplayController from "#flow/stages/identification/controllers/Ca
 import RememberMeController from "#flow/stages/identification/controllers/RememberMeController";
 import WebauthnController from "#flow/stages/identification/controllers/WebauthnController";
 import Styles from "#flow/stages/identification/styles.css";
+import {
+    compareLoginSource,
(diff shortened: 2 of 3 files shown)


BeryJu (approved):
lgtm

GirlBossRush (approved):

PR #23522, opened 2026-06-29
core/api: add PrivilegedFieldsSerializerMixin

emilburzo:
This is meant for free-form `JSONField` settings-style fields that can hold sensitive values, where marking the whole field `write_only` would be too coarse (it would also hide the value from the user who legitimately manages the object).

### DIFF
--- a/authentik/core/api/utils.py
+++ b/authentik/core/api/utils.py
@@ -81,6 +81,27 @@ class ModelSerializer(BaseModelSerializer):
         return instance
 
 
+class PrivilegedFieldsSerializerMixin:
+    """Serializer mixin that empties credential-bearing fields for callers who can
+    only use the object, not manage it."""
+
+    privileged_fields: list[str] = []
+
+    def to_representation(self, instance: Model) -> dict:
+        data = super().to_representation(instance)
+        request = self.context.get("request")
+        if not request or not self.privileged_fields:
+            return data
+        meta = instance._meta
+        permission = f"{meta.app_label}.view_{meta.model_name}"
+        if request.user.has_perm(permission, instance) or request.user.has_perm(permission):
+            return data
+        for field in self.privileged_fields:
+            if field in data:
+                data[field] = {}
+        return data
+
+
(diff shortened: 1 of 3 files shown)


authentik/core/api/utils.py
```diff
@@ -81,6 +81,27 @@ def update(self, instance: Model, validated_data):
         return instance
 
 
+class PrivilegedFieldsSerializerMixin:
+    """Serializer mixin that empties credential-bearing fields for callers who can
+    only use the object, not manage it."""
+
+    privileged_fields: list[str] = []
```
BeryJu, line 87: To align with how other things in DRF work I'd probably make this both a variable but also add a callable that by default just returns `self.privileged_fields`

authentik/core/api/utils.py
```diff
@@ -81,6 +81,27 @@ def update(self, instance: Model, validated_data):
...
+        request = self.context.get("request")
+        if not request or not self.privileged_fields:
+            return data
+        meta = instance._meta
+        permission = f"{meta.app_label}.view_{meta.model_name}"
+        if request.user.has_perm(permission, instance) or request.user.has_perm(permission):
+            return data
+        for field in self.privileged_fields:
+            if field in data:
+                data[field] = {}
```
BeryJu, line 104: instead of `{}` being the censored value, I think it should either be some sort of tombstone type, a fixed string or configurable

authentik/core/tests/test_api_utils.py
```diff
@@ -33,3 +37,52 @@ def test_all_serializers_descend_from_custom(self):
...
+    """Serializer that treats `attributes` as privileged, for testing the mixin"""
+
+    privileged_fields = ["attributes"]
+
+    class Meta:
+        model = Group
+        fields = ["name", "attributes"]
+
+
+class TestPrivilegedFieldsSerializerMixin(APITestCase):
```
BeryJu: new test case = new file (imo, not a strong requirement)

BeryJu (commented):
lgtm, some nits

PR #23060, opened 2026-06-15
web/flows: fix race condition in continuous login and support source stages in authentication flows (2026.2)

ryanpesek:
same as https://github.com/goauthentik/authentik/pull/23049 but for 2026.2

---

## Checklist

-   [ ] The project has been linted, built, and tested (`make all`)
-   [ ] The documentation has been updated and formatted (`make docs`)

PR #23557, opened 2026-06-30
web/flows: fix race condition in continuous login and support source stages in authentication flows (cherry-pick #23049 to version-2026.2)

GirlBossRush:
This PR is a backport of #23049 to `version-2026.2`, rebuilt from the merged commit on `main` (`42960643`) rather than the earlier PR revision the previous backport branch used in #23060

## Adaptations for version-2026.2
- `version-2026.2` lacks the flow controllers refactor (#20898), so `FlowMultitabController`'s logic is ported inline into `FlowExecutor`'s `ak-multitab-continue` listener. It uses the merged design's same-origin suppress/leave logic for `next` (the SP-initiated SAML resume fix) but keeps the `blank://blank` app-hiding guard, which 2026.2 still relies on (the dashboard-hide option only landed in 2026.5).
- The multi-tab files (`broadcast.ts`, `orchestrator.ts`, `TabID.ts`, new `events.ts`/`messages.ts`) are identical to `main`.
- `RedirectStage.ts` carries only the #23049 changes (`final_redirect` handling), without unrelated `main` drift.
- `test_auth_n_request.py` `setUp` creates the `test-app` Application (already present on `main`).
- Generated API clients (go/rust/ts) come from `schema.yml` at release time; 2026.2 tracks none of them, so only the schema contract is committed.

### DIFF
--- a/web/src/flow/tabs/broadcast.ts
+++ b/web/src/flow/tabs/broadcast.ts
@@ -1,63 +1,32 @@
+import { AKMultiTabEvent, AKMultiTabExitEvent } from "#flow/tabs/events";
+import { BroadcastMessage, BroadcastMessageType } from "#flow/tabs/messages";
 import { TabID } from "#flow/tabs/TabID";
 
 import { ConsoleLogger, Logger } from "#logger/browser";
 
-export const BROADCAST_CHANNEL_NAME = "authentik";
-
-enum BroadcastMessageType {
-    discover = "discover",
-    continue = "continue",
-    exit = "exit",
-    discoverReply = "discoverReply",
-}
-
-export interface BroadcastMessage {
-    type: BroadcastMessageType;
-    sender: string;
-    [key: string]: unknown;
-}
+import { match } from "ts-pattern";
 
 export class Broadcast extends BroadcastChannel implements Disposable {
-    static shared = new Broadcast();
+    public static readonly ChannelName = "authentik";
+    public static readonly SessionStorageKey = "authentik_tab_resume_id";
+
+    public static readonly shared = new Broadcast();
 
     protected discoveredTabIDs = new Set<string>();
-    public exitedTabIDs: string[] = [];
+    protected shouldSuppressExit = false;
 
     protected logger: Logger;
+    protected storageKey: string;
 
-    protected messageListener = (ev: MessageEvent<BroadcastMessage>) => {
-        this.logger.debug("broadcast event", ev.data);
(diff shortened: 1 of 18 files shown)


web/src/flow/tabs/broadcast.ts
```diff
@@ -1,79 +1,115 @@
...
+     */
+    public async discoverTabs(): Promise<Set<string>> {
         this.discoveredTabIDs.clear();
 
-        this.postMessage({
-            type: BroadcastMessageType.discover,
+        this.dispatchMessage({
+            type: BroadcastMessageType.Discover,
             sender: TabID.shared.current,
         });
```
kensternberg-authentik, line 114: Immediately after this is a delay of 20ms to await the transfer of discovered Tabs. If a tab is busy in its main thread, it could easily miss this window.  Is 20ms really long enough?

web/src/flow/tabs/orchestrator.ts
```diff
@@ -1,68 +1,149 @@
...
-        timeout = self.setTimeout(() => {
-            logger.warn("Timed out waiting for tab to exit, moving on", tab);
-            cancelAnimationFrame(checker);
-            done.resolve();
-        }, TAB_EXIT_TIMEOUT_MS);
+        logger.debug("Telling tab to continue", tab);
+        Broadcast.shared.resumeTab(tab, resumeID);
 
-        await done.promise;
+        await done;
```
kensternberg-authentik, line 146: Okay, dumb question. `for` loops with an `await` are a code smell. Why isn't this an array fed to `Promise.allSettled()`?  You're waiting on a _per tab_ basis, serially, sequentially.  That seems... wrong, somehow.  Maybe I don't understand the problem.

kensternberg-authentik (approved):
Mixed feelings... This seems esoteric enough that I have trouble following along, but the basic code makes sense.  The window seems a little small (20ms) for something that's looking for multiple tabs; in a heavily-loaded system (like my little toy Surface 1), chrome can bog down badly and getting a response back in that short a period seems unlikely.

Also, there's a `for/await` loop here that's gathering information from a bunch of tabs; the order doesn't seem to matter, so why isn't this "launch a bunch of promises and gather them with `allSettled()`?" If they're loaded tabs, the user could be waiting on a frozen screen as each one runs or times out (15 seconds each!).  

I'm not going to block approval; this is a fair backport. I just want it to be one that doesn't generate service calls.

gergosimonyi (approved):

PR #23578, opened 2026-07-01
web/admin: refactor user detail page layout

dominic-r:
The user detail page had grown into a single large component that owned page loading, tab structure, overview cards, credentials, roles, applications, notes, and user actions. This made the overview hard to reason about and left several awkward UI states, especially sparse user profile data, notes, and recovery actions.

BEFORE:

image

AFTER:

image


web/src/admin/users/UserOverviewTab.ts
```diff
@@ -0,0 +1,84 @@
...
+import PFCard from "@patternfly/patternfly/components/Card/card.css";
+import PFContent from "@patternfly/patternfly/components/Content/content.css";
+import PFGrid from "@patternfly/patternfly/layouts/Grid/grid.css";
+
+@customElement("ak-user-overview-tab")
+export class UserOverviewTab extends AKElement {
+    @property({ attribute: false })
+    public user?: User;
+
+    @property({ type: Number })
```
kensternberg-authentik, line 24: I feel like this is a mistake; all of the information other than the user's primary key is derivative. Sure, you're just passing it around, this is more a "display this stuff" utility component, but it's still a worthwhile thought experiment to say, "In the context of a running session, could this component be complete with just that primaryKey?"

Not something to change, just something to consider.

web/src/admin/users/UserInfoCard.ts
```diff
@@ -0,0 +1,297 @@
...
+                    grid-template-columns: minmax(0, 1fr);
+                }
+            }
+        `,
+    ];
+
+    public override connectedCallback() {
+        super.connectedCallback();
+
+        this.#resizeObserver = new ResizeObserver(([entry]) => {
```
kensternberg-authentik: It's 2026.  Container queries are a thing. https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Containment/Container_queries. `resizeObserver` should no longer be necessary.  And the `512` conflicts with the CSS `max-width: 768`.  :-). Pick one.

web/src/admin/users/UserRolesTab.ts
```diff
@@ -0,0 +1,86 @@
...
+            return;
+        }
+
+        this.activatedTabs = new Set([...this.activatedTabs, tab]);
+    }
+
+    protected renderWhenActive(tab: string, content: unknown) {
+        return this.activatedTabs.has(tab) ? content : null;
+    }
+
```
kensternberg-authentik: The tab management here is present in three different places: UserRolesTable, UserViewTab, and UserCredentialsTab.  That's sufficient duplication to justify extracting it somewhere else.

kensternberg-authentik (approved):
Still had thoughts.

PR #23592, opened 2026-07-02
root, web, docs, packages/*-config: Fix pnpm symlinks, version ranges

GirlBossRush:
## Details

This PR is a follow up to #22826.

### Dependabot clean up

- fixed some issues that dependabot carried over from its PRs that were created before the pnpm migration
- fixed some deprecation warnings stemming from latest pnpm update

## Package version ranges 

Tightened up the prettier version, preventing prettier's new algo from counting the line length differently, resulting in a lot of unnecessary diffs

## Symlink fixes

`packages/eslint-config` and `packages/prettier-config` mutually `link:`-depended on each other so each could lint/format its own source. Under npm's flat hoisting that was harmless. Under pnpm's isolated linker the two get directly nested `node_modules` symlinks, forming an infinite path that any symlink-following recursive walk hits with `ENAMETOOLONG`.

That broke the website build with a misleading "Sidebars file failed to be loaded", whose real cause was a deep `scandir ENAMETOOLONG` under the theme's nested `node_modules`.

This PR drops `@goauthentik/eslint-config` from `prettier-config`'s `devDependencies` to cut one edge of the cycle. Its `eslint.config.mjs` still resolves the config via the root `node_modules` symlink (workspace v2.0.0), so linting is unaffected.


dominic-r (approved):

kensternberg-authentik (approved):
It gets us back to sanity.

dewi-tik (approved):

rissson (approved):

Issue #22823, opened 2026-06-03
Expiring users/offboarding users (ent)


BeryJu:
Two main use-cases
- Automatically expire users after a configurable relative duration in which a user has logged in
- Offboarding, per-user absolute datetime where account is deactivated.
- Potentially also password expiry (side-task)

Common:
- Both actions need revoke all tokens/sessions/etc

Differences:
- Offboarding date is saved per-user
- Auto-expiry is saved per group...?

#### Use Object Lifecycle system 
 - add option to select user
 - Add type to lifecycle rules of {review, action}
 - add action of expire
 - figure out how to deal with offboarding of a single user

#### User state
Maybe we need to get rid of `is_active` and replace it with a `state`/`status` field of  {active, deactivated, locked}? This would allow us to show better UI in flows for locked accounts (mainly applies to auto-expiry), to tell users they might need to reset passwords/etc.

Migration path:
- Add `status` + `deactivation_reason` (release 1), which writes to both
- Remove `is_active` (backfill from logic above) (release 2)

Very rough idea: does any of the lockdown stuff help here? Might be helpful when locking a user.

cc @atereshkin for opinions/thoughts

atereshkin:
This could work within Object Lifecycle system, but there's one more mismatch in semantics that we'd need to deal with. Lifecycle rules have intervals, but those are relative to the previous review. Repurposing them to mean "time since last login" could be rather confusing (and would make the two types of lifecycle rules function in entirely different ways).
I think it would make more sense to have an additional field for "time since login" on expire-type lifecycle rules. That would leave the user with the need to specify a review interval while they probably just want users to expire right away, but that sounds like an acceptable tradeoff to me. 
Now we could take this one step further and instead of "action" rules introduce automatic/manual review types. The automatic reviews would have policy attached to determine whether an object passes review and an action to take if it doesn't.

BeryJu:
from @gergosimonyi 

I don't like doing it with object lifecycle
Should just be a task IMO
I'm always and forever a REST enjoyer, so I'd create a new resource called UserExpiration, have a time field and other fields that inform what to do when the time comes
Task runs every minute: get all UserExpiry objects where time has come, if there are any, potentially spin each off to another task that actually does the expiration

BeryJu:
The conclusion to the structure for this

have a higher-level thing above current object lifecycle (which will probably be called object lifecycle from now on)
move the current one to something named "audit lifecycle"/"review lifecycle"/etc
[...]

Issue #5375, opened 2023-04-25
User expiration (off boarding)


guyguy333:
**Is your feature request related to a problem? Please describe.**

We would like to set a user expiration date for off boarding . When a user ends its "mission" on an end date, we would like to set this end date ahead of end date to avoid manual edition of user to set as "inactive".

**Describe the solution you'd like**

I expect:
- a field in user to set expiration date. 
- all generated tokens/sessions to not be after expiration date (ex: if user session if 24h and user login at 10, we don't want session to end 24 later, but instead at midnight of end date). 
- Not sure if possible but a way to automatically force logout all existing sessions. 
- It would be also great to have an option to automatically remove user from all groups after expiration date.
- A web hook to trigger external off boarding

**Describe alternatives you've considered**

Running all previous steps manually

**Additional context**

BeryJu:
Even though it's not fully intended for this, you can use two attributes on the user:

```yaml
goauthentik.io/user/generated: true
goauthentik.io/user/expires: 1682499574
```

This will delete the user at the specified time, removing them from all groups, revoking all sessions and all tokens. To trigger this externally you can use the user delete API https://goauthentik.io/developer-docs/api/browser#delete-/core/users/-id-/

guyguy333:
Thank you @BeryJu, indeed it could work. However, currently if we delete users then we lose username associated to event logs (after it's Anonymous User) so it's hard to track something in the history if we loose original username if we want to investigate a potential issue. Do you think there could be a kind of "archive user" or "set to inactive"  removing them from all groups, revoking all sessions and all tokens ?

BeryJu:
Events are retained when a user is deleted and include username and email. However https://goauthentik.io/docs/installation/configuration#authentik_gdpr_compliance is enabled by default which causes those events to be deleted

BeryJu:
Things required for implementing this:

- Alter user deactivate API to include datetime for when deactivation should happen (null for instantly)
- A separate task queue for "realtime" tasks (otherwise if we schedule a task for a specific timestamp and a scim/ldap/etc sync happens right before, the task might be delayed)

melizeche:
closed by #23624 :)

This feature will be released in 2026.8.0

PR #23624, opened 2026-07-02
enterprise/core: add scheduled user offboarding

melizeche:
## Details

Adds a way to schedule deactivating or deleting a user at a future time, instead of doing it by hand when someone leaves.

From a user's page (Admin > Users) there's a new "Schedule Offboarding" button. You pick an action (deactivate or delete), a date/time, and whether to also revoke their sessions and tokens. 

A worker checks every 5 minutes and runs anything that's due. While one is pending, the button flips to "Cancel Offboarding", which asks for confirmation and keeps the record as canceled. 

You can't schedule one for yourself or an internal service account, and a user can only have one pending at a time.

The actual deactivate/delete logic lives in a small `offboard_user()` helper so it can be reused later by the policy-driven action lifecycle. Every run emits a `USER_OFFBOARDED` audit event, attributed to the admin who scheduled it.

image

image

image

image

## Report UI

https://github.com/goauthentik/authentik/pull/24266

## Related

#5375
#22823

---

## Checklist


authentik/enterprise/lifecycle/offboarding.py
```diff
@@ -0,0 +1,78 @@
...
+from structlog.stdlib import get_logger
+
+from authentik.core.models import AuthenticatedSession, Token, User
+from authentik.enterprise.lifecycle.models import OffboardingAction
+from authentik.events.models import Event, EventAction
+
+LOGGER = get_logger()
+
+
+def revoke_sessions_for_user(user: User) -> int:
```
BeryJu: Could probably be shared with account lockdown, might make sense to refactor into a shared function (...?) that deletes related things to a user when they're deactivated...maybe this could be a signal too?

authentik/enterprise/lifecycle/apps.py
```diff
@@ -11,12 +11,19 @@ class ReportsConfig(EnterpriseConfig):
...
             ScheduleSpec(
                 actor=apply_lifecycle_rules,
                 crontab=f"{fqdn_rand('lifecycle_apply_lifecycle_rules')} "
                 f"{fqdn_rand('lifecycle_apply_lifecycle_rules', 24)} * * *",
-            )
+            ),
+            ScheduleSpec(
+                actor=execute_due_offboardings,
+                crontab=f"{fqdn_rand('lifecycle_execute_due_offboardings', 5)}-59/5 * * * *",
+            ),
```
BeryJu, line 26: Marc says this looks fine, i personally prefer just scheduling the task at a future time and deleting it when the offboarding is cancelled

authentik/enterprise/lifecycle/api/offboarding.py
```diff
@@ -0,0 +1,104 @@
...
+    queryset = UserOffboarding.objects.select_related("user", "created_by").all()
+    serializer_class = UserOffboardingSerializer
+    search_fields = ["user__username"]
+    ordering = ["scheduled_for"]
+    ordering_fields = ["scheduled_for", "created_at", "status"]
+    filterset_fields = ["user__uuid", "status", "action"]
+    # Offboarding records are immutable: they can be scheduled (POST), cancelled
+    # (DELETE → soft-cancel), and read — but never edited. PUT/PATCH would let a
+    # terminal audit row's action/schedule/user be rewritten after the fact.
+    http_method_names = ["get", "post", "delete", "head", "options"]
```
gergosimonyi: This seems to still allow `PUT`/`PATCH`, just not put it in the schema. Is an additional guard required?

gergosimonyi (commented):
Generally lovely, I have a design comment. This adds offboarding as kind of a coupled sibling of the thing currently called "lifecycle". I believe that design is okay for the short-term, but I foresee more lifecycle things being added soon.

I believe the following restructure should be done:

1. The thing currently called "lifecycle" should be called something else. Maybe "object review" or something along those lines.
2. No-op: the thing currently called "user offboarding" is nicely named.
3. These two should use common stuff from a higher-order thing, that can after the rename in (1.) be called "lifecycle".
4. Minor detail: these two siblings should probably have separate folders within `/authentik/enterprise/lifecycle/`.

authentik/enterprise/core/revocation.py
```diff
@@ -0,0 +1,108 @@
...
+    """
+    if not sessions and not tokens:
+        return
+    with atomic():
+        for queryset in _revocable_querysets(user, sessions=sessions, tokens=tokens):
+            queryset.delete()
+    while has_revocable_artifacts(user, sessions=sessions, tokens=tokens):
+        with atomic():
+            for queryset in _revocable_querysets(user, sessions=sessions, tokens=tokens):
+                queryset.delete()
```
BeryJu, line 105: probably do this in the same transaction?

authentik/enterprise/lifecycle/api/offboarding.py
melizeche: http_method_names was actually enforced at dispatch (405, covered by the test), but switched to mixin composition anyway so the update handlers don't exist at all

melizeche:
@gergosimonyi @BeryJu changes since review:

- Session/token revocation moved to a shared `revoke_user_access()` in `authentik/enterprise/core/revocation.py`, used by both offboarding and account lockdown so the two can't drift on what "revoke access" means. The delete-and-retry loop now runs in a single transaction.
- `UserOffboardingViewSet` is composed from mixins instead of `ModelViewSet` with `http_method_names`, so the update handlers don't exist at all (records stay immutable, same as before, just structural now)
- Restructured the lifecycle app into an umbrella with two sibling packages: the existing access-review feature moved to `enterprise/lifecycle/review/`, offboarding lives in `enterprise/lifecycle/offboarding/`. Internal-only: app label, migrations, and API paths are unchanged (no schema diff). Dramatiq actor names change with the module paths; stale schedules are cleaned up by the reconciler on startup

authentik/enterprise/lifecycle/models.py
```diff
@@ -1,357 +1,40 @@
...
+
+The lifecycle app is an umbrella for sibling features that each live in their
+own package: `review/` (object access reviews) and `offboarding/` (user
+offboarding). Django loads models through this module, so every feature's
+models must be imported here.
+"""
+
+# Import order sets model registration order, which the API schema's ModelEnum
+# follows; review stays first to match the pre-split layout.
+# isort: off
```
melizeche, line 11: The # isort: off here pins review before offboarding so ModelEnum keeps its old order and this stays a zero-diff move in schema.yml and the generated clients. No problem to drop it and eat the reorder diff if you'd rather

authentik/enterprise/lifecycle/offboarding/models.py
```diff
@@ -0,0 +1,134 @@
...
+
+
+class UserOffboarding(SerializerModel):
+    """Scheduled deactivation/deletion of a single user at an absolute point in time."""
+
+    id = models.UUIDField(primary_key=True, default=uuid4)
+    user = models.ForeignKey(
+        "authentik_core.User", on_delete=models.CASCADE, related_name="offboardings"
+    )
+    scheduled_for = models.DateTimeField(
```
gergosimonyi:
```suggestion
    scheduled_at = models.DateTimeField(
```

(?)

authentik/enterprise/lifecycle/offboarding/models.py
```diff
@@ -0,0 +1,134 @@
...
+    revoke_tokens = models.BooleanField(
+        default=True, help_text=_("Revoke all of the user's tokens when offboarding.")
+    )
+
+    status = models.TextField(choices=OffboardingStatus.choices, default=OffboardingStatus.PENDING)
+    created_by = models.ForeignKey(
+        "authentik_core.User", on_delete=models.SET_NULL, null=True, related_name="+"
+    )
+    created_at = models.DateTimeField(auto_now_add=True)
+    executed_on = models.DateTimeField(null=True, default=None)
```
gergosimonyi:
```suggestion
    executed_at = models.DateTimeField(null=True, default=None)
```

(?)

gergosimonyi (approved):
Works for me, you may or may not want to wait for @BeryJu's review.

authentik/enterprise/lifecycle/offboarding/tasks.py
```diff
@@ -0,0 +1,59 @@
...
+    due_pks = UserOffboarding.objects.filter(
+        status=OffboardingStatus.PENDING,
+        scheduled_at__lte=timezone.now(),
+    ).values_list("pk", flat=True)
+    if not due_pks:
+        return
+    schedule = Schedule.objects.get(actor_name=execute_due_offboardings.actor_name)
+    for pk in due_pks:
+        # rel_obj groups each execution under the sweeper's schedule in the UI.
+        execute_offboarding.send_with_options(args=(str(pk),), rel_obj=schedule)
```
rissson:
```suggestion
    task = CurrentTask.get_task()
    # Only the pk is dispatched, so fetch pks alone rather than whole rows.
    due_pks = UserOffboarding.objects.filter(
        status=OffboardingStatus.PENDING,
        scheduled_at__lte=timezone.now(),
    ).values_list("pk", flat=True)
    for pk in due_pks:
        # rel_obj groups each execution under the sweeper's schedule in the UI.
        execute_offboarding.send_with_options(args=(str(pk),), rel_obj=task.rel_obj)
```

authentik/enterprise/lifecycle/review/tasks.py
```diff
@@ -0,0 +1,46 @@
...
+from authentik.events.models import Event, Notification, NotificationTransport
+from authentik.tasks.schedules.models import Schedule
+
+
+@actor(description=_("Dispatch tasks to apply lifecycle rules."))
+def apply_lifecycle_rules():
+    for rule in LifecycleRule.objects.all():
+        apply_lifecycle_rule.send_with_options(
+            args=(rule.id,),
+            rel_obj=Schedule.objects.get(actor_name=apply_lifecycle_rules.actor_name),
```
rissson: Same here, this should use task.rel_obj

rissson (changes requested):

authentik/enterprise/lifecycle/apps.py
rissson: The issue with future schedules is that offboarding is planned for a very far date, and authentik is updated in between, we might never find that out.

Regardless, these are the sort of tasks that should not be tasks but housekeeping routines run in the background, which we'll do in 2-3 releases, so I think this is fine for now.

rissson (approved):

BeryJu (approved):

PR #23664, opened 2026-07-03
tasks: add pre_delete for TasksModel to avoid OOM when deleting object with many tasks linked

rissson:
### Linked issues

---

## Checklist


melizeche (approved):
looks good to me. Do you think adding tests for task deletion is worth it here or nah?

rissson:
> looks good to me. Do you think adding tests for task deletion is worth it here or nah?

They will be deleted regardless. This is just us doing it a bit earlier to avoid the OOM

PR #23670, opened 2026-07-04
web: add additional context to event log messages

### LINKED ISSUE #23193: Events logs should be more descriptive
- The event log for a failed login should specify why it failed to make troubleshooting and reporting easier.
- Model created, updated, and deleted events should specify which kind of model (user, group, application, provider, source etc) was created, updated, and deleted.

Benricheson101:
## Details

Updates event log messages to be more descriptive

image

### What does this PR change?

Adds context to event log messages
- Model created
- Model updated
- Model deleted
- Login failed

The Model messages just add the model type to the title. "Login failed" events add a new `context.reason` field in their API response containing the reason the login failed.

It's hard to tell what's being modified or the reason for login failures from a glance. This adds it front and center so it's easier for admins to see what was changed or why a login failed without expanding each logged event.

### How was this tested?

Updating existing tests to test new API fields.

### Linked issues

closes #23193  
---

## Checklist

### DIFF
--- a/authentik/events/models.py
+++ b/authentik/events/models.py
@@ -144,9 +144,14 @@ class EventAction(models.TextChoices):
     CUSTOM_PREFIX = "custom_"
 
 
-def event_actions():
-    # Wrapper used in models to prevent migrations constantly changing when actions are added
-    return EventAction.choices
+class LoginFailedReason(models.TextChoices):
+    """All possible reasons a login failed, to be shown in event log"""
+
+    INCORRECT_PASSWORD = "incorrect_password"
+    ACCOUNT_INACTIVE = "account_inactive"
+    MFA_INVALID_OTP = "mfa_invalid_otp"
+    MFA_WEBAUTHN_FAILED = "mfa_webauthn_failed"
+    MFA_DUO_DENIED = "mfa_duo_denied"
 
 
 class Event(SerializerModel, ExpiringModel):
--- a/authentik/stages/password/stage.py
+++ b/authentik/stages/password/stage.py
@@ -16,6 +16,7 @@ from structlog.stdlib import get_logger
 
 from authentik.core.models import User
 from authentik.core.signals import login_failed
+from authentik.events.models import LoginFailedReason
 from authentik.flows.challenge import (
     Challenge,
     ChallengeResponse,
@@ -68,6 +69,7 @@ def authenticate(
         credentials=_clean_credentials(credentials),
         request=request,
         stage=stage,
+        context={"reason": LoginFailedReason.INCORRECT_PASSWORD},
     )
 
 
@@ -115,6 +117,15 @@ class PasswordChallengeResponse(ChallengeResponse):
             del auth_kwargs["password"]
             # User was found, but permission was denied (i.e. user is not active)
             self.stage.logger.debug("Denied access", **auth_kwargs)
+            login_failed.send(
+                sender=__name__,
+                credentials={"username": auth_kwargs["username"]},
+                request=self.stage.request,
+                stage=self.stage.executor.current_stage,
+                context={
+                    "reason": LoginFailedReason.ACCOUNT_INACTIVE,
+                },
+            )
             raise StageInvalidException("Denied access") from exc
         except ValidationError as exc:
             del auth_kwargs["password"]
--- a/authentik/stages/password/tests.py
+++ b/authentik/stages/password/tests.py
@@ -6,6 +6,7 @@ from django.core.exceptions import PermissionDenied
 from django.urls import reverse
 
 from authentik.core.tests.utils import create_test_admin_user, create_test_brand, create_test_flow
+from authentik.events.models import Event, EventAction, LoginFailedReason
 from authentik.flows.markers import StageMarker
 from authentik.flows.models import FlowDesignation, FlowStageBinding
 from authentik.flows.planner import PLAN_CONTEXT_PENDING_USER, FlowPlan
@@ -111,6 +112,9 @@ class TestPasswordStage(FlowTestCase):
             self.flow,
             response_errors={"password": [{"string": "Invalid password", "code": "invalid"}]},
         )
+        event = Event.objects.filter(action=EventAction.LOGIN_FAILED, user__pk=self.user.pk).first()
+        self.assertIsNotNone(event)
+        self.assertEqual(event.context["reason"], LoginFailedReason.INCORRECT_PASSWORD.value)
 
     def test_invalid_password(self):
         """Test with a valid pending user and invalid password"""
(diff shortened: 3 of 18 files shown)


rissson (approved):

authentik/events/models.py
```diff
@@ -144,9 +144,14 @@ class EventAction(models.TextChoices):
     CUSTOM_PREFIX = "custom_"
 
 
-def event_actions():
-    # Wrapper used in models to prevent migrations constantly changing when actions are added
-    return EventAction.choices
```
rissson, line 149: this shouldn't be removed

rissson:
approved in principle, the tests still need to pass

web/src/common/labels.ts
```diff
@@ -54,9 +83,27 @@ export const EventActionLabelRecord: Record<EventActions, MessageFormatter<strin
...
+
+        return msg(str`Model updated (${context.model.model_name})`);
+    },
+    [EventActions.ModelDeleted]: (context?: EventContext) => {
+        if (!isContextWithModel(context)) {
+            return msg("Model deleted");
+        }
+
+        return msg(str`Model deleted (${context.model.model_name})`);
+    },
```
kensternberg-authentik, line 106: None of these actually name the object.  They name the _class_.  The database table.  Which is a fine thing, but it it's not the whole story.  If it's not considered PII, what else would we say about the model?  It's PK, perhaps? Event contexts carry a lot of information, and while you probably don't want to put all of it on the page, having more my help.

web/src/common/labels.ts
```diff
@@ -28,9 +36,30 @@ export function formatIntentLabel(intent: IntentEnum = IntentEnum.Api): string {
...
+    EventActions,
+    MessageFormatter<string, [context?: EventContext]>
+> = {
     [EventActions.Login]: () => msg("Login"),
-    [EventActions.LoginFailed]: () => msg("Failed login"),
+    [EventActions.LoginFailed]: (context?: EventContext) => {
+        if (!isEventContextWithReason(context)) {
+            return msg("Failed login");
+        }
+        return msg(str`Failed login: (${formatEventContextReason(context.reason)})`);
```
kensternberg-authentik, line 61: I'm not entirely sure how this would get translated, or how word order would be preserved in other languages, but nesting `msg/str/msg/gettext` strings looks like a solid way to make translators' lives difficult.

authentik/stages/password/stage.py
```diff
@@ -115,6 +117,15 @@ def validate_password(self, password: str) -> str | None:
...
             self.stage.logger.debug("Denied access", **auth_kwargs)
+            login_failed.send(
+                sender=__name__,
+                credentials={"username": auth_kwargs["username"]},
+                request=self.stage.request,
+                stage=self.stage.executor.current_stage,
+                context={
+                    "reason": LoginFailedReason.ACCOUNT_INACTIVE,
+                },
+            )
```
rissson, line 128: This actually isn't correct. This will end up in two `login_failed` signals being sent, which will decrease the reputation by 2 instead of 1. We need to find another way to propagate the failure reason

rissson (changes requested):

authentik/stages/password/stage.py
rissson: @BeryJu if you have any ideas

authentik/stages/password/stage.py
```diff
@@ -68,6 +69,7 @@ def authenticate(
         credentials=_clean_credentials(credentials),
         request=request,
         stage=stage,
+        context={"reason": LoginFailedReason.INCORRECT_PASSWORD},
```
PeshekDotDev, line 72:
```suggestion
        context={"reason": LoginFailedReason.INCORRECT_PASSWORD, "reason_details": str(exc)},
```

Something like this would be a nice little touch to get a lot more information with little effort

authentik/stages/password/tests.py
```diff
@@ -111,6 +112,9 @@ def test_valid_password_inactive(self):
             self.flow,
             response_errors={"password": [{"string": "Invalid password", "code": "invalid"}]},
         )
+        event = Event.objects.filter(action=EventAction.LOGIN_FAILED, user__pk=self.user.pk).first()
+        self.assertIsNotNone(event)
+        self.assertEqual(event.context["reason"], LoginFailedReason.INCORRECT_PASSWORD.value)
```
PeshekDotDev, line 117: Since this test will "Test with a valid pending user and valid password (with an inactive user)", the "incorrect password" value should not be rendered here, and instead should be "inactive user". It looks like the flagging for inactive user above might of had some issues so maybe that's why this test  incorrectly passed?

PeshekDotDev (changes requested):

web/src/common/labels.ts
Benricheson101: A similar pattern is used in `IconTokenCopyButton.ts`.

https://github.com/goauthentik/authentik/blob/ffaf524175afe433e0663a6fe137f970a66f80d3/web/src/elements/buttons/IconTokenCopyButton.ts#L23-L28

Would it be better to have each one here be the full string instead of the injection:
```ts
[FailedLoginReason.IncorrectPassword]: () => msg("Failed login: incorrect password"),
[FailedLoginReason.AccountInactive]: () => msg("Failed login: account inactive"),
```

PeshekDotDev:
@Benricheson101 do you have time to look at some of the feedback we've left?

Benricheson101:
Apologies, this one fell off my radar. I'm working through it now.

Do you have any suggestion re https://github.com/goauthentik/authentik/pull/23670#discussion_r3683572023 ?

authentik/stages/password/stage.py
Benricheson101: Could you clarify this a bit?

There's no `exc` in scope here, and the backends don't indicate why the authentication failed, just whether they succeeded. They return `User | None`, where `None` could mean either actual authn failure or that the backend doesn't apply to the user (e.g. LDAPBackend if no dn). Getting the reason why a backend rejected the credentials would require changing the return value of each backend.

`login_failed` being raised here also only happens when all of the backends failed, not just one, so it would contain potentially several failure reasons.

authentik/stages/password/tests.py
Benricheson101: Good catch, this looks like it was probably a typo on my end that's accidentally passing.

The reason it's erroneously giving INCORRECT_PASSWORD is the same reason as the above reply.

Cross-posting from Slack:

> it looks like the rest of the review feedback is related to same issue, which is that the overloaded `authenticate` functions on the backends just return `User | None`, and don't indicate if/why they've failed, just that it wasn't successful. `None` could either mean the backend rejected the credentials, or that the backend isn't applicable to the current user and it should try the next one (the LDAP directory backend works this way).
>
> one solution that I think could work would be raising an exception instead of returning `None` if a backend fails to authenticate, that way we could tell the reason a backend failed. not sure if that would be fully in scope of this ticket or what other implications it would have.

PR #23773, opened 2026-07-06
providers/oauth2: run invalidation flow for unauthenticated end-session requests

severin:
### What does this PR change?

Currently the Oauth2 end-session endpoint requires authentication. This PR changes it to make authentication optional: if no user is authenticated then the end-session endpoint still executes the associated invalidation flow and does a redirection to the `post_logout_redirect_uri`.

Code-wise this is implemented as follows:

A new flag on the policy view is introduced: `require_authentication`. If the flag is set to `False` then the view can be accessed even if no authenticated user is present. 
The end-session view is changed to only do token revocation etc. if a authenticated user is present; if not it just executes the associated invalidation flow.

Currently when the end-session endpoint is called with no authenticated user then the endpoint prompts the user to log in. 

This is unexpected behaviour in my opinion: when I want to log out then I should be redirected to the post-logout URL and not asked to log in, even if I'm actually not logged in anymore.

**Our specific use-case**: We have a mobile app and log users in via an Authentik Oauth2 flow in a web view. The mobile app then gets access/refresh tokens and the user stays logged in via these tokens. When the app crashes or the OS reboots then the Authentik session is lost but the tokens are still valid; so we show the option to log out to the user. When a user then clicks "log out" we want to end their session (if it still exists) and have them redirected to a specific post-logout URL. 

**Note:** The OIDC spec (https://openid.net/specs/openid-connect-rpinitiated-1_0.html#ValidationAndErrorHandling) is not 100% clear on this:

> Note that because RP-Initiated Logout Requests are intended to be idempotent, it is explicitly not an error for an RP to request that a logout be performed when the OP does not consider that the End-User is logged in with the OP at the requesting RP
 
### How was this tested?

Unit tests. Hitting the end-session endpoint on a local instance with a valid session and without.

---

## Checklist

### Disclaimer

I used AI when working on this PR, but I reviewed all the changes before pushing. I didn't find any "no AI" policy in your contribution guidelines (in fact, there are AGENTS.md/CLAUDE.md files in the repo), so I assume AI use is no problem. If this is violating some rules though then please feel free to close this PR.


authentik/policies/views.py
```diff
@@ -48,6 +48,11 @@ class PolicyAccessView(AccessMixin, View):
 
     provider: Provider | None = None
     application: Application | None = None
+    # When True (default), unauthenticated requests are redirected to the authentication
+    # flow. Views that should be reachable without a session (e.g. logout/end-session,
+    # where an unauthenticated request is a no-op) can set this to False, in which case
+    # anonymous requests are dispatched directly without an access-policy check.
+    require_authentication: bool = True
```
BeryJu: Instead of adding this `require_authentication`, I think I'd prefer overriding `handle_no_permission` in the child class

BeryJu (commented):
one implementation nit otherwise LGTM, i'll defer to @PeshekDotDev for the OIDC SLO side

authentik/policies/views.py
severin: Makes sense. Adjusted!

severin:
@BeryJu @PeshekDotDev could you have a look at this again?

BeryJu (approved):
LGTM implementation wise

therour:
@BeryJu  may I know when this is planned to be released? currently this features blocks me to replace auth0 with authentik. our RP-app currently does refresh token in the backend so the RP session is longer and extendable than the authentik session

PeshekDotDev:
Hey @severin , did you run into this issue on a release build of authentik, or when running it locally?

PeshekDotDev:
There was an accidental regression with single logout that was fixed in this pr #23919 . After adding that, I don't think this PR works as intended. Let me test some things

PeshekDotDev:
Apologies, all is good with this pr

PeshekDotDev (approved):

severin:
Thanks for checking @PeshekDotDev.

How does merging PRs work? Do you do that when you see fit?

PR #23774, opened 2026-07-06
root: in-process per-IP rate throttle (#23015) (2026.2)

gergosimonyi:
#23015


rissson:
given that this is a change in behaviour, I'd be against cherry-picking it

rissson:
> given that this is a change in behaviour, I'd be against cherry-picking it

actually, since the throttling options aren't documented, I'm ok with this

BeryJu (approved):

PR #23779, opened 2026-07-06
events: migrate top_per_user to serializer for params

### LINKED ISSUE #23770: GET /api/v3/events/events/top_per_user/ returns 500 on non-numeric or negative top_n query parameter
### Describe the bug

The `top_per_user` endpoint crashes with an unhandled server error when the `top_n` query parameter is not a positive integer. Both a non-numeric value and a negative number trigger HTTP 500 responses.

1. **Non-numeric value causes unhandled ValueError** — `GET /api/v3/events/events/top_per_user/?top_n=abc` returns 500. The string `"abc"` cannot be converted to an integer, and the endpoint does not handle the conversion failure.

2. **Negative value causes unhandled AssertionError** — `GET /api/v3/events/events/top_per_user/?top_n=-5` returns 500. The negative integer parses fine but the resulting negative slice operation on the database query is not supported.

For comparison, `top_n=15` (default), `top_n=1`, and `top_n=999999999` all return 200 — the endpoint only fails on non-numeric or negative values.

### How to reproduce

**Setup:**

Authentik self-initializes on first boot. All requests use the bootstrap token. No additional setup needed — the endpoint requires only an authenticated user.

**Control — valid top_n:**

```
GET /api/v3/events/events/top_per_user/?top_n=15
Authorization: Bearer dokkimi-bootstrap-api-token
```

Returns 200 with a JSON array of user event counts. Also works with `top_n=1` and `top_n=999999999`.

**Bug 1 — non-numeric top_n:**

```
GET /api/v3/events/events/top_per_user/?top_n=abc
Authorization: Bearer dokkimi-bootstrap-api-token
```

Returns 500 Internal Server Error.

**Bug 2 — negative top_n:**

```
GET /api/v3/events/events/top_per_user/?top_n=-5
Authorization: Bearer dokkimi-bootstrap-api-token
```

Returns 500 Internal Server Error.

**Reproducing with Dokkimi:**

[Dokkimi](https://dokkimi.com) is an open source testing tool that stands up isolated Docker environments from declarative YAML definitions — services, databases, seed data, test steps, and assertions all in one file. The definition for this bug is in [dokkimi/dokkimi-in-the-wild/.dokkimi/authentik](https://github.com/dokkimi/dokkimi-in-the-wild/tree/main/.dokkimi/authentik) — `dokkimi run` reproduces it from scratch.

### Expected behavior

Non-numeric and negative `top_n` values should return 400 with a validation message, not crash with 500. The endpoint should validate the query parameter before using it.

### Screenshots

N/A

### Additional context

- Any authenticated user can trigger repeated 500 errors, polluting error-monitoring systems
- Automated dashboards or monitoring integrations that pass user-controlled values into `top_n` will crash silently if a non-numeric or negative value is introduced
- In deployments where error responses include stack traces (debug mode), the 500 could leak internal implementation details

### Deployment Method

Docker

### Version

2026.5.3

### Relevant log output

```shell
N/A
```

BeryJu:
closes https://github.com/goauthentik/authentik/issues/23770

### DIFF
--- a/authentik/events/tests/test_api.py
+++ b/authentik/events/tests/test_api.py
@@ -51,12 +51,17 @@ class TestEventsAPI(APITestCase):
     def test_top_n(self):
         """Test top_per_user"""
         event = Event.new(EventAction.AUTHORIZE_APPLICATION)
+        event.context["authorized_application"] = {"name": "foo"}
         event.save()  # We save to ensure nothing is un-saveable
         response = self.client.get(
             reverse("authentik_api:event-top-per-user"),
-            data={"filter_action": EventAction.AUTHORIZE_APPLICATION},
+            data={"action": EventAction.AUTHORIZE_APPLICATION},
         )
         self.assertEqual(response.status_code, 200)
+        self.assertJSONEqual(
+            response.content,
+            [{"application": {"name": "foo"}, "counted_events": 1, "unique_users": 0}],
+        )
 
     def test_actions(self):
         """Test actions"""
--- a/authentik/events/api/events.py
+++ b/authentik/events/api/events.py
@@ -13,9 +13,7 @@ from django.utils.text import slugify
 from django.utils.timezone import now
 from djangoql.schema import DateTimeField as QLDateTimeFIeld
 from djangoql.schema import IntField, StrField
-from drf_spectacular.types import OpenApiTypes
-from drf_spectacular.utils import OpenApiParameter, extend_schema
-from guardian.shortcuts import get_objects_for_user
+from drf_spectacular.utils import extend_schema
 from rest_framework.decorators import action
 from rest_framework.fields import (
     CharField,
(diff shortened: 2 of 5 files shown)


GirlBossRush (approved):

authentik/events/tests/test_api.py
```diff
@@ -51,12 +51,17 @@ def test_filter_model_pk_int(self):
     def test_top_n(self):
         """Test top_per_user"""
         event = Event.new(EventAction.AUTHORIZE_APPLICATION)
+        event.context["authorized_application"] = {"name": "foo"}
         event.save()  # We save to ensure nothing is un-saveable
         response = self.client.get(
             reverse("authentik_api:event-top-per-user"),
-            data={"filter_action": EventAction.AUTHORIZE_APPLICATION},
+            data={"action": EventAction.AUTHORIZE_APPLICATION},
```
rissson, line 58: we're still sending `action` here but I don't see it used in the handler

BeryJu: it's a filter on the entire eventviewset thats used by self.filter_queryset

rissson: ahhh fair enough

PR #23785, opened 2026-07-06
packages/truncator: structure-aware string truncation

GirlBossRush:
Revives the old URL truncator and generalizes it into `@goauthentik/truncator`, a small TypeScript package for structure-aware truncation: keep the readable parts of a value, ellipsize the rest.

**Core (pure, framework-free).** Per-type truncators (url, hash, uuid, mac-address, ip-address, email, user-agent, string) built on a few shared primitives: `endEllipsis`, `middleEllipsis`, `segmentEllipsis`, a `reducePipeline` runner, and `shrinkHost`. Width comes from an injectable `Measurer`; the default counts characters (pure, SSR-safe, and what the 52 tests cover), while `createCanvasMeasurer` measures pixel width for the DOM.

**Lit layer (`@goauthentik/truncator/lit`, `lit` as an optional peer dependency).** A dedicated element per type (`<ak-truncate-ip-address>`, `<ak-truncate-url>`, …) over a shared `AKTruncateBase`. Each measures its own rendered width through a shared `ResizeObserver`, reads its source from the `value` property or slotted text, and re-truncates on resize and once web fonts settle. It truncates off-screen rows too, so a table doesn't reflow as you scroll. Either the element or its `createTruncatorFC` function form works:

```ts
html`<ak-truncate-ip-address value=${ip}></ak-truncate-ip-address>`;
TruncateIPAddress(ip); // renders the element, or a "-" fallback when ip is empty
```

The core barrel stays Lit-free, so importing a truncator never pulls in Lit.

A couple of notes:

- `user-agent` parses to a summary like `Chrome 120 · Windows` rather than ellipsizing the `Mozilla/5.0 (…)` boilerplate.
- The package carries its own vitest suite (52 tests, exercised against the char-count measurer only).

First consumer is the event tables (client-IP cells in `ObjectChangelog`), with a Storybook story under `Elements / Truncate`. Design and plan notes live in `docs/superpowers/`. Thanks for reading!


BeryJu (commented):
why is this +2600 lines

PR #23827, opened 2026-07-07
internal/outpost/ak: handle empty outpost list during refresh

### LINKED ISSUE #22673: panic: nil pointer dereference in outpost/ak.OnRefresh via startIntervalUpdater and startEventHandler (possible 2026.5.0 regression)
### Describe the bug

I'm not sure if this is really a bug, but as it worked until the upgrade it might. If you need further infos, I'm happy to provide them.

When running the server and worker as separate containers (recommended deployment), the server crashes with a nil pointer dereference in the embedded outpost's `OnRefresh` method approximately 3-5 minutes after the worker starts. Without the worker running the server is stable indefinitely.

### How to reproduce

1. Deploy authentik 2026.5.0 with server and worker as separate containers/processes
2. Start the server — runs stably
3. Start the worker — server crashes within ~5 minutes

### Expected behavior

Server should not crash when the worker is running.

### Screenshots

_No response_

### Additional context

- Regression from 2026.2.x — identical setup was stable on that version
- The crash is triggered from two different code paths: `startIntervalUpdater` (periodic interval) and `startEventHandler` (WebSocket events), both hitting the same line `api.go:224`
Line 224 is `a.Outpost = outposts.Results[0]` — `outposts` is nil despite no error returned from `OutpostsInstancesList(...).Execute()`
- Root cause appears to be the new Unix socket IPC path introduced in 2026.5.0: when `akURL.Scheme == "unix"`, the URL is mutated to `http://localhost` for the API client config, but under certain timing conditions `Execute()` returns `(nil, nil)` instead of a proper error, bypassing the `err != nil` guard
- The immediate fix would be a nil-check on `outposts` before accessing `outposts.Results[0]`
Also noted: worker now binds to the same HTTP port as the server (`9000`) which is new in 2026.5.0 — previously the worker had no HTTP listener. This required separate `AUTHENTIK_LISTEN__HTTP` configs per container and may be related.

### Deployment Method

Other (please specify)
Podman quadlets, server and worker as separate containers in a shared pod, listening on localhost behind Traefik

### Version

2026.5.0

### Relevant log output

```shell
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x30 pc=0xbe43e9]

goroutine 75 [running]:
goauthentik.io/internal/outpost/ak.(*APIController).OnRefresh(0x2caeab054400)
        /go/src/goauthentik.io/internal/outpost/ak/api.go:224 +0x89
goauthentik.io/internal/outpost/ak.(*APIController).startIntervalUpdater(0x2caeab054400)
        /go/src/goauthentik.io/internal/outpost/ak/api_event.go:235 +0x1d2
goauthentik.io/internal/outpost/ak.(*APIController).StartBackgroundTasks.func3()
        /go/src/goauthentik.io/internal/outpost/ak/api.go:270 +0x57

goroutine 105 [running]:
goauthentik.io/internal/outpost/ak.(*APIController).OnRefresh(0x387093b48600)
        /go/src/goauthentik.io/internal/outpost/ak/api.go:224 +0x89
goauthentik.io/internal/outpost/ak.(*APIController).startEventHandler(0x387093b48600)
        /go/src/goauthentik.io/internal/outpost/ak/api_event.go:170 +0x7ab
goauthentik.io/internal/outpost/ak.(*APIController).StartBackgroundTasks.func1()
        /go/src/goauthentik.io/internal/outpost/ak/api.go:262 +0x57
```

rabindra789:
Fixes #22673.

### What does this PR change?

This PR adds a guard in `OnRefresh()` before accessing
`outposts.Results[0]`.

If the API returns no outposts, the function now returns an error
instead of accessing the first element.

`OnRefresh()` assumes the API always returns at least one outpost.

If the response is empty, accessing `outposts.Results[0]` can cause a
runtime panic.

Returning an error allows the refresh to fail gracefully while keeping
the existing configuration unchanged.

### How was this tested?

- Reviewed the complete `OnRefresh()` call chain.
- Verified that all callers already handle refresh errors and continue
  running.
- Compared the implementation with the existing LDAP and RADIUS
  outpost controllers to follow the same error handling pattern.
- Verified the modified package compiles.
- The full project build on Windows is currently blocked by a
  pre-existing platform-specific `syscall.SIGUSR1` build error, which is
  unrelated to this change.

### Linked issues

Closes #22673.

---

## Checklist

- [ ] The project has been linted, built, and tested (`make all`)
- [ ] The documentation has been updated and formatted (`make docs`)


rissson (approved):

rabindra789:
Sorry for the mistake i accidentally clicked the request for re-reiew.

rissson (approved):

PR #23837, opened 2026-07-07
web/maint: harmonize and make routine the display of tasks (part 1: cards)

kensternberg-authentik:
web/bug: harmonize and make routine the display of tasks (part 1: cards)

## What

This commit:

- replaces the cut-and-paste “card” version of `ak-task-list` with a function call, `taskCard()`, that fully implements the card as a drop-in replacement.

- removes the requirement that the `ModelEnum` entry be parsed for `ak-task-list`

- moves the parse and definition of the `ModelEnum` to a const outside the object.

- In some cases, the container function was updated to take the task-related object when it was known to never be `undefined`. This removes the `?` operator from a number of places.

## Why

- While working on the details list feature, I noticed that `ak-task-list` is both implemented and used with little to no consistency. This is the first of a series of PRs to improve the consistency of `ak-task-list`. This commit isolates one of three common idioms around `ak-task-list`, the “card in an overview” idiom.

- The `ModelEnum` never changes during the lifecycle of a session, and does not need to be re-allocated with every instance. Extracting it makes sense.

  - In most cases, the parse is left in-place in the existing files because `ak-schedule-list` also uses them. This has been scheduled to a future task.

- *Every* instance of `ak-task-list` (and `ak-schedule-list`) starts with the parse; this sort of repetition is a common source of bugs. Moving it to the function eliminates this source of bugs.

- Where the containing function takes a related object, these were cases where the object was known not to be `undefined`. The `?` in these functions is unnecessary, and may give future maintainers pause to wonder why it’s even there when there’s no doubt the object *is* defined.

- -19 LOC. 

## Not done

The `GoogleWorkspaceProviderViewPage.ts` is the only one of these in which the `renderOverview()` section comes _after_ the `render()` method.  I may soon have PR to enforce some more consistency on this later.

## How to test

In every case, this is a 1:1 substitution of a large block of HTML with, well, a function that produces the same large block of HTML. The test is that there’s no visual change between implementations. Because this replaces a Lit de-reference with actual function calls, the compiler and linter will correctly identify if there has been a mis-type, unwanted substitution, or last-line effect.

---

## Checklist

-   [🦋] The project has been linted, built, and tested (`make all`)
-   [🦋] The documentation has been updated and formatted (`make docs`)


web/src/admin/providers/google_workspace/GoogleWorkspaceProviderViewPage.ts
```diff
@@ -148,13 +156,8 @@ export class GoogleWorkspaceProviderViewPage extends AKElement {
...
     }
 
-    renderTabOverview(): SlottedTemplateResult {
-        if (!this.provider) {
-            return nothing;
-        }
-        const [appLabel, modelName] =
-            ModelEnum.AuthentikProvidersGoogleWorkspaceGoogleworkspaceprovider.split(".");
-        return html`${!this.provider?.assignedBackchannelApplicationName
+    renderTabOverview(provider: GoogleWorkspaceProvider): SlottedTemplateResult {
```
GirlBossRush, line 159:
```suggestion
    protected renderTabOverview(provider: GoogleWorkspaceProvider): SlottedTemplateResult {
```

web/src/admin/providers/scim/SCIMProviderViewPage.ts
```diff
@@ -48,6 +52,9 @@ import PFPage from "@patternfly/patternfly/components/Page/page.css";
 import PFGrid from "@patternfly/patternfly/layouts/Grid/grid.css";
 import PFStack from "@patternfly/patternfly/layouts/Stack/stack.css";
 
+const PROVIDER_TYPE = ModelEnum.AuthentikProvidersScimScimprovider;
```
GirlBossRush, line 55: @kensternberg-authentik I like your pattern quite a bit! Maybe something like a typed tuple helper function?

```suggestion
function pluckModelAspects(model: ModelEnum): [modelLabel: string, modelName] {
  // …
}
```

GirlBossRush (approved):
Small tweaks but LGTM 😊

PR #23840, opened 2026-07-08
web/bug: harmonize and make routine the display of tasks (part 3: problem children)

kensternberg-authentik:
web/bug: harmonize and make routine the display of tasks (part 3: problem children)

## What

This commit:

- replaces two more instances of `ak-task-list` with a function call, `taskCard()`, that fully implements the task list as a drop-in replacement.

- removes the requirement that the `ModelEnum` entry be parsed for `ak-task-list`

- moves the parse and definition of the `ModelEnum` to a const outside the object.

- Replaces two more “switch/case” expression with a, you know, an actual *expression* expression

## Why

This is the same work as [harmonize display of tasks, part 1: cards](https://github.com/goauthentik/authentik/pull/23837). In this case, these two looked like cut-and-paste errors (a grid of one cell nested inside a grid? Really?) and I decided to keep them separate to ensure that they got closer attention.

In both cases, making the substitution *did not change* the visual presentation.

- Using `tspattern` because using statements as expressions annoys me.

## How to test

See [prior branch](https://github.com/goauthentik/authentik/pull/23837) for the same test routine.

---

## Checklist

-   [🦋] The project has been linted, built, and tested (`make all`)
-   [🦋] The documentation has been updated and formatted (`make docs`)


web/src/admin/providers/oauth2/OAuth2ProviderViewPage.ts
```diff
@@ -196,9 +192,8 @@ export class OAuth2ProviderViewPage extends AKElement {
         </main>`;
     }
 
-    renderTabOverview(): SlottedTemplateResult {
-        const [appLabel, modelName] = ModelEnum.AuthentikProvidersOauth2Oauth2provider.split(".");
-        return html`${this.provider?.assignedApplicationName
+    renderTabOverview(provider: OAuth2Provider): SlottedTemplateResult {
```
kensternberg-authentik, line 199: Yep, this can only be run when the provider is concrete, so may as well lock it down.

GirlBossRush (approved):

PR #23850, opened 2026-07-08
Add configurable gunicorn worker timeout and graceful shutdown

joaocfernandes:
### What does this PR change?
Adds two configurable web worker timeouts: `AUTHENTIK_WEB__GUNICORN_TIMEOUT` (Gunicorn worker silent timeout) and `AUTHENTIK_WEB__GRACEFUL_SHUTDOWN_TIMEOUT` (Uvicorn graceful shutdown timeout). Both default to sensible values (30s / 30s) via code fallbacks and are documented under Advanced settings.

When a worker hits Gunicorn's `max_requests`, Uvicorn begins shutdown. Its graceful shutdown wait was unbounded (`timeout_graceful_shutdown=None`), so a worker could stop heartbeating long enough for the arbiter to hit `WORKER TIMEOUT` and `SIGABRT` it. Bounding Uvicorn's shutdown (and exposing the paired Gunicorn worker timeout) prevents these avoidable `SIGABRT` worker aborts.

### How was this tested?
- Verified config resolution: defaults (`30`/`30`), env overrides (`45`/`30`), and `null` → `None` (restores prior unbounded behavior).
- Confirmed Gunicorn loads the module-level `timeout` from `-c ./lifecycle/gunicorn.conf.py` (same mechanism as `max_requests`).
- `ruff check` and `ruff format --check` pass on the changed lifecycle files.

## Checklist

### DIFF
--- a/lifecycle/gunicorn.conf.py
+++ b/lifecycle/gunicorn.conf.py
@@ -45,6 +45,8 @@ preload_app = True
 max_requests = CONFIG.get_int("web.max_requests", 1000)
 max_requests_jitter = CONFIG.get_int("web.max_requests_jitter", 50)
 
+timeout = CONFIG.get_int("gunicorn.timeout", 30)
+
 logconfig_dict = get_logger_config()
 
 workers = CONFIG.get_int("web.workers", 2)
--- a/lifecycle/worker.py
+++ b/lifecycle/worker.py
@@ -2,6 +2,8 @@
 
 from uvicorn.workers import UvicornWorker
 
+from authentik.lib.config import CONFIG
+
 
 class DjangoUvicornWorker(UvicornWorker):
     """Custom configured Uvicorn Worker without lifespan"""
@@ -11,6 +13,10 @@ class DjangoUvicornWorker(UvicornWorker):
         "http": "httptools",
         "lifespan": "off",
         "ws": "wsproto",
+        "timeout_graceful_shutdown": CONFIG.get_optional_int(
+            "gunicorn.timeout_graceful_shutdown",
+            10,
+        ),
     }
 
     _worker_id: int
--- a/website/docs/install-config/configuration/configuration.mdx
+++ b/website/docs/install-config/configuration/configuration.mdx
@@ -800,6 +800,20 @@ Defaults to `120s`
 
 ## Advanced settings
 
+##### `AUTHENTIK_GUNICORN__TIMEOUT`
+
+The number of seconds a gunicorn worker may go without notifying the gunicorn master before it is considered unresponsive, killed, and restarted. Maps to gunicorn's `timeout` setting.
+
+Defaults to `30`.
+
+##### `AUTHENTIK_GUNICORN__TIMEOUT_GRACEFUL_SHUTDOWN`
+
+The maximum number of seconds the gunicorn uvicorn worker waits for in-flight connections and background tasks to finish when shutting down (for example during worker recycling triggered by `AUTHENTIK_WEB__MAX_REQUESTS`) before remaining tasks are cancelled and the worker exits. Passed to uvicorn's `timeout_graceful_shutdown` option. Set to `0` or `null` to wait indefinitely.
+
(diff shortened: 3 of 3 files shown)


website/docs/install-config/configuration/configuration.mdx
```diff
@@ -800,6 +800,20 @@ Defaults to `120s`
 
 ## Advanced settings
 
+##### `AUTHENTIK_GUNICORN__TIMEOUT`
```
rissson, line 803: Let's name this `AUTHENTIK_WEB__GUNICORN__TIMEOUT`

website/docs/install-config/configuration/configuration.mdx
```diff
@@ -800,6 +800,20 @@ Defaults to `120s`
...
+
+The number of seconds a gunicorn worker may go without notifying the gunicorn master before it is considered unresponsive, killed, and restarted. Maps to gunicorn's `timeout` setting.
+
+Defaults to `30`.
+
+##### `AUTHENTIK_GUNICORN__TIMEOUT_GRACEFUL_SHUTDOWN`
```
rissson, line 809: Let's name this `AUTHENTIK_WEB__GRACEFUL_SHUTDOWN_TIMEOUT`

lifecycle/gunicorn.conf.py
```diff
@@ -45,6 +45,8 @@
 max_requests = CONFIG.get_int("web.max_requests", 1000)
 max_requests_jitter = CONFIG.get_int("web.max_requests_jitter", 50)
 
+timeout = CONFIG.get_int("gunicorn.timeout", 30)
```
rissson, line 48: I'd also set `graceful_timeout` and `dirty_graceful_timeout` to `AUTHENTIK_WEB__GRACEFUL_SHUTDOWN_TIMEOUT`

lifecycle/worker.py
```diff
@@ -11,6 +13,10 @@ class DjangoUvicornWorker(UvicornWorker):
         "http": "httptools",
         "lifespan": "off",
         "ws": "wsproto",
+        "timeout_graceful_shutdown": CONFIG.get_optional_int(
+            "gunicorn.timeout_graceful_shutdown",
+            10,
```
rissson, line 18: I'd default this to 30 seconds, to keep the current behaviour.

Let's also update `packages/ak-common/src/arbiter.rs`, in the `do_graceful_shutdown` method on the `Arbiter` struct, the timeout will need to be the same value as here + 5 seconds (and update the comment to specify that this file needs to be updated too).

rissson (changes requested):

website/docs/install-config/configuration/configuration.mdx
joaocfernandes: Ok!

website/docs/install-config/configuration/configuration.mdx
joaocfernandes: Ok!

lifecycle/worker.py
joaocfernandes: Agree on the 30 seconds to keep the current behavior. The 10 seconds came from my experiments. Will explore the arbiter.rs to see how it fits here.

lifecycle/gunicorn.conf.py
joaocfernandes: Agree .

For my reference: 
https://gunicorn.org/reference/settings/#graceful_timeout
https://gunicorn.org/reference/settings/#dirty_graceful_timeout

website/docs/install-config/configuration/configuration.mdx
joaocfernandes: I would flatten to `AUTHENTIK_WEB__GUNICORN_TIMEOUT` as the others. So we have `web.gunicorn_timeout ` . Wdyt?

dewi-tik:
Can you please rebase this on main?

PR #23851, opened 2026-07-08
enterprise/providers/ws_fed: add SAML 1.1 support

BeryJu:
thanks microsoft

heavily claude assisted


dominic-r (approved):
lgtm docs

rissson (approved):
don't know much about saml, but the code is clean

PR #23359, opened 2026-06-24
ci: better image building

rissson:
### Linked issues

---

## Checklist

-   [ ] The project has been linted, built, and tested (`make all`)
-   [ ] The documentation has been updated and formatted (`make docs`)

PR #23853, opened 2026-07-08
tests/e2e: run against full authentik installation

rissson:
## Details

Needs https://github.com/goauthentik/authentik/pull/23359

### What does this PR change?

Run e2e tests against a real authentik installation instead of the django live test server.

### Linked issues

---

## Checklist


tests/src/lib.rs
```diff
@@ -0,0 +1,502 @@
...
+
+        stack.compose = Some({
+            let mut compose = DockerCompose::with_local_client(&[
+                concat!(env!("CARGO_MANIFEST_DIR"), "/compose/compose.authentik.yml"),
+                concat!(env!("CARGO_MANIFEST_DIR"), "/compose/compose.utils.yml"),
+                concat!(env!("CARGO_MANIFEST_DIR"), "/compose/compose.outposts.yml"),
+                concat!(env!("CARGO_MANIFEST_DIR"), "/compose/compose.proxies.yml"),
+            ])
+            .with_env("PG_PASS", "password")
+            .with_env("AUTHENTIK_SECRET_KEY", "secret_key")
```
BeryJu: might as well generate these dynamically, no?

tests/src/lib.rs
```diff
@@ -0,0 +1,502 @@
...
+#[derive(Default)]
+pub struct AuthentikStackBuilder {
+    blueprint_paths: Vec<String>,
+    selenium: bool,
+    mailpit: bool,
+    whoami: bool,
+    caddy_single: bool,
+    envoy_single: bool,
+    nginx_single: bool,
+    traefik_single: bool,
```
BeryJu: maybe an emum of the options and have a `profiles: Vec<Profile>` here?

tests/src/lib.rs
```diff
@@ -0,0 +1,502 @@
...
+            compose = compose.with_env("COMPOSE_PROFILES", stack.compose_profiles.join(","));
+
+            compose
+        });
+
+        stack.compose().up().await?;
+
+        stack.driver = if self.selenium {
+            let mut caps = DesiredCapabilities::chrome();
+            caps.set_browser_log_level(thirtyfour::LoggingPrefsLogLevel::All)?;
```
BeryJu, line 203: should probably add these from python

```
        opts.add_argument("--disable-search-engine-choice-screen")
        # This breaks selenium when running remotely...?
        # opts.set_capability("goog:loggingPrefs", {"browser": "ALL"})
        opts.add_experimental_option(
            "prefs",
            {
                "profile.password_manager_leak_detection": False,
            },
        )
```

BeryJu (commented):
some nits

tests/src/lib.rs
rissson: yes, same with akadmin password and token

tests/src/lib.rs
rissson: yeah that sounds cleaner. I'd still want to keep the `.with_selenium` (for example) interface tho

tests/src/lib.rs
rissson: agreed, will look into them, although I didn't need them for this PR

BeryJu: everything but the extension should be required, without those I had tests randomly fail

rissson: done in 2dfb7faa15318ab00b74c05d3af1b652a0776eaf

tests/src/lib.rs
rissson: done in 9399e0bf953b7eb27fdc305618aab0471bb9ee77

tests/src/lib.rs
rissson: done in b9f6702975ba0dc03c8e077ed0f02b9ba900c6ee. I actually removed the `.with_selenium` interfaces, but that's fine

.github/workflows/ci-main.yml
BeryJu, line 302: Probably remove this too since its covered below?

rissson: not all flows tests have been rewritten, so I'd say no

tests/e2e/flows_login.rs
```diff
@@ -0,0 +1,79 @@
...
+
+    stack
+        .assert_user(
+            stack.akadmin_username(),
+            "authentik Default Admin",
+            "root@example.com",
+        )
+        .await?;
+
+    stack.quit().await
```
BeryJu, line 33: I think since the stack implements Drop we maybe should just not have the explicit quit? Makes it look like manual quit is required when its not...?

tests/e2e/provider_proxy.rs
```diff
@@ -0,0 +1,333 @@
...
+        None,
+        None,
+        None,
+        None,
+        None,
+        None,
+        None,
+        None,
+        Some(stack.akadmin_username()),
+        None,
```
BeryJu, line 74: This is awful

BeryJu (commented):
Also I'd still prefer having coverage for e2e tests in containers even if its less accurate since it'll measure startup tasks too

otherwise LGTM

tests/e2e/flows_login.rs
rissson: it's faster to quit like this than to attach to the current runtime (since you're blocking one of its threads while the Drop impl runs)

tests/e2e/provider_proxy.rs
rissson: yes, gotta love generated API clients.

Maybe we hide if as a method on `AuthentikStack`?

rissson:
> Also I'd still prefer having coverage for e2e tests in containers even if its less accurate since it'll measure startup tasks too

I'll see what I can do, but it might involve an extra env variable to know whether coverage should be enabled in gunicorn or not

BeryJu:
> > Also I'd still prefer having coverage for e2e tests in containers even if its less accurate since it'll measure startup tasks too
> 
> I'll see what I can do, but it might involve an extra env variable to know whether coverage should be enabled in gunicorn or not

Or we start build actual "dev" images that ship with all the dev-deps and produce coverage I suppose

rissson:
> > > Also I'd still prefer having coverage for e2e tests in containers even if its less accurate since it'll measure startup tasks too
> > 
> > I'll see what I can do, but it might involve an extra env variable to know whether coverage should be enabled in gunicorn or not
> 
> Or we start build actual "dev" images that ship with all the dev-deps and produce coverage I suppose

Yes I'd like to do that as well, just not now

BeryJu:
> > > > Also I'd still prefer having coverage for e2e tests in containers even if its less accurate since it'll measure startup tasks too
> > > 
> > > 
> > > I'll see what I can do, but it might involve an extra env variable to know whether coverage should be enabled in gunicorn or not
> > 
> > 
> > Or we start build actual "dev" images that ship with all the dev-deps and produce coverage I suppose
> 
> Yes I'd like to do that as well, just not now

Ok fair, how about we keep the python tests for the time being and add these in addition until we have it completely ported over? they can both run in parallel

rissson:
> > > > > Also I'd still prefer having coverage for e2e tests in containers even if its less accurate since it'll measure startup tasks too
> > > > 
> > > > 
> > > > I'll see what I can do, but it might involve an extra env variable to know whether coverage should be enabled in gunicorn or not
> > > 
> > > 
> > > Or we start build actual "dev" images that ship with all the dev-deps and produce coverage I suppose
> > 
> > 
> > Yes I'd like to do that as well, just not now
> 
> Ok fair, how about we keep the python tests for the time being and add these in addition until we have it completely ported over? they can both run in parallel

sure. But I've added coverage support for the worker now, and am in the process of doing it for the server. Do I discard it or what? Also, we need to remove the proxy provider e2e tests in Python 'cause the websocket implementation there fails with https://github.com/goauthentik/authentik/pull/21856

rissson:
drafting 'till we get coverage working or debug images

PR #23899, opened 2026-07-09
lib/sync/outgoing: only run discover once per provider for outgoing syncs

ryanpesek:
## Details
Move outgoing provider discovery out of individual page tasks so remote users and groups are each discovered only once per full sync. Manual single-object syncs retain discovery behavior.
This prevents large SCIM/Entra/GWS directories from being enumerated repeatedly for every local sync page.

---

## Checklist

### DIFF
--- a/authentik/lib/sync/outgoing/tasks.py
+++ b/authentik/lib/sync/outgoing/tasks.py
@@ -39,7 +39,7 @@ class SyncTasks:
         self,
         current_task: Task,
         provider: OutgoingSyncProvider,
-        sync_objects: Actor[[str, int, int, bool], None],
+        sync_objects: Actor[[str, int, int, bool, bool], None],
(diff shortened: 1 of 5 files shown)


authentik/lib/sync/outgoing/tasks.py
```diff
@@ -101,7 +102,9 @@ def sync(
                         object_type=Group,
                     )
                 )
+                self._discover(provider, User)
```
BeryJu, line 105: I think we should either only sync here or in L189, not in both

ryanpesek: I left it at L189 to preserve a manual sync of a single object still doing a discovery first. But it doesn't really make a difference to me if that happens for manual syncs or not

PR #23901, opened 2026-07-10
core: honor expiry date for API tokens

### LINKED ISSUE #14093: Expiry date for API tokens not honored
**Describe the bug**
The expiry date input is ignored when creating or modifying an API token.

**To Reproduce**

1. Create a token, (I chose API Token)
2. Ensure "Expiring" is checked
3. Change the date or time in the "Expires on" field to a valid date in the future
4. Save
5. Observe that the "Expiry date" column shows a date 30 minutes in the future regardless of date selected earlier.

**Expected behavior**
The expires on field in the table shows the selected date/time (if valid)

**Screenshots**

**Logs**

-   authentik version: [2025.2.4]

**Additional context**

### LINKED ISSUE #19045: Bug when creating an API token
### Describe the bug

When creating an API token, it is not possible to choose a custom expiry date. The expiry date taken into account is always the one in System > Settings > Default token duration. The token is created with a replaced expiry date, without any message.
When performing the same operation to create an application password, if the expiry date is greater than the default setting, an error message appears, but it is possible to choose a shorter expiry date.

### How to reproduce

1. Go to `System > Settings` and check the `Default token duration` parameter. (30min)

**Error with a expiry date greater than the default value**
1. Go to `Directory > Tokens and App passwords`
2. Create an API token with an expiry date one day forward
The token is created with a duration of 30min. No messages

**Error with a expiry date less than the default value**
1. Go to `Directory > Tokens and App passwords`
2. Create an API token with an expiry date 5 min forward
The token is created with a duration of 30min. No messages

Repeat same operation with App passwords

### Expected behavior

- If it is a default lifetime, it should be possible to choose a longer or shorter duration.
- If the lifetime chosen for the API token is incorrect, an error message should be displayed, as is the case for the application password.
-  If it is a maximum lifetime and not a default lifetime that is set in the settings, the name sould be changed

### Screenshots

_No response_

### Additional context

_No response_

### Deployment Method

Docker

### Version

2025.10.3

### Relevant log output

```shell
❯ date -u +"%Y-%m-%dT%H:%M:%SZ" && \
curl -kL '<domain>/api/v3/core/tokens/' \
-H 'Content-Type: application/json' \
-H 'Accept: application/json' \
-H 'Authorization: Bearer <token>' \
-d '{
  "identifier": "Foo",
  "intent": "api",
  "user": 6,
  "expires": "2025-12-24T14:30:00.000Z",
  "expiring": true
}'

2025-12-24T14:25:02Z
{
   "pk":"a9edf8d3-4c06-47b4-90d8-134894f65e2a",
   "managed":null,
   "identifier":"Foo",
   "intent":"api",
   "user":6,
   "description":"",
   "expires":"2025-12-24T14:55:07.031072Z",
   "expiring":true
}
```

### LINKED ISSUE #25282: API token create/edit ignores the chosen expiry
## What

creating/editing an API token ignores the chosen expiry and always applies the system default token duration.

## Want

- Creating/editing an API token ignores the chosen expiry
- Always applies the system default token duration

## Ideas this came from

- [#14093 Expiry date for API tokens not honored](https://github.com/goauthentik/authentik/issues/14093) — The expiry date input is ignored when creating or modifying an API token. To Reproduce Steps to reproduce the behavior: 1. Create a token, (I chose API Token) 2
- [#19045 Bug when creating an API token](https://github.com/goauthentik/authentik/issues/19045) — When creating an API token, it is not possible to choose a custom expiry date. The expiry date taken into account is always the one in System Settings Default t

sebastian-mora:
## Details

When creating or modifying an Tokens with  **Expiring** enabled, the
**Expires on** date was ignored. The token's expiry was overwritten
with the default duration regardless of the value submitted. 

From a superuser's perspective, these limitations also applied generating tokens with an expiry longer than the system default in the admin UI.

It feels a bit non-intuitive to treat the system token timeline as the authoritative max expiry but, it is called out in the docs as the behavior so I opted not to change that existing behavior https://docs.goauthentik.io/users-sources/user/user_ref/#goauthentikiousertoken-maximum-lifetime

### What does this PR change?

Expiry handling in `TokenSerializer.validate` is now uniform across token
intents and honors the submitted date within the acting user's limits:

- A submitted *expires* is honored as long as it does not exceed the acting user's
  maximum token lifetime (*token_maximum_lifetime* attribute, falling back to
  the system default if not specified).
- Superusers and users with *token_expiring* set to `false` are not subject to
  the limit, so expires greater than the system default or token_maximum_lifetime are allowed.
- The check reads the acting user's own attributes rather than group
attributes to align with existing check in `perform_create`

The **Expires on** field is presented as editable in the admin UI, so a user
who sets a future expiry reasonably expects it to be honored. Instead the value
was silently discarded and set to the system default, making longer-lived expiring API tokens impossible to create

### How was this tested?

Automated tests were added in `authentik/core/tests/test_token_api.py` covering:
- superuser allowed to create expiring token > system default
- API token creation is rejected when expires exceeds the user's maximum lifetime
- Omitting expires falls back to the user attribute maximum lifetime
- Omitting expires with no user attribute falls back to the system default

### Linked issues

closes #14093
closes #19045

closes #25282
---

## Checklist

### DIFF
--- a/authentik/core/api/tokens.py
+++ b/authentik/core/api/tokens.py
@@ -62,10 +62,12 @@ class TokenSerializer(ManagedSerializer, ModelSerializer):
         if attrs.get("intent") not in [TokenIntents.INTENT_API, TokenIntents.INTENT_APP_PASSWORD]:
             raise ValidationError({"intent": f"Invalid intent {attrs.get('intent')}"})
 
-        if attrs.get("intent") == TokenIntents.INTENT_APP_PASSWORD:
-            # user IS in attrs
-            user: User = attrs.get("user")
-            max_token_lifetime = user.group_attributes(request).get(
+        actor: User = request.user if request else attrs.get("user")
+        token_expiring = actor.attributes.get(USER_ATTRIBUTE_TOKEN_EXPIRING, True)
+        expires = attrs.get("expires")
+
+        if not actor.is_superuser and token_expiring:
+            max_token_lifetime = actor.attributes.get(
                 USER_ATTRIBUTE_TOKEN_MAXIMUM_LIFETIME,
             )
             max_token_lifetime_dt = default_token_duration()
@@ -75,7 +77,7 @@ class TokenSerializer(ManagedSerializer, ModelSerializer):
                 except ValueError:
                     pass
 
-            expires = attrs.get("expires")
+            # if expires is provided, ensure it does not exceed the maximum resolved token lifetime
             if expires is not None and expires > max_token_lifetime_dt:
                 raise ValidationError(
                     {
@@ -84,9 +86,10 @@ class TokenSerializer(ManagedSerializer, ModelSerializer):
                         )
                     }
                 )
-        elif attrs.get("intent") == TokenIntents.INTENT_API:
-            # For API tokens, expires cannot be overridden
-            attrs["expires"] = default_token_duration()
+
+            # if expires is None, set it to the allowed token lifetime
+            if expires is None:
+                attrs["expires"] = max_token_lifetime_dt
 
         return attrs
 
--- a/authentik/core/tests/test_token_api.py
+++ b/authentik/core/tests/test_token_api.py
@@ -13,6 +13,7 @@ from authentik.core.models import (
     USER_ATTRIBUTE_TOKEN_MAXIMUM_LIFETIME,
     Token,
     TokenIntents,
+    default_token_duration,
 )
 from authentik.core.tests.utils import create_test_admin_user, create_test_user
 from authentik.lib.generators import generate_id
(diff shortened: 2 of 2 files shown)


abhisheksharma2411 (commented):
Honouring a user-supplied `expires` on API tokens is clearly the right direction — silently overwriting it with `default_token_duration()` meant the field was accepted and then ignored, which is worse than rejecting it. The new tests are thorough, and I particularly like `test_token_create_no_expires_with_max_lifetime` bracketing the result between a `before` and `after` rather than asserting an exact timestamp.

Two things in the refactor look like they change the security model rather than just unify the two branches, and I don't think either is intended.

**1. Group- and brand-level maximum lifetimes stop applying.**

```python
-            max_token_lifetime = user.group_attributes(request).get(
+            max_token_lifetime = actor.attributes.get(
```

`group_attributes()` is not a synonym for `attributes` — it merges three layers (`authentik/core/models.py:472`):

```python
if request and hasattr(request, "brand"):
    always_merger.merge(final_attributes, request.brand.attributes)
for group in self.all_groups().order_by("name"):
    always_merger.merge(final_attributes, group.attributes)
always_merger.merge(final_attributes, self.attributes)
```

Reading `actor.attributes` directly keeps only the last of those. So an admin who sets `goauthentik.io/user/token-maximum-lifetime` on a **group** — which is how you'd administer this for anything above a handful of users — finds the cap silently stops being enforced after this lands. Same for a brand-level default.

`USER_ATTRIBUTE_TOKEN_EXPIRING` is read the same way a line above, so a group-level `token-expiring: false` would also be dropped.

The existing tests wouldn't catch it: every one of them sets the attribute directly on the user.

**2. The cap now binds to the caller rather than the token's owner.**

```python
-            user: User = attrs.get("user")   # the token's owner
+            actor: User = request.user if request else attrs.get("user")   # the caller
```

These differ whenever an admin creates a token *for* another user — `attrs["user"]` is set explicitly and is not `request.user`. After this change the restricted user's maximum lifetime is never consulted; the admin's is. Combined with the new `not actor.is_superuser` guard, a superuser creating a token on behalf of a capped user bypasses that user's cap entirely, and `test_token_create_superuser_ignores_lifetime_limits` locks that in as intended behaviour.

That may well be the intent — "superusers aren't constrained" is a defensible rule — but it's a different rule from "this user's tokens may not outlive their cap", and it's worth being explicit about which one the project wants, because the second is the one an admin setting a group attribute thinks they're getting.

There's also an inconsistency worth tidying either way: when `request` is None the code falls back to `attrs.get("user")`, so the same validator binds the cap to the owner on one path and the caller on the other.

**A suggestion, if the group/brand behaviour was accidental:**

```python
owner: User = attrs.get("user")
actor: User = request.user if request else owner
if not actor.is_superuser and owner.group_attributes(request).get(
    USER_ATTRIBUTE_TOKEN_EXPIRING, True
):
    max_token_lifetime = owner.group_attributes(request).get(
        USER_ATTRIBUTE_TOKEN_MAXIMUM_LIFETIME,
    )
```

That keeps the superuser escape you've added while leaving the cap attached to whoever the token is for, and preserves inheritance. A test with the attribute on a group rather than the user would pin it.

**One smaller note on scope.** Flipping `test_token_create_expiring_custom_api` from `assertNotEqual` to `assertEqual` is the actual behaviour change in this PR, and it's a good one — but it's easy to miss in the diff. Worth calling out in the PR description that user-supplied expiry on `INTENT_API` tokens now takes effect where it was previously discarded, since anyone relying on the old force-override gets different tokens after upgrading.

None of this is a blocker on the idea — the feature is right, and the branch unification is a genuine simplification. It's the two attribute-resolution changes riding along with it that I'd want a maintainer to sign off on deliberately rather than inherit 🙂

abhisheksharma2411 (commented):
Re-read at `0fc64741`. Apologies for the nine-day gap — you pushed again today and I should have come back sooner.

Both points from my earlier review still stand as written, so I won't repeat the reasoning. But re-reading the whole validator rather than the diff turned up a third thing that I think is the most consequential of the three, and it is new since the refactor.

## `not actor.is_superuser` now gates the default expiry, not just the cap

```python
if not actor.is_superuser and token_expiring:
    ...
    # if expires is None, set it to the allowed token lifetime
    if expires is None:
        attrs["expires"] = max_token_lifetime_dt

return attrs
```

The `expires is None` fallback sits *inside* the guard, so for a superuser nothing assigns `expires` at all. Before this PR that path was unconditional:

```python
elif attrs.get("intent") == TokenIntents.INTENT_API:
    # For API tokens, expires cannot be overridden
    attrs["expires"] = default_token_duration()
```

There is no model default to catch it — `ExpiringModel.expires` is `DateTimeField(default=None, null=True)` — so the row persists with `expires = NULL`. And a NULL never satisfies the expiry predicates, in either direction:

```python
# ExpiringManager.get_queryset
.exclude(expires__lt=now(), expiring=True)     # NULL never matches -> never excluded
# cleanup
.filter(Q(expires__lt=now(), expiring=True))   # NULL never matches -> never deleted
```

So a superuser who creates an API token without specifying `expires` gets a **permanent credential**, where previously they got `default_token_duration()`. `intent` defaults to `INTENT_API` and superusers are who create API tokens, so this is the common path rather than a corner.

"Superusers are not subject to the maximum-lifetime cap" is a defensible rule. "Superuser tokens default to never expiring" is a different and much larger one, and I don't think the second was intended. Hoisting the default out of the guard would keep the first without the second:

```python
if expires is None:
    expires = attrs["expires"] = default_token_duration()

if not actor.is_superuser and token_expiring:
    ...cap check, and lower the default if the cap is shorter...
```

## The two earlier points, briefly

**Group and brand attributes are still dropped** — `actor.attributes.get(...)` in place of `user.group_attributes(request).get(...)`, for both `USER_ATTRIBUTE_TOKEN_MAXIMUM_LIFETIME` and `USER_ATTRIBUTE_TOKEN_EXPIRING`. `group_attributes()` merges brand, then groups, then user; reading `.attributes` keeps only the last. An admin who sets `token-maximum-lifetime` on a group finds the cap stops being enforced. The existing tests can't see it because they all set the attribute directly on the user.

**The cap still binds to the caller rather than the token's owner** — `actor = request.user if request else attrs.get("user")`. When an admin creates a token *for* a capped user, that user's cap is never consulted. Note the two paths disagree with each other: with no `request` the same validator binds to the owner instead.

None of this is an objection to the feature — honouring a user-supplied `expires` instead of silently overwriting it is plainly right, and the bracketing style in `test_token_create_no_expires_with_max_lifetime` is nice. It's the three cases above where the refactor moves the security model further than the description suggests.

Happy to look again quickly whenever — I'll be faster this time.

Issue #22817, opened 2026-06-03
Add option for user to resend email OTP


artur-motyka:
### Is your feature request related to a problem?

For now when user is asked to type otp from their email client they have no option to resend it other than logging out and logging in back again or selecting different mfa method and clicking email again (but that option is displayed only if user has multiple mfas).
In situation where we are enforcing MFA for everyone (email by default) it's just annoying for user in case where he didn't received an email.

I also let claude analyse the codebase and haven't found anything helpful for me in workarounding this issue:
```
⏺ Here's my analysis:                                                                                                         
                                                                                                                              
  Resend OTP - Feasibility Assessment                                                                                         
                                                                                                                              
  Short answer: Not doable via Terraform/configuration alone — it requires UI changes to Authentik's frontend.              
                                                                                                                              
  What the code shows:
                                                                                                                              
  Backend already supports resend — in challenge.py:137-141, submitting selected_challenge with the email device triggers   
  select_challenge_email() which calls device.generate_token() and resends. This is the existing resend mechanism.

  Frontend has NO resend button — AuthenticatorValidateStageCode.ts only renders a code input field and a "Continue" button.
  No resend UI exists.

  The "back" button is only shown when a user has 2+ MFA methods — base.ts:34-46, showBackButton is only true when
  deviceChallenges.length > 1. For typical users who only have Email OTP, this button never appears. Re-selecting the email
  challenge from the picker would technically trigger a resend, but most users never see the picker.
```

### Describe the solution you'd like

_No response_

### Describe alternatives that you've considered

None tbh

### Additional context

Authentik 2026.5.2

xyb:
Same here, I'd love to see authentik support this too.

dominic-r:
We're closing this in favor of this parent issue: #25301

PR #23911, opened 2026-07-10
stages/authenticator_email: add resend button with cooldown

xyb:
### What does this PR change?

Adds a "resend code" action to the email authenticator, on both the email setup stage and the validation stage's already-registered-device path.

The resend is rate limited so it cannot be used to flood an address:

- a 60-second cooldown enforced **server-side, keyed by email address**, using `cache.add` (Redis `SETNX`) so a burst of concurrent requests cannot race past it;
- the countdown shown in the web UI is cosmetic — a client that skips it still hits the server-side limit;
- a failed send releases the reserved window, so a delivery error does not lock the address out for the full interval.

New file `web/src/flow/stages/resend-cooldown.ts` holds the small client-side countdown helper shared by both stages.

Today the email authenticator sends a single code and offers no way to request another. If that code never arrives (spam folder, delivery delay, mistyped inbox), the user is stuck and has to abandon the flow. A resend button is the obvious fix, but a naive one turns the send endpoint into a mail bomb, so the throttle is part of the same change rather than a follow-up.

### How was this tested?

- New unit tests in `authentik/stages/authenticator_email/tests.py` (`TestAuthenticatorEmailResendThrottle`) cover the throttle: first send allowed, resend within the window refused with the remaining seconds, the countdown shrinking as the window elapses, the window reopening after `release_send`, per-address isolation, case-insensitive address matching, and the empty-address no-op. All pass against a real Django cache backend.
- `ruff check` and `ruff format --check` pass on the changed Python.

I was not able to run the full `make all` (web build/lint) in my local environment; CI will exercise those.

### Linked issues

- https://github.com/goauthentik/authentik/issues/22817

## Screenshot

---

## Checklist

### DIFF
--- a/authentik/stages/authenticator_email/stage.py
+++ b/authentik/stages/authenticator_email/stage.py
@@ -1,5 +1,8 @@
 """Email Setup stage"""
 
+from time import time
+
+from django.core.cache import cache
 from django.db.models import Q
 from django.http import HttpRequest, HttpResponse
 from django.http.request import QueryDict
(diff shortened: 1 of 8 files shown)


xyb:
@BeryJu @melizeche @GirlBossRush @kensternberg-authentik could one of you give this a review when you get a chance? It implements what #22817 is asking for.

xyb:
I attached a screenshot of this feature.

dewi-tik:
We appreciate the contribution. However, we already have an email throttling feature built-in. We'd rather re-use the existing throttle vs adding another one. Would you be interested in refactoring this PR?

PR #23918, opened 2026-07-12
website/docs: update flow-executor docs

PeshekDotDev:
### What does this PR change?

This PR adds more information to the flow executor docs to help customers implement their own UI and handle the flow via api calls

We have had customers request more information on how to handle this usecase

### How was this tested?

I manually used curl calls and tested all of the example added

### Linked issues

---

## Checklist


website/api/flow-executor.md
```diff
@@ -3,25 +3,41 @@ title: Flow executor (backend)
 sidebar_position: 2
 ---
 
-A big focus of authentik is the flows system, which allows you to combine and build complex conditional processes using stages and policies. Normally, these flows are automatically executed in the browser using authentik's [standard browser-based flow executor (/if/flows)](/docs/add-secure-apps/flows-stages/flow/executors/if-flow).
+A big focus of authentik is the flows system, which allows you to combine and build complex conditional processes using stages and policies. Normally, these flows are automatically executed in the browser using authentik's [standard browser-based flow executor (/if/flows)](../add-secure-apps/flows-stages/flow/executors/if-flow).
```
dominic-r: You must use the previous URL as it's cross-site (api --> docs)

PeshekDotDev: Ah, alright, fixed

website/api/flow-executor.md
```diff
@@ -5,23 +5,39 @@ sidebar_position: 2
 
 A big focus of authentik is the flows system, which allows you to combine and build complex conditional processes using stages and policies. Normally, these flows are automatically executed in the browser using authentik's [standard browser-based flow executor (/if/flows)](/docs/add-secure-apps/flows-stages/flow/executors/if-flow).
 
-However, any flow can be executed via an API from anywhere, in fact that is what every flow executor does. With a few requests you can execute flows from anywhere, and integrate authentik even better.
+However, any flow can be executed via an API from anywhere; in fact, that is what every flow executor does. With a few requests you can execute flows from anywhere, and integrate authentik even better — for example, to build a fully custom login UI in your own application.
```
dewi-tik:
```suggestion
A core feature of authentik is its flow system, which lets you design complex, conditional processes by combining stages and policies.

Normally, these flows are automatically executed in the browser using authentik's [standard browser-based flow executor (/if/flows)](/docs/add-secure-apps/flows-stages/flow/executors/if-flow).

However, flows are not limited to the browser. Every flow can also be executed via the API — the same mechanism used by flow executors internally. With just a few API requests, you can trigger flows from anywhere and integrate authentik more deeply into your systems. For example, you could build a fully custom login interface within your own application while still relying on authentik’s flow logic.
```

website/api/flow-executor.md
```diff
@@ -5,23 +5,39 @@ sidebar_position: 2
...
 
 :::info
```
dewi-tik:
```suggestion
:::info Cookie persistence
```

website/api/flow-executor.md
```diff
@@ -43,11 +59,15 @@ Below is the response, for example for an Identification stage.
 }
 ```
 
+Note that the stage-specific fields describe what to render (for example, which kinds of user identifiers the stage accepts) — they are not the field names of the response body. The response body layout for each stage is defined in the API schema (see [Discovering stage payloads](#discovering-stage-payloads) below).
```
dewi-tik:
```suggestion
Note that stage-specific fields describe what should be rendered — for example, which types of user identifiers a stage accepts. They do not correspond to the field names in the response body. The structure of each stage's response body is defined in the API schema; see [Discovering stage payloads](#discovering-stage-payloads) below.
```

website/api/flow-executor.md
```diff
@@ -59,14 +79,91 @@ With this body
 }
 ```
 
-Depending on the flow, you'll either get a 200 Response with another challenge, or a 302 redirect, which should be followed.
+Depending on the flow, you'll either get a 200 response with another challenge, or a 302 redirect. The redirect typically points back at the executor endpoint itself and has an empty body; follow it (preserving cookies and the `Accept: application/json` header) and the response will contain the next challenge, or the final result. Command-line HTTP clients often don't follow redirects or show empty bodies by default, so an apparently "empty" response after a POST usually means a redirect that wasn't followed.
```
dewi-tik:
```suggestion
Depending on the flow, the server will either return a 200 response with another challenge or a 302 redirect. The redirect usually points back to the executor endpoint and has an empty body. Follow this redirect (preserving cookies and the `Accept: application/json` header) to receive either the next challenge or the final result. Command-line HTTP clients often do not follow redirects or display empty response bodies by default, so an apparently "empty" response after a POST usually indicates a redirect that was not followed.
```

website/api/flow-executor.md
```diff
@@ -59,14 +79,91 @@ With this body
...
+    "component": "ak-stage-password",
+    "response_errors": {
+        "password": [{ "string": "Invalid password", "code": "invalid" }]
+    }
+}
+```
+
+## Discovering stage payloads
+
+The `component` string acts as a discriminator: each challenge component maps to a corresponding response schema in the API (for example, `ak-stage-identification` maps to `IdentificationChallengeResponseRequest`). To see the data layout for every possible stage, see the [API Browser](../reference/flows-executor-get), or fetch the OpenAPI schema of your own instance at `/api/v3/schema/`.
```
dewi-tik:
```suggestion
The `component` string acts as a discriminator: each challenge component maps to a corresponding response schema in the API. For example, `ak-stage-identification` maps to `IdentificationChallengeResponseRequest`. To inspect the payload structure for all possible stages, refer to the [API Browser](../reference/flows-executor-get) or fetch the OpenAPI schema from your own instance at `/api/v3/schema/`.
```

website/api/flow-executor.md
```diff
@@ -59,14 +79,91 @@ With this body
...
+
+```shell
+curl -sS -L -c cookies.txt -b cookies.txt \
+  'https://authentik.company/api/v3/flows/executor/default-authentication-flow/' \
+  -H 'Content-Type: application/json' -H 'Accept: application/json' \
+  -d '{"component": "ak-stage-authenticator-validate", "code": "123456"}'
+# -> challenge: xak-flow-redirect
+```
+
+If instead the user has no device and the validation stage is configured to force enrollment, the injected TOTP setup stage returns an `ak-stage-authenticator-totp` challenge containing a `config_url` — a standard `otpauth://` provisioning URI. Render this value as a QR code for the user to scan (this is exactly what authentik's own UI does client-side), then confirm with the first generated code:
```
dewi-tik:
```suggestion
If the user does not have an MFA device enrolled and the validation stage is configured to require enrollment, the injected TOTP setup stage returns an `ak-stage-authenticator-totp` challenge containing a `config_url`, which is a standard `otpauth://` provisioning URI. Render this value as a QR code for the user to scan (this is the same behavior as authentik's built-in UI), then verify enrollment using the first generated code:
```

website/api/flow-executor.md
```diff
@@ -59,14 +79,91 @@ With this body
...
+```
 
 ## Result
 
 If a stage with the component `ak-stage-access-denied` is returned, the flow has been denied.
 
-If a stage with the component `xak-flow-redirect` is returned, the flow has been executed successfully.
+If a stage with the component `xak-flow-redirect` is returned, the flow has been executed successfully. The `to` field contains the URL the user should be redirected to.
+
+:::info
```
dewi-tik:
```suggestion
:::info Resulting authenticated session
```

Not sure about this title, but it needs something

dewi-tik (changes requested):

website/api/flow-executor.md
PeshekDotDev: Added manually

dewi-tik (approved):

PR #23969, opened 2026-07-13
website/docs: cleanup 07-12: refresh contributor guidance

dominic-r:


website/docs/developer-docs/hackathon/index.md
```diff
@@ -1,63 +1,27 @@
 ---
-title: Hackathon 2023
+title: 2023 hackathon archive
```
dewi-tik, line 2: We don't need to update this file

dewi-tik (approved):
I think that updating the hackathon file isn't necessary. But it doesn't matter much either way

PR #23987, opened 2026-07-13
core: Serve interface shell for any if/admin and if/user subpath.

GirlBossRush:
### What does this PR change?

Adds anchored catch-all routes in `authentik/core/urls.py` so any subpath under `/if/admin/` and `/if/user/` renders the same interface template as the exact prefix. Deep links like `/if/admin/identity/users/42` currently 404; with this change they load the SPA shell with identical access control and injected context (`BrandDefaultRedirectView` is reused).

Each catch-all registers after its exact sibling, so `reverse("authentik_core:if-admin")`, flow, websocket, and API routes resolve exactly as before. Guard tests cover the ordering and non-shadowing cases.

Server-side prerequisite for moving the web UI from hash-based routing (`/if/admin/#/core/applications`) to path-based routing (`/if/admin/core/applications`). The router itself lands in separate web PRs; until the interfaces start emitting path URLs, this PR has no user-visible effect.

### How was this tested?

New `TestInterfaceCatchAll` suite (13 tests): exact prefixes, deep subpaths, reversibility, route ordering, non-shadowing of flow/ws/API routes, and `web.path` propagation into the shell context. `uv run pytest authentik/core/tests/test_interface_views.py` passes; `ruff check` and `ruff format --check` are clean.

### Linked issues

---

## Checklist


BeryJu (commented):
looks good but for the time being, as long as we aren't switching our frontend router to use this, i think i'd like this to redirect from `/if/admin/foo/bar` to `/if/admin/#/foo/bar`..?

authentik/core/urls.py
```diff
@@ -50,11 +50,21 @@
         BrandDefaultRedirectView.as_view(template_name="if/admin.html"),
         name="if-admin",
     ),
+    re_path(
+        r"^if/admin/(?P<path>.*)$",
```
BeryJu:
```suggestion
    path(
        "if/admin/<path:path>,
```

BeryJu (approved):

PR #24008, opened 2026-07-14
lifecycle/container: drop curl and runit

emilburzo:
Alternative to https://github.com/goauthentik/authentik/pull/23953

### Differences:

- it drops `runit` (by changing from `chpst` to `setpriv`), which no longer pulls in a lot of perl dependencies
- it leaves `perl-base` installed since it's marked as an "essential" package in debian
- since we are not breaking dpkg, we can just use `apt-get purge` to remove `curl`

The tradeoff is that we accept some `perl-base` CVEs (3 critical, 6 high), but leave dpkg/apt in a working state.

### Numbers

Before:

```
 ✔ Scanned for vulnerabilities     [276 vulnerability matches]
   ├── by severity: 26 critical, 54 high, 60 medium, 5 low, 95 negligible (36 unknown)
   └── by status:   5 fixed, 271 not-fixed, 0 ignored
```

After:

```
 ✔ Scanned for vulnerabilities     [209 vulnerability matches]
   ├── by severity: 5 critical, 24 high, 52 medium, 5 low, 84 negligible (39 unknown)
   └── by status:   5 fixed, 204 not-fixed, 0 ignored
```

### DIFF
--- a/lifecycle/ak
+++ b/lifecycle/ak
@@ -55,7 +55,6 @@ function check_if_root_and_run {
         return
     fi
     SOCKET="/var/run/docker.sock"
-    GROUP="authentik"
     if [[ -e "$SOCKET" ]]; then
         # Get group ID of the docker socket, so we can create a matching group and
         # add ourselves to it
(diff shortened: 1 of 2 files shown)


lifecycle/ak
```diff
@@ -55,23 +55,19 @@ function check_if_root_and_run {
...
-        # since the name of the group might not be docker, we need to lookup the group id
-        GROUP_NAME=$(getent group "${DOCKER_GID}" | sed 's/:/\n/g' | head -1)
-        GROUP="authentik:${GROUP_NAME}"
     fi
     # Fix permissions of certs and media
     chown -R authentik:authentik /data /certs "${PROMETHEUS_MULTIPROC_DIR}"
     chmod ug+rwx /data
     chmod ug+rx /certs
-    exec chpst -u authentik:"${GROUP}" env HOME=/authentik $(run_authentik "$@")
+    exec setpriv --reuid authentik --regid authentik --init-groups -- env HOME=/authentik $(run_authentik "$@")
```
emilburzo, line 70: `--init-groups` replaces the need for the group discovery above (now removed), since it reads `/etc/groups` directly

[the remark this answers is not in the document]
emilburzo: it's in `util-linux`, marked as an essential package. this is probably overkill

BeryJu: neat

BeryJu (approved):

rissson:
We've previously added curl specifically because it's so useful for debugging. I don't really want to get rid of it.

thmo:
Removing `curl` in a patch release is, from a container consumer's pov, unexpected.
It broke healthchecks in my setup (polling `/-/health/ready/` for both, server and worker) and thus prevented proper container start.

SCBionicle:
I agree. This should've been listed as breaking changes and possibly pushed to the minor release schedule rather being pushed as a patch.

emilburzo:
@thmo @SCBionicle sorry for the breakage, I was too eager to see the security improvements make it in the next patch release and I didn't consider people might be relying on `curl` as opposed to external health checks. lesson learned for the future

PR #24014, opened 2026-07-14
crypto: Add key type safeguards

PeshekDotDev:
### What does this PR change?

Although there was some safeguarding on the front-end, this PR adds back-end safeguards as well as more front-end safeguards for selecting unsupported certificate types

Many certificates are limited on their use based on the libraries that we are using. Go has decided that they will not support ed448 for TLS, so we would need to use something besides the crypto/tls library. Our xml library does not support ed certs, so we would need to change that. Just want to log that we could support ed certificates in more locations, but there isn't much of a reason to due to PQ certs coming

This helps prepare for when we add post quantum certificates, preventing people from selecting them as an option before they are supported, even though we will allow for generating, uploading, and storing them

Preferably PR #24013 is merged before this so that the filtering for ed448 is accurate

### How was this tested?

Generating rsa, ecdsa, ed448, ed25519, and trying to select on the front-end (unusable certs properly filtered)

Generating certs and attempting to post to back-end (properly prevented)

### Linked issues

---

## Checklist


authentik/crypto/validators.py
```diff
@@ -0,0 +1,53 @@
...
+
+# Key types that have an XML-DSIG signature transform in libxmlsec1.
+XML_SIGNING_KEY_TYPES = [KeyType.RSA, KeyType.EC, KeyType.DSA]
+
+# Key types the Go outposts can serve TLS with. Ed448 will never be supported
+# See https://github.com/golang/go/issues/29390#issuecomment-614175576
+TLS_KEY_TYPES = [KeyType.RSA, KeyType.EC, KeyType.ED25519]
+
+
+def validate_key_type(keypair: CertificateKeyPair | None, allowed: list[KeyType]) -> None:
```
BeryJu: Instead of this function this should be a django validator class

authentik/providers/ldap/api.py
```diff
@@ -30,6 +32,10 @@ class LDAPProviderSerializer(ProviderSerializer):
 
     outpost_set = ListField(child=CharField(), read_only=True, source="outpost_set.all")
 
+    def validate_certificate(self, keypair: CertificateKeyPair) -> CertificateKeyPair:
+        validate_key_type(keypair, TLS_KEY_TYPES)
+        return keypair
+
```
BeryJu: instead of calling the function it should be set as a validator

website/docs/sys-mgmt/certificates.md
```diff
@@ -20,6 +20,20 @@ While this certificate can be used for SAML providers/sources, remember that it'
 
 For SAML use-cases, you can generate a certificate with a longer validity period (at your own risk).
 
+## Supported key algorithms
+
+Not every key algorithm can be used for every purpose. The algorithm is detected from the certificate when it is generated or imported, and authentik only offers a certificate where its key algorithm is usable.
```
dewi-tik:
```suggestion
Not every key algorithm is suitable for every use case. When a certificate is generated or imported, authentik determines its key algorithm and will only present that certificate for operations where its algorithm is supported.
```

dewi-tik (approved):

web/src/admin/common/certificate-key-types.ts
```diff
@@ -0,0 +1,26 @@
...
+/**
+ * Key algorithms accepted for each certificate purpose.
+ *
+ * These mirror the allowlists enforced server-side in `authentik/crypto/validators.py`. They are
+ * duplicated here only so the certificate pickers can filter out keypairs that would be rejected;
+ * the server remains the source of truth.
+ */
+
+/** Signing a JWT. Bounded by the algorithms `JWTAlgorithms.from_private_key` can map to. */
+export const JWTSigningKeyTypes = [
```
kensternberg-authentik, line 12: This information is also encoded in `./authentik/crypto/validators.py`, creating a "two sources of truth" problem.  Could `JWTSigningKeyTypes` be made available via the API, so that we have a single source of truth?

GirlBossRush (approved):

web/src/admin/common/certificate-key-types.ts
```diff
@@ -0,0 +1,26 @@
...
+    KeyTypeEnum.Rsa,
+    KeyTypeEnum.Ec,
+    KeyTypeEnum.Ed25519,
+    KeyTypeEnum.Ed448,
+];
```
kensternberg-authentik, line 17: Does this change keys that the user could have previously selected?  Like, does it broaden the list of filtered keys? (I'm thinking, if someone had a provider with a weird key that's now not in the "allowed" list, they won't be able to edit that provider because now their chosen key won't show up in the drop-down field; it's been filtered out.)

web/src/admin/common/certificate-key-types.ts
```diff
@@ -0,0 +1,26 @@
...
+
+/** Encrypting a JWT. `RSA-OAEP-256` is hardcoded, and has no non-RSA counterpart. */
+export const JWEEncryptionKeyTypes = [KeyTypeEnum.Rsa];
```
kensternberg-authentik, line 20: So, the thing that bugs me here is that, now that we've got these distinctions, you've created _mysteries_.  The user might upload an Ed448 key, then go look for it in OAuth2's **Advanced Options** section and, mysteriously, it isn't there, because we _silently_ filter it out.  This patch creates a situation where user-provided materials can just not show up where an naive user might expect them to, and without explanation as to why.

web/src/admin/providers/oauth2/OAuth2ProviderFormForm.ts
```diff
@@ -324,6 +326,7 @@ export function renderForm({
                         label=${msg("Signing Key")}
                         placeholder=${msg("Select a signing key...")}
                         certificate=${ifPresent(provider.signingKey)}
+                        .allowedKeyTypes=${JWTSigningKeyTypes}
```
kensternberg-authentik, line 329: Lotta shotgun surgery going on here.  Feels like there's a missing idiom.  Not sure how to deal with it, but it bugs me.

kensternberg-authentik (approved):
I have misgivings: 

- there's a big UX problem is hiding stuff the user might generate or upload without telling them _why_
- there's duplication-of-knowledge across the client and server, creating a maintenance headache
- there's something subtly wrong with how the filters are handled in the various drop-downs, but I can't put my finger on it. It's a DX problem, not a UX or functional one.

The "uploaded a weird key and now it's missing from an existing provider's form" thing is trying to compensate for a very silly user error; it's not even worth addressing, really, but it's just something that occurred to me.

The underlying functionality is solid. It does what it says. We can address these issues in a future PR, if we think we should. None of them block merging.

authentik/crypto/validators.py
PeshekDotDev: Fixed

web/src/admin/common/certificate-key-types.ts
PeshekDotDev: This should be fixed when PR #24770 gets merged

web/src/admin/common/certificate-key-types.ts
PeshekDotDev: It could be possible that before someone selected a wrong cert, and would have been able to save it. but they most likely would have gotten a runtime error using one of the invalid certs. so this prevents them from ever selecting it. If they had selected a wrong cert before, they'd just select a new one now. Authentik self-signed cert always shows as a valid option because its a correct key type and has a private key, although it shouldn't be selected for something like encrypting messages to an SP. but it could be used to clear weird state and would be unlikely to be a problem in production

authentik/providers/ldap/api.py
PeshekDotDev: Fixed

authentik/providers/oauth2/tests/test_api.py
```diff
@@ -34,6 +80,59 @@ def setUp(self) -> None:
         self.user = create_test_admin_user()
         self.client.force_login(self.user)
 
+    def set_signing_key(self, cert: CertificateKeyPair):
```
BeryJu, line 85: I know the fixture cert is valid for a long time, but the more correct way would be `from freezegun import freeze_time` `@freeze_time("2022-10-14T14:15:00Z")`

web/src/admin/common/certificate-key-types.ts
BeryJu: I dont think easily

BeryJu (approved):
lgtm backend wise with 1 nit

authentik/providers/oauth2/tests/test_api.py
PeshekDotDev: fixed :D

PR #24024, opened 2026-07-14
web, locale: Normalize legacy locale file names

GirlBossRush:
## Details

`web/xliff/` carried two naming conventions at once. The current lit-localize pipeline writes hyphenated BCP-47 filenames (`de-DE.xlf`, `pt-BR.xlf`), while an older Transifex config wrote underscores (`de_DE.xlf`, `pt_BR.xlf`). That Transifex config was fixed months ago, so the underscore files are just leftovers. Each one is either an exact duplicate of a file that already exists in the correct form, or a one-off translation someone contributed for a locale that was never enabled.

This clears them out:

- Eleven underscore files that already had a hyphenated counterpart are removed (`cs_CZ`, `de_DE`, `es_ES`, `fi_FI`, `fr_FR`, `it_IT`, `ja_JP`, `ko_KR`, `pt_BR`, `ru_RU`, `tr_TR`). lit-localize reads the hyphenated versions; the underscore copies were dead weight.
- Three had no hyphenated counterpart, so they were orphaned contributions for locales that were never turned on. Those get renamed to their BCP-47 form and enabled in `lit-localize.json`: `bg_BG` becomes `bg-BG`, `no_NO` becomes `nb-NO` (the file is Bokmål, so `nb` is the right language subtag), and `ar_AA` becomes `ar` (the `AA` region is a private-use placeholder, so this keeps the plain language code). `locale-codes.ts` and the locale definitions regenerate to match.

The Django gettext catalogs under `locale/` keep their underscore convention, but two directories carried language codes that no longer line up with the web tags. `locale/ar_AA` becomes `locale/ar` and `locale/no_NO` becomes `locale/nb_NO`, mirroring the `ar` and `nb-NO` decisions above (Django uses `nb` for Bokmål, and neither `ar_AA` nor `no_NO` is a code Django recognizes, so these load where they previously might not have). The `.po` `Language` headers move with them. Every other directory, including `bg_BG` and the Chinese catalogs, is left alone.

Duplicate files drift apart. A translator updates `de-DE.xlf`, the stale `de_DE.xlf` sits behind it, and it's not obvious which one is live. Clearing them leaves one file per locale. Folding in the three orphaned contributions also means translation work people already did actually ships, instead of sitting unreferenced in the tree.

### How was this tested?

`build-locales` regenerates cleanly and precommit passes; the source locale (`en`) was not renamed or removed. On the Django side, `ar` and `nb_NO` compile and load after the rename, and nothing in authentik's config referenced the old `ar_AA` / `no_NO` codes.


kensternberg-authentik (approved):
Straightforward and mechanical.

PR #24031, opened 2026-07-14
website/docs: Remove Public Release Docs.

dominic-r:
Better Tracked in Notion.

cc @PeshekDotDev 
cc @dewi-tik 

### Linked issues

---

## Checklist


PeshekDotDev:
@dewi-tik curious what you think. but the public docs release process page feels weird when we have strong internal steps. and it kind of begs the question as to who it's for. I guess we could make all the internal docs external. but it creates multiple sources of truth

dominic-r:
I'm kind of against making the internal docs public. As you said, " it kind of begs the question as to who it's for"

dominic-r:
Plus most of our internal workflows, planning, and such are in Notion, so it's much easier to just keep it there.

dewi-tik (approved):

PR #24032, opened 2026-07-14
web: Add the path-router outlet.

GirlBossRush:
## Details

This PR adds `<ak-router-view>`, the outlet that renders the new path-based route table, plus the two core hooks it needs. It strips the interface prefix from `location.pathname`, matches through the router core, intercepts in-interface clicks, re-matches on popstate, redirects `/` to a configurable `default-path`, falls back to `<ak-router-404>` with the URL preserved, and owns the loading and error states, so a rejecting route renders an error state instead of escaping as an unhandled rejection. Sentry pageload and navigation spans mirror the legacy hash outlet.

Two supporting changes come with it: `Route.resolve()` exposes a route's render callback as a promise so the outlet can own the async states, and `navigate()` falls back to a full-page assign for cross-origin destinations instead of letting `pushState` throw.

Nothing mounts the element yet. The user-interface flip lands on top of this.

### DIFF
--- a/web/src/elements/router/core/RouterView.ts
+++ b/web/src/elements/router/core/RouterView.ts
@@ -1,0 +1,246 @@
+/**
+ * @file Path-router outlet for the new route table.
+ *
+ * Renders the route matched from `location.pathname` (with the interface
+ * `prefix` stripped per the matcher's leading-slash contract), owns the
+ * loading and error states, and claims in-interface anchor clicks. Ships
+ * inert: nothing imports it until the interface flip in Plan 3b.
+ *
+ * App-context-free: imports only the router core, the reused 404/empty-state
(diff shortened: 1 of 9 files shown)


web/src/elements/router/core/RouterView.ts
```diff
@@ -0,0 +1,246 @@
...
+
+    //#region Sentry
+
+    #sentryClient = getClient();
+    #pageLoadSpan: Span | null = null;
+
+    constructor() {
+        super();
+
+        if (process.env.NODE_ENV !== "production" && this.#sentryClient) {
```
BeryJu: imo this should be a helper function
```
enableSentry() {
  if NODE_ENV !== "production" && not queryArg("disable-sentry")
    return true
  return server-side enable
}
```

kensternberg-authentik (approved):

PR #24038, opened 2026-07-15
web/maint: upgrade ak-toggle-group

kensternberg-authentik:
web/maint: upgrade ak-toggle-group

## What

This commit upgrades ak-toggle-group with the following improvements:

- Moves ToggleGroup to *Elements*, since it is not API-aware
- Fixes the margins between the buttons so that they are now correct
- Correctly implements `compact` mode
- Updates ak-toggle-group to monitor the options collection and update itself when they change.
- Correctly sets the `aria-pressed` and `disabled` attributes as needed.
- Can correctly handle multi-select
- Can correctly interpret when an option is marked `disabled`
- Uses Patternfly 5 settings (which are almost unchanged)
- Uses a vendored, owned copy of its own CSS, local to the implementation
- Updates `value` correctly, but no longer changes the attribute: that’s an anti-idiom

This is still a *controlled* component: the parent element is responsible for updating its value when it acknowledges the change.

I have updated all components that use ak-toggle-group to point to the correct implementation.

I have updated the Storybook to make `compact` mode visible.

## Why

This is the better way to move to P5: Bring the elements *we already have* up-to-date with vendored solutions. Having the `:root` and `:host` in a global-vs-local solution solves so many of the conundrums we worried about.

It also fixes that really, really ugly margin bug that was driving me out of my mind.

## Note

Yes, I’m trying out a new layout for the implementation folder.

I actually tend to prefer @beryju’s “ToggleGroup” naming scheme over the `ak-whatever` variant I was using earlier, but I also like splitting up the view into “The element’s most common use case” with the implementation, which can be imported, inherited and extended into new variants *without* triggering the registry, in a related folder. @GirlBossRush, if you think the `_impl` tag is ugly, let me know and we’ll work something better out.

---

## Checklist

-   [🎢] The project has been linted, built, and tested (`make all`)
-   [🎢] The documentation has been updated and formatted (`make docs`)


kensternberg-authentik (commented):
I _technically_ didn't touch any of the files in `admin`, Perl did, so... I didn't add `@file` comments there. Sorry.  :-)

web/src/elements/ToggleGroup_impl/ToggleGroup.stories.ts
```diff
@@ -65,3 +59,17 @@ export const ToggleGroup = () => {
         </ak-toggle-group>`,
     );
 };
+
+export const Compact = () => {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const displayChange = (ev: any) => {
```
GirlBossRush: Custom event?

```ts
export class AKToggleGroupEvent extends Event {
    public static readonly eventName = "ak-toggle-group";
    public value: string;

    constructor(value: string) {
        super(AKToggleGroupEvent.eventName, { bubbles: true, composed: true });
        this.value = value;
    }
}

declare global {
    interface WindowEventMap {
        [AKToggleGroupEvent.eventName]: AKToggleGroupEvent;
    }
}
```

web/src/elements/ToggleGroup_impl/ToggleGroup.styles.ts
```diff
@@ -0,0 +1,152 @@
...
+        background-color: var(--pf-c-toggle-group__button--BackgroundColor);
+        border: 0;
+    }
+
+    [part~="button"]::before {
+        position: absolute;
+        inset-block-start: 0;
+        inset-block-end: 0;
+        inset-inline-start: 0;
+        inset-inline-end: 0;
```
GirlBossRush, line 75:
```suggestion
        inset: 0;
```

web/src/elements/ToggleGroup_impl/ToggleGroup.ts
```diff
@@ -0,0 +1,110 @@
...
+                { disabled: option.hasAttribute("disabled") },
+            ],
+        );
+    }
+
+    get values(): Set<string> {
+        return new Set(this.multi ? this.value.split(this.separator) : [this.value]);
+    }
+
+    render() {
```
GirlBossRush, line 90:
```suggestion
    protected override render() {
```

web/src/elements/ToggleGroup_impl/ToggleGroup.ts
```diff
@@ -0,0 +1,110 @@
...
+ */
+
+import Styles from "./ToggleGroup.styles";
+
+import { CustomEmitterElement } from "#elements/utils/eventEmitter";
+
+import { html, LitElement } from "lit";
+import { property } from "lit/decorators.js";
+
+type Option = [string, string, { disabled?: boolean }];
```
GirlBossRush, line 10:
```suggestion
type Option = [key: string, label: string, state: { disabled?: boolean }];
```

web/src/elements/ToggleGroup_impl/ToggleGroup.ts
```diff
@@ -0,0 +1,110 @@
...
+ *
+ * @element ak-toggle-group
+ *
+ * @fires ak-toggle - Fired when someone clicks on a toggle option. Carries the value of the option.
+ */
+
+// MYNIS:
+// A 'name' property so that the event carries *which* toggle group emitted the event.
+
+export class ToggleGroup extends CustomEmitterElement(LitElement) {
```
GirlBossRush: We've got about 9 instance of `CustomEmitterElement` remaining and they've steadily dwindled with the introduction of extending the `Event` object. IMO we can knock out another here :) 

```suggestion
export class ToggleGroup extends LitElement {
```

web/src/elements/ToggleGroup_impl/ToggleGroup.ts
```diff
@@ -0,0 +1,110 @@
...
+     */
+
+    @property({ type: String })
+    value = "";
+
+    @property({ type: Boolean, reflect: true })
+    multi = false;
+
+    @property({ type: String })
+    separator = ",";
```
GirlBossRush, line 48:
```suggestion
    @property({ type: String })
    public value = "";

    @property({ type: Boolean, reflect: true })
    public multi = false;

    @property({ type: String })
    public separator = ",";
```

web/src/elements/ToggleGroup_impl/ToggleGroup.ts
```diff
@@ -0,0 +1,110 @@
...
+        this.optionsObserver.observe(this, {
+            subtree: true,
+            childList: true,
+            attributes: true,
+            attributeFilter: ["value", "disabled"],
+            characterData: true, // Watch for label changes, too.
+        });
+    }
+
+    public disconnectedCallback() {
```
GirlBossRush, line 66:
```suggestion
    public override disconnectedCallback(): void {
```

web/src/elements/ToggleGroup_impl/ToggleGroup.ts
```diff
@@ -0,0 +1,110 @@
...
+        super.disconnectedCallback();
+        this.optionsObserver?.disconnect();
+        this.optionsObserver = null;
+    }
+
+    get rawOptions(): HTMLOptionElement[] {
+        return Array.from(this.querySelectorAll("option") ?? []);
+    }
```
GirlBossRush, line 74:
```suggestion
    public get rawOptions(): readonly HTMLOptionElement[] {
        return Array.from(this.querySelectorAll("option"));
    }
```

web/src/elements/ToggleGroup_impl/ToggleGroup.ts
```diff
@@ -0,0 +1,110 @@
...
+
+    get options(): Option[] {
+        return Array.from(this.rawOptions).map(
+            (option: HTMLOptionElement): Option => [
+                option.getAttribute("value") ?? "",
+                option.textContent ?? "",
+                { disabled: option.hasAttribute("disabled") },
+            ],
+        );
+    }
```
GirlBossRush, line 84:
```suggestion
    public get options(): readonly Option[] {
        return Array.from(this.rawOptions, (option: HTMLOptionElement): Option => [
                option.getAttribute("value") ?? "",
                option.textContent ?? "",
                { disabled: option.hasAttribute("disabled") },
            ],
        );
    }
```

web/src/elements/ToggleGroup_impl/ToggleGroup.ts
```diff
@@ -0,0 +1,110 @@
...
+
+    render() {
+        const values = this.values;
+        const mkClick = (v: string) => () => {
+            this.dispatchCustomEvent("ak-toggle", { value: v });
+        };
+
+        return html` <div role=${this.multi ? "group" : "radiogroup"} part="toggle-group">
+            ${this.options.map(
```
GirlBossRush, line 97: potential use for Lit's `map` iterator?

web/src/elements/ToggleGroup_impl/ToggleGroup.styles.ts
```diff
@@ -0,0 +1,152 @@
+/**
+ * @file ShadowDOM CSS for the ToggleGroup component
+ */
+
+import { css } from "lit";
+
+export const styles = css`
```
GirlBossRush, line 7: I'm hesitant to re-introduce css-in-js files as a pattern. lit-analyzer seems to be withering on the vine and doesn't play nice with some newer properties. And if when we do the switch to oxlint/fmt I'd like to take advantage of the synergy that comes with stylelint too -- one of your PF5 goodies I really liked

GirlBossRush (approved):
A welcomed addition to our pantheon of elements. 

Approving as my suggestions aren't blocking. 🥂

PR #24074, opened 2026-07-15
packages/django-dramatiq-postgres/broker: use positive state filter for pending messages

EasyAsABC123:
## Details
- Change the Postgres task consumer pending-message lookup from a negative terminal-state exclusion to a positive consumable-state filter.
- Let PostgreSQL use the existing `(queue_name, state, eta)` index for queue polling.
- Add a regression test that asserts the consumer applies the positive state filter.

### What does this PR change?
Changes query to be positive and thereby near instant

Seek over all rows in tasks, including completed tasks is slow.

### How was this tested?

- `git diff --check`
- Built and published a container image from this branch.
- Deployed the image to a Kubernetes Authentik worker/server pair.
- Verified the running worker contains `.filter(state__in=CONSUMABLE_TASK_STATES)`.
- Verified PostgreSQL plans the positive-state query with `authentik_t_queue_n_099d0d_idx`.

On a production-like task table with about 409k rows, mostly completed tasks, the old query shape planned as a parallel sequential scan:

```text
Parallel Seq Scan on authentik_tasks_task
cost ~= 253678
```

The new query shape uses the existing index:

```text
Index Scan using authentik_t_queue_n_099d0d_idx on authentik_tasks_task
cost ~= 346
```

### Linked issues

None opened

---

## Checklist

### DIFF
--- a/packages/django-dramatiq-postgres/django_dramatiq_postgres/broker.py
+++ b/packages/django-dramatiq-postgres/django_dramatiq_postgres/broker.py
@@ -48,6 +48,12 @@ DATABASE_ERRORS = (
     OperationalError,
 )
 
+CONSUMABLE_TASK_STATES: set[TaskState] = set(TaskState) - {
+    TaskState.DONE,
+    TaskState.REJECTED,
+    TaskState.WAITING_FOR_DEPENDENCIES,
+}
+
 
 def channel_name(queue_name: str, identifier: ChannelIdentifier) -> str:
     return f"{CHANNEL_PREFIX}.{queue_name}.{identifier.value}"
@@ -375,13 +381,7 @@ class _PostgresConsumer(Consumer):
         pending = set(
             self.query_set.exclude(message_id__in=self.in_processing)
             .filter(queue_name=self.queue_name)
-            .exclude(
-                state__in=(
-                    TaskState.DONE,
-                    TaskState.REJECTED,
-                    TaskState.WAITING_FOR_DEPENDENCIES,
-                )
-            )
+            .filter(state__in=CONSUMABLE_TASK_STATES)
             .exclude(eta__gte=timezone.now() + timedelta(seconds=self.timeout))
             .order_by(F("eta").asc(nulls_first=True))
             .values_list("message_id", flat=True)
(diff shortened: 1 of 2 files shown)


packages/django-dramatiq-postgres/django_dramatiq_postgres/broker.py
```diff
@@ -48,6 +48,14 @@
     OperationalError,
 )
 
+CONSUMABLE_TASK_STATES: tuple[TaskState, ...] = (
```
rissson: Could we instead do `set(TaskState) - {TaskState.DONE, TaskState.REJECTED, TaskState.WAITING_FOR_DEPENDENCIES}`, please?

packages/django-dramatiq-postgres/django_dramatiq_postgres/broker.py
```diff
@@ -375,13 +383,7 @@ def _fetch_pending_messages(self) -> set[str]:
...
             self.query_set.exclude(message_id__in=self.in_processing)
             .filter(queue_name=self.queue_name)
-            .exclude(
-                state__in=(
-                    TaskState.DONE,
-                    TaskState.REJECTED,
-                    TaskState.WAITING_FOR_DEPENDENCIES,
-                )
-            )
+            .filter(state__in=CONSUMABLE_TASK_STATES)
```
rissson, line 384: The ~same filter is used a bit later in `_consume_one`, but I suppose it's less of an issue there since we also filter on message_id.

rissson (changes requested):

packages/django-dramatiq-postgres/django_dramatiq_postgres/broker.py
EasyAsABC123: This completely makes sense

packages/django-dramatiq-postgres/django_dramatiq_postgres/broker.py
EasyAsABC123: We can reuse here, but not sure it matters as much for the performance more for DRY ;)

EasyAsABC123:
@rissson all resolved, ready for another pass, when you have time

rissson (approved):

rissson:
@EasyAsABC123 you do not need to keep rebasing on main. When you do we need to re-run the CI every time for no good reason, and we need to re-check your changes. Thanks.

EasyAsABC123:
👍heard

rissson:
Thank you for the contribution!

PR #24080, opened 2026-07-15
core: add endpoint for revoking all user sessions

Benricheson101:
### What does this PR change?

Adapts existing API endpoint for bulk revoking user sessions to revoke all sessions when a list of `user_pks` is not provided.

`DELETE /api/v3/core/authenticated_sessions/bulk_delete/`

Shortcut to calling the bulk delete endpoint with every user ID.

### How was this tested?

Unit tests.

## Checklist

### DIFF
--- a/authentik/core/api/authenticated_sessions.py
+++ b/authentik/core/api/authenticated_sessions.py
@@ -16,6 +16,7 @@ from rest_framework.serializers import (
     DateTimeField,
     IPAddressField,
     ListField,
+    PrimaryKeyRelatedField,
 )
 from rest_framework.viewsets import GenericViewSet
 from ua_parser import user_agent_parser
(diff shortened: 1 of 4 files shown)


authentik/core/api/authenticated_sessions.py
```diff
@@ -66,10 +66,19 @@ class UserAgentDict(TypedDict):
 
 
 class BulkDeleteSessionSerializer(PassiveSerializer):
-    """Serializer for bulk deleting authenticated sessions by user"""
+    """Serializer for bulk deleting authenticated sessions"""
 
     user_pks = ListField(
-        child=serializers.IntegerField(), help_text="List of user IDs to revoke all sessions for"
+        child=serializers.IntegerField(),
```
BeryJu: i dont know why this wasn't here from the start but this should be a PrimaryKeyRelatedField to `User.objects.all()`

PR #24104, opened 2026-07-16
web: rework ak-search-select dropdown as an anchored popover (retire ak-portal)

GirlBossRush:
## Details

This PR reworks the dropdown to render its `ak-list-select` menu **in place as a native popover**, so it sits in the top layer and escapes the dialog cleanly, positioned with native CSS anchor positioning (with a JS fallback for Firefox — see below).

These changes are necessary after our recent-ish move towards native `<dialog>` elements. `ak-search-select`'s positioning via `ak-portal` relocates the menu to the top of the DOM and positions it with floating-ui, but inside a modal's top layer (with `overflow: hidden`). This results in the menu being clipped depending on the modal and viewport's dimensions.

## Changes

- `ak-list-select` renders in place with `popover`; positioned with CSS anchor positioning where it's reliable and imperatively otherwise (below by default, flips above when there's no room, width matches the input, long labels wrap).
- Close when the anchor input scrolls out of view (`IntersectionObserver`);
- Menu overscroll forwards to the scroll container behind it — fixed/top-layer popovers don't chain to it natively, which otherwise felt "stuck."
- Removes `<ak-portal />` and the `@floating-ui/dom` dependency; updates the e2e `FormFixture` selector to the new DOM.

## A note on CSS anchor positioning

Similar to our application card components, this uses on CSS anchor positioning + `position-try`, gated behind feature detection. Unfortunately Firefox (through 152) reports `CSS.supports` as true, but mis-renders it when the anchor is inside a `<dialog>`. A fallback for Firefox handles it for now.

## Testing

- Chrome and Firefox (Playwright): placement, width match, wrapping, flip, scroll-tracking through a shadow-DOM dialog body, close-on-scroll-out, and wheel chaining.
- Existing Storybook stories.


web/src/elements/forms/SearchSelect/ak-search-select-view.ts
```diff
@@ -273,13 +295,191 @@ export class SearchSelectView extends AKElement implements ISearchSelectView {
...
+    /** Timestamp of the last browser-driven popover close (see reopen guard). */
+    #lastLightDismiss = -Infinity;
+
+    #clickListener = (event: Event) => {
         if (this.readOnly) return;
 
-        this.open = !this.open;
+        // If this same click just light-dismissed the open popover, treat it as a
+        // close: leave `open` false instead of toggling it back on.
+        const dismissedByThisClick = event.timeStamp - this.#lastLightDismiss < 250;
```
kensternberg-authentik: Suggestion: 

1. put the `250` in a constant at the top, something like `DEFAULT_REFOCUS_DELAY`
2. replace this with `getComputedStyle(this)?.getPropertyValue("--ak-search-select--RefocusDelay") ?? DEFAULT_REFOCUS_DELAY` so this display-oriented delay behavior can be configured via CSS.  :-)

kensternberg-authentik (approved):
Just one suggestion, really.

web/src/elements/dialogs/positioning.ts
```diff
@@ -0,0 +1,17 @@
...
+
+/**
+ * Whether the browser supports the `anchor-size()` function for sizing an element
+ * against its anchor (e.g. `width: anchor-size(width)`).
+ *
+ * @remarks
+ * This is a *separate* capability from {@link AnchorPositionSupported}: Firefox
+ * (through at least 152) ships `position-anchor` and `anchor()` but not
+ * `anchor-size()`, so a consumer that sizes against its anchor must check this too
+ * or the sizing declaration is silently dropped.
```
kensternberg-authentik, line 14: There will, one hopes, come a day when this will be true even for Firefox.  Put a TODO with a date here to "check in some future time whether or not Firefox works."

web/src/elements/forms/SearchSelect/ak-search-select-view.ts
kensternberg-authentik: Clicking an option closes the selector, right?  What if someone clicks an option, decides they were wrong about it, then clicks the selector to re-open?  That can happen within a quarter of a second, "oops," hands on mice are pretty fast.  Should you differentiate between "close for option select" and "close for other reasons?"  Consider recording a different delay for non-light-dismissed events.

web/src/elements/forms/SearchSelect/ak-search-select-view.css
```diff
@@ -0,0 +1,88 @@
...
+            opacity: 0;
+        }
+        to {
+            opacity: 1;
+        }
+    }
+
+    ak-list-select[popover]:popover-open {
+        animation: ak-search-select-menu-in 120ms ease-out;
+    }
```
kensternberg-authentik, line 87: No corresponding fade-out for close?

PR #24118, opened 2026-07-16
ak: obscure password for hashing

PeshekDotDev:
### What does this PR change?

Currently, getting a hashed password to use requires you to pass in the password plaintext as part of the command argument, exposing it in your bash history. This pr obfuscates the plaintext password.

It also asks you to verify the password, a good safety measure to make sure there are no mistakes in inputting the password.

Having your password in plaintext in the command argument keeps it in your bash history, which can compromise the password.

### How was this tested?

Running the command before and after implementation.

### Linked issues

---

## Checklist

### DIFF
--- a/authentik/core/management/commands/hash_password.py
+++ b/authentik/core/management/commands/hash_password.py
@@ -1,5 +1,8 @@
 """Hash password using Django's password hashers"""
 
+import sys
+from getpass import getpass
+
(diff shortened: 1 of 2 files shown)


PeshekDotDev:
@dominic-r Can you please review this and make sure it looks good?

PR #24126, opened 2026-07-16
core: prompt securely when hashing passwords

dominic-r:
Generating a password hash currently requires placing the plaintext password on the command line, where it can be retained in shell history and exposed through process inspection. The command now prompts twice without echoing when standard input is a terminal, or reads one password from standard input for automation.

Follow up to: https://github.com/goauthentik/authentik/pull/24118

### DIFF
--- a/authentik/core/management/commands/hash_password.py
+++ b/authentik/core/management/commands/hash_password.py
@@ -18,10 +18,7 @@ class Command(BaseCommand):
             "password",
             type=str,
             nargs="?",
-            help=_(
-                "Password to hash. Supplying this argument exposes it in shell history and "
-                "the process list; omit it to be prompted securely."
-            ),
+            help=_("Password to hash. If omitted, prompt for the password."),
         )
 
     def handle(self, *args, **options):
--- a/authentik/core/management/commands/hash_password.py
+++ b/authentik/core/management/commands/hash_password.py
@@ -17,23 +17,14 @@ class Command(BaseCommand):
 
     def handle(self, *args, **options):
         if sys.stdin.isatty():
-            try:
-                password = getpass("Password: ")
-                password_again = getpass("Password (again): ")
-            except (EOFError, KeyboardInterrupt) as exc:
-                raise CommandError("Aborted") from exc
+            password = getpass("Password: ")
+            password_again = getpass("Password (again): ")
             if password != password_again:
                 raise CommandError("Passwords do not match")
         else:
-            try:
-                password = input()
-            except EOFError as exc:
-                raise CommandError("Password cannot be empty") from exc
+            password = input()
 
         if not password:
             raise CommandError("Password cannot be empty")
-        try:
-            hashed = make_password(password)
-            self.stdout.write(hashed)
-        except ValueError as exc:
-            raise CommandError(f"Error hashing password: {exc}") from exc
+        hashed = make_password(password)
+        self.stdout.write(hashed)
--- a/authentik/core/management/commands/hash_password.py
+++ b/authentik/core/management/commands/hash_password.py
@@ -1,28 +1,47 @@
 """Hash password using Django's password hashers"""
 
+import sys
+from getpass import getpass
+
 from django.contrib.auth.hashers import make_password
 from django.core.management.base import BaseCommand, CommandError
+from django.utils.translation import gettext as _
 
 
 class Command(BaseCommand):
     """Hash a password using Django's password hashers"""
 
-    help = "Hash a password for use with AUTHENTIK_BOOTSTRAP_PASSWORD_HASH"
+    help = _("Hash a password for use with AUTHENTIK_BOOTSTRAP_PASSWORD_HASH")
 
     def add_arguments(self, parser):
         parser.add_argument(
             "password",
             type=str,
-            help="Password to hash",
+            nargs="?",
+            help=_(
+                "Password to hash. Supplying this argument exposes it in shell history and "
+                "the process list; omit it to be prompted securely."
+            ),
         )
 
     def handle(self, *args, **options):
         password = options["password"]
 
+        if password is None:
+            if not sys.stdin.isatty():
+                raise CommandError(_("Password prompting requires an interactive terminal"))
+            try:
+                password = getpass(_("Password: "))
+                password_again = getpass(_("Password (again): "))
+            except (EOFError, KeyboardInterrupt) as exc:
+                raise CommandError(_("Aborted")) from exc
+            if password != password_again:
+                raise CommandError(_("Passwords do not match"))
+
         if not password:
-            raise CommandError("Password cannot be empty")
+            raise CommandError(_("Password cannot be empty"))
         try:
             hashed = make_password(password)
             self.stdout.write(hashed)
         except ValueError as exc:
-            raise CommandError(f"Error hashing password: {exc}") from exc
+            raise CommandError(f"{_('Error hashing password')}: {exc}") from exc
--- a/authentik/core/management/commands/hash_password.py
+++ b/authentik/core/management/commands/hash_password.py
@@ -11,22 +11,13 @@ from django.utils.translation import gettext as _
 class Command(BaseCommand):
     """Hash a password using Django's password hashers"""
 
-    help = _("Hash a password for use with AUTHENTIK_BOOTSTRAP_PASSWORD_HASH")
-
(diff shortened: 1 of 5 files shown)


PeshekDotDev:
@emilburzo Last release, we added a feature to put hashed passwords in docker composes, send to api, etc instead of sending plaintext. You generate the hashed password by running `docker compose run --rm server hash_password 'your-password'`. This exposes the plaintext password in the bash history. My feeling is that we should never allow a user to run the command like this, and instead the command should be ran as `docker compose run --rm server hash_password`, and then the password is collected after, to keep bash history clean

Do you feel that it is okay for the password to be plaintext in bash history optionally, or should this never be allowed?

dominic-r:
(If we remove support, we would need to include this as a breaking change in the release notes due to how it might affect existing scripts and tooling. Just adding a reminder for myself)

emilburzo:
@PeshekDotDev 

agree that it's not good practice, besides `bash_history`, it can also appear in the running process list (albeit shortly) or logged in places like CI/CD pipelines

for human use, I think the interactive prompt is the way to go

but I assume we likely want/should also support automation? then I think supporting an environment variable and/or standard input would work -- does that make sense?

website/docs/install-config/automated-install.mdx
```diff
@@ -14,10 +14,16 @@ Configure the default password for the `akadmin` user using a pre-hashed Django
 
 This stores the hash directly as authentik's local password verifier. Because authentik never sees the raw password, hashed-password imports do not propagate the password to LDAP or Kerberos integrations, even when password writeback is enabled.
 
-To generate a hash, run this command before your initial deployment:
+To generate a hash without exposing the password in your shell history, run this command before your initial deployment and enter the password when prompted:
```
PeshekDotDev: I think we can skip noting the part of "without exposing the password in your shell history". I think that should be in the next section around using -T, saying something like "you can use -T to read from standard input. Pass the password in an env variable to avoid exposing the password in your shell history"

website/docs/customize/blueprints/v1/models.mdx
```diff
@@ -49,10 +49,16 @@ For example:
...
 
-Use `password_hash` when you need to import or bootstrap an existing hash without exposing the raw password to authentik:
+Use `password_hash` when you need to import or bootstrap an existing hash without exposing the raw password to authentik. Generate the hash without exposing the password in your shell history by entering it when prompted:
 
 ```bash
-docker compose run --rm server hash_password 'your-password'
+docker compose run --rm server hash_password
+```
+
+For automation, pipe the password through standard input. The `-T` option disables pseudo-TTY allocation so that the command reads from standard input:
```
PeshekDotDev: Similar to the other comment, I think the "Generate the hash without exposing the password in your shell history" can probably be removed from that spot, and something here saying like -T allows you to read the command from standard input. Pass the password as an env variable here ot prevent leaking your password in bash history"

PeshekDotDev (commented):
The only other thing, I think we should document how to run the command with the sed filter I had mentioned. IE for when we need to prepend all $'s with a $ for docker compose files

dewi-tik (approved):
Agree with Connor's comment, otherwise LGTM

website/docs/customize/blueprints/v1/models.mdx
```diff
@@ -49,10 +49,16 @@ For example:
...
 
 ```bash
-docker compose run --rm server hash_password 'your-password'
+docker compose run --rm server hash_password
+```
+
+For automation, pass the password in the `PASSWORD` environment variable and pipe it through standard input to prevent exposing the password in your shell history. The `-T` option disables pseudo-TTY allocation so that the command reads from standard input:
+
+```bash
+printf '%s\n' "$PASSWORD" | docker compose run --rm -T server hash_password
```
emilburzo: is the `-T` really needed? it seems to work without it

```bash
$ printf '%s\n' "$PASSWORD" | docker compose run --rm server hash_password
pbkdf2_sha256$1000000$IfRvgi6VkGkwXXXXXXXXXXXXXXta/T8f8tHriB9xacQcI8p0Gpko=
```

website/docs/install-config/automated-install.mdx
```diff
@@ -14,10 +14,16 @@ Configure the default password for the `akadmin` user using a pre-hashed Django
...
 
 ```bash
-docker compose run --rm server hash_password 'your-password'
+docker compose run --rm server hash_password
+```
+
+For automation, pass the password in the `PASSWORD` environment variable and pipe it through standard input to prevent exposing the password in your shell history. The `-T` option disables pseudo-TTY allocation so that the command reads from standard input:
+
+```bash
+printf '%s\n' "$PASSWORD" | docker compose run --rm -T server hash_password
```
emilburzo: NIT: `echo "$PASSWORD"` instead of `printf` seems cleaner, since you need a newline anyway, or am I missing some detail?

authentik/core/management/commands/hash_password.py
```diff
@@ -1,28 +1,37 @@
...
+        if sys.stdin.isatty():
+            try:
+                password = getpass(_("Password: "))
+                password_again = getpass(_("Password (again): "))
+            except (EOFError, KeyboardInterrupt) as exc:
+                raise CommandError(_("Aborted")) from exc
+            if password != password_again:
+                raise CommandError(_("Passwords do not match"))
+        else:
+            password = sys.stdin.readline().rstrip("\r\n")
```
gergosimonyi: Why is this not just


```suggestion
            password = input()
```

?

authentik/core/management/commands/hash_password.py
```diff
@@ -1,28 +1,37 @@
 """Hash password using Django's password hashers"""
 
+import sys
+from getpass import getpass
+
 from django.contrib.auth.hashers import make_password
 from django.core.management.base import BaseCommand, CommandError
+from django.utils.translation import gettext as _
```
gergosimonyi: What is the use for translations in a management command?

authentik/core/tests/test_hash_password_command.py
```diff
@@ -11,18 +12,69 @@
...
         self.assertIn("Password cannot be empty", str(ctx.exception))
+
+    def test_hash_password_prompt_errors(self):
+        """Test errors raised while prompting for a password."""
+        for side_effect, message in (
+            (["", ""], "Password cannot be empty"),
+            (["test123", "different"], "Passwords do not match"),
+            (KeyboardInterrupt, "Aborted"),
+        ):
+            with self.subTest(message=message), self.assertRaises(CommandError) as ctx:
```
gergosimonyi: Generally authentik's codebase doesn't use `subTest` for cases this small. The `(KeyboardInterrupt, "Aborted"),` is overkill, the two others should just be unabstracted.

authentik/core/tests/test_hash_password_command.py
```diff
@@ -11,18 +12,69 @@
...
-        call_command("hash_password", "test123", stdout=out)
+        self.call_prompt(["test123", "test123"], stdout=out)
         hashed = out.getvalue().strip()
 
         self.assertTrue(hashed.startswith("pbkdf2_sha256$"))
         self.assertTrue(check_password("test123", hashed))
 
-    def test_hash_password_empty_fails(self):
-        """Test that empty password raises error."""
+    def test_hash_password_argument_fails(self):
```
gergosimonyi: This is good to test at the time of this PR, but it won't ever be needed in the future. Remove please.

authentik/core/tests/test_hash_password_command.py
```diff
@@ -11,18 +12,69 @@
 class TestHashPasswordCommand(TestCase):
     """Test hash_password management command."""
 
-    def test_hash_password(self):
-        """Test hashing a password."""
+    def call_prompt(self, side_effect, **options):
```
gergosimonyi: This abstraction is most likely not needed. As far as I can tell, there should be exactly one test that uses this function: the one that tests whether the prompt works. Every other test should just stdin.

website/docs/install-config/automated-install.mdx
gergosimonyi: (Also possibly add an example like

```
[...] < passwordfile
```
)

dominic-r:
:)

image

emilburzo (approved):

rissson (approved):

PR #24132, opened 2026-07-16
sources/oauth: allow long authorization URLs on OAuth sources

FarukZahiragic:
## Details

`OAuthSource.authorization_url` (and the sibling `request_token_url`, `access_token_url`, `profile_url` fields) are `CharField(max_length=255)`. That is too short when the Authorization URL includes static query parameters — for example CILogon's `idphint` with multiple SAML EntityIDs to pre-select / shortlist institutional IdPs.

This switches those four URL fields to `TextField`, consistent with the existing longer URL fields on the same model (`oidc_well_known_url`, `oidc_jwks_url`).

### Use case

Self-hosted authentik federating to CILogon. We need something like:

```
https://cilogon.org/authorize?idphint=<EntityID1>,<EntityID2>,...
```

With ~10 Swiss institutional EntityIDs URL-encoded, this exceeds 255 characters and the admin UI rejects the save with "Ensure this field has no more than 255 characters."

## Changes

- `authentik/sources/oauth/models.py`: `request_token_url`, `authorization_url`, `access_token_url`, `profile_url` → `TextField`
- `authentik/sources/oauth/migrations/0015_oauthsource_url_textfields.py`: `AlterField` migration for the four fields
- `authentik/sources/oauth/tests/test_api.py`: test saving an authorization URL longer than 255 characters
- `schema.yml`: regenerated to drop the `maxLength: 255` constraint on these fields


FarukZahiragic:
## Manual verification

In addition to the automated test (`test_patch_long_url`) and the local checks that mirror what CI runs (`schema.yml` / `blueprints/schema.json` diffed byte-for-byte against `make gen-build`, `Ruff`/`Black` clean, `makemigrations --check --dry-run`: no pending migrations, full `authentik.sources.oauth` suite: 49 passed), this change was manually tested end-to-end against a local dev environment (Postgres + Django backend + the actual built web UI) for **all four** widened fields:

- **`authorization_url`**, **`access_token_url`**, **`profile_url`**: created/edited an OpenID Connect OAuth source through the real admin web UI (not just the API), set each field to a value well over 255 characters, saved, and confirmed the full value persisted and re-rendered correctly on reload.
- **`request_token_url`**: this field isn't exposed by the current web UI for any provider type — its visibility in `OAuthSourceForm.ts` is gated on the provider type's `requestTokenUrl` default, which no built-in provider (including legacy OAuth1 ones) sets. So this one was verified directly against the API: `PATCH`'d a 431-character URL, then confirmed the full value round-tripped both in the API response and via a direct read from Postgres.

All four fields save and persist correctly past the old 255-character limit.

BeryJu (approved):

FarukZahiragic:
Updated branch to main — workflows are awaiting approval again. @BeryJu could you approve the workflows so CI can re-run? Thanks!

PR #24134, opened 2026-07-16
web/elements: update the progress bar

kensternberg-authentik:
web/maint: upgrade ak-progress

## Note

This PR depends on _both_ #24082 and #24170, and since those two _don't_ build on each other (they're separate issues!), I can't properly create a dependency graph.  When those two are committed, the "files changed" on this should drop significantly.

## What

This commit upgrades ak-progress with the following improvements:

- Merges our variant with the Patternfly 5 CSS
- Tightens the margins on labels and displays
- Fixes the “compact” variant
- Fixes the “large” variant

I have updated provide a comprehensive Storybook story

## Checklist

-   [🚧] The project has been linted, built, and tested (`make all`)
-   [🚧] The documentation has been updated and formatted (`make docs`)


web/src/elements/types.ts
```diff
@@ -336,3 +336,33 @@ export type SlottedTemplateResult =
     | HTMLElement;
 
 export type Spread = { [key: string]: unknown };
+
+/**
+ * For builder functions, valid data attributes that can be placed on custom components.
+ */
+
+export type DataRecord = {
```
GirlBossRush, line 344:
```suggestion
export type DataRecord<K extends string = string> = {
```

web/src/elements/types.ts
```diff
@@ -336,3 +336,33 @@ export type SlottedTemplateResult =
...
+    [key: `data-${string}`]: string | boolean;
```
GirlBossRush, line 345:
```suggestion
    [key: `data-${K}`]: string | boolean;
```

web/src/elements/types.ts
```diff
@@ -336,3 +336,33 @@ export type SlottedTemplateResult =
...
+export type AriaRecord = {
+    [key: `aria-${string}`]: string;
+};
+
+/**
+ * This may be overkill.
+ */
+
+export type ElementRest = DataRecord &
+    AriaRecord & {
```
GirlBossRush: Recommend pulling the final type into an interface. I think React calls these something like `ElementIntrinsicAttributes`

GirlBossRush (approved):

web/src/elements/types.ts
kensternberg-authentik: "An index signature parameter type cannot be a literal type or generic type. Consider using a mapped object type instead."  Converting it to a mapped object means it cannot be generic over strings; you would HAVE to specify all the `data` keys in the type.  I'm gonna have to leave it as-is; otherwise, it becomes a DX headache.  We _want_ users to be able to add their own `data-` keys if they have their own purposes.

PR #24160, opened 2026-07-17
lifecycle: prevent version skipping upgrades

gergosimonyi:


lifecycle/migrate.py
```diff
@@ -73,6 +75,43 @@ def release_lock(conn: Connection, cursor: Cursor):
...
+    if (
+        db_version.major == previous_code_version.major
+        and db_version.minor == previous_code_version.minor
+    ) or (
+        db_version.major == current_code_version.major
+        and db_version.minor == current_code_version.minor
+    ):
+        return
+
+    message = "Major version skips are not allowed. Please upgrade one major version at a time."
```
BeryJu: I think we should add a link to the docs and or the release notes overview page to make it easier for people to figure out which versions they need to upgrade to

BeryJu (approved):

.github/workflows/release-branch-off.yml
```diff
@@ -72,6 +72,12 @@ jobs:
           dependencies: "system,python,go,node,runtime,rust-nightly"
       - name: Run migrations
         run: make migrate
+      - name: Update previous version
+        run: |
+          VERSION="$(<"${PWD}/internal/constants/VERSION")"
+          sed -i -E \
+            "s/^VERSION_PREVIOUS_MAJOR = .*/VERSION_PREVIOUS_MAJOR = \"${VERSION}\"/" \
+            "${PWD}/authentik/__init__.py"
```
rissson, line 80: This should be done by `make bump` imo

authentik/__init__.py
```diff
@@ -4,6 +4,7 @@
 from os import environ
 
 VERSION = "2026.8.0-rc1"
+VERSION_PREVIOUS_MAJOR = "2026.5.0-rc1"
```
rissson: Should we just store `2026.5` instead? Maybe call it `VERSION_FAMILIY_PREVIOUS`?

lifecycle/migrate.py
```diff
@@ -73,6 +75,43 @@ def release_lock(conn: Connection, cursor: Cursor):
...
+    LOGGER.error(message, upgrading_from=db_version, upgrading_to=current_code_version)
+    raise RuntimeError(message)
```
rissson, line 121: should we add a hidden config setting `AUTHENTIK_MIGRATIONS__ALLOW_MULTIPLE_MAJOR_VERSION_UPGRADE__YES_I_ACCEPT_THAT_I_MAY_BREAK_MY_DATABASE` or something to allow bypassing that?

.github/workflows/release-branch-off.yml
BeryJu: bump runs everytime we bump version but this should only run when doing branch off

lifecycle/migrate.py
gergosimonyi: `insert into authentik_version_history [...]` can bypass stuff, but yeah, I'll add this

authentik/__init__.py
gergosimonyi: This is the lazy way, may or may not be optimal. Parsing semver in bash (or: writing python that is then called from bash) seemed overkill.

.github/workflows/release-branch-off.yml
gergosimonyi: If it's in `make bump`, we'd have to do something like

```
if it's a patch release or prerelease
    do nothing
else
    do the thing
```

In bash. In a makefile. (Or in a script called by bash).

Long story short, this is again the lazy way.

authentik/__init__.py
gergosimonyi: I'll revert my position, I now believe yours is the lazy way. I'll change it.

lifecycle/migrate.py
```diff
@@ -73,6 +75,52 @@ def release_lock(conn: Connection, cursor: Cursor):
...
+    if (
+        db_version.major == lowest_acceptable_version.major
+        and db_version.minor == lowest_acceptable_version.minor
+    ) or (
+        db_version.major == current_code_version.major
+        and db_version.minor == current_code_version.minor
+    ):
+        return
+
+    message = f"Major version skips are not allowed. See: https://docs.goauthentik.io/install-config/upgrade/?from={db_version}&to={current_code_version}"
```
gergosimonyi, line 115: (Currently the querystring does nothing, but the page is correct.)

PR #24170, opened 2026-07-18
web/admin: enterprise status card properties

kensternberg-authentik:
web/maint: clean-up ak-enterprise-status-card

## What

This commit updates ak-enterprise-status-card with the following improvements:

- `@state()` replaced with `@property({ attribute: false })`, which gets us better linting and typechecking for client code
- `renderSummaryBadge()` uses a map expression instead of switch statement. Removes some boilerplate and clutter
- Slightly adjusted some margins, cleaning up the card. It’s a little shorter now, too.
- Fixed some misconceptions around what `msg(str` actually does. The remaining `msg` strings are all that lit-localize actually uses.
- Added the previously unhandled "User Count Exceeded" condition
- Duplicate progress bars replaced with a single function that does the percentage and severity in a new scope, creating a tighter and more visually appealing codebase.
- Includes a complete Storybook story file with multiple stories for Valid, Expiring, Exceeded, Unlicensed and Loading

## Why

This mostly started as a test-case for ak-progress-bar. I needed a case where I could see the progress bar in use, and we didn’t have many of them. I decided to add a Storybook story for the Enterprise Status card, because it was a standalone component that used two of them, so it would be a good place to see how they looked stacked one atop another.

While I was rigging the Storybook, the linter complained that `@state()` wasn’t appropriate for properties that were being passed to the component. So I cleaned those. While trying to balance the whitespace around the `∞` symbol, I got frustrated with all the repetition and unbalanced code, and, well, things sorta spiraled out from there.

Anyway, it’s nicer now.  Some samples from the Storybook:

image

image

## Checklist

-   [🌪️] The project has been linted, built, and tested (`make all`)
-   [🌪️] The documentation has been updated and formatted (`make docs`)


web/src/admin/enterprise/EnterpriseStatusCard.ts
```diff
@@ -5,40 +5,46 @@ import { PFColor } from "#elements/Label";
...
+import { css, CSSResult, html, nothing } from "lit";
+import { customElement, property } from "lit/decorators.js";
+import { classMap } from "lit/directives/class-map.js";
 
 import PFCard from "@patternfly/patternfly/components/Card/card.css";
 import PFDescriptionList from "@patternfly/patternfly/components/DescriptionList/description-list.css";
 import PFProgress from "@patternfly/patternfly/components/Progress/progress.css";
 import PFSplit from "@patternfly/patternfly/layouts/Split/split.css";
 
+const badgeDetails = new Map<LicenseSummaryStatusEnum | undefined, [PFColor, string]>([
```
GirlBossRush:
```suggestion
const badgeDetails = new Map<LicenseSummaryStatusEnum, [PFColor, string]>([
```

Alternatively, a Record object to enforce type safety on each enum.

web/src/admin/enterprise/EnterpriseStatusCard.ts
```diff
@@ -5,40 +5,46 @@ import { PFColor } from "#elements/Label";
...
+    [LicenseSummaryStatusEnum.Expired, [PFColor.Red, msg("Expired")]],
+    [LicenseSummaryStatusEnum.ExpirySoon, [PFColor.Orange, msg("Expiring soon")]],
+    [LicenseSummaryStatusEnum.Unlicensed, [PFColor.Gray, msg("Unlicensed")]],
+    [LicenseSummaryStatusEnum.ReadOnly, [PFColor.Red, msg("Read Only")]],
+    [LicenseSummaryStatusEnum.LimitExceededAdmin, [PFColor.Orange, msg("User Count Exceeded")]],
+    [LicenseSummaryStatusEnum.LimitExceededUser, [PFColor.Red, msg("User Count Exceeded")]],
+    [LicenseSummaryStatusEnum.Valid, [PFColor.Green, msg("Valid")]],
+]);
+
+const Style = css`
```
GirlBossRush:
```suggestion
const Styles = css`
```

To match our imports elsewhere

web/src/admin/enterprise/EnterpriseStatusCard.ts
```diff
@@ -5,40 +5,46 @@ import { PFColor } from "#elements/Label";
...
+    .pf-l-split {
+        --pf-l-split--m-gutter--MarginRight: 3rem;
+    }
+`;
+
 @customElement("ak-enterprise-status-card")
 export class EnterpriseStatusCard extends AKElement {
-    @state()
+    @property({ attribute: false })
```
GirlBossRush, line 53: `public`

web/src/admin/enterprise/EnterpriseStatusCard.ts
```diff
@@ -5,40 +5,46 @@ import { PFColor } from "#elements/Label";
...
     forecast?: LicenseForecast;
 
-    @state()
+    @property({ attribute: false })
     summary?: LicenseSummary;
 
-    static styles: CSSResult[] = [PFDescriptionList, PFCard, PFSplit, PFProgress];
+    static styles: CSSResult[] = [PFDescriptionList, PFCard, PFSplit, PFProgress, Style];
```
GirlBossRush: Recommend moving static members towards the top of the class.  
```suggestion
    public static styles: CSSResult[] = [PFDescriptionList, PFCard, PFSplit, PFProgress, Styles];
```

GirlBossRush (approved):
Minor nits but a solid improvement to the original.

web/src/admin/enterprise/EnterpriseStatusCard.ts
kensternberg-authentik: Accepted, but the patch is bigger than this.  The `| undefined` there is an escape in case `this.summary?.status` was undefined.  Putting check into `renderSummaryBadge()` to ensure it is defined was also necessary.

PR #24177, opened 2026-07-18
internal/config: make storage file paths overwritable via env vars

alexeyabel:
Now *__FILE__PATH environemnt variables have an effect and can overwrite values from default.yml. Without this change uploaded files can not be accessed if the folder is different from the default /data because the GoLang part assumes a /data prefix and e.g. ignores AUTHENTIK_STORAGE__FILE__PATH.

### How was this tested?

I have not tested this. I am on NixOS and `make all` does not work out-of-the-box due to dynamic linking. I presume that this minor change will not break anything.

### Linked issues

I have not created an issue, since I already (hopefully) have the solution.

P.S. AFAICS your [package list](https://docs.goauthentik.io/developer-docs/contributing/#authentiks-structure) does not include Go modules. Therefore, I was not able to chose a fitting prefix and came up with `config:`. Feel free to change if you need. I just want this to be fixed.

### DIFF
--- a/internal/config/struct.go
+++ b/internal/config/struct.go
@@ -69,7 +69,7 @@ type StorageConfig struct {
 }
 
 type StorageFileConfig struct {
-	Path string `yaml:"path" env:"AUTHENTIK_STORAGE__FILE__PATH"`
+	Path string `yaml:"path" env:"AUTHENTIK_STORAGE__FILE__PATH, overwrite"`
 }
 
 type StorageMediaConfig struct {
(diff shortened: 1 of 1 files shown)


BeryJu (approved):

dominic-r (approved):

alexeyabel:
@BeryJu I don't understand all these workflows and approval processes and thus how to proceed. Am I supposed to merge this myself or will any of the maintainers do it?

rissson:

PR #24130, opened 2026-07-16
core: validate imported password hashes at the API boundary

dominic-r:
Malformed hashes submitted to the password-hash API previously reached the model before format validation. This moves structural validation into `authentik.lib.validators` and attaches it directly to the API serializer field, while leaving the user model as a storage primitive and user-authored blueprints unrestricted.

`AUTHENTIK_BOOTSTRAP_PASSWORD_HASH` is validated explicitly at its environment boundary because bootstrap is implemented through a system blueprint. Bootstrap apply failures leave setup incomplete. This PR is stacked on #24126. Validated with focused Django tests, `make gen`, `make lint-fix`, `make docs`, and `./manage.py makemigrations --check`.

PR #24234, opened 2026-07-20
core: restrict imported password hashers

dominic-r:
Django can identify password hashes that do not meet authentik's current security policy. The API and `AUTHENTIK_BOOTSTRAP_PASSWORD_HASH` now default to PBKDF2-SHA256, bcrypt-SHA256, scrypt, or Argon2 with current parameters, sufficient salt entropy, and structurally valid digests.

API clients can set `override` to import another structurally valid Django hash; bootstrap has no override, while blueprint hashes remain unrestricted. This PR is stacked on #24130.


authentik/core/api/password_hashes.py
```diff
@@ -1,22 +1,70 @@
...
+)
 from django.utils.translation import gettext_lazy as _
 from rest_framework.exceptions import ValidationError
 from rest_framework.fields import CharField
 
 INVALID_PASSWORD_HASH_MESSAGE = _(
-    "Invalid password hash format. Must be a valid Django password hash."
+    "Invalid password hash. Use a supported algorithm with current parameters."
+)
+SUPPORTED_PASSWORD_HASHERS = (
```
rissson: let's add a comment that this should match the list we have in settings.py

authentik/lib/validators.py
```diff
@@ -1,21 +1,92 @@
...
+    must_update_salt,
+)
 from django.utils.translation import gettext_lazy as _
 from rest_framework.exceptions import ValidationError
 from rest_framework.serializers import Serializer
 from rest_framework.utils.representation import smart_repr
 
+INVALID_PASSWORD_HASH_MESSAGE = _("Invalid password hash encoding.")
+
+# Keep this allowlist in sync with the non-legacy hashers in settings.PASSWORD_HASHERS.
```
rissson: sry I should've mentioned, but a reverse comment in settings.py as well, so whenever each is updated we know to update the other one.

dominic-r: lol all good. I completely forgot but it was the logical thing to do.

dominic-r: I can't seem to find it, do you have a specific Line?

authentik/core/api/users.py
```diff
@@ -108,6 +108,12 @@
 
 LOGGER = get_logger()
 
+PASSWORD_HASH_REQUIRES_OVERRIDE_MESSAGE = _(
+    "This password hash does not use its current work factor and sufficient salt entropy. "
```
gergosimonyi: This is "or" instead of "and", right? I'd prefer two errors for the two separate issues.

I'd also prefer the algorithm-side error message with much more detail, including all the parameters of the algorithm used. Let the user know exactly where they fall short, something like

```
Provided:
Algorithm: argon2
Time cost: 2
Memory cost: 2048000
Parallelism: 1

Expected:
Algorithm: argon2
Time cost: 2
Memory cost: 102400
Parallelism: 8
```

gergosimonyi: Could also be a static error message if you like with all possible expectations.

gergosimonyi (commented):
Also, can I just import `this-cannot-be-parsed-haha` as a hash if I set `override=true`? Don't think that should be possible.

authentik/lib/validators.py
```diff
@@ -1,21 +1,91 @@
...
-def validate_password_hash(password_hash: str) -> None:
-    """Validate the format of an encoded Django password."""
-    try:
-        hasher = identify_hasher(password_hash)
-        hasher.decode(password_hash)
-    except (AssertionError, TypeError, ValueError) as exc:
-        raise ValidationError(
-            _("Invalid password hash format. Must be a valid Django password hash.")
-        ) from exc
+class PasswordHashRequiresOverride(Exception):
```
gergosimonyi: Should probably be


```suggestion
class PasswordHashRequiresOverride(ValidationError):
```

, no?

authentik/core/api/users.py
```diff
@@ -959,9 +989,12 @@ def set_password_hash(
         """
         user: User = self.get_object()
         try:
-            user.set_password_from_hash(body.validated_data["password"], request=request)
+            user.set_password_from_hash(
+                body.validated_data["password"],
+                request=request,
+            )
```
gergosimonyi: Weird noop change

authentik/lib/validators.py
```diff
@@ -1,21 +1,91 @@
...
+        hasher, decoded = self._decode(password_hash)
+        importable = [
+            configured_hasher
+            for configured_hasher in get_hashers()
+            if configured_hasher.algorithm != "pbkdf2_sha1"
+        ]
+        messages: list[str] = []
+        salt_stale = must_update_salt(decoded["salt"], hasher.salt_entropy)
+        # hasher.must_update() is also true for a short salt.
+        if hasher not in importable or (hasher.must_update(password_hash) and not salt_stale):
```
gergosimonyi: `hasher not in importable` compares identity and can result in stuff like

```
BCryptSHA256PasswordHasher() not in [BCryptSHA256PasswordHasher()]
```

being `True`. We should compare on something else, probably `hasher.algorithm` is good enough.

authentik/core/api/users.py
```diff
@@ -498,7 +502,33 @@ class UserPasswordSetSerializer(PassiveSerializer):
...
+        """Validate the password hash against the selected import policy."""
+        validator = PasswordHashValidator() if attrs["override"] else PasswordHashImportValidator()
+        try:
+            validator(attrs["password"])
+        except PasswordHashRequiresOverride as exc:
+            raise ValidationError(
+                {
+                    "password": [
+                        _('%(reason)s Set "override" to true to import it anyway.')
+                        % {"reason": reason}
```
gergosimonyi, line 524: Seems like this can't handle two messages. If I have a pbkdf2 hash with low entropy salt and non-standard iterations, I'll first get an error for the salt, then only after fixing it will I get an error for the number of iterations.

authentik/lib/validators.py
```diff
@@ -1,21 +1,91 @@
...
+                    "rounds",
+                    "time_cost",
+                    "memory_cost",
+                    "work_factor",
+                    "block_size",
+                    "parallelism",
+                )
+                if hasattr(hasher, attr)
+            )
+            expected.append(f"{hasher.algorithm} ({params})")
```
gergosimonyi: Instead of a dump of all algorithms and their default parameters, only output the relevant algorithm and its default parameters.

authentik/lib/validators.py
```diff
@@ -1,21 +1,91 @@
...
+        importable = [
+            configured_hasher
+            for configured_hasher in get_hashers()
+            if configured_hasher.algorithm != "pbkdf2_sha1"
+        ]
+        messages: list[str] = []
+        salt_stale = must_update_salt(decoded["salt"], hasher.salt_entropy)
+        # hasher.must_update() is also true for a short salt.
+        if hasher not in importable or (hasher.must_update(password_hash) and not salt_stale):
+            messages.append(self._policy_message(importable))
```
gergosimonyi: Passing only `importable` to `_policy_message` means setting a non-importable `pbkdf2_sha1` hash doesn't get any help from the error message.

authentik/core/setup/signals.py
```diff
@@ -34,7 +34,8 @@ def post_startup_setup_bootstrap(sender, **_):
             continue
         with tenant:
             if password_hash := getenv("AUTHENTIK_BOOTSTRAP_PASSWORD_HASH"):
-                validate_password_hash(password_hash)
+                validator = PasswordHashImportValidator()
+                validator(password_hash)
```
gergosimonyi: The error here is quite noisy and still is of type `PasswordHashRequiresOverride` which a bit confusing since it is not possible to set `override` while bootstrapping. Please make it more friendly.

authentik/lib/validators.py
```diff
@@ -1,21 +1,91 @@
...
+
+class PasswordHashImportValidator(PasswordHashValidator):
+    """Validate a password hash against authentik's import policy."""
+
+    def __call__(self, password_hash: str) -> None:
+        hasher, decoded = self._decode(password_hash)
+        importable = [
+            configured_hasher
+            for configured_hasher in get_hashers()
+            if configured_hasher.algorithm != "pbkdf2_sha1"
```
gergosimonyi: Magic string. Instead, please create something like a non-importable list of algorithms and check against that.

authentik/root/settings.py
```diff
@@ -390,6 +390,16 @@
 # Password validation
 # https://docs.djangoproject.com/en/2.1/ref/settings/#auth-password-validators
 
+# The first hasher is used for new passwords. The rest remain available so existing
```
gergosimonyi, line 390: This should start 4 lines above. The current state is that you've separated the comment above from `AUTH_PASSWORD_VALIDATORS` below.

authentik/root/settings.py
```diff
@@ -390,6 +390,16 @@
...
+# hashes still verify.
+PASSWORD_HASHERS = [
+    "django.contrib.auth.hashers.PBKDF2PasswordHasher",
+    "django.contrib.auth.hashers.PBKDF2SHA1PasswordHasher",
+    "django.contrib.auth.hashers.Argon2PasswordHasher",
+    "django.contrib.auth.hashers.BCryptSHA256PasswordHasher",
```
gergosimonyi, line 396: Seems like [`bcrypt` requires a package](https://docs.djangoproject.com/en/5.2/topics/auth/passwords/#using-bcrypt-with-django). It's currently installed transitively because it's required by `paramiko` which we require, but we should probably require it explicitly instead.

authentik/core/tests/test_users_api.py
```diff
@@ -31,7 +35,7 @@
 from authentik.stages.email.models import EmailStage
 
 INVALID_PASSWORD_HASH = "not-a-valid-hash"
-INVALID_PASSWORD_HASH_ERROR = "Invalid password hash format. Must be a valid Django password hash."
+INVALID_PASSWORD_HASH_ERROR = "Invalid password hash encoding."
```
gergosimonyi: Please write a more helpful message.

authentik/core/api/users.py
```diff
@@ -959,9 +989,12 @@ def set_password_hash(
...
         user: User = self.get_object()
         try:
-            user.set_password_from_hash(body.validated_data["password"], request=request)
+            user.set_password_from_hash(
+                body.validated_data["password"],
+                request=request,
+            )
             user.save()
-        except (ValidationError, IntegrityError) as exc:
+        except IntegrityError as exc:
```
gergosimonyi: This should probably have some helpful message instead of just returning HTTP 400 with no body.

gergosimonyi (commented):
Also, there should be some way we put the information that `override` was used for a password hash update into the audit log.

authentik/core/api/users.py
```diff
@@ -992,11 +992,15 @@ def set_password_hash(
...
+                password_hash_override=body.validated_data["override"],
             )
             user.save()
         except IntegrityError as exc:
             LOGGER.debug("Failed to set password hash", exc=exc)
-            return Response(status=400)
+            return Response(
+                {"detail": _("The password hash could not be saved due to a database constraint.")},
+                status=400,
+            )
```
gergosimonyi: Actually, when can this `IntegrityError` even happen?

authentik/lib/validators.py
```diff
@@ -13,19 +13,29 @@
 from rest_framework.serializers import Serializer
 from rest_framework.utils.representation import smart_repr
 
+_NON_IMPORTABLE_PASSWORD_HASHER_ALGORITHMS = ("pbkdf2_sha1",)
```
gergosimonyi: Weird placing. The information this constant carries is not for `authentik/lib/validators.py` Perhaps this should even be next to the importable hashers in `root/settings.py`? Or is that too much?

authentik/lib/validators.py
```diff
@@ -13,19 +13,29 @@
...
+_PASSWORD_HASHER_PARAMETERS = (
+    ("iterations", "iterations", _("Iterations")),
+    ("work_factor", "rounds", _("Work factor")),
```
gergosimonyi: Huh? Why is this different?

authentik/core/setup/signals.py
```diff
@@ -34,8 +37,16 @@ def post_startup_setup_bootstrap(sender, **_):
...
-                validator(password_hash)
+                try:
+                    PasswordHashImportValidator()(password_hash)
+                except ValidationError as exc:
+                    raise ImproperlyConfigured(
+                        gettext_lazy(
+                            "AUTHENTIK_BOOTSTRAP_PASSWORD_HASH does not match authentik's current "
+                            "password hashing policy. Generate a new hash with authentik's current "
+                            "settings."
+                        )
```
gergosimonyi: Why is this translated?

authentik/core/api/users.py
```diff
@@ -498,7 +502,33 @@ class UserPasswordSetSerializer(PassiveSerializer):
...
+                {
+                    "password": [
+                        _('%(reason)s Set "override" to true to import it anyway.')
+                        % {"reason": reason}
+                        for reason in exc.detail
+                    ]
+                }
+            ) from exc
+        except ValidationError as exc:
+            raise ValidationError({"password": exc.detail}) from exc
```
gergosimonyi, line 530: When does this happen? Can it not be removed?

authentik/core/models.py
```diff
@@ -569,7 +569,14 @@ def set_password(self, raw_password, signal=True, sender=None, request=None):
...
         return super().set_password(raw_password)
 
-    def set_password_from_hash(self, password_hash: str, signal=True, sender=None, request=None):
+    def set_password_from_hash(
+        self,
+        password_hash: str,
+        signal=True,
+        sender=None,
+        request=None,
+        password_hash_override: bool | None = None,
```
gergosimonyi: Unclear name. Something like `hasher_defaults_overridden` may convey better what this is.

authentik/lib/validators.py
```diff
@@ -1,21 +1,107 @@
...
+                % {"algorithm": hasher.algorithm}
+            )
+
+        parameters = [
+            (label, decoded[decoded_name], getattr(hasher, configured_name))
+            for decoded_name, configured_name, label in _PASSWORD_HASHER_PARAMETERS
+            if decoded_name in decoded and hasattr(hasher, configured_name)
+        ]
+        if any(provided != expected for _, provided, expected in parameters):
+            messages.append(self._policy_message(hasher, parameters))
```
gergosimonyi: I don't like this, we're using django internals here, not public API. We should depend on `must_update`, not our own shenanigans. `_PASSWORD_HASHER_PARAMETERS` _may_ be okay for descriptive error messages, but not for determining hash validity.

authentik/lib/validators.py
```diff
@@ -1,21 +1,107 @@
...
+
+        if must_update_salt(decoded["salt"], hasher.salt_entropy):
```
gergosimonyi: I didn't notice this before: django internals may fail us here. The source for this is

```python
def must_update_salt(salt, expected_entropy):
    # Each character in the salt provides log_2(len(alphabet)) bits of entropy.
    return len(salt) * math.log2(len(RANDOM_STRING_CHARS)) < expected_entropy
```

, but salt from an external source may have used a different alphabet. I think the code should stay (since we have no way of knowing the alphabet anyway), but it's worth calling this out in the docs. Something that carries the information "if your hash salt wasn't generated by django and was flagged for low entropy, that may be a false positive; we estimate entropy simply by `length * log2(62)`".

authentik/lib/validators.py
gergosimonyi: Case in point: argon2i hashes are not flagged by this.

authentik/lib/validators.py
```diff
@@ -1,21 +1,107 @@
...
+            configured_hasher.algorithm
+            for configured_hasher in get_hashers()
+            if configured_hasher.algorithm not in _NON_IMPORTABLE_PASSWORD_HASHER_ALGORITHMS
+        }
+        messages: list[str] = []
+        if hasher.algorithm not in importable_algorithms:
+            messages.append(
+                _("Password hash algorithm %(algorithm)s is not accepted by the import policy.")
+                % {"algorithm": hasher.algorithm}
+            )
```
gergosimonyi: Not actionable. Provide a list of accepted algorithms.

authentik/lib/validators.py
```diff
@@ -1,21 +1,107 @@
...
+        hasher: BasePasswordHasher,
+        parameters: list[tuple[str, Any, Any]],
+    ) -> str:
+        """Describe the password hash parameters accepted for import."""
+        provided = "\n".join(f"{label}: {value}" for label, value, _ in parameters)
+        expected = "\n".join(f"{label}: {value}" for label, _, value in parameters)
+        return _(
+            "Password hash parameters do not match authentik's current policy.\n\n"
+            "Provided:\nAlgorithm: %(algorithm)s\n%(provided)s\n\n"
+            "Expected:\nAlgorithm: %(algorithm)s\n%(expected)s"
```
gergosimonyi: Bit weird to me, since there's no real "expected algorithm". How about


```suggestion
            "Password hash parameters do not match authentik's current policy for algorithm %(algorithm)s.\n\n"
            "Provided:\n%(provided)s\n\n"
            "Expected:\n%(expected)s"
```

?

gergosimonyi (commented):
We keep saying "password hashing policy" which is a bit confusing to me. Policy is an overloaded word. Maybe: password hashing settings(/defaults)?

authentik/core/api/users.py
dominic-r: offline discussion, keeping

gergosimonyi: Because a `ValidationError` can be raised as a result of a hash that can't be decoded.

authentik/root/settings.py
```diff
@@ -396,6 +396,12 @@
     "django.contrib.auth.hashers.BCryptSHA256PasswordHasher",
     "django.contrib.auth.hashers.ScryptPasswordHasher",
 ]
+PASSWORD_HASH_IMPORT_ALLOWED_ALGORITHMS = (
+    "pbkdf2_sha256",
+    "argon2",
+    "bcrypt_sha256",
+    "scrypt",
+)
```
gergosimonyi: Bit weird to me that we have hashers above, but algorithms (in string form) below. Why not

```python
PASSWORD_HASHERS_IMPORT_ALLOWED = [
    "django.contrib.auth.hashers.PBKDF2PasswordHasher",
    "django.contrib.auth.hashers.Argon2PasswordHasher",
    "django.contrib.auth.hashers.BCryptSHA256PasswordHasher", 
    "django.contrib.auth.hashers.ScryptPasswordHasher",
]
```

?

authentik/core/models.py
```diff
@@ -569,7 +569,14 @@ def set_password(self, raw_password, signal=True, sender=None, request=None):
...
         return super().set_password(raw_password)
 
-    def set_password_from_hash(self, password_hash: str, signal=True, sender=None, request=None):
+    def set_password_from_hash(
+        self,
+        password_hash: str,
+        signal=True,
+        sender=None,
+        request=None,
+        hasher_defaults_overridden: bool | None = None,
```
gergosimonyi:
```suggestion
        hasher_defaults_overridden: bool = False,
```

?

authentik/events/signals.py
```diff
@@ -124,10 +124,14 @@ def on_password_changed(
...
     password: str | None = None,
     request: HttpRequest | None = None,
+    hasher_defaults_overridden: bool | None = None,
     **_,
 ):
     """Log password change"""
-    Event.new(EventAction.PASSWORD_SET).from_http(request, user=user)
+    context = {}
+    if hasher_defaults_overridden is not None:
+        context["hasher_defaults_overridden"] = hasher_defaults_overridden
```
gergosimonyi, line 133: Following the previous suggestion, not sure what we want. Either

```python
    context = {}
    if hasher_defaults_overridden:
        context["hasher_defaults_overridden"] = hasher_defaults_overridden
```

or


```python
    context = { hasher_defaults_overridden: hasher_defaults_overridden }
```

authentik/lib/validators.py
```diff
@@ -1,21 +1,129 @@
...
+        if "params" in decoded:
+            expected = hasher.params()
+            parameters.extend(
+                (
+                    (_("Variety"), decoded["variety"], f"argon2{expected.type.name.lower()}"),
+                    (_("Version"), decoded["version"], expected.version),
+                    (_("Hash length"), decoded["params"].hash_len, expected.hash_len),
+                )
+            )
+        return parameters
```
gergosimonyi: As far as I can tell, this code does what's expected, but cognitive complexity is too high (at least to my brain). Can we refactor somewhat to clearly tell a story?

Issue #21849, opened 2026-04-27
Swap OpenLayers with MapLibre + optional self hosted Protomaps


GirlBossRush:
In addition to MapLibre’s support for performant vector tiles, the ability to self host tile data is desirable for airgapped setups.

Ideally a fresh authentik instance can fallback to a public vector tile source, alternatively using a XYZ url provided by the server config. A `@goauthentik/tile-server’ package defining a Cloudflare worker compatible tile server provides a simple setup that can illustrate how this all comes together. Also included, a Dockerfile which can be included in our compose setup.

PR #24253, opened 2026-07-21
brands: add branding_map_tiles for the events map tile source

GirlBossRush:
### What does this PR change?

Adds a `branding_map_tiles` field to Brands: empty (the default) keeps the events map on its bundled basemap, while a `pmtiles://` archive URL or XYZ template points it at a self-hosted tile source. Includes the migration, schema, regenerated clients, the web-side brand plumbing, and a **Map tiles** input under **System** > **Brands** so the field is settable from the Admin interface rather than only via the API or a blueprint.

Groundwork for the events map refresh (refs #21849) — deployments should be able to bring their own basemap without touching code.

### How was this tested?

`make test authentik/brands` passes (18 tests). `make gen` reproduces `schema.yml` and the Go/Rust/TS clients with no drift, `ak makemigrations --check` reports nothing pending, and the migration graph has a single leaf. Web type-check, lint, and spellcheck pass. The default (empty) path is exercised end-to-end by the events map against a dev instance; a non-empty override has not yet been pointed at a live tile server.

### Linked issues

refs #21849


kensternberg-authentik (approved):
Hey, that was very clear and readable.  Thanks.

PR #24254, opened 2026-07-21
web, packages/geo, server/static: Air Gapped Maps Merge Branch

GirlBossRush:
```
 ╭┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄╮
 ┆ This PR is a merge branch! ┆
 ╰┄┄┄┄┄┄┄┄┄┬┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄╯

      　　　   ╱╲ ╱╲ 
      　   ╱￣(°㇀°⚟)￣╱╲
         ╱┌──じ㇀し───┐╲╱
          │          │
          │      ǝpoɔ│ ╱
          └─────┄───┄
``` 
---

- #24253 
- #24257
- #24258

---

## Rust changes

The bundled basemap is read with HTTP byte serving. Nothing currently guards that contract, so a later change to the static handler's compression could quietly break the map — which is the exact failure this PR originally set out to fix.

Adds Rust tests covering byte-range requests for static assets, and moves the static handler's compression layer behind a named constructor so a test can exercise the same configuration `build_router` builds.

- `ServeDir` serves ranges itself, returning a 206 with `Content-Range`.
- tower-http never compresses a response carrying `Content-Range`, so the gzip half is handled upstream.
- No ETag middleware survived the rewrite, so there is nothing to bypass.

### Linked issues

refs #21849


rissson (approved):

kensternberg-authentik (approved):
Pretty ordinary Axum-oriented testing.  Using `http_body` instead of `axum_body` is a good call; I've had that headache before.

rissson (approved):
just pushed some nits to the Rust code, but otherwise LGTM both on the Rust and Python side

rissson:
@GirlBossRush please merge main one more time before merging this

PR #24255, opened 2026-07-21
packages/geo: add @goauthentik/geo and the hexworld event map

GirlBossRush:
### What does this PR change?

A standalone Lit + MapLibre package for the events map: a tilted globe that bins events into H3 cells and raises them as action-colored pie columns, over a hex-grid basemap bundled as a PMTiles archive. Zoom bands cross-fade into each other and columns animate between re-bins. Ships the committed archive (22 MB, Natural Earth land + OSM-derived labels), the generator CLI to rebuild it, and node tests for the geometry, styling, and tiling plan.

Stacked on #24253; will retarget to `main` when it merges.

The map draws with no tile server and no external requests, so it works in air-gapped deployments — the main constraint that ruled out hosted basemaps.

### How was this tested?

56 node tests across binning, wedge geometry, style building, label handling, and the tiling plan; visual verification in the admin UI across zoom levels and both themes.

### Linked issues

refs #21849


packages/geo/scripts/build-hexworld.ts
```diff
@@ -0,0 +1,413 @@
...
+            ...borderPieces,
+            `${outDir}/places.pmtiles`,
+        ]);
+    }
+
+    return plan;
+}
+
+// Deliberately synchronous. An earlier async createWriteStream + .end() version
+// looked fine but bit the places geojsonl step: the following spawnSync
```
kensternberg-authentik, line 182: "... but bit the places geojsonl step" is hard to parse after only one cup of coffee.

kensternberg-authentik (approved):
It's just huge, and other than building it and testing to make sure it looks right, all I can say is that the code is clean and well-documented.

PR #24258, opened 2026-07-21
website/docs: document the hexworld event map

GirlBossRush:
### What does this PR change?

Documents the bundled event map — what the columns and colors mean, how clicking filters the list, and how to point a brand at a self-hosted basemap via the **Map tiles** field. The section is marked `:ak-enterprise`, matching the license gate on the map in the events view, and carries an explicit `{#event-map}` anchor so the air-gapped page can link straight to it.

Also removes the OSM tile server from the air-gapped outbound-connections list, since the bundled basemap makes no outbound requests, and replaces it with a note on what a custom basemap implies for an air-gapped deployment.

Keeps the events and air-gapped documentation in step with the map swap (refs #21849). The old outbound-connections entry told air-gapped operators to expect a call to `tile.openstreetmap.org` that no longer happens.

### How was this tested?

`make docs` builds clean and spellcheck passes. Verified in the built output that the heading renders `id=event-map` (so the air-gapped cross-link resolves) and that the enterprise badge is emitted.

### Linked issues

refs #21849

### DIFF
--- a/website/docs/sys-mgmt/events/index.md
+++ b/website/docs/sys-mgmt/events/index.md
@@ -23,3 +23,15 @@ The event retention setting is configured in the **System > Settings** area of t
 If you want to forward these events to another application, forward the log output of all authentik containers. Every event creation is logged with the log level "info". For this configuration, it is also recommended to set the internal retention time period to a short time frame (for example, `days=1`).
 
 If you want to forward authentik events to another system, see [Log forwarding](./log-forwarding.mdx).
+
+## Event map :ak-enterprise {#event-map}
+
+The events overview renders login and audit event locations on a built-in map. By default authentik uses a bundled "hexworld" basemap: land drawn as a hexagonal grid at three zoom bands, with country, region, and locality labels layered on top. The archive ships as part of the web bundle, so the map draws with no tile server, no external service, and no online request, and it works in air-gapped deployments.
+
+Event locations come from GeoIP and are city-accurate at best. The hex bands are sized to match: about 1000 km at world view, 400 km mid-range, and 130 km at closest zoom. Events raise their hex cell into a column — the fuller a cell, the taller the column — on a tilted globe. When a cell contains more than one kind of event, the column splits into pie wedges colored by kind: logins (green), failed logins (red), logouts (blue), application authorizations (purple), and everything else (gray). Clicking a column filters the event list below to that cell's events. Custom basemaps configured under **System** > **Brands** show the same event columns over the configured tiles.
+
+Label data is derived from [OpenStreetMap](https://openstreetmap.org/copyright) (© OpenStreetMap contributors, ODbL). Land shapes are from Natural Earth (public domain).
+
+### Using a conventional basemap instead
+
+To replace the hexworld default with a conventional vector basemap, set a tile URL template on the brand under **System** > **Brands** — the **Map tiles** field takes either a `pmtiles://` archive URL or an XYZ endpoint like `/tiles/{z}/{x}/{y}.mvt`. Any non-empty value overrides the bundled default; leave it blank to keep hexworld.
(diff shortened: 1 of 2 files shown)


website/docs/sys-mgmt/events/index.md
```diff
@@ -23,3 +23,15 @@ The event retention setting is configured in the **System > Settings** area of t
 If you want to forward these events to another application, forward the log output of all authentik containers. Every event creation is logged with the log level "info". For this configuration, it is also recommended to set the internal retention time period to a short time frame (for example, `days=1`).
 
 If you want to forward authentik events to another system, see [Log forwarding](./log-forwarding.mdx).
+
+## Event map
+
+The events overview renders login and audit event locations on a built-in map. By default authentik uses a bundled "hexworld" basemap: land drawn as a hexagonal grid at three zoom bands, with country, region, and locality labels layered on top. The archive ships as part of the web bundle, so the map draws with no tile server, no external service, and no online request, and it works in air-gapped deployments.
```
dewi-tik, line 29:
```suggestion
The events overview renders login and audit event locations on an integrated map. By default, authentik uses the bundled "hexworld" basemap: land is rendered as a hexagonal grid across three zoom levels, with country, region, and locality labels overlaid. Because the map archive is included in authentik, it requires no external services or network requests, and therefore also works in air-gapped deployments.
```

website/docs/sys-mgmt/events/index.md
```diff
@@ -23,3 +23,15 @@ The event retention setting is configured in the **System > Settings** area of t
...
+
+Event locations come from GeoIP and are city-accurate at best. The hex bands are sized to match: about 1000 km at world view, 400 km mid-range, and 130 km at closest zoom. Events raise their hex cell into a column — the fuller a cell, the taller the column — on a tilted globe. When a cell contains more than one kind of event, the column splits into pie wedges colored by kind: logins (green), failed logins (red), logouts (blue), application authorizations (purple), and everything else (gray). Clicking a column filters the event list below to that cell's events. Custom basemaps configured under **System > Brands** show the same event columns over the configured tiles.
```
dewi-tik:
```suggestion
Event locations are derived from GeoIP data and are accurate to city level at best. The hex-grid sizes reflect that limitation: approximately 1,000 km at the world view, 400 km at mid-range, and 130 km at the closest zoom level. Events appear as raised columns on a tilted globe: the more events a cell contains, the taller its column.

If a cell contains multiple event types, its column is divided into colored pie segments:

- **Logins**: green
- **Failed logins**: red
- **Logouts**: blue
- **Application authorizations**: purple
- **All other events**: grey

Select a column to filter the event list below to events from that cell. Custom basemaps configured under **System** > **Brands** display the same event columns over the configured tiles.
```

website/docs/sys-mgmt/events/index.md
```diff
@@ -23,3 +23,15 @@ The event retention setting is configured in the **System > Settings** area of t
...
+
+Label data is derived from [OpenStreetMap](https://openstreetmap.org/copyright) (© OpenStreetMap contributors, ODbL). Land shapes are from Natural Earth (public domain).
+
+### Using a conventional basemap instead
+
+To replace the hexworld default with a conventional vector basemap, set a tile URL template on the brand under **System > Brands** — the **Map tiles** field takes either a `pmtiles://` archive URL or an XYZ endpoint like `/tiles/{z}/{x}/{y}.mvt`. Any non-empty value overrides the bundled default; leave it blank to keep hexworld.
```
dewi-tik:
```suggestion
To replace the default hexworld basemap with a conventional vector basemap, set the **Map tiles** field for the brand under **System > Brands**. This field takes either a `pmtiles://` archive URL or an XYZ endpoint like `/tiles/{z}/{x}/{y}.mvt`. Any non-empty value overrides the bundled default; leave it blank to keep the hexworld basemap.
```

dewi-tik (approved):

dominic-r (approved):

PR #24265, opened 2026-07-21
web/maint: upgrade ak-label

kensternberg-authentik:
web/maint: upgrade ak-label

## What

This commit upgrades ak-label with the following improvements:

- Upgrades our code to use the **Patternfly 5** CSS with our changes
- Provides a large number of semantic `status` synonyms
- Fixes the “compact” variant
- Fixes the “outline” variant

What the current "status" tokens look like (standard mode, so not "compact" or "outline"):

image

There is an ad-hoc `font-awesome.css` (now in the bare `Icons_impl` folder) to make the font-awesome icons accessible to the labels. This mirrors the previous solutions, but it should not be the permanent solution.

I have added the FontAwesome font files to the list of assets we have vendored.

I have provided a comprehensive Storybook story.

I have updated every location in our code where `<ak-label>` is used.

## Checklist

-   [🚧] The project has been linted, built, and tested (`make all`)
-   [🚧 The documentation has been updated and formatted (`make docs`)


web/src/admin/events/DataExportListPage.ts
```diff
@@ -84,8 +83,8 @@ export class DataExportListPage extends TablePage<DataExport> {
             >`,
             Timestamp(item.requestedOn),
             html`${item.completed
-                ? html`<ak-label color=${PFColor.Green}>${msg("Finished")}</ak-label>`
-                : html`<ak-label color=${PFColor.Gray}>${msg("Queued")}</ak-label>`}`,
+                ? html`<ak-label status="success">${msg("Finished")}</ak-label>`
+                : html`<ak-label>${msg("Queued")}</ak-label>`}`,
```
kensternberg-authentik, line 87: `gray` is the default, and there is no corresponding CSS to be triggered, so it doesn't need to be specified.

PR #24273, opened 2026-07-21
web/admin: drop misleading delete consequences from user activation review

### LINKED ISSUE #23777: Associations not removed from user upon deactivation
### Describe the bug

When deactivating a user, there is now a modal popup, that informs you about the consequences of deactivating that user. This usually includes removing them from groups or deleting MFA devices. When confirming the deactivation the user gets deactivated, but the outlined consequences are not executed (i.e. the user is still in those groups, MFA devices not deleted). However, the session

Also, the modal shows every currently associated group twice.

### How to reproduce

1. Go to the admin area
2. Select an active user with groups, an MFA device and an active session
3. Click 'Deactivate'
4. See _Review User Deactivation modal_
5. Notice every group is listed twice
6. Click _Deactivate User_
7. Click on the _Groups_ tab
8. See the groups are still there
9. Click on the _Credentials / Tokens_ tab
10. See that session is gone, but MFA device is still there

### Expected behavior

Upon deactivation the connection to the groups should have been removed and the MFA devices deleted (plus any other consequences that may be listed in the deactivation review modal).

### Screenshots

Image

### Additional context

_No response_

### Deployment Method

Docker

### Version

2026.5.3

### Relevant log output

```shell

```

### LINKED ISSUE #23778: User activation popup claims activation would delete something
### Describe the bug

When activating a user, the popup claims that connections to groups or MFA devices would be deleted. This does not actually happen upon activation and is probably only a display logic error from the related deactivation review popup. However, a consequence of 'connection gets created' would also not make sense, as the user is already part of those groups.

### How to reproduce

1. Go to admin area
2. Select deactivated user, which is part of at least one group
3. Click Activate
4. See popup that claims consequences of deleting group associations or MFA devices
5. Click Activate user
6. See that user is active
7. See that none of the listed consequences actually happened (e.g. user is still in groups)

### Expected behavior

The review popup should inform an admin what consequences the activation will have. Group memberships might be interesting, but if they already exist, the activation does not change anything. More relevant would be the effective access resulting from these memberships (i.e. which applications can be accessed).

### Screenshots

Image

### Additional context

_No response_

### Deployment Method

Docker

### Version

2026.5.3

### Relevant log output

```shell

```

melizeche:
## Details

The activate/deactivate review reused the destructive form's used-by table, whose "Consequence" column claims associated objects "will be deleted". Toggling `is_active` deletes nothing, so that column was false.

~Add a `showConsequences` opt-out to the shared form and set it on the user activation toggle, so the review still lists the user's associations but without the deletion framing.~

now just shows the user to be activated/deactivated

Before:
image

poposal:
image

final:
image

Closes #23778
Closes #23777

---

## Checklist

### DIFF
--- a/web/src/admin/users/UserActiveForm.ts
+++ b/web/src/admin/users/UserActiveForm.ts
@@ -9,7 +9,7 @@ import { DestructiveModelForm } from "#elements/forms/DestructiveModelForm";
 import { WithLocale } from "#elements/mixins/locale";
 import { SlottedTemplateResult } from "#elements/types";
 
-import { CoreApi, User } from "@goauthentik/api";
+import { CoreApi, UsedBy, User } from "@goauthentik/api";
 
 import { str } from "@lit/localize";
 import { msg } from "@lit/localize/init/install";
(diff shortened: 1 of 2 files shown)


melizeche:
first I tried showing no related objects but the empty dialog just looked wrong imho

melizeche:
lastest:
image

GirlBossRush (approved):

PR #24290, opened 2026-07-22
events: add email notifications to users about security events on their account

### LINKED ISSUE #23124: Email notifications for user events
Notify user via email when certain events happen on their account:

- Password changed
- MFA device added/removed
- Attempted impossible travel login
- Account is locked
- Account expires
- Welcome email on account creation

atereshkin:
### What does this PR change?

Adds the ability to email users about security-relevant events on their own account, and ships a set of default (disabled) notification rules that cover the events requested in #23124.

**Notification engine**

- New `NotificationRule.destination_event_subject` flag: sends the notification to the user *affected* by the event, as recorded in the event's new `subject_uuid` context field. This is distinct from the existing `destination_event_user` (the actor) — e.g. when an administrator resets a user's password, the email goes to the user whose password was changed, not the administrator.
- New `NotificationRule.enabled` flag, so default rules can ship disabled and be turned on per-install. Disabled rules are skipped at dispatch time.
- Email transports can now render a per-template subject line: a `<template>_subject.txt` shipped next to the body template is rendered with the event context in the recipient's locale, falling back to the generic title as before.

**New events**

- `user_created` — emitted when a user account is created.
- `mfa_device_added` / `mfa_device_removed` — emitted generically from the `authenticator` stage's device signals. The static authenticator's ad-hoc `static_authenticator_disable` event is removed in favor of these.
- `login_blocked` — emitted when a login is prevented by policies on the user login stage (via `ReevaluateMarker`). GeoIP policy results now carry machine-readable `reasons` (e.g. `impossible_travel`), so rules can match on the specific block cause.
- Events gain an `origin` context value (`http`, `blueprint`, `source_sync`, …) via a new `event_origin()` context manager, so rules can distinguish how an account or password change came about. `password_set` events additionally record whether the password hash was cached from an external source.

**Default rules** (new `blueprints/default/events-user-security-notifications.yaml`, all **disabled by default**)

- `default-notify-user-password-change` — password changed (excluding hashes cached from source logins and blueprint-set passwords)
- `default-notify-user-mfa-device-change` — MFA device added or removed
- `default-notify-user-impossible-travel` — login blocked due to an impossible travel pattern
- `default-notify-user-account-lockdown` — account locked down
- `default-notify-user-welcome` — welcome email on account creation (excluding service accounts and source-synced users)
- `default-notify-user-account-offboarding` — "account expired" email on scheduled onboarding

The security rules use a new **Security Notification** email template; the welcome rule uses a new **Account Welcome** template.

**Other changes**

- GDPR cleanup on user deletion now also removes events where the deleted user is the subject (`context.subject_uuid`), not only the actor.
- Admin UI: notification rule form and list support the new *Enabled* and *Send notification to affected user* settings.
- Docs: new [User security notifications](website/docs/sys-mgmt/events/user-security-notifications.md) page, plus updates to event actions, notifications, transports, email stage docs, and the 2026.8 release notes.
- The user_pk context field that #23624 introduced on user_offboarded is removed — it is fully superseded by subject.pk, has no consumers, and offboarding hasn't shipped in a release yet, so this is the last point at which it can be dropped without a breaking change.

Per feature request described in #23124

### How was this tested?

- Unit tests added/extended for: the new event actions and `subject_uuid` routing, notification rule `enabled` handling, subject-line template rendering in email transports, `login_blocked` emission from `ReevaluateMarker`, MFA device signals, account lockdown subject routing, GDPR cleanup of subject events, and origin filtering of welcome/password events. Plus "end-to-end" (request-to-email) tests for all new notification types.
- Manually verified the default blueprint applies cleanly, the notifications are sent and the emails render (HTML + plaintext).

### Linked issues

Closes #23124.

---

## Checklist


authentik/events/signals.py
```diff
@@ -86,6 +87,22 @@ def on_user_write(sender, request: HttpRequest, user: User, data: dict[str, Any]
...
+@receiver(post_save, sender=User)
+def on_user_created(sender, instance: User, created: bool, raw: bool, **_):
+    """Emit a dedicated event when a user account is created, labelled with the
+    channel that created it ("http" for admin interface/API/enrollment flows,
+    "blueprint", "source_sync", ...) so notification rules can match on it."""
+    if raw or not created:
+        return
+    Event.new(
+        EventAction.USER_CREATED,
+        subject_uuid=instance.uuid,
```
BeryJu: Since this standardises the subject field in an event, I'd prefer `subject=instance` which will be cleaned up the cleanse_dict

authentik/events/models.py
```diff
@@ -92,12 +95,17 @@ class EventAction(models.TextChoices):
 
     LOGIN = "login"
     LOGIN_FAILED = "login_failed"
+    LOGIN_BLOCKED = "login_blocked"
     LOGOUT = "logout"
 
+    USER_CREATED = "user_created"
```
BeryJu, line 101: I'd rather not have this in general but yeah from an admin perspective it probably makes sense to have these compared to just filtering model_created

authentik/events/models.py
```diff
@@ -92,12 +95,17 @@ class EventAction(models.TextChoices):
...
     USER_WRITE = "user_write"
     SUSPICIOUS_REQUEST = "suspicious_request"
     PASSWORD_SET = "password_set"  # noqa # nosec
 
+    MFA_DEVICE_ADDED = "mfa_device_added"
+    MFA_DEVICE_REMOVED = "mfa_device_removed"
```
BeryJu: should be _created and _deleted to match the other ones

BeryJu: For both of these I'm tempted to say we should use something like `user.created` or `directory.user.created`, this is a discussion point thats come up to basically have some way of preventing us from having a giant list of actions, we could have it nested

authentik/events/models.py
```diff
@@ -562,6 +581,21 @@ def send_webhook_slack(self, notification: Notification) -> list[str]:
             response.text,
         ]
 
+    def render_email_subject(self, context: dict[str, Any], language: str) -> str:
```
BeryJu, line 590: I forget, do we do this anywhere else? it seems familiar

authentik/events/models.py
```diff
@@ -674,6 +715,10 @@ class NotificationRule(TasksModel, SerializerModel, PolicyBindingModel):
     """Decide when to create a Notification based on policies attached to this object."""
 
     name = models.TextField(unique=True)
+    enabled = models.BooleanField(
```
BeryJu, line 724: In the migration this should probably be set dynamically based on the previous rules (destination_group or destination_event_user set)

authentik/events/signals.py
```diff
@@ -122,7 +139,12 @@ def on_password_changed(
     **_,
 ):
     """Log password change"""
-    Event.new(EventAction.PASSWORD_SET).from_http(request, user=user)
+    Event.new(
+        EventAction.PASSWORD_SET,
+        subject_uuid=user.uuid,
+        synced_from_source=isinstance(sender, Source),
+        origin=get_event_origin(),
```
BeryJu, line 151: If this is being added everywhere we could just have it in Event.new?

authentik/flows/tests/test_markers.py
```diff
@@ -0,0 +1,112 @@
...
+
+    def setUp(self):
+        self.request_factory = RequestFactory()
+        self.flow = create_test_flow()
+        self.user = create_test_user()
+
+    def _request(self, user=None):
+        request = self.request_factory.get("/")
+        request.user = user or AnonymousUser()
+        return request
```
BeryJu, line 31: with the requestfactory from authentik.core.tests.utils you can just pass a `user=` argument so this isn't needed

authentik/flows/tests/test_markers.py
```diff
@@ -0,0 +1,112 @@
...
+
+    def _binding(self, stage):
+        return FlowStageBinding.objects.create(
+            target=self.flow, stage=stage, order=0, re_evaluate_policies=True
+        )
+
+    def _plan(self, **context):
+        plan = FlowPlan(flow_pk=self.flow.pk.hex)
+        plan.context.update(context)
+        return plan
```
BeryJu, line 41: These tests should use the flow planner imo

authentik/flows/markers.py
```diff
@@ -72,4 +72,15 @@ def process(
...
+        from authentik.stages.user_login.models import UserLoginStage
+
+        if isinstance(binding.stage, UserLoginStage):
+            Event.new(
+                EventAction.LOGIN_BLOCKED,
+                message="; ".join(str(message) for message in result.messages)
+                or "Login blocked by policy.",
+                reasons=sorted(result.reasons),
+                subject_uuid=user.uuid if user.is_authenticated else None,
+            ).from_http(http_request, user)
```
BeryJu, line 85: yeah not a fan of this, this should ideally be in the UserLoginStage

authentik/stages/authenticator/apps.py
```diff
@@ -10,3 +13,18 @@ class AuthentikStageAuthenticatorConfig(ManagedAppConfig):
     label = "authentik_stages_authenticator"
     verbose_name = "authentik Stages.Authenticator"
     default = True
+
+    def ready(self):
+        from .signals import device_post_save_event, device_pre_delete_event
```
BeryJu: should be absolute import

authentik/stages/authenticator_static/signals.py
```diff
@@ -1,16 +0,0 @@
...
-from authentik.stages.authenticator_static.models import StaticDevice
-
-
-@receiver(pre_delete, sender=StaticDevice)
-def pre_delete_event(sender, instance: StaticDevice, **_):
-    """Create event before deleting Static Devices"""
-    # Create event with email notification
-    event = Event.new("static_authenticator_disable", message="User disabled Static OTP Tokens.")
-    event.set_user(instance.user)
-    event.save()
```
BeryJu, line 16: wtf I dont remember this code at all

authentik/stages/email/templates/email/security_notification.html
```diff
@@ -0,0 +1,94 @@
...
+      {% if "impossible_travel" in event_context.reasons %}
+      {% trans "A sign-in from an unusual location was blocked" %}
+      {% else %}
+      {% trans "A sign-in to your account was blocked" %}
+      {% endif %}
+      {% elif event_action == "user_write" and event_context.action_id == "account_lockdown" %}
+      {% trans "Your account has been locked" %}
+      {% else %}
+      {% trans "Security notification" %}
+      {% endif %}
```
BeryJu: should probably have some indentation

authentik/events/signals.py
atereshkin: Done.

authentik/events/models.py
atereshkin: Done.

authentik/events/models.py
atereshkin: This is very similar to what `authentik.stages.email.utils.TemplateEmailMessage` does to render html and text email parts, although I couldn't find an elegant way to unify the two.

authentik/events/models.py
atereshkin: That was my first instinct but then I realized that those rules will still be inactive, will still be marked as such in the UI (no destination set gets a special "No recipients" designation in the Enabled column) so the only difference that migration would make is that the admin will have to both set a destination and enable the rule. If you still think that's the right thing to do, just say a word and I'll update that.

authentik/flows/markers.py
atereshkin: Yep, but the problem is UserLoginStage is never visited if the policies decide against that. The option you've suggested on Slack (user is unauthenticated && pending user in flow plan && flow designation is set to authentication) works and seems better than the above, but it still requires recording policy failure reason in ReevaluateMarker. Should I update the PR with that approach?

authentik/stages/authenticator/apps.py
atereshkin: Fixed.

authentik/stages/authenticator_static/signals.py
atereshkin: Doesn't seem to be used anywhere either. But I've included a mention in the release notes just in case.

authentik/stages/email/templates/email/security_notification.html
atereshkin: Fixed.

authentik/events/models.py
atereshkin: Do you think this is something we should do right away, or just thinking out loud? Seems like a total switch to hierarchical system would be a huge and disruptive migration, while having just some of enum values use new notation doesn't solve the original problem if I understand it correctly. 

Also, what if we kept the string constants as they are, but introduced a hierarchy at the code level and then used it to make the UI more ergonomic? Meanwhile the database and all public contracts would remain untouched.

atereshkin:
Added account expiration (offboarding) notification and updated the description. The noteworthy part being:
>The user_pk context field that #23624 introduced on user_offboarded is removed — it is fully superseded by subject.pk, has no consumers, and offboarding hasn't shipped in a release yet, so this is the last point at which it can be dropped without a breaking change.

authentik/events/models.py
atereshkin: @BeryJu I've made a quick and dirty prototype just to illustrate what I was talking about in the previous comment: https://github.com/goauthentik/authentik/compare/main...atereshkin:authentik:event-hierarchy

What do you think about that general direction?

PR #24292, opened 2026-07-22
core: object attributes - minor fixes

BeryJu:
- fix incorrect default group for address
- fix loading for attributes whose key is dotted (like the default settings.locale)

### DIFF
--- a/web/src/admin/object-attributes/renderAttributes.ts
+++ b/web/src/admin/object-attributes/renderAttributes.ts
@@ -29,20 +29,21 @@ export type AttributesMixin = {
     attributes?: Record<string, unknown>;
 };
 
-export function getPath(
-    root: Record<string, unknown>,
+/**
+ * Given a path of tokens with a separator, walk through a nested object to return whatever is at that path, or a default value if the path doesn't exist
+ */
+export function getValueAtPath(
     path: string,
-    defaultValue: unknown = null,
-    sep = ".",
+    from: Record<string, unknown>,
+    failure: unknown = null,
+    separator = ".",
 ): unknown {
-    // Recursively walk through `root`, checking each part of `path` separated by `sep`.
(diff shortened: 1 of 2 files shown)


web/src/admin/object-attributes/renderAttributes.ts
```diff
@@ -29,6 +29,25 @@ export type AttributesMixin = {
     attributes?: Record<string, unknown>;
 };
 
+export function getPath(
+    root: Record<string, unknown>,
+    path: string,
+    defaultValue: unknown = null,
+    sep = ".",
+): unknown {
+    // Recursively walk through `root`, checking each part of `path` separated by `sep`.
```
kensternberg-authentik: It's iterative, not recursive.  If it was recursive, it would look more like:

```
    const next = (path: Array<string>, walk: Record<string, unknown>): Record<string, unknown> | unknown =>
        Array.length === 0
            ? walk
            : typeof walk === "object" && walk !== null && path[0] in walk
              ? next(path.slice(1), (walk as Record<string, unknown>)[path[0]])
              : defaultValue;

    return next(path.split(sep), root);
```

Besides, the comment tells you what it does functionally, which is pretty clear in the code.  The real question is, "What is this _for_?"  Something like:

"Given a path of tokens with a separator, walk through a nested object to return whatever is at that path, or a default value if the path doesn't exist."

Or rename it: `getValueAtPath(path: string, from: Record<string, unknown>, failure: unknown=null, separator=".")`

kensternberg-authentik (approved):
Nothing functionally wrong with it, but...

web/src/admin/object-attributes/renderAttributes.ts
BeryJu: Yeah the comment is copied straight from the python version of this

rissson (approved):

PR #24302, opened 2026-07-22
stages/password: lock password login after failed attempts

### LINKED ISSUE #23572: Lock account after x incorrect passwords
Add an option to the password stage to lock an account after x number of incorrect passwords.

### LINKED ISSUE #25264: Lock the account after N failed password attempts
## What

lock the account after N incorrect passwords (lockout-attempts policy / password-stage option). Also: lockout/reputation: reset reputation to 0 on successful login, vs lock account after N failures. Same brute-force-mitigation feature area.

## Want

- Lock the account after N incorrect passwords (lockout-attempts policy / password-stage option)
- Related lockout/reputation: reset reputation to 0 on successful login, vs lock account after N failures

## Ideas this came from

- [#11762 Lockout attempts policy](https://github.com/goauthentik/authentik/issues/11762) — To reduce the effectiveness of attempts to compromise a user account, things like preventing an account from logging in if enough failed login attempts have bee
- [#20148 Option to automatically set reputation score to 0 upon successful login](https://github.com/goauthentik/authentik/issues/20148) — An option/flag to automatically set reputation score back to 0 for a user/IP upon successful login. Rather than just increasing reputation score by one
- [#23572 Lock account after x incorrect passwords](https://github.com/goauthentik/authentik/issues/23572) — Add an option to the password stage to lock an account after x number of incorrect passwords

dominic-r:
Closes: https://github.com/goauthentik/authentik/issues/23572
Discussed in: https://authentiksecurity.slack.com/archives/C08ABMS9TRP/p1784736557410369
Requested by: https://github.com/goauthentik/internal-customer-ref/issues/13

Some screenshots:

image

image

image

Closes #25264


web/src/admin/stages/password/PasswordStageForm.ts
```diff
@@ -46,7 +49,91 @@ export class PasswordStageForm extends BaseStageForm<PasswordStage> {
...
+        }
+
+        return html`<ak-form-element-horizontal
+                label=${msg("Failed attempts before lockout", {
+                    id: "password-stage.lockout-threshold.label",
+                })}
+                required
+                name="failedAttemptsBeforeLockout"
+            >
+                <input
```
kensternberg-authentik: We have a whole suite of inputs in the `components` section that work with forms to support this sort of thing. I'm curious to know why you didn't use them here.

web/src/admin/stages/password/PasswordStageForm.ts
```diff
@@ -46,7 +49,91 @@ export class PasswordStageForm extends BaseStageForm<PasswordStage> {
...
+                name="showLastAttemptWarning"
+                label=${msg("Show last-attempt warning", {
+                    id: "password-stage.last-attempt-warning.label",
+                })}
+                ?checked=${this.instance?.showLastAttemptWarning ?? false}
+                help=${msg("Show a warning when the user has one password attempt remaining.", {
+                    id: "password-stage.last-attempt-warning.description",
+                })}
+            ></ak-switch-input>
+            <ak-form-element-horizontal
```
kensternberg-authentik: Consider showing the custom message inputs _only_ when the `show` is active.

An admin might fill in this field assuming that just doing that is all that's needed, and then wonder why it doesn't work when they test it out.  By making the custom message field show up only when the `show` is active, you reinforce that "turning it on" and "customizing it" are two very different acts.

web/src/admin/stages/password/PasswordStageForm.ts
```diff
@@ -46,37 +52,89 @@ export class PasswordStageForm extends BaseStageForm<PasswordStage> {
...
-            },
-            {
-                name: BackendsEnum.AuthentikSourcesKerberosAuthKerberosBackend,
-                label: msg("User database + Kerberos password"),
-            },
-        ];
+    protected renderLockoutSettings(): SlottedTemplateResult {
+        if (!this.hasEnterpriseLicense) {
+            return null;
+        }
```
kensternberg-authentik: Possibly controversial opinion: maybe this section should be greyed out instead.

Not controversial opinion: `aki(StagesApi).stagesPasswordUpdate({ stageUuid, passwordStageRequest }),` is a full update, not a patch.  

I think, if I'm reading the code correctly, that should a licensed admin update a PasswordStage in good faith, the license expires, and then they update it again, _all_ of these fields will get sent back as `undefined`, and the server will reset the threshold to zero and delete all custom messages.

web/src/admin/users/UserListPage.ts
```diff
@@ -361,6 +363,11 @@ export class UserListPage extends WithLicenseSummary(
                 <dd class="pf-c-description-list__description">
                     <div class="pf-c-description-list__text">
                         ${ToggleUserActivationButton(item)}
+                        ${this.hasEnterpriseLicense
```
kensternberg-authentik: `instance.unlock_password_login` doesn't have an `is_enterprise` flag on it in the server-side of the patch, so it shouldn't here. Again, the problem is license expiration: with the `hasEnterpriseLicense` condition, if the license expires the conditional will prevent admins from unlocking users who have been locked in the meantime.

There's a big difference between "You've lost important features because your license lapsed" and "We're taking hostages until you pay us."  😉

dominic-r:
Marking as draft as we are gonna be doing a few changes first to make this cleaner. See: https://app.notion.com/p/authentiksecurity/2026-11-Lock-out-after-X-Failed-Attempts-Dependants-Considerations-way-of-doing-it-3a6aee05b24e80af8648c300bc7228e8

GirlBossRush (changes requested):
Looks good. Needs E2Etests.

web/src/admin/users/UserInfoCard.ts
```diff
@@ -74,6 +75,12 @@ export class UserInfoCard extends AKElement {
             </button>
 
             ${ToggleUserActivationButton(user, { className: "pf-m-block" })}
+            ${this.hasEnterpriseLicense
```
kensternberg-authentik: Same issue as the other mention about the lock button; the server doesn't require this be behind a "is enterprise" check, and we shouldn't either; otherwise, users locked during a licensed period will be unlockable if the license expires.

web/src/admin/users/UserPasswordLoginLockForm.ts
```diff
@@ -0,0 +1,129 @@
...
+import { CoreApi, User, UserTypeEnum } from "@goauthentik/api";
+
+import { msg, str } from "@lit/localize";
+import { html } from "lit";
+import { guard } from "lit-html/directives/guard.js";
+import { customElement } from "lit/decorators.js";
+
+@customElement("ak-user-password-login-lock-toggle-form")
+export class UserPasswordLoginLockToggleForm extends WithLocale(DestructiveModelForm<User>) {
+    protected coreAPI = aki(CoreApi);
```
kensternberg-authentik: This was a good habit, but now it's mostly a neutral one, since `aki()` caches it's connections.  The CoreAPI you get here is the same one shared across the whole UI.  It's not really necessary to cache it on a per-instance basis.

**Don't change this.** It's fine. I just want you to be aware of the caching so you don't feel compelled to do it in the future.

GirlBossRush (approved):

web/src/admin/users/UserPasswordLoginLockForm.ts
```diff
@@ -0,0 +1,129 @@
...
+    user: User,
+    { className = "", currentUserPk }: ToggleUserPasswordLoginLockButtonProps = {},
+): SlottedTemplateResult {
+    const locked = !!user.passwordLoginLockedAt;
+
+    const shouldRender =
+        (!user.isActive ||
+            user.type === UserTypeEnum.InternalServiceAccount ||
+            user.pk === currentUserPk) &&
+        !locked;
```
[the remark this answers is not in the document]
GirlBossRush: @dominic-r

dominic-r:
Creating stack for this

PR #24318, opened 2026-07-23
website/integrations: add m365 ws-fed guide

dewi-tik:
### What does this PR change?

Add WS-Fed integration doc for M365. 
Update SAML doc for M365. 
Add redirect

2026.8 adds SAML1.1 support to the WS-Fed provider which is required by Entra ID

### How was this tested?

Against Entra ID

---

## Checklist


BeryJu (commented):
should probably add a note about why you'd want to use either one of these

dominic-r (approved):

PR #24319, opened 2026-07-23
ci: add working dir prefix to file paths

emilburzo:


rissson (approved):

rissson:
@emilburzo does this need backporting?

emilburzo:
@rissson yes, but just one line is needed. let's see how it goes

PR #24320, opened 2026-07-23
root: rewrite server in rust

rissson:
### Linked issues

---

## Checklist

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/.github/workflows/ci-main.yml
+++ b/.github/workflows/ci-main.yml
@@ -105,6 +105,9 @@ jobs:
         uses: ./.github/actions/setup
         with:
           dependencies: ${{ matrix.deps }}
+      - name: Create required files
+        run: |
+          touch web/robots.txt web/security.txt web/dist/standalone/loading/startup.html
       - name: run job
         run: make ci-lint-${{ matrix.job }}
   test-gen:
@@ -437,6 +440,9 @@ jobs:
       - name: Prepare database
         run: |
           uv run make migrate
+      - name: Create required files
+        run: |
+          touch web/robots.txt web/security.txt web/dist/standalone/loading/startup.html
       - name: run tests
         run: |
           cargo llvm-cov --no-report nextest --workspace
--- a/.github/workflows/ci-main.yml
+++ b/.github/workflows/ci-main.yml
@@ -108,7 +108,9 @@ jobs:
       - name: Create required files
         run: |
           mkdir -p web/dist/standalone/loading
-          touch web/robots.txt web/security.txt web/dist/standalone/loading/startup.html
+          for f in web/robots.txt web/security.txt web/dist/standalone/loading/startup.html; do
+            echo empty > "$f"
+          done
       - name: run job
         run: make ci-lint-${{ matrix.job }}
   test-gen:
@@ -444,7 +446,9 @@ jobs:
       - name: Create required files
         run: |
           mkdir -p web/dist/standalone/loading
-          touch web/robots.txt web/security.txt web/dist/standalone/loading/startup.html
+          for f in web/robots.txt web/security.txt web/dist/standalone/loading/startup.html; do
+            echo empty > "$f"
+          done
       - name: run tests
         run: |
           cargo llvm-cov --no-report nextest --workspace
--- a/.github/workflows/ci-main.yml
+++ b/.github/workflows/ci-main.yml
@@ -107,6 +107,7 @@ jobs:
           dependencies: ${{ matrix.deps }}
       - name: Create required files
         run: |
+          mkdir -p web/dist/standalone/loading
           touch web/robots.txt web/security.txt web/dist/standalone/loading/startup.html
       - name: run job
         run: make ci-lint-${{ matrix.job }}
@@ -442,6 +443,7 @@ jobs:
           uv run make migrate
       - name: Create required files
         run: |
+          mkdir -p web/dist/standalone/loading
           touch web/robots.txt web/security.txt web/dist/standalone/loading/startup.html
       - name: run tests
         run: |
--- a/.github/workflows/ci-main.yml
+++ b/.github/workflows/ci-main.yml
@@ -433,7 +433,10 @@ jobs:
       - name: Setup authentik env
         uses: ./.github/actions/setup
         with:
-          dependencies: rust,runtime
+          dependencies: system,python,rust,runtime
+      - name: Prepare database
+        run: |
+          uv run make migrate
       - name: run tests
         run: |
           cargo llvm-cov --no-report nextest --workspace
--- a/authentik/root/monitoring.py
+++ b/authentik/root/monitoring.py
@@ -5,18 +5,17 @@ from django.db.utils import OperationalError
 from django.dispatch import Signal
 from django.http import HttpRequest, HttpResponse
 from django.views import View
-from django_prometheus.exports import ExportToDjangoView
 
 monitoring_set = Signal()
 
 
 class MetricsView(View):
-    """Wrapper around ExportToDjangoView with authentication, accessed by the authentik router"""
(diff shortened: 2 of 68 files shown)


.github/workflows/ci-main.yml
```diff
@@ -101,6 +101,12 @@ jobs:
         uses: ./.github/actions/setup
         with:
           dependencies: ${{ matrix.deps }}
+      - name: Create required files
```
BeryJu, line 108: I am confused why we need these but I'm sure it'll make sense

authentik/root/monitoring.py
```diff
@@ -5,18 +5,17 @@
...
 
 
 class MetricsView(View):
-    """Wrapper around ExportToDjangoView with authentication, accessed by the authentik router"""
+    """View for metrics monitoring_set signal, accessed by the authentik router"""
 
     def get(self, request: HttpRequest) -> HttpResponse:
         """Check for HTTP-Basic auth"""
         monitoring_set.send_robust(self)
-        return ExportToDjangoView(request)
```
BeryJu, line 19: so where do the django metrics come from now

src/server/core.rs
```diff
@@ -0,0 +1,503 @@
...
+            .to_string(),
+        )
+        .expect("infallible")
+});
+
+static STARTUP_RESPONSE_HTML: LazyLock<Response<String>> = LazyLock::new(|| {
+    Response::builder()
+        .status(StatusCode::SERVICE_UNAVAILABLE)
+        .header(CONTENT_TYPE, "text/html")
+        .body(include_str!("../../web/dist/standalone/loading/startup.html").to_owned())
```
BeryJu, line 48: ah yes that does explain the CI, but maybe we just gate this behind `#[cfg(not(any(debug_assertions, test)))]`

src/server/mod.rs
```diff
@@ -91,34 +154,219 @@ impl Server {
...
+    }
+
+    #[instrument(skip(self))]
+    pub(crate) async fn health_ready(&self) -> Result<bool> {
+        trace!("sending health ready request to server");
+        let req = Request::builder()
+            .method("GET")
+            .uri("http://localhost:8000/-/health/ready/")
+            .header(HOST, "localhost")
+            .body(Body::from(""))?;
```
BeryJu: technically these shouldn't require a body at all, GET requests can't have a body

src/server/static.rs
```diff
@@ -0,0 +1,253 @@
...
+        CACHE_CONTROL,
+        HeaderValue::from_static("public, no-transform"),
+    );
+    response.headers_mut().insert(
+        "X-authentik-version",
+        HeaderValue::from_static(env!("CARGO_PKG_VERSION")),
+    );
+    response
+        .headers_mut()
+        .insert(VARY, HeaderValue::from_static("X-authentik-version, Etag"));
```
BeryJu: yeah this is incorrect, there's a GH Issue #24078 , we were using this wrong in go too

src/server/static.rs
```diff
@@ -0,0 +1,253 @@
...
+                        set_csp_header: false,
+                    },
+                    storage_middleware,
+                ));
+        router = router.nest("/files/reports/", reports_router);
+    }
+
+    router = router.route(
+        "/robots.txt",
+        any(async || include_str!("../../web/robots.txt")),
```
BeryJu, line 238: Same as above

website/docs/install-config/reverse-proxy.md
```diff
@@ -8,29 +8,55 @@ Use this page when authentik is exposed through a reverse proxy or load balancer
...
+If the connection is not trusted, or the above is missing, authentik will look at whether the connection was made over plaintext or TLS.
+
+### Host
+
+Required for various security checks, correct URL handling, WebSocket handshake, and communication with outposts and proxy providers.
+
+The Host is grabbed as follows. If the incoming connection is from a trusted proxy, the following is considered:
+
+- `X-Forwarded-Host` header,
+- `Forwarded` header, as defined in [RFC 7239](https://datatracker.ietf.org/doc/html/rfc7239). If multiple `host=` stanzas are present, only the first one is retained.
```
BeryJu: the main consideration from this is if someone can spoof the host header, do people need to update their reverse proxy config to drop one of these if the reverse proxy doesn't send it

BeryJu (approved):
Couple questions, couple ideas, lgtm

.github/workflows/ci-main.yml
rissson: I'll resolve this for now, and keep the discussion in one of these comments only

authentik/root/monitoring.py
rissson: Just like the worker, the rust server grabs them directly from the filesystem by running a bit of Python https://github.com/goauthentik/authentik/blob/rust-server2/src/metrics/handlers.rs#L54

src/server/static.rs
rissson: I'll resolve this for now, and keep the discussion in one of these comments only

src/server/mod.rs
rissson: Changed it to Body::empty() to be a bit more correct. Unfortunately, we can't use `.body(())` here because of types on the hyper unix socket client. So we are half correct with the change I made, since we don't render a body at all (https://docs.rs/http-body-util/0.1.3/src/http_body_util/empty.rs.html#25-44) even though the types don't show "no body" but "empty body".

In the future, this code goes away once we have freethreading.

website/docs/install-config/reverse-proxy.md
rissson: So for current setups, people don't need to change anything, because we use the first of those header lists, which is always the one that Go used, so the one we expect people to have set previously.

However, for new setups, that is indeed the case. If you were to setup authentik with a `Forwarded` header, you'd need to ensure to drop the ones before, so `X-Real-IP` and `X-Forwarded-For` for example. I'll add a note about that in the docs

rissson: So I've added the note, however what I said is incorrect but only for the Host header. We previously didn't look at the `X-Forwarded-Host` header.

What do you propose about this?

src/server/static.rs
rissson: Should I just remove it then?

src/server/core.rs
rissson: Yeah I don't know about this tbh.

The issue about doing `cfg(not(debug_assertions))` is that when running locally, those won't be served anymore.

If in the future we embed the frontend assets directly in the binary we'll need to build the frontend in CI anyway, so maybe we shouldn't worry about this too much?

BeryJu: fair point yeah, I dont mind this too much, its just giving me flashbacks to the empty files we had to create for when we generated the API clients on the fly.

src/server/static.rs
BeryJu: yeah

website/docs/install-config/reverse-proxy.md
BeryJu: I'd prefer to make it as compatible and identical as the current implementation and save adding new features for a future PR

src/server/core.rs
rissson: yeah same x) we'll get it figured out some way, I just didn't want to build the whole frontend just to run clippy in ci

website/docs/install-config/reverse-proxy.md
rissson: will do!

rissson: this is now identical to what was there before

PR #24324, opened 2026-07-23
stages/authenticator: protect devices from stage deletion

dominic-r:
Deleting an SMS, email, or Duo, authenticator stage currently cascades into its associated devices or connections, allowing an administrator to unintentionally remove enrolled authenticators.

### Linked issues

---

## Checklist


BeryJu (approved):
can probably leave out the GDTC stage since that's deprecated anyways otherwise yeah exactly this

PR #24325, opened 2026-07-23
core: store passwords as authenticator devices

dominic-r:
Passwords currently live as fields on the User model, which prevents password-specific authenticator state and policy from sharing the device model used by other authenticators. This moves local password hashes and change timestamps into a PasswordDevice owned by the password stage, while keeping Django's password APIs as integration adapters and excluding password devices from MFA discovery.


dominic-r (commented):
.

authentik/core/migrations/0064_remove_user_password.py
```diff
@@ -0,0 +1,18 @@
...
+    dependencies = [
+        ("authentik_core", "0063_actor"),
+        ("authentik_stages_password", "0012_migrate_user_passwords"),
+    ]
+
+    operations = [
+        migrations.RemoveField(
+            model_name="user",
+            name="password",
+        ),
```
BeryJu: can't remove this yet, needs to be removed in a future field

authentik/core/models.py
BeryJu: move to password device class too

authentik/enterprise/providers/ssf/signals.py
```diff
@@ -157,7 +157,7 @@ def ssf_device_post_save(sender: type[Model], instance: Device, created: bool, *
 
 @receiver(post_delete)
 def ssf_device_post_delete(sender: type[Model], instance: Device, **_):
-    if not isinstance(instance, Device):
+    if not isinstance(instance, Device) or not instance.is_mfa:
```
BeryJu: I'd prefer checking against device class instead of a special flag

authentik/stages/authenticator/__init__.py
```diff
@@ -124,13 +124,13 @@ def user_has_device(user, confirmed=True):
...
+    Returns an iterable of all loaded device models usable as a second factor.
     """
     from django.apps import apps
 
     from authentik.stages.authenticator.models import Device
 
     for config in apps.get_app_configs():
         for model in config.get_models():
-            if issubclass(model, Device):
+            if issubclass(model, Device) and model.is_mfa:
```
BeryJu: same

authentik/stages/password/migrations/0012_migrate_user_passwords.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+def create_password_devices(apps: Apps, schema_editor: BaseDatabaseSchemaEditor):
+    User = apps.get_model("authentik_core", "User")
+    PasswordDevice = apps.get_model("authentik_stages_password", "PasswordDevice")
+
+    db_alias = schema_editor.connection.alias
+    users = User.objects.using(db_alias).values_list("pk", "password").iterator(chunk_size=1000)
+    PasswordDevice.objects.using(db_alias).bulk_create(
+        (PasswordDevice(user_id=pk, name="Password", password=password) for pk, password in users),
```
BeryJu: should also carry over last_use date

authentik/stages/password/migrations/0012_migrate_user_passwords.py
```diff
@@ -0,0 +1,26 @@
...
+from django.db import migrations
+from django.db.backends.base.schema import BaseDatabaseSchemaEditor
+
+
+def create_password_devices(apps: Apps, schema_editor: BaseDatabaseSchemaEditor):
+    User = apps.get_model("authentik_core", "User")
+    PasswordDevice = apps.get_model("authentik_stages_password", "PasswordDevice")
+
+    db_alias = schema_editor.connection.alias
+    users = User.objects.using(db_alias).values_list("pk", "password").iterator(chunk_size=1000)
```
BeryJu: `progress_bar` from authentik.lib.migrations

BeryJu: also `chunked_queryset` from authentik.lib.utils.db

BeryJu (changes requested):
With the password stage as is it kinda breaks how authenticators work where a specific type configures the type of authenticator and the validation stage validates any of them, where as this one is configured via prompt.

in an ideal world, the authenticator_* stages would both configure and validate authenticators and the authenticator_validation stage would only exist to choose a type, and you could just go backwards within a flow, but that's a big change and a lot of consequences

authentik/stages/password/models.py
```diff
@@ -88,3 +91,18 @@ def ui_user_settings(self) -> UserSettingSerializer | None:
     class Meta:
         verbose_name = _("Password Stage")
         verbose_name_plural = _("Password Stages")
+
+
+class PasswordDevice(Device):
```
gergosimonyi, line 96: I don't really like this here, but at this point all the other `Device` subclasses are in their respective stages' models, so I guess it's fine.

authentik/stages/password/models.py
```diff
@@ -88,3 +91,18 @@ def ui_user_settings(self) -> UserSettingSerializer | None:
...
+    """A user's password, stored as an authenticator device.
+
+    A password is a knowledge factor rather than a second factor, so this device is kept
+    out of MFA discovery, validation and the device APIs."""
+
+    user = models.OneToOneField(User, on_delete=models.CASCADE, related_name="password_device")
+    password = models.CharField(max_length=128)
+    password_change_date = models.DateTimeField(default=now)
+
```
gergosimonyi, line 105: Add `__str__` method?

authentik/core/migrations/0065_remove_user_password.py
```diff
@@ -0,0 +1,40 @@
...
+        migrations.AlterField(
+            model_name="user",
+            name="password",
+            field=models.CharField(max_length=128, null=True),
+        ),
+        migrations.AlterField(
+            model_name="user",
+            name="password_change_date",
+            field=models.DateTimeField(auto_now_add=True, null=True),
+        ),
```
gergosimonyi, line 26: Why is this needed?

authentik/stages/password/migrations/0012_migrate_user_passwords.py
```diff
@@ -0,0 +1,44 @@
+from django.apps.registry import Apps
```
gergosimonyi: Merge with previous migration.

authentik/stages/password/models.py
```diff
@@ -88,3 +91,18 @@ def ui_user_settings(self) -> UserSettingSerializer | None:
...
+class PasswordDevice(Device):
+    """A user's password, stored as an authenticator device.
+
+    A password is a knowledge factor rather than a second factor, so this device is kept
+    out of MFA discovery, validation and the device APIs."""
+
+    user = models.OneToOneField(User, on_delete=models.CASCADE, related_name="password_device")
+    password = models.CharField(max_length=128)
+    password_change_date = models.DateTimeField(default=now)
```
gergosimonyi, line 104: There was an index on `User.password_change_date`. Is there no reason to replicate that here? (Chesterton's fence: then, why was it on `User.passwordpassword_change_date`?)

authentik/core/tests/test_users.py
```diff
@@ -40,6 +41,43 @@ def test_user_ak_groups_event(self):
         user.ak_groups.all()
         self.assertEqual(Event.objects.count(), 1)
 
+    def test_password_stored_on_device(self):
```
gergosimonyi: Is there a reason for this test? It tests implementation for no reason IMO.

authentik/core/tests/test_users.py
```diff
@@ -40,6 +41,43 @@ def test_user_ak_groups_event(self):
...
+        """Test a new user's password is written to their password device"""
+        user = User.objects.create_user(username=generate_id(), password="initial")  # nosec
+        self.assertEqual(user.password, PasswordDevice.objects.get(user=user).password)
+        self.assertTrue(User.objects.get(pk=user.pk).check_password("initial"))
+
+    def test_password_change_updates_device(self):
```
gergosimonyi, line 86: Possibly there could be another test (or one replacing this) that checks that only one `PasswordDevice` is ever allowed per `User`.

Side question: is there a db constraint for this? If not, perhaps there should be, but I'm less sure about this.

authentik/core/tests/test_users.py
```diff
@@ -40,6 +41,43 @@ def test_user_ak_groups_event(self):
...
+
+    def test_password_staged_until_save(self):
+        """Test a password is only written to the device once the user is saved"""
+        user = User.objects.create(username=generate_id())
+        user.set_password("staged")
+        self.assertFalse(PasswordDevice.objects.filter(user=user).exists())
+        user.save()
+        self.assertTrue(User.objects.get(pk=user.pk).check_password("staged"))
+
+    def test_password_unusable_without_device(self):
```
gergosimonyi, line 110: Possibly also add an API test for this. (Possibly replacing this?)

authentik/core/tests/test_users.py
```diff
@@ -40,6 +41,43 @@ def test_user_ak_groups_event(self):
...
+        """Test a user without a password device cannot authenticate with a password"""
+        user = User.objects.create(username=generate_id())
+        self.assertFalse(PasswordDevice.objects.filter(user=user).exists())
+        self.assertFalse(user.has_usable_password())
+        self.assertFalse(user.check_password("anything"))
+
+    def test_session_auth_hash_follows_password(self):
```
gergosimonyi, line 117: Is this wanted behavior?

If so, this is also possibly better as an API test.

authentik/core/models.py
```diff
@@ -407,6 +411,16 @@ class Meta:
...
 
+    def save(self, *args, **kwargs):
+        if self._pending_password_hash is None:
+            super().save(*args, **kwargs)
+            return
+        # The user and their password device hold what used to be a single row, so they
+        # have to be written together.
+        with transaction.atomic():
+            super().save(*args, **kwargs)
+            self._save_pending_password()
```
gergosimonyi: Why?

authentik/enterprise/providers/ssf/signals.py
```diff
@@ -130,7 +131,7 @@ def ssf_password_changed_cred_change(signal, sender, user: User, password: str |
 
 @receiver(post_save)
 def ssf_device_post_save(sender: type[Model], instance: Device, created: bool, **_):
-    if not isinstance(instance, Device):
+    if not isinstance(instance, Device) or isinstance(instance, PasswordDevice):
```
gergosimonyi, line 137: Please explain.

authentik/core/models.py
```diff
@@ -369,6 +370,11 @@ class User(SerializerModel, AttributesMixin, AbstractUser):
     # (This knowingly violates the Liskov substitution principle. It is better to fail loudly.)
     user_permissions = None
 
+    # Hash staged by the `password` setter, written to the password device on save.
+    _pending_password_hash: str | None = None
+    # Change date staged alongside the hash, written to the password device on save.
+    _pending_password_change_date: datetime | None = None
+
```
gergosimonyi: I'm not a big fan of storing these on the `User` (or anywhere, really). Feels like too much magic. Why not just update the device?

authentik/stages/password/migrations/0012_migrate_user_passwords.py
```diff
@@ -0,0 +1,44 @@
...
+
+def create_password_devices(apps: Apps, schema_editor: BaseDatabaseSchemaEditor):
+    User = apps.get_model("authentik_core", "User")
+    PasswordDevice = apps.get_model("authentik_stages_password", "PasswordDevice")
+
+    db_alias = schema_editor.connection.alias
+    users = User.objects.using(db_alias).only(
+        "pk", "password", "password_change_date", "last_login"
+    )
+    print("\nMigrating user passwords to password devices, this might take a couple of minutes...")
```
gergosimonyi: An SQL command should be written instead. It's a simple copy.

authentik/stages/password/migrations/0012_migrate_user_passwords.py
```diff
@@ -0,0 +1,44 @@
...
+    batch = []
+    for user in progress_bar(chunked_queryset(users), total=users.count()):
+        batch.append(
+            PasswordDevice(
+                user_id=user.pk,
+                name="Password",
+                password=user.password,
+                password_change_date=user.password_change_date,
+                last_used=user.last_login,
```
gergosimonyi: This is not necessarily true, is it? Perhaps we should just leave it blank and advise people about it in documentation.

authentik/stages/password/migrations/0012_migrate_user_passwords.py
```diff
@@ -0,0 +1,44 @@
...
+        PasswordDevice.objects.using(db_alias).bulk_create(batch)
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("authentik_stages_password", "0011_passworddevice"),
+    ]
+
+    # Irreversible on purpose: the User.password column is no longer written once devices
```
gergosimonyi: Reversing a migration is not technically supported by authentik, it's only ever included for (developer or operator) convenience. This should be easy to reverse, so let's include it.

gergosimonyi (commented):
Generally I believe this is too much noise in the `User` model. Just as `User` doesn't know about `TOTPDevice`, I don't think it should know about `PasswordDevice`.

PR #24333, opened 2026-07-24
enterprise/stages/source: optional resume oidc source stage

ryanpesek:
### Summary

Adds an opt-in Source Stage setting to resume an embedded OIDC flow when user matching fails because a required property is missing.
When enabled, the parent flow resumes with the userinfo and structured matching failure in its context, without creating a source connection. Existing denial behavior remains the default. Token expiry, source/stage validation, and session cleanup are enforced.

Closes https://github.com/goauthentik/authentik/issues/19705

---

## Checklist


authentik/enterprise/stages/source/models.py
```diff
@@ -24,6 +24,13 @@ class SourceStage(Stage):
         ),
     )
 
+    resume_on_missing_match_property = models.BooleanField(
+        default=False,
+        help_text=_(
+            "Resume the flow when source matching fails because a required property is missing."
+        ),
+    )
+
```
BeryJu: Instead of a boolean field I think it would be better to have an ArrayField with a TextField with choices where the choices are the different Failure reasons (yes I know there's only one right now, but that's more expandable)

authentik/sources/oauth/views/callback.py
```diff
@@ -131,6 +140,45 @@ class OAuthSourceFlowManager(SourceFlowManager):
...
+        plan.context.pop(PLAN_CONTEXT_SOURCE_STAGE_RESUME_ON_MISSING_PROPERTY, None)
+        plan.context.update(self.policy_context)
+        plan.context[PLAN_CONTEXT_SOURCE_MATCH_FAILURE] = {
+            "reason": failure.reason.value,
+            "property": failure.property,
+            "source": self.source.slug,
+        }
+        plan.context[PLAN_CONTEXT_IS_RESTORED] = session_token
+        response = plan.to_redirect(self.request, token.flow)
+        token.delete()
```
BeryJu: I'd prefer not having this much logic specific to the oauth source, I think a lot of this can be generalised

web/xliff/ar.xlf
```diff
@@ -12525,6 +12525,12 @@ Bindings to groups/users are checked against the user of the event.</target>
 <trans-unit id="sae72e1569d34fb02">
   <source>External user usage</source>
 </trans-unit>
+<trans-unit id="stages.source.resume-on-missing-match-property.label">
```
BeryJu: These changes shouldn't be included

ryanpesek: removed

authentik/sources/oauth/views/callback.py
ryanpesek: moved logic to flow_manager so it can be reused across different source types

authentik/enterprise/stages/source/models.py
ryanpesek: makes sense

BeryJu:
replaced by https://github.com/goauthentik/authentik/pull/24963

PR #24351, opened 2026-07-24
fix: persist /data/media by making the media symlink runtime-aware instead of baked into the image

mvanhorn:
### What does this PR change?

Moves provisioning of `/data/media` out of the container image build and into the runtime entrypoint so it can tell a real legacy `/media` mount apart from a fresh install.

- `lifecycle/container/Dockerfile`: drops the build-time `ln -s /media /data/media` and removes `/data/media` from the adjacent `chown` list. The `mkdir -p /certs /data /media /blueprints` is kept. This stops a dangling `/data/media` symlink from being baked into the image and copied into an empty `/data` volume.
- `lifecycle/ak`: adds a `prepare_media_dir` step, invoked from `check_if_root_and_run`, that sets `/data/media` up correctly at startup:
  - If `/media` is a real, populated legacy mount, it (re)creates the `/data/media -> /media` symlink so existing deployments keep working, and it preserves any real `/data/media` directory that already contains data rather than deleting it.
  - Otherwise it ensures `/data/media` is a real directory: a leaked build-time symlink is converted to a directory (only when it resolves to `/media`), migrating any content that was written through it, so uploads land inside the persistent `/data` volume. Symlinks pointing elsewhere are left untouched.
  - The step runs before the privilege-drop path so it also applies in the default non-root (`USER 1000`) container, and it is fail-open: if `/data` is not writable it logs and continues rather than blocking startup.

When authentik runs with a named Docker volume mounted at `/data` (rather than a host bind mount) and that volume has no `media` subdirectory, uploaded media is silently lost on container recreation.

The image baked a build-time symlink `ln -s /media /data/media`. For an empty `/data` volume, Docker seeds the volume from the image tree, so the symlink is copied into the volume while the ephemeral `/media` contents in the container's writable layer are not. Uploads written through `/data/media` therefore never persist. The symlink was only ever created at build time, not at runtime, which is why it survives into the volume as a dangling pointer.

The symlink still exists to support legacy `/media` mounts from older versions, so it cannot simply be removed. Handling it at runtime keeps legacy `/media` users working while making new installs persist media inside `/data`.

### How was this tested?

- The entrypoint script was validated with `bash -n` (syntax) and `shellcheck` (no new findings introduced by the change).
- The logic was traced through the affected scenarios: fresh install on an empty named `/data` volume, an existing legacy `/media` mount, an upgrade whose `/data` volume already carries the leaked `/data/media` symlink, and the default non-root (`USER 1000`) container where the entrypoint takes the non-root path.

End-to-end acceptance test (the original report's reproduction), which should now keep media across a recreate:

1. Use the official Compose file with a named volume mounted at `/data` for the server and worker (e.g. `authentik_data:/data`), no `/media` host mount, and no pre-existing `media` dir in the volume.
2. `docker compose up -d`, complete initial setup, and confirm `/data` has no `media` entry.
3. Log in and upload an icon/image under Customization > Files.
4. In the server container, `ls -la /data` now shows `media` as a real directory (not a symlink into the ephemeral container layer).
5. `docker compose down` then `docker compose up -d`; the uploaded media is still present.

Legacy deployments with a real `/media` mount continue to resolve `/data/media` to `/media`, and existing content is left in place.

### Linked issues

closes #20270

---

## Checklist


rissson:
Hello, I'd rather not, and simply remove legacy symlink support.

mvanhorn:
Done — dropped the runtime `prepare_media_dir` migration entirely as you suggested. `/data/media` is now just a directory created in the image alongside the other data dirs, with no legacy `/media` symlink handling at all. Net effect on this PR is −70 lines: the only remaining change is that the dangling `/data/media` symlink is no longer baked into the image.

PR #24361, opened 2026-07-24
web/elements: key the message list so stacked toasts don't share state

melizeche:
The toast container renders messages with an unkeyed `.map`, so Lit reuses `<ak-message>` elements across messages. Stateful bits then bleed: the newer toast loses its level icon (e.g. the success checkmark), and the auto-dismiss timer set in `firstUpdated` sticks to whichever message rendered first.

Render with `repeat()` keyed on the message so each toast keeps its own element.

before:
image

after:
image

---

## Checklist


kensternberg-authentik (approved):
What an excellent fix.

PR #24367, opened 2026-07-25
stages/prompt: relative prompt order & improved editor

BeryJu:
- add complete prompt preview
- move order to binding of stage and pormpt
- allow creating inline

image


kensternberg-authentik (commented):
That's sorta ugly:

Image

Builder modal is cut off and inaccessible:

Image

Can't create decorative elements like separators without providing unused info:

Image

Not sure what providing a sort order to a hidden field accomplishes:

Image

PR #24368, opened 2026-07-25
web/admin: better table filter

BeryJu:
From 
image

To 
image


kensternberg-authentik (approved):
An idiom is an unrecognized component.  Excellent.

PR #24416, opened 2026-07-27
website: enterprise CTA

dominic-r:
image


kensternberg-authentik (approved):
Surprised (annoyed may be more accurate) that we need a React component for a couple of static buttons.

PR #24422, opened 2026-07-28
web/elements: Add location to sessions

PeshekDotDev:
### What does this PR change?

Currently, the list of sessions shows a flag to denote location, but doesn't show city or anything else. This adds a city name next to the device information.

Just a flag doesn't always help show that a session is one that you don't know. I may see a second session in the US and assume it's me on a different device, but if I see it's in a city I've never been to, it helps me identity a bad session faster.

### How was this tested?

Built and ran locally

### Linked issues

---

## Checklist


kensternberg-authentik (approved):
A bit magical (all those Boolean filters), but obvious at a second glance.

PR #24434, opened 2026-07-28
core: add admin_authenticated_session_deleted signal for admin deleted session

BeryJu:
currently ssf listens on pre_delete of AuthenticatedSession, which also fires if a user logs out manually, which then propagates and causes ssf to log out out of everything even if it still has refresh tokens

### DIFF
--- a/authentik/core/api/authenticated_sessions.py
+++ b/authentik/core/api/authenticated_sessions.py
@@ -6,7 +6,7 @@ from drf_spectacular.utils import (
     extend_schema,
     inline_serializer,
 )
-from rest_framework import mixins, serializers
+from rest_framework import mixins, serializers, status
 from rest_framework.decorators import action
 from rest_framework.fields import SerializerMethodField
 from rest_framework.request import Request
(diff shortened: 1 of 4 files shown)


rissson (approved):

authentik/core/api/authenticated_sessions.py
```diff
@@ -152,6 +154,16 @@ class AuthenticatedSessionViewSet(
...
+        for session in chunked_queryset(AuthenticatedSession.objects.filter(user_id__in=user_pks)):
+            admin_authenticated_session_deleted.send(self, session=session, request=request)
+            session.delete()
+            count += 1
+
+        return Response({"deleted": count}, status=200)
+
+    def destroy(self, request, *args, **kwargs):
+        instance = self.get_object()
+        admin_authenticated_session_deleted.send(self, session=instance, request=request)
```
rissson:
```suggestion
        admin_authenticated_session_deleted.send(self, instance=instance, request=request)
```

rissson (changes requested):

rissson (approved):

PR #24437, opened 2026-07-28
website/docs: add object attributes doc

dewi-tik:
### What does this PR change?

Adds doc for object attributes change

Added feature in 2026.8

---

## Checklist


website/docs/sys-mgmt/object-attributes.mdx
```diff
@@ -0,0 +1,111 @@
+---
+title: Object attributes
+description: "Define reusable object attributes for authentik objects"
```
dominic-r:
```suggestion
description: "Define reusable object attribute fields for authentik objects"
```

website/docs/sys-mgmt/object-attributes.mdx
```diff
@@ -0,0 +1,111 @@
...
+
+Before authentik 2026.8, an object's attributes were edited as free-form YAML or JSON in an object's **Attributes** field. These attributes had no discoverable keys, human-readable labels, or validation. Object attributes improve that experience without changing how attributes are stored: they remain in an object's **Attributes** field.
+
+Object attributes let you define structured, reusable attribute fields for Users, Groups, Application entitlements, and Device access groups. When you define an object attribute, authentik displays it as a labeled field in the relevant create and edit forms. You can group related attribute fields and optionally validate submitted values.
+
+Attributes remain editable in the **Attributes** field on an object. However, when an attribute has an object attribute definition, its generated form field is authoritative. When you save the object, authentik overwrites any value entered directly in **Attributes** for the same key.
+
+## Default object attribute definitions
+
+authentik includes definitions for common object attributes. These default object attribute definitions are currently disabled by default.
```
dominic-r:
```suggestion
authentik includes definitions for common attributes. These defaults are disabled by default; [enable a definition](#enable-or-disable-an-object-attribute-definition) when you want its labeled field on forms.
```

website/docs/sys-mgmt/object-attributes.mdx
```diff
@@ -0,0 +1,111 @@
...
+| Group                 | No       | Groups attributes under a shared heading on forms. Attributes without a group are rendered ungrouped        |
+| Enabled               | Yes      | Controls whether the attribute is shown on forms for the selected object type                               |
+| Type                  | Yes      | The attribute type: Text, Number, or Boolean                                                                |
+| Object type           | Yes      | The object type that the attribute applies to: User, Group, Application entitlement, or Device access group |
+| Attribute is required | No       | Enforces whether a value is required for the attribute on the selected object type                          |
+| Attribute is unique   | No       | Enforces whether the value must be unique across all instances of the selected object type                  |
+| Regex                 | No       | A regular expression that the value of the attribute is validated against                                   |
+
+## Using object attributes
```
dominic-r:
```suggestion
## Use object attributes
```

website/docs/sys-mgmt/object-attributes.mdx
```diff
@@ -0,0 +1,111 @@
...
+### Enable or disable an object attribute definition
+
+Disable an object attribute definition when you want to remove the attribute field on forms without losing the values already stored on objects.
+
+1. Log in to authentik as an administrator and open the Admin interface.
+2. Navigate to **Directory** > **Object attributes**.
+3. Select an object attribute definition and click **Edit**.
+4. Toggle **Enabled**, then click **Update**.
+
+Disabling a definition removes the field from forms and stops Regex and uniqueness validation for that key. When you re-enable an object attribute definition stored values remain present and validation resumes on subsequent saves.
```
dominic-r:
```suggestion
Disabling a definition removes the field from forms and stops Regex and uniqueness validation for that key. When you re-enable an object attribute definition, stored values remain present and validation resumes on subsequent saves.
```

website/docs/sys-mgmt/object-attributes.mdx
```diff
@@ -0,0 +1,111 @@
...
+
+### Examples of using object attributes in property mappings
```
dominic-r:
```suggestion
### Use object attributes in property mappings
```

website/docs/sys-mgmt/object-attributes.mdx
```diff
@@ -0,0 +1,111 @@
...
+
+SAML property mapping:
```
dominic-r:
```suggestion
**Example**: SAML property mapping:
```

website/docs/sys-mgmt/object-attributes.mdx
```diff
@@ -0,0 +1,111 @@
...
+
+```python
+return user.attributes.get("given_name", user.name)
+```
+
+Nested attribute value, for example a locale claim in an OAuth2 scope mapping:
```
dominic-r:
```suggestion
**Example**: Nested attribute value in an OAuth2 scope mapping:
```

website/docs/sys-mgmt/object-attributes.mdx
```diff
@@ -0,0 +1,111 @@
...
+}
+```
+
+Always use `.get()` with a fallback. An object attribute definition does not guarantee that every object has a value for that key.
+
+## Validation
+
+### Uniqueness
+
+If an object attribute definition is set as `Attribute is unique`, authentik rejects a save when another object already stores the same value for that key. Uniqueness is only enforced while the object attribute definition is enabled.
```
dominic-r:
```suggestion
If an object attribute definition has **Attribute is unique** enabled, authentik rejects a save when another object already stores the same value for that key. Uniqueness is only enforced while the object attribute definition is enabled.
```

website/docs/sys-mgmt/object-attributes.mdx
```diff
@@ -0,0 +1,111 @@
...
+
+### Existing data
+
+Creating an object attribute definition does not validate values that are already stored on existing objects. Validation is applied when an object is saved. An object that was created before the object attribute definition existed can therefore hold a value that does not match the object attribute definition's regex.
```
dominic-r:
```suggestion
Creating an object attribute definition does not validate values already stored on existing objects. Validation runs when an object is saved, so an object created before the definition existed can hold a value that does not match the definition's regex.
```

dominic-r (approved):
once applied

website/docs/sys-mgmt/object-attributes.mdx
```diff
@@ -0,0 +1,111 @@
...
+
+```python
+return user.attributes.get("given_name", user.name)
+```
+
+**Example**: Nested attribute value in an OAuth2 scope mapping:
+
+```python
+return {
+    "locale": user.attributes.get("settings", {}).get("locale", "en-US"),
```
BeryJu: Note: use `ak_obj_attr`, reminder for @BeryJu to change fallback behaviour in `ak_obj_attr`

website/docs/sys-mgmt/object-attributes.mdx
```diff
@@ -0,0 +1,111 @@
...
+
+If an object attribute definition has **Attribute is unique** enabled, authentik rejects a save when another object already stores the same value for that key. Uniqueness is only enforced while the object attribute definition is enabled.
+
+### Existing data
+
+Creating an object attribute definition does not validate values already stored on existing objects. Validation runs when an object is saved, so an object created before the definition existed can hold a value that does not match the definition's regex.
+
+## Considerations and limitations
+
+- Array values are not currently supported. Only Text, Number, or Boolean values are supported. Support for array values is tracked in [this issue](https://github.com/goauthentik/authentik/issues/24250).
```
BeryJu: Maybe more explicitly state that this also affects direct API changes to the affected objects

website/docs/sys-mgmt/object-attributes.mdx
```diff
@@ -0,0 +1,109 @@
+---
+title: Object attributes
+description: "Define reusable object attribute fields for authentik objects"
+sidebar_label: Object attributes
+authentik_version: "2026.8"
+authentik_preview: true
```
BeryJu, line 6: @BeryJu add preview banner in UI

BeryJu (approved):

PR #24440, opened 2026-07-28
root: add .nvmrc

GirlBossRush:
## Details

Adds a single file at the repository root: `.nvmrc`, containing `24`.

With it in place, `nvm install` and `nvm use` resolve the version on their own instead of needing it spelled out each time.

Follows up #24439, which points the setup docs at [nvm](https://github.com/nvm-sh/nvm) as the preferred way to install Node.js. The two PRs touch disjoint files and can merge in either order.

The setup docs now tell contributors to install Node.js with nvm, but nothing in the repo told nvm which version to use.

### Linked issues

refs #24439


kensternberg-authentik (approved):
Exciting!

PR #24441, opened 2026-07-28
providers/scim: ignore key casing in SCIM responses

### LINKED ISSUE #24431: SCIM provider sync to Bitwarden fails with `KeyError: 'Resources'` during discovery (regression since 2026.5.4)
### Describe the bug

After upgrading from 2026.5.3 to 2026.5.4, the outgoing SCIM provider sync targeting Bitwarden fails during the discovery phase with `KeyError: 'Resources'`. The error is still present in 2026.5.6. Rolling back to 2026.5.3 restores a working sync.

**Observed behavior (verified on my deployment)**

- 2026.5.3: SCIM sync to Bitwarden works.
- 2026.5.4 and 2026.5.6: sync fails immediately with the exception below, raised in `discover()` at `authentik/providers/scim/clients/users.py:128`.
- When I locally patched the failing lookup to `res.get("Resources", [])` as a workaround, the sync task no longer crashed but instead ran forever and never completed. This is consistent with the analysis below (pagination loop makes no progress when the resource list reads as empty).

**Presumed root cause (verified against a live Bitwarden response)**

The `discover()` method (added in #20658, first effectively active for SCIM in 2026.5.4 — presumably via the `can_discover` fix #23035) reads the list response with case-sensitive keys:

```python
res = self._request("GET", "/Users")
expected_items = int(res["totalResults"])
while True:
    for user in res["Resources"]:
        ...
```

Bitwarden serializes its SCIM list responses in camelCase, i.e. the resource list key is `resources` (lowercase), while `totalResults` happens to match authentik's lookup. Verified against my self-hosted Bitwarden instance (sanitized):

```json
// GET /scim/v2/{orgId}/Users?startIndex=1&count=1
{
    "totalResults": 2,
    "startIndex": 1,
    "itemsPerPage": 1,
    "resources": [
        {
            "id": "<redacted-guid>",
            "meta": { "resourceType": "User", "created": null, "lastModified": null },
            "userName": "user@example.com",
            "name": { "formatted": "Example User", "givenName": null, "middleName": null, "familyName": null },
            "emails": [ { "primary": true, "value": "user@example.com", "type": "work" } ],
            "primaryEmail": "user@example.com",
            "workEmail": "user@example.com",
            "displayName": "Example User",
            "active": true,
            "groups": [],
            "externalId": "user@example.com",
            "schemas": [ "urn:ietf:params:scim:schemas:core:2.0:User" ]
        }
    ],
    "schemas": [ "urn:ietf:params:scim:api:messages:2.0:ListResponse" ]
}
```

Note that #23357 ("allow failures during discovery") does not cover this: it wraps the per-object `_discover_user_single()` call in `try/except ValidationError`, but the `res["Resources"]` access happens outside that guard. `groups.py` has the same pattern.

A related quirk that may affect discovery against Bitwarden independently: bitwarden/server#4034 (SCIM `/Users` listing returns no users unless `startIndex` and `count` are specified) — the initial discovery request is sent without query parameters.

### How to reproduce

1. Set up a Bitwarden organization with SCIM enabled.
2. In authentik ≥ 2026.5.4, create a SCIM provider (default user and group mappings) targeting the Bitwarden SCIM base URL and assign it as backchannel provider to an application.
3. Run the provider sync.
4. Sync fails during discovery with `KeyError: 'Resources'`.

### Expected behavior

Discovery tolerates the camelCase attribute names in Bitwarden's list responses (per RFC 7643 §2.1) and the sync completes. One observation that may be useful when fixing this: simply tolerating a missing `Resources` key is not sufficient on its own — when I suppressed the `KeyError` locally, the pagination loop never terminated because `seen_items` never reached `expected_items`. Both `users.py` and `groups.py` contain the same pattern.

### Screenshots

_No response_

### Additional context

The code-level analysis in this report (identifying the case-sensitive `res["Resources"]` lookup and Bitwarden's camelCase response serialization as the mismatch) was done with the assistance of an AI tool. The error, the affected/working versions, and the live SCIM response shown above were reproduced and verified by me on my real deployment.

### Deployment Method

Docker

### Version

2026.5.4, 2026.5.6

### Relevant log output

```shell
{
    "exception": [
        {
            "frames": [
                {
                    "name": "process_message",
                    "lineno": 540,
                    "filename": "/ak-root/.venv/lib/python3.14/site-packages/dramatiq/worker.py"
                },
                {
                    "name": "__call__",
                    "lineno": 194,
                    "filename": "/ak-root/.venv/lib/python3.14/site-packages/dramatiq/actor.py"
                },
                {
                    "name": "scim_sync",
                    "lineno": 20,
                    "filename": "/authentik/providers/scim/tasks.py"
                },
                {
                    "name": "sync",
                    "lineno": 87,
                    "filename": "/authentik/lib/sync/outgoing/tasks.py"
                },
                {
                    "name": "_discover",
                    "lineno": 123,
                    "filename": "/authentik/lib/sync/outgoing/tasks.py"
                },
                {
                    "name": "discover",
                    "lineno": 128,
                    "filename": "/authentik/providers/scim/clients/users.py"
                }
            ],
            "exc_type": "KeyError",
            "is_cause": false,
            "is_group": false,
            "exc_notes": [],
            "exc_value": "'Resources'",
            "exceptions": [],
            "syntax_error": null
        }
    ]
}
```

BeryJu:
closes https://github.com/goauthentik/authentik/issues/24431

I dont like this

Spec defines some keys as `Resources`: https://datatracker.ietf.org/doc/html/rfc7644#section-3.4.2

However in https://datatracker.ietf.org/doc/html/rfc7643#section-2.1 they are defined as case-insensitive...


rissson (approved):
eww, but what can you do, postel's law and all that

(approval in principle, tests still need to pass)

PR #24450, opened 2026-07-29
stages/authenticator_webauthn: remove prevent_duplicate_devices

PeshekDotDev:
### Linked issues

Closes https://github.com/goauthentik/authentik/issues/24352

---

## Checklist

-   [ ] The project has been linted, built, and tested (`make all`)
-   [ ] The documentation has been updated and formatted (`make docs`)

Issue #24352, opened 2026-07-24
Retire “Prevent duplicate keys” webauthn toggle unless we can find a workaround


dewi-tik:
See https://github.com/goauthentik/authentik/issues/23047#issuecomment-4903959228

And https://github.com/goauthentik/authentik/pull/23823

PR #24451, opened 2026-07-29
stages/authenticator_webauthn: prevent duplicate enrollment via excludeCredentials

PeshekDotDev:
### What does this PR change?

authentik now sends the user's already-enrolled credentials in the `excludeCredentials` parameter of the WebAuthn registration options. An authenticator that already holds a credential for the account refuses to create a second one, and the browser surfaces that as an `InvalidStateError`. This is a similar behavior to what the Prevent Duplicate Keys behavior was going for

Without it, a user can enroll the same key twice and end up with junk credentials on their account. The authenticator overwrites the existing credential for that account rather than storing a second one, so the device row authentik already has points at a credential the key no longer holds. The user sees two entries in their settings and one of them silently doesn't work.

### How was this tested?

Via added unit and e2e tests

### Linked issues

refs #24352
refs #24450

---

## Checklist


dewi-tik (approved):

authentik/stages/authenticator_webauthn/stage.py
```diff
@@ -170,8 +170,17 @@ def get_challenge(self, *args, **kwargs) -> Challenge:
             elif hint_values <= platform:
                 authenticator_attachment = AuthenticatorAttachment.PLATFORM
 
+        rp_id = get_rp_id(self.request)
+
+        # Let the authenticator itself refuse to register a credential it already holds for this
+        # user.
+        exclude_credentials = [
+            device.descriptor
+            for device in WebAuthnDevice.objects.filter(user=user, rp_id=rp_id).order_by("pk")
```
BeryJu, line 179: this needs to also filter on confirmed=True

BeryJu (changes requested):
Kinda on the fence whether this should be backported or not

PeshekDotDev:
I am okay not backporting and announcing this more officially in the next release. If that's the desire just let me know and I'll remove the release notes changes. I feel like they'll be missed now that the release has happened long enough ago

authentik/stages/authenticator_webauthn/stage.py
PeshekDotDev: fixed

authentik/stages/authenticator_webauthn/stage.py
```diff
@@ -170,8 +170,20 @@ def get_challenge(self, *args, **kwargs) -> Challenge:
...
+
+        # Let the authenticator itself refuse to register a credential it already holds for this
+        # user. Only confirmed devices count; an unconfirmed one is not usable and must not block
+        # re-enrolling the same authenticator.
+        exclude_credentials = [
+            device.descriptor
+            for device in WebAuthnDevice.objects.filter(
+                user=user, rp_id=rp_id, confirmed=True
+            ).order_by("pk")
+        ]
```
BeryJu, line 187: `.values_list("descriptor", flat=True)`

PeshekDotDev: How's this?

PR #24454, opened 2026-07-29
crypto: cache parsed certificates and private keys across model instances

### LINKED ISSUE #24453: OAuth2 endpoints repeatedly validate unchanged RSA signing keys
## Describe the bug

OIDC requests repeatedly parse and validate the same persisted RSA private key. `OAuth2Provider.jwt_key` and `CertificateKeyPair.private_key` cache the parsed key only on the current ORM model instance, so a fresh model instance performs the expensive RSA validation again.

On an official arm64 FIPS Docker deployment using an RSA-4096 signing key, profiling measured approximately:

- `provider.jwt_key`: 1.53 seconds
- OIDC discovery response: 1.85 seconds
- The same request with a process-local, cross-instance parsed-key cache: 0.25–0.30 seconds when warm

The first parse remains fully validated; only unchanged PEM content is reused within the server process.

## How to reproduce

1. Configure an OAuth2/OpenID provider with an RSA-4096 signing certificate.
2. Send repeated OIDC requests handled by the same long-lived server process.
3. Profile `OAuth2Provider.jwt_key`, or access the same `CertificateKeyPair` through two separately fetched ORM instances.
4. Observe that `load_pem_private_key` runs for each fresh model instance even though the PEM content is unchanged.

## Expected behavior

An unchanged private key should be parsed and fully validated once per server process, then reused across ORM model instances. Changed PEM content must cause a new validated parse, and parsing failures must not be cached.

## Additional context

The existing cached properties avoid repeated work only while a particular provider/key model instance remains alive. Requests generally load fresh ORM instances, so that cache does not cover the repeated-request case.

## Deployment Method

Docker

## Version

2026.5.6

jakosky:
### What does this PR change?

Adds bounded process-local caches for parsed certificates and fully parsed private keys, keyed by PEM content. `CertificateKeyPair.certificate` and `CertificateKeyPair.private_key` keep their existing instance caches and now reuse parsed objects across separately loaded ORM instances.

The regression tests verify that:

- Two ORM instances containing the same certificate or private key invoke the corresponding cryptography loader once and return the same parsed object.
- Different PEM content causes another parse.

OIDC requests commonly load fresh provider and certificate model instances. The existing instance-level caches therefore repeat certificate parsing and the expensive RSA private-key validation even when the stored PEM has not changed.

Each cache is bounded to 128 entries. Changed content is a cache miss, and parse failures are not cached.

### How was this tested?

- Added the regression test red-first; it failed before the cache and passed afterward.
- `authentik/crypto/tests.py`: 27 passed.
- Black, Ruff, mypy, and Bandit passed.
- `make all` completed all lint, build, generation, and web checks. The server suite reported 3,883 passed and 27 skipped; three unrelated Kerberos sync tests could not load the native `kadm5` library in the macOS test environment.

### Linked issues

Closes #24453

---

## Checklist

-   [x] Documentation is not applicable; this changes no public behavior, API, or configuration.


rissson (approved):

rissson:
I'm happy with this. cc. @PeshekDotDev @BeryJu if you guys have different thoughts about this

BeryJu:
yeah looks good if it doesnt cause any issues, although I'd add support for certificates too

jakosky:
Added cross-instance certificate parsing cache and regression coverage in dcd327b4c1, using the same bounded-cache pattern as private keys.

The prior CI attempt's only test failure was unrelated: `TestWSFedSignIn.test_token_gen` could not fetch the imported OASIS WS-Security XSD. The new commit has triggered a fresh CI run.

PR #24465, opened 2026-07-29
providers/rac: parse IPv6 literal endpoint hosts correctly

chughes741:
ConnectionToken.get_settings() splits the endpoint host on the first colon. When an IPv6 address is passed such as [2001:1234::1]:5900 it produces "[2001" and "1234::1]:5900" which causes an error when it's passed to guacd. This made it functional impossible to use IPv6 for RAC endpoints.

Replaced with the standard urllib parser that gracefully handles IPv4, IPv6, and FQDNs. Added unit tests to ensure it continues to work correctly.


rissson:
@chughes741 no need to merge main like this, it just adds unnecessary noise and force us to restart the CI

PR #24479, opened 2026-07-30
packages/ak-axum: Correctly canonicalize IP when checking for trusted reverse proxy

BeryJu:
at least in istio client IPs show up as `::ffff:10.2.0.229`


packages/ak-axum/Cargo.toml
```diff
@@ -23,6 +23,7 @@ tokio.workspace = true
 tower-http.workspace = true
 tower.workspace = true
 tracing.workspace = true
+ipnet.workspace = true
```
rissson: Let's keep this list sorted

rissson (approved):

PR #24480, opened 2026-07-30
ci: add rust proxy binary build to release CI

rissson:
This is a breaking change in the way outpost binaries are named, but I think it's fine.

Tested by hijacking the CI on main: https://github.com/goauthentik/authentik/actions/runs/30542610801/job/90870892705?pr=24480

### DIFF
--- a/.github/workflows/release-tag.yml
+++ b/.github/workflows/release-tag.yml
@@ -174,9 +174,6 @@ jobs:
     needs:
       - bump-version
     timeout-minutes: 120
-    # aws-lc-sys/aws-lc-fips-sys compile C via cc/cmake, so every target needs a matching
-    # native toolchain. Build each one on a runner of the same OS/arch instead of cross-compiling.
-    runs-on: "${{ matrix.target.runs-on }}"
     permissions:
       # Needed for checkout
       contents: read
@@ -194,6 +191,7 @@ jobs:
             runs-on: macos-15-intel
           - triple: aarch64-apple-darwin
             runs-on: macos-latest
+    runs-on: "${{ matrix.target.runs-on }}"
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v5
         with:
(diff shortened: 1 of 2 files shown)


.github/workflows/release-tag.yml
```diff
@@ -171,6 +170,55 @@ jobs:
...
+      - name: Setup rust
+        uses: actions-rust-lang/setup-rust-toolchain@166cdcfd11aee3cb47222f9ddb555ce30ddb9659 # v1
+        with:
+          rustflags: ""
+          target: "${{ matrix.target.triple }}"
+      - name: Setup go
+        uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e # v7.0.0
+        with:
+          go-version-file: "go.mod"
+          cache: false
```
BeryJu, line 208: I dont think this is needed?

.github/workflows/release-tag.yml
```diff
@@ -171,6 +170,55 @@ jobs:
...
+      - name: Build outpost
```
BeryJu, line 209: this is missing I think

```
      - uses: ./.github/actions/setup-node
        with:
          working-directory: web
      - name: Build web
        working-directory: web/
        run: |
          npm run build-proxy
```

.github/workflows/release-tag.yml
rissson: It is, because go is required for building aws-lc-rs

.github/workflows/release-tag.yml
rissson: It is not, because we don't include any static files in the proxy outpost anymore. We were including them in the go outpost, but not actually using them

.github/workflows/release-tag.yml
BeryJu: ah right yeah

.github/workflows/release-tag.yml
BeryJu: how do we do the error page? it used to be setup in the go implementation only in proxy mode, if the backend was down for example

rissson: it's all embedded since #17683, which should have also removed the embedded static files, but never did

BeryJu: oh yeah I forget

BeryJu (approved):

PR #24524, opened 2026-07-30
web/elements: compile ak-mdx content from the API as CommonMark

kah-ja:
## What

`<ak-mdx>` renders two kinds of source:

- documentation authored in-tree and bundled with the web interface, passed as `url`,
- content that comes back from the API, passed as `content` — `user.attributes.notes` (`UserNotesCard.ts`), `group.attributes.notes` (`GroupViewPage.ts`) and blueprint descriptions (`BlueprintListPage.ts`).

Both were compiled with the same `compile()` call, i.e. as MDX. MDX parses expressions (`{...}`), JSX and ESM `import`/`export` in addition to Markdown, and those are evaluated when the compiled module is run by `run()` in `delegateRender()`. So a source that came from the API was evaluated as code in the admin interface.

The `BrandedHTMLPolicy` DOMPurify pass in front of the compiler does not cover this: it operates on HTML, and neither `{...}` nor an `export` statement is HTML — both reach `compile()` unchanged.

Reported privately as GHSA-5vfm-49hj-h95w.

## Change

Thread the compile format through `delegateRender()`:

- `url` (bundled documentation) keeps `format: "mdx"` — unchanged behaviour,
- `content` (from the API) gets `format: "md"`, i.e. CommonMark, so nothing in it is evaluated.

This is a parser-level distinction rather than a filter, so it does not depend on enumerating constructs. The DOMPurify pass stays where it is.

## Behaviour

Measured against the pinned versions (`@mdx-js/mdx` 3.1.1, `dompurify` 3.4.12, `react` 19.2.8), rendering through the same `compile()`/`run()` pipeline the element uses:

| Source in `content` | before | after |
| --- | --- | --- |
| `{...}` expression | evaluated | rendered as text |
| `export const … = …` | evaluated | rendered as text |
| `import "data:text/javascript,…"` | evaluated | rendered as text |
| `<div>{…}</div>` | evaluated | rendered as text |
| headings, bold/italic, lists, GFM tables, fenced code with highlighting, links | rendered | rendered, unchanged |

One intentional difference: raw HTML in `content` (for example `<strong>hi</strong>` in a note) now renders as text instead of as an element. Markdown syntax is unaffected.

## Tests

`web/src/elements/ak-mdx/content-format.unit.test.ts` covers that a source is evaluated under `mdx` and inert under `md`, and that CommonMark still renders.

A note on running them locally: `web/vite.config.js` points the `Unit Tests` project at `./tsconfig.unit.json`, which is not in the tree, so `vitest --project "Unit Tests"` fails during transform before any test runs — including for the existing `custom-elements-get-name.unit.test.ts`. I therefore ran the assertions of the new test as a plain node script against `web/node_modules` (all four pass) rather than through vitest. Happy to look at the missing tsconfig in a separate PR if that is useful.

### DIFF
--- a/web/src/elements/ak-mdx/ak-mdx.tsx
+++ b/web/src/elements/ak-mdx/ak-mdx.tsx
@@ -55,6 +55,17 @@ const highlightThemeOptions: HighlightOptions = {
  */
 export type Replacer = (input: string) => string;
 
+/**
+ * The syntax a source is compiled as.
+ *
+ * - `mdx` additionally parses JSX, expressions (`{...}`) and ESM
+ *   imports/exports, all of which are evaluated when the compiled module runs.
+ * - `md` is CommonMark only, i.e. no part of the source is evaluated.
+ *
+ * @see {@linkcode https://mdxjs.com/packages/mdx/#optionsformat}
+ */
+type ContentFormat = "md" | "mdx";
+
 @customElement("ak-mdx")
 export class AKMDX extends AKElement {
     // HACK: Fixes Lit Analyzer's parsing of TSX files with decorators.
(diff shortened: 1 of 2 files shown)


GirlBossRush:
Hi @kah-ja,

Thank you for the excellent PR. We're reworking our in-app markdown compilation to remove React from our dependency list. There's also some ongoing planning surrounding Content Security Policy support and trusted inputs.

You can track progress on this effort at #21949

PR #24530, opened 2026-07-30
outpost/proxy: fix issuer validation

rissson:
follow-up on https://github.com/goauthentik/authentik/pull/24489

this might not be valid. Don't merge

### Linked issues

---

## Checklist


rissson:
This does not seem to be required. Leaving open for now just in case.

Issue #9972, opened 2024-06-04
broken ldap cached session after "failed to execute flow"


calbrecht:
A note upfront, the instance is deployed using the https://github.com/nix-community/authentik-nix module. If you think the issue is related to the installation method in use feel free to close it, although it seems to me the cause is an oversight in refactoring. But then again, what do i know,

**Describe the bug**
Suddenly authentication via ldap provider fails on client with ```Invalid credentials```. The failure is reproducible.
In authentik ldap logs at the time of the first client authentication failure, is this error ```"failed to submit challenge 502 Bad Gateway"``` and event ```"failed to execute flow"```, followed by timespaced recurring events ```"authenticated from session"```.
Changing the bind method of the ldap provider to direct bind, authenticating, and changing back to cached bind fixes the failure.

**To Reproduce**
Don't know how to reproduce the bind failure "failed to execute flow". See logs. 

**Expected behavior**
The ldap cached bind should not cache broken sessions, i guess.

**Logs**
```
Jun 03 15:40:47 authentik ldap[479322]: {"bindDN":"cn=service,ou=users,dc=goauthentik,dc=io","client":"same::ipv:6","error":"failed to submit challenge 502 Bad Gateway","event":"failed to execute flow","level":"warning","requestId":"same-redacted-req-id","timestamp":"2024-06-03T15:40:47Z"}
Jun 03 15:40:47 authentik ldap[479322]: {"bindDN":"cn=service,ou=users,dc=goauthentik,dc=io","client":"same::ipv:6","event":"Bind request","level":"info","requestId":"same-redacted-req-id","timestamp":"2024-06-03T15:40:47Z","took-ms":164}
```
From now on, the client fails to authenticate, despite time-spaced recurring lines like
```
Jun 03 15:40:49 authentik ldap[479322]: {"bindDN":"cn=service,ou=users,dc=goauthentik,dc=io","event":"authenticated from session","level":"info","logger":"authentik.outpost.ldap.binder.session","timestamp":"2024-06-03T15:40:49Z"}
Jun 03 15:40:49 authentik ldap[479322]: {"bindDN":"cn=service,ou=users,dc=goauthentik,dc=io","client":"same::ipv:6","event":"Bind request","level":"info","requestId":"different-redacted-req-id","timestamp":"2024-06-03T15:40:49Z","took-ms":0}
```

**Version and Deployment (please complete the following information):**

-   authentik version: 2024.4.2
[...]

PR #24582, opened 2026-07-31
internal/outpost/ldap: only cache successful binds in SessionBinder

NikolaiBessonov:
## Summary

- `DirectBinder.Bind()` never returns a non-nil `error` for a rejected bind (invalid credentials, access denied, flow execution/bruteforce errors) — every outcome is encoded solely as an `ldap.LDAPResultCode`, with `err` always `nil`.
- `SessionBinder` (the `bind_mode: cached` wrapper) only checked `err == nil` before caching a bind result. Since `err` is always `nil`, it ended up caching **every** outcome, including transient failures.
- A single rejected bind (e.g. a bruteforce lockout) then got served back from the session cache as a false success/failure for the full session TTL (hours), even after the underlying cause was gone and the credentials were re-verified correct — the outpost logged "authenticated from session" while actually returning Invalid credentials (49) to the client.

**Fix**: only cache when the bind actually succeeded (`err == nil && result == ldap.LDAPResultSuccess`), and split the log message so a cached non-success result isn't logged as "authenticated from session".

**Behavioral note**: After this fix, a rejected bind always runs the authentication flow again. It does not use the cached result anymore. Before this fix, a rejected bind was cached for the session TTL. This meant that even a new attempt, with the correct password, could still get the old rejected result from the cache. This change is intentional: the old behavior let a rate-limit or bruteforce rejection get "stuck" in the cache, and it also let attempts skip the flow's bruteforce and rate-limit checks. Now, every failed attempt costs one flow execution, just like any other bind that is not cached as successful yet — this is the correct place for that cost.

refs #9972

## Test plan

- [x] `go build ./internal/outpost/ldap/...`
- [x] `go vet ./internal/outpost/ldap/...`
- [x] `go test ./internal/outpost/ldap/...`
- [x] Unit test for `SessionBinder.Bind()` covering non-success results not being cached (not yet added)


NikolaiBessonov:
@BeryJu / @goauthentik/backend — fix for the same root cause as #9972 (session cache stores every bind outcome, not just successes, since `DirectBinder.Bind()` never returns a non-nil error). Reproduced live, confirmed via logs + LDAP response mismatch. No unit test yet — let me know if that's a blocker. Would appreciate a review 🙏

NikolaiBessonov:
Hi @rissson! I rebased this branch on the latest `main` (no code changes, just picked up upstream commits), so the workflows need approval again.

Also, the earlier CI failure ([test_authorization_logout](https://github.com/goauthentik/authentik/actions/runs/30645772403/job/91219169154?pr=24582)) is not related to my change. The test itself passed, but the test database flush failed due to a PostgreSQL deadlock — looks like a race condition in test setup. If needed to fix that, let me know.

Could you approve the workflows? Thanks!

NikolaiBessonov:
@rissson gentle ping — still waiting on workflow approval since Aug 3. Could you take a look? cc @BeryJu

NikolaiBessonov:
@rissson Two checks are failing (cargo-deny and test-migrations-from-stable), but both are unrelated to my fix — they come from an outdated base commit, not from my code. They're already fixed on the latest main. All LDAP-related checks pass. Is it OK to merge like this, or should I rebase? cc @BeryJu

abcdegorov:
@rissson, @BeryJu, hi! please take a look at this whenever possible - this is blocking some important LDAP outpost functionality

abhisheksharma2411 (commented):
Verified the load-bearing claim rather than taking the comment's word for it, and it holds exactly. Every return in `DirectBinder.Bind` (`internal/outpost/ldap/bind/direct/bind.go`) pairs a result code with a `nil` error:

```
 49:  return ldap.LDAPResultInvalidCredentials, nil
 59:  return ldap.LDAPResultInvalidCredentials, nil
 73:  return ldap.LDAPResultInsufficientAccessRights, nil
 83:  return ldap.LDAPResultOperationsError, nil
 97:  return ldap.LDAPResultOperationsError, nil
106:  return ldap.LDAPResultSuccess, nil
```

So `err == nil` was true for **every** outcome and the old guard cached all of them. Not a subtle mis-scoping — the condition never excluded anything.

`LDAPResultOperationsError` is the one that makes this more than tidiness: it is the transient case (flow execution failed), and caching it meant a user whose credentials were *correct* was refused for the remainder of the session TTL after one blip, keyed on `{DN, Password}` so retrying the same correct password kept hitting the poisoned entry. Fixing the log line matters as much as the cache write — the old code reported `"authenticated from session"` while returning `InvalidCredentials`, which is precisely the log you would be staring at during the incident.

## The new `else` branch is unreachable

`sb.sessions.Set(...)` at line 75 is the only write into the cache, and after this change it is reached only when `result == ldap.LDAPResultSuccess`. So nothing can ever store a non-success value, and:

```go
} else {
    sb.log.WithField("bindDN", req.BindDN).WithField("result", item.Value()).Info("returning cached non-success bind result")
}
```

cannot fire. The cache is in-process, so it does not even survive a restart carrying old entries. Harmless, and I can see the instinct — but it reads as "this state is expected", which is the opposite of what the PR establishes. I would either drop it, or keep it with a line saying it is a belt-and-braces assertion that should be unreachable, so a future reader does not infer that non-success entries are a thing that happens.

## One consequence worth naming in the description

Every rejected bind now re-executes the flow instead of being served from cache. I think that is **correct** and would not change it — the bruteforce and rate-limit policies live in the flow, so caching a denial meant the limiter stopped seeing repeat attempts, and the same cache entry locked out the legitimate user. Both of those are fixed by this.

It is still a load change: credential spraying now costs a flow execution per attempt where it previously cost a map lookup. That is the intended place for that cost to land, but it is the kind of thing worth being deliberate about in the PR body rather than discovered later.

## Test

There is no test in the diff, and no `*_test.go` under `internal/outpost/ldap/bind/` at all today — the nearest are `ldap_test.go`, `entries_test.go`, `search/request_test.go`. Given the failure mode is "a transient error is cached and replayed as an authentication decision", a small table test over `SessionBinder.Bind` with a stubbed `DirectBinder` returning each of the five non-success codes, asserting nothing lands in `sessions`, would pin exactly the thing that regressed — and would have failed against the old condition. Worth it if you are willing; the change is right either way.

*AI usage disclosure, per [AI_POLICY.md](https://github.com/goauthentik/authentik/blob/main/AI_POLICY.md): written with AI assistance (Claude). I read `bind.go`, `memory.go` and the cache write path myself and can defend each point without it.*

NikolaiBessonov:
@abhisheksharma2411 

Thanks for the review. I fixed all three points:
1. You are right, it can never run. I've removed dead code
2. added a note in the PR description about it.
3. I've added test. It checks that a failed bind is not cached, and a cached success is served without calling the flow again. And it fails on the old code (it even panics there, because the session cookie is still nil in that case)

Could you run workflow, review and merge it ?

abhisheksharma2411 (commented):
Sorry for the delay coming back — you pushed both of these on 26 Aug and I left it sitting.

You've addressed everything, and the behavioural note in the description is better than what I suggested: naming the bruteforce/rate-limit skip as a *second* bug the old cache caused, rather than just as a cost, is the right framing.

The dead `else` is gone and the guard now reads `err == nil && result == ldap.LDAPResultSuccess` with a comment that explains why `err == nil` alone was never a filter. That comment is the part that stops this regressing.

The tests do the job. `assert.Equal(t, 0, sb.sessions.Len())` fails directly against the old condition, and asserting `calls == 2` on the second attempt pins the "flow must re-execute" half, which is the behaviour the fix trades for. Checking the cached-success path in the other test means a future tightening can't quietly disable caching altogether and still pass.

Two small things, neither worth another round on its own.

**The transient code is the one not covered.** The table walks `LDAPResultInvalidCredentials`; `LDAPResultOperationsError` is the outcome that actually motivated this — a flow execution failure, i.e. the case where the user's credentials are *correct* and a blip locked them out for the session TTL. Adding it (and `LDAPResultInsufficientAccessRights`) to the same table is a two-line change and covers the failure the PR is named for. Genuinely optional, since the guard is `== LDAPResultSuccess` and so treats all of them identically by construction — it's about what the test documents, not about coverage of a distinct code path.

**Stale checkbox.** The test plan still reads:

```
- [ ] Unit test for `SessionBinder.Bind()` covering non-success results not being cached (not yet added)
```

which is now done.

Looks good to me — the change is right and the test earns its place.

*AI usage disclosure, per [AI_POLICY.md](https://github.com/goauthentik/authentik/blob/main/AI_POLICY.md): written with AI assistance (Claude). I read `memory.go`, the new `memory_test.go` and `bind/direct/bind.go` myself and can defend each point without it.*

NikolaiBessonov:
@abhisheksharma2411 thanks a lot for the fair review
Could you check the latest changes? 

I really hope this is the final change and this PR will be merged asap :)

abhisheksharma2411 (commented):
Both picked up, including the one I explicitly said was optional. Thanks.

The table now covers `LDAPResultOperationsError` and `LDAPResultInsufficientAccessRights` alongside `LDAPResultInvalidCredentials`. `OperationsError` is the one worth having: it is the transient case where the user's credentials are *correct* and a flow blip locked them out for the session TTL, which is the failure this PR is named for. The test now documents that directly rather than by implication from the `== LDAPResultSuccess` guard.

Checkbox is ticked too, though the label still reads "(not yet added)" — trailing text rather than anything that matters.

Nothing further from me; this looks complete.

NikolaiBessonov:
@abhisheksharma2411 @rissson @BeryJu  could you run workflow and approve PR?

rissson:
yeah we don't approve PRs. We review them. Hopefully that ends up in an approval. We'll get to this when we can. cc. @dewi-tik for tracking

NikolaiBessonov:
FYI: We deployed that fix in our production env and it works good. No more invalid cached credentials

PR #24598, opened 2026-07-31
enterprise/agents: add

BeryJu:


authentik/enterprise/agents/models.py
```diff
@@ -0,0 +1,56 @@
...
+        name: str = "",
+        expiring: bool = False,
+        expires=None,
+        policy_behavior: str = ActorPolicyInheritance.MIRROR,
+    ) -> Agent:
+        # An agent is a machine identity, not a login user: mark it as a service account
+        # and disable password auth. It authenticates through its issued API token. By default
+        # it MIRRORs its parent, so it can never exceed the access of the user it acts for.
+        agent = cls.objects.create(
+            username=f"agent-{generate_id()}",
```
rissson:
```suggestion
            username=f"{user.username}-agent-{generate_id()}",
```
Same as for actors. Dunno if that got changed there or not

rissson (approved):
Relatively straightforward

authentik/enterprise/agents/models.py
BeryJu: it did yeah, I forgot to change it here too

web/src/admin/agents/AgentListPage.ts
```diff
@@ -0,0 +1,99 @@
...
+
+@customElement("ak-agent-list")
+export class AgentListPage extends TablePage<Agent> {
+    public override checkbox = true;
+    public override clearOnRefresh = true;
+    public override searchPlaceholder = msg("Search for a agent...");
+    public override pageTitle = msg("Agents");
+    public override pageDescription = msg(
+        "Admin-provisioned delegate identities that access can be requested and granted for, separately from their parent user.",
+    );
```
kensternberg-authentik, line 28: "Admin provisioned delegate identities that can request and be granted access independent of their parent user."

Is 'parent' a standard term here?

web/src/user/agents/UserAgentForm.ts
```diff
@@ -0,0 +1,102 @@
...
+                    { id: "agent.token.description" },
+                )}
+            </p>
+            <form class="pf-c-form pf-m-horizontal">
+                <ak-hidden-text-input
+                    label=${msg("Token", { id: "agent.token.label" })}
+                    value=${ifDefined(this.result?.token)}
+                    input-hint="code"
+                    readonly
+                ></ak-hidden-text-input>
```
kensternberg-authentik, line 87: Needs both `copy` and `download` buttons.

kensternberg-authentik (commented):
Observation: Nowhere on the "User" page in Admin is there any indication the User has a delegated agent.

There's not a lot of information on the Create Agent page that indicates how and when the `responseForm()` would get triggered or shown; I find the criteria and explanation for this page vague.

The managed-role for the user shows up in Admin ⮕ User ⮕ [User] ⮕ Roles... is it there forever, long after the agent has expired?  Can it/should it be hidden (by default) if the associated agent has expired?

Should superusers _always_ be able to create Agents from their User page, even if the flag isn't set?

Should "Can create agent accounts" be a _Permission_ rather than a _Setting_?

BeryJu:
> Observation: Nowhere on the "User" page in Admin is there any indication the User has a delegated agent.

Fixed, there's now a tab in the credentials section

> There's not a lot of information on the Create Agent page that indicates how and when the `responseForm()` would get triggered or shown; I find the criteria and explanation for this page vague.

Its the same as in the service account form

> The managed-role for the user shows up in Admin ⮕ User ⮕ [User] ⮕ Roles... is it there forever, long after the agent has expired?  Can it/should it be hidden (by default) if the associated agent has expired?

The managed role will stay, but the permissions expire, and we purposefully dont hide it

> Should superusers _always_ be able to create Agents from their User page, even if the flag isn't set?

Previously no, now yes

> Should "Can create agent accounts" be a _Permission_ rather than a _Setting_?

Yeah good point actually, moved it to RBAC with a global role

web/src/admin/agents/AgentListPage.ts
BeryJu: I think so? I dont like calling it owner and parent matches the db

GirlBossRush (approved):

PR #24614, opened 2026-08-01
web/flows: complete Plex source sign-in with a forwardUrl redirect instead of a popup

### LINKED ISSUE #24613: Plex source is unusable on mobile: the popup is blocked and the sign-in hand-off strands the user on plex.tv
### Describe the bug

Signing in through a Plex source fails in practice on mobile browsers, and is degraded on desktop. Four related defects in the web flow stage combine to cause this.

**1. The popup is opened outside the user gesture, so it is blocked.**

`popupCenterScreen` in `web/src/common/helpers/plex.ts` wraps `window.open` in a `setTimeout`, which discards transient user activation:

```ts
return new Promise((resolve) => {
    setTimeout(() => {
        const popup = window.open(url, title, `scrollbars=yes,...`);
        resolve(popup);
    });
});
```

Measured blocked on 100 percent of runs on iOS Safari 26.5.2 (iOS 18.7) and on desktop Firefox. Every user therefore takes the "If no Plex popup opens, click the button below" fallback path, on desktop as well as mobile.

**2. When the flow auto-advances, there is no gesture available to fix this with.**

`AutoRedirectController.hostUpdate()` auto-dispatches the source challenge when the identification stage has a single source, no user fields, and no passwordless URL. A Plex-only login flow (a common shape for a Plex portal) reaches the Plex stage with no click at all, so the extra button press is structural: there is no user activation to open a popup from.

**3. The hand-off runs entirely in the opener, and Plex is given no way back.**

`PlexAPIClient.getPin` builds `https://app.plex.tv/auth#!?clientID=...&code=...` and never sets Plex's `forwardUrl`, so the second browsing context has nowhere to return to. On mobile, `window.open` produces a tab rather than a popup: after signing in, the user is left sitting on plex.tv with no indication that anything is waiting on them. Polling the pin, closing the auth window, and advancing the flow all happen in the tab they just left.

**4. `pinPoll` never times out, and the stage never handles its rejection.**

`pinStatus` throws once the pin expires (about 15 minutes), and `pinPoll` recurses on a 500 ms `setTimeout` with no deadline. `PlexLoginInit.firstUpdated` attaches only a `.then`, so the eventual rejection is unhandled and a stale tab shows "Waiting for authentication..." forever.

Minor, in the same code path: the redeem error handler does `window.location.assign("/")`. On deployments that expose only the flow paths through a reverse proxy allowlist, `/` is not reachable, so any error strands the user a second time.

### How to reproduce

1. authentik 2026.5.6 (the relevant code is unchanged on current main).
2. Create a Plex source. Create an authentication flow whose identification stage has no user fields, no password stage, and only that Plex source. Point an application's authentication flow at it.
3. Open the application's login page on iOS Safari (a regular window, not Private Browsing).
4. The flow auto-advances to the Plex stage. No popup opens. Pressing the fallback button opens a new tab; after signing in to Plex, the user stays on plex.tv. The original tab never advances unless the user manually switches back to it while the pin is still being polled.

### Expected behavior

Completing the Plex sign-in returns the user to authentik and finishes the flow, without depending on a popup that current browsers block.

Plex supports exactly this: `app.plex.tv/auth` accepts a `forwardUrl` parameter and navigates back to it once the pin is authorized. I have this working as a local patch on my own deployment: the stage redirects the current tab to Plex with `forwardUrl` set to the flow URL (carrying the pin id and a retry counter in the query string, since Safari partitions web storage in Private Browsing), then on return polls the pin with a bounded deadline, redeems the token, and continues the flow. Verified end to end on desktop and iOS Safari, including the auto-advance case, which needs no gesture at all once it is a redirect.

Two notes from getting that working:

* The forward URL must preserve the flow's existing query string. `next=` is what routes the completed flow back to `/application/o/authorize/`; dropping it makes the flow succeed and then land external users on "Permission denied: Interface can only be accessed by internal users".
* Plex's documented form is `auth#?clientID=...`; the `forwardUrl` behavior was verified with that form, not with the `#!?` form the helper currently builds.

I will open a PR with this change. If maintainers would rather keep the popup and fix `popupCenterScreen` to open the window inside the gesture, that helps the clicked cases but cannot help the auto-advance case, where no gesture exists, so the redirect appears necessary regardless.

### Deployment Method

Docker

### Version

2026.5.6

MrRussLuther:
## Details

Closes #24613

The Plex flow stage opened its popup from inside a `setTimeout`, which has no transient user activation, so current browsers block it and every sign-in fell through to the manual button. When the identification stage auto-redirects to a sole Plex source there is no gesture at all, so a popup can never open in that shape of flow. And because `getPin` never set Plex's `forwardUrl`, users who signed in from the fallback tab were left sitting on plex.tv while the backgrounded opener did the polling.

This replaces the popup on the flow stage with a same-tab redirect:

- The stage sends the tab to app.plex.tv with `forwardUrl` set to the current flow URL, query string included, so `next=` still routes the finished flow back to the application. The pin id and an attempt counter live in namespaced sessionStorage keys rather than the query string: a pin id in the URL would let a crafted link hand a victim an attacker-authorized pin and sign them in as the attacker's Plex account. A blocked storage write reports that and does not redirect, since a trip whose return leg cannot be recognized has no way to complete.
- On return the stage does a single `pinStatus` check instead of polling, since Plex only follows `forwardUrl` once the pin is authorized. An unauthorized pin mints a fresh one, capped at two automatic attempts so a broken return path cannot loop. The counter clears as soon as a return leg arrives, because reaching that point proves the return path works.
- A cancelled sign-in, a failed redeem, a failed mint, blocked storage, and the attempt cap each render an actionable retry state instead of leaving a permanent spinner above a button that does nothing.

The two callers that keep the popup, the admin source form and user source settings, both awaited the pin request before opening the window, so the click's activation was already spent and the popup was blocked regardless of where `window.open` sat. They now open `about:blank` inside the click and navigate the returned window once the pin arrives. `popupCenterScreen` loses its `async` signature with that, since an awaitable signature is what invited the problem, and both callers now handle a `pinPoll` rejection, which is where an expired pin and an unopened popup both end up.

The auth URL uses the `#?` form that app.plex.tv documents, which is the form `forwardUrl` is known to work with.

Tested with this logic running against 2026.5.6 on my own deployment: desktop Chrome and Firefox, and iOS Safari including a Private window over cellular, on a flow whose identification stage auto-advances to Plex with no interaction. The return trip resumes the flow and completes the `/application/o/authorize/` redirect back to the application. sessionStorage does survive the cross-origin round trip in an iOS Safari Private window, which I had assumed would be the weak point of carrying state there. The popup change is reasoned rather than exercised, since my deployment does not use the admin or user-settings paths.

Known limitation, not addressed here: on a flow with multiple sources or user fields, source selection is client-side and the server-side flow plan stays on identification, so a returning user lands back on identification and has to click Plex again. That completes immediately, since the stored pin is already authorized. Tracking the left-off stage server-side is out of scope.


MrRussLuther:
@GirlBossRush & @kensternberg-authentik please let me know if anything is needed for this PR. 🙏

MrRussLuther:
@GirlBossRush @BeryJu I've rebased this onto current main (the conflicts were only the oxfmt reformat) and added the AI disclosure and checklist the template asks for now. Could one of you approve the workflows and take a look when you get a chance? Plex sign-in still breaks the same way on main.

PR #24615, opened 2026-08-01
ci: add names to release-tag input steps for readability

MGPOCKY:
## Problem

In `.github/workflows/release-tag.yml`, the `check` and `changelog-url` steps are unnamed long `run` blocks, so the Actions UI shows generic labels while validating the release version and resolving the changelog URL.

## Changes

Semantics are unchanged: same step `id`s, same `GITHUB_OUTPUT` keys, and the same job/step output wiring.

- Add `name: Validate release version input` to `id: check`
- Add `name: Resolve changelog URL` to `id: changelog-url`

## Test plan
- [ ] Confirm `check-inputs.outputs.version` / `major-version` / `changelog-url` still resolve from the same step ids
- [ ] Confirm the version regex and changelog URL logic are unchanged

### DIFF
--- a/.github/workflows/release-tag.yml
+++ b/.github/workflows/release-tag.yml
@@ -29,14 +29,16 @@ jobs:
     name: Check inputs validity
     runs-on: ubuntu-latest
     steps:
-      - id: check
+      - name: Validate release version input
+        id: check
         env:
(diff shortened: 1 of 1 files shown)


rissson:
Thank you but not required.

PR #24619, opened 2026-08-01
providers/scim: recognize primary application assignments

dominic-r:
authentik explicitly supports SCIM providers as either primary or backchannel providers, but the SCIM detail endpoint and UI only considered backchannel assignments.

This caused valid primary assignments to appear unassigned and exposed an action that could only fail because the application already existed.


kensternberg-authentik (approved):
Nothing particularly surprising here.

PR #24664, opened 2026-08-03
root: don't include debug info for rust release profile

emilburzo:
Because:
- `debug = 2` tells the compiler to make full debug info
- `strip = true` then removes that info from the binary

It seems pointless to keep the debug info, as builders with less RAM have trouble building it in the link optimizer step due the amount of data:
- `debug = 2`: 0.83 GiB
- `debug = 0`: 0.58 GiB

---

`lto = "thin"` might also be something worth exploring, but that can be later on.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -275,7 +275,7 @@ verbose_file_reads = "warn"
 opt-level = 3
 
 [profile.release]
-debug = 0
+debug = false
 lto = "fat"
 strip = true
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -275,7 +275,6 @@ verbose_file_reads = "warn"
 opt-level = 3
 
 [profile.release]
-debug = false
 lto = "fat"
 strip = true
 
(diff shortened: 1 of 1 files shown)


rissson:
There was a reason why I left it like that, but I can't remember why. Could it have something to do with line number for logs or tracebacks?

emilburzo:
> There was a reason why I left it like that, but I can't remember why. Could it have something to do with line number for logs or tracebacks?

Since `strip = true` is in there, I think there's no way that would be possible since there is no debug info? Unless I'm missing some magic

Cargo.toml
```diff
@@ -275,7 +275,7 @@ verbose_file_reads = "warn"
 opt-level = 3
 
 [profile.release]
-debug = 2
+debug = 0
```
rissson:
```suggestion
```

Default for release profile is `debug = false`. https://doc.rust-lang.org/cargo/reference/profiles.html#release

rissson (approved):
Approving in principle, with a small nit

rissson (approved):

PR #24679, opened 2026-08-03
web, packages/geo, server/static: revert Air Gapped Maps Merge Branch (cherry-pick #24254 to version-2026.8)

BeryJu:
Reverts goauthentik/authentik#24602


rissson:
drafting until we reach a consensus on whether to revert or not

PR #24699, opened 2026-08-03
core/sources: reuse group property mapping manager across groups

### LINKED ISSUE #24694: Source group sync rebuilds PropertyMappingManager once per group, adding ~24ms and one query per group to login
### Describe the bug

`SourceFlowManager.__init__` rebuilds the `PropertyMappingManager` once **per synced group** instead of once per login, because it calls `SourceMapper.build_object_properties()` in a loop without passing the `manager` argument that method already accepts.

Each rebuild runs `group_property_mappings.all().select_subclasses()` and recompiles the expression set. The dominant cost is `select_subclasses()` (django-model-utils), not the DB round trip. This makes source group sync **linear in group count with a large constant**, and it applies even when the source has **zero** group property mappings attached — the queryset is evaluated regardless.

`PropertyMappingManager`'s own docstring states its purpose is to "Pre-compile and cache property mappings when an identical set is used multiple times", and it already guards compilation with `__has_compiled`. The source login path never reuses an instance, so that caching is defeated.

https://github.com/goauthentik/authentik/blob/main/authentik/core/sources/flow_manager.py#L112-L121

```python
self.groups_properties = {
    group_id: self.mapper.build_object_properties(
        object_type=Group,
        request=request,
        user=None,
        group_id=group_id,
        **self.user_info,
    )
    for group_id in self.user_properties.setdefault("groups", [])
}
```

`build_object_properties()` then does, per call:

```python
if not manager:
    manager = self.get_manager(object_type, list(kwargs.keys()))
```

The `context_keys` are identical for every iteration (`["group_id"] + user_info keys`), and the mapping queryset does not change mid-login, so one manager could serve the whole loop.

### How to reproduce

1. Configure an OAuth source with group sync — e.g. an Entra ID source with the `GroupMember.Read.All` scope, so `EntraIDType.get_base_user_properties()` returns a populated `groups` list. Any source whose `get_base_user_properties` returns N groups reproduces it.
2. Have a user who is a member of a large number of groups (measured below with 69).
3. Log in as that user and time the group-sync portion of the flow, or instrument `SourceMapper.build_object_properties`.

Measured on 2026.5.5, **zero group property mappings attached to the source**, timing inside the server container:

| Operation | Time/call | Queries/call |
|---|---|---|
| `build_object_properties()` as the login path calls it (no shared manager) | **24.47 ms** | **1.00** |
| `build_object_properties()` with a reused `manager` | **0.01 ms** | **0.00** |

Isolating where that goes:

| Step | Time/call | Queries/call |
|---|---|---|
| `get_manager()` alone (queryset still lazy) | 0.41 ms | 0.0 |
| `get_manager()` + `compile()` — what each group pays | **23.23 ms** | 1.0 |
| `group_property_mappings.all().select_subclasses()` evaluated | 19.21 ms | 1.0 |
| same queryset **without** `select_subclasses()` | 1.08 ms | 1.0 |
| trivial control query (round-trip baseline) | 0.70 ms | 1.0 |

So ≈18.1 ms of the ≈23 ms per group is `select_subclasses()`, and only ≈0.7–1.1 ms is the database round trip. Multiplied by group count, a 69-group user spends **≈1.69 s** of login latency rebuilding an object that could be built once.

For scale, the rest of that user's group sync is ≈0.25 s (`SourceMatcher.get_group_action`, 2.00 queries/group) — i.e. the per-group manager rebuild is roughly **85% of total group-sync cost**.

### Expected behavior

One `PropertyMappingManager` per object type per login, reused across all groups — `build_object_properties()` already supports this via its `manager` parameter. Expected effect from the numbers above: group-sync cost for a 69-group user drops from ≈1.94 s to ≈0.25 s, and the per-group query count drops by one.

Suggested fix — hoist the manager in `SourceFlowManager.__init__`:

```python
groups = self.user_properties.setdefault("groups", [])
group_manager = self.mapper.get_manager(Group, ["group_id"] + list(self.user_info.keys()))
self.groups_properties = {
    group_id: self.mapper.build_object_properties(
        object_type=Group,
        manager=group_manager,
        request=request,
        user=None,
        group_id=group_id,
        **self.user_info,
    )
    for group_id in groups
}
```

A narrower alternative, if callers should not have to know this: memoize the manager on the `SourceMapper` instance keyed by `(object_type, tuple(context_keys))`, so any caller in a loop benefits without changing call sites.

### Additional context

Only the group loop is affected in practice; the `User` path calls `build_object_properties()` once per login, so it pays this cost a single time.

Not specific to Entra ID — any source with `get_base_user_properties()` returning multiple groups hits it. Entra ID is simply where it is most visible, since `EntraIDClient` fetches all direct memberships from `/me/memberOf`.

### Deployment Method

docker-compose

### Version

2026.5.5 — the code at `flow_manager.py#L112-L121` is unchanged on `main` at the time of filing.

cbmn1987:
Closes #24694.

`SourceFlowManager.__init__` builds group properties in a loop without passing the `manager` argument that `SourceMapper.build_object_properties()` already accepts, so every group re-creates the `PropertyMappingManager` — re-evaluating `group_property_mappings.all().select_subclasses()` and recompiling every expression once per group.

`PropertyMappingManager` is explicitly built for reuse ("Pre-compile and cache property mappings when an identical set is used multiple times", and it guards compilation behind `__has_compiled`), so this hoists the instance out of the loop. The `context_keys` are identical on every iteration (`["group_id"] + user_info keys`) and the mapping queryset does not change mid-login, so one manager serves the whole loop.

### Measurements

**⚠️ The numbers in #24694 understated this by roughly 30×.** I filed it having measured with *zero* group property mappings attached, where `compile()` walks an empty queryset. Once **even one** group property mapping exists — the normal case for anyone using them — the per-group cost jumps from ~24 ms to ~615 ms.

`SourceFlowManager.__init__` with **69 groups and 1 group property mapping**, stock 2026.5.5 server image, 5 runs after a warm-up:

| | min | median | max |
|---|---|---|---|
| before | 46,259 ms | **46,868 ms** | 47,423 ms |
| after | 671 ms | **714 ms** | 1,146 ms |

Queries for the same constructor: **28 → 9** for 20 groups (9 for 1 group in both cases, i.e. constant after the change).

Where the per-group cost goes — `select_subclasses()`, not expression compilation:

| Measured, per call | 0 mappings | 1 mapping | 2 mappings |
|---|---|---|---|
| `get_manager()` + `compile()` | 21.02 ms | **615.69 ms** | 612.01 ms |
| `group_property_mappings.all().select_subclasses()` | 19.95 ms | **712.21 ms** | 729.05 ms |
| same queryset without `select_subclasses()` | 1.11 ms | 2.61 ms | 2.88 ms |
| `PropertyMappingEvaluator(...)` construction | — | 0.08 ms | — |
| construction + `.compile()` | — | 0.27 ms | — |

Two things worth noting from that table:

- Constructing and compiling the evaluator is **0.27 ms** — the cost is not expression compilation. It is iterating the `select_subclasses()` queryset.
- The cost is **fixed once the result set is non-empty** rather than per-row (1 mapping ≈ 2 mappings), which is why a single property mapping is enough to trigger it.

That ~710 ms for a one-row `select_subclasses()` looks worth a separate look on its own merits; this PR only stops the login path from paying it once per group.

With zero group property mappings the same change takes 69 groups from ~24.5 ms/group (~1.7 s) to ~0.01 ms/call.

### Test

`test_group_properties_reuse_mapping_manager` asserts the query count for 1 group equals the count for 20, and that every group still receives its own properties from both the base properties and the property mapping. Comparing counts rather than asserting an absolute number keeps it robust against unrelated query changes.

Verified it discriminates: on unpatched code the two counts are 9 vs 28 and the assertion fails; with the change they are both 9 and it passes.

I also confirmed reuse is semantically safe before writing this — a mapping whose output derives from `group_id` returns per-group-correct, byte-identical results across 8 groups with a shared manager (`iter_eval()` re-runs `set_context()` per call, so nothing is captured at compile time). That was the main correctness risk, since a stale context would have returned the first group's values for every group.

### DIFF
--- a/authentik/core/sources/flow_manager.py
+++ b/authentik/core/sources/flow_manager.py
@@ -109,9 +109,15 @@ class SourceFlowManager:
         self.user_properties = self.mapper.build_object_properties(
             object_type=User, request=request, user=None, **self.user_info
         )
+        # Build the group property mapping manager once and reuse it for every group.
+        # Letting build_object_properties() create it per group re-evaluates the mapping
+        # queryset (including select_subclasses()) and recompiles every expression for
+        # each group, which makes group sync scale poorly for users with many groups.
+        groups_manager = self.mapper.get_manager(Group, ["group_id", *self.user_info.keys()])
         self.groups_properties = {
             group_id: self.mapper.build_object_properties(
                 object_type=Group,
+                manager=groups_manager,
                 request=request,
                 user=None,
                 group_id=group_id,
--- a/authentik/core/sources/flow_manager.py
+++ b/authentik/core/sources/flow_manager.py
@@ -109,10 +109,6 @@ class SourceFlowManager:
         self.user_properties = self.mapper.build_object_properties(
             object_type=User, request=request, user=None, **self.user_info
         )
-        # Build the group property mapping manager once and reuse it for every group.
-        # Letting build_object_properties() create it per group re-evaluates the mapping
-        # queryset (including select_subclasses()) and recompiles every expression for
-        # each group, which makes group sync scale poorly for users with many groups.
         groups_manager = self.mapper.get_manager(Group, ["group_id", *self.user_info.keys()])
         self.groups_properties = {
             group_id: self.mapper.build_object_properties(
--- a/authentik/core/tests/test_source_flow_manager.py
+++ b/authentik/core/tests/test_source_flow_manager.py
@@ -1,7 +1,9 @@
 """Test Source flow_manager"""
 
 from django.contrib.auth.models import AnonymousUser
+from django.db import connection
 from django.test import TestCase
+from django.test.utils import CaptureQueriesContext
 from django.urls import reverse
 from guardian.shortcuts import get_anonymous_user
 
@@ -15,7 +17,11 @@ from authentik.lib.generators import generate_id
 from authentik.policies.denied import AccessDeniedResponse
 from authentik.policies.expression.models import ExpressionPolicy
 from authentik.policies.models import PolicyBinding
-from authentik.sources.oauth.models import OAuthSource, UserOAuthSourceConnection
+from authentik.sources.oauth.models import (
+    OAuthSource,
+    OAuthSourcePropertyMapping,
+    UserOAuthSourceConnection,
+)
 from authentik.sources.oauth.views.callback import OAuthSourceFlowManager
 
 
@@ -255,3 +261,38 @@ class TestSourceFlowManager(TestCase):
         self.assertIsInstance(response, AccessDeniedResponse)
 
         self.assertEqual(response.error_message, "foo")
+
+    def build_group_properties(self, group_count: int) -> int:
+        """Build group properties for `group_count` groups, check them, and return the
+        number of queries it took"""
+        group_ids = [f"group-{index}" for index in range(group_count)]
+        with CaptureQueriesContext(connection) as queries:
+            flow_manager = OAuthSourceFlowManager(
+                self.source,
+                self.request_factory.get("/", user=AnonymousUser()),
+                self.identifier,
+                {"info": {"groups": group_ids}},
+                {},
+            )
+        self.assertEqual(len(flow_manager.groups_properties), group_count)
+        for group_id in group_ids:
+            # Each group must get its own properties, from both the base properties and
+            # the property mapping
+            self.assertEqual(
+                flow_manager.groups_properties[group_id],
+                {"name": group_id, "attributes": {"group_id": group_id}},
+            )
+        return len(queries.captured_queries)
+
+    def test_group_properties_reuse_mapping_manager(self):
+        """Test that building group properties doesn't re-fetch and re-compile the
+        source's group property mappings for every single group"""
+        self.source.group_property_mappings.add(
+            OAuthSourcePropertyMapping.objects.create(
+                name=generate_id(),
+                expression="""return {"attributes": {"group_id": group_id}}""",
+            )
+        )
+        # The property mappings are fetched and compiled once per flow manager, so the
+        # query count must not grow with the number of groups
+        self.assertEqual(self.build_group_properties(1), self.build_group_properties(20))
--- a/authentik/core/tests/test_source_flow_manager.py
+++ b/authentik/core/tests/test_source_flow_manager.py
@@ -262,7 +262,7 @@ class TestSourceFlowManager(TestCase):
 
         self.assertEqual(response.error_message, "foo")
 
-    def build_group_properties(self, group_count: int) -> int:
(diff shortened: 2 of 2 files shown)


authentik/core/sources/flow_manager.py
```diff
@@ -109,9 +109,15 @@ def __init__(
         self.user_properties = self.mapper.build_object_properties(
             object_type=User, request=request, user=None, **self.user_info
         )
+        # Build the group property mapping manager once and reuse it for every group.
+        # Letting build_object_properties() create it per group re-evaluates the mapping
+        # queryset (including select_subclasses()) and recompiles every expression for
+        # each group, which makes group sync scale poorly for users with many groups.
```
rissson:
```suggestion
```

AI slop no thank you

authentik/core/tests/test_source_flow_manager.py
```diff
@@ -255,3 +261,38 @@ def test_error_non_applicable_flow(self):
...
+        for group_id in group_ids:
+            # Each group must get its own properties, from both the base properties and
+            # the property mapping
+            self.assertEqual(
+                flow_manager.groups_properties[group_id],
+                {"name": group_id, "attributes": {"group_id": group_id}},
+            )
+        return len(queries.captured_queries)
+
+    def test_group_properties_reuse_mapping_manager(self):
```
rissson, line 287: I'm not sure I want to add a test for this specific thing. @BeryJu thoughts?

rissson (changes requested):

authentik/core/tests/test_source_flow_manager.py
BeryJu: I dont mind, but `build_group_properties` should be called `calculate_group_property_mapping_queries`, and it probably shouldn't be compared equal but compared against a threshold (there might be other random queries counted into this that cause the test to fail)

rissson: Yeah that's the reason why I don't want a test here :/ But I guess a threshold is good enough

BeryJu: threshold is what we do for policy engine tests to ensure they aren't O(2n)

authentik/core/sources/flow_manager.py
cbmn1987: Removed.

authentik/core/tests/test_source_flow_manager.py
cbmn1987: Renamed the helper to `calculate_group_property_mapping_queries` and switched the assertion to `assertLess` against a threshold — building 100 groups must take fewer than 100 queries, same shape as the policy engine tests.

rissson (approved):

PR #24716, opened 2026-08-03
providers/saml: Metadata parser to only pull usable options

### LINKED ISSUE #21009: SAML Provider metadata import selects first endpoint even if binding is unsupported

PeshekDotDev:
### What does this PR change?

Currently, the metadata parser for saml providers pulls the first available option, even if we don't support it. This instead pulls the first valid option from provided metadata.

Without it, we can easily import improper options to a saml provider

### Linked issues

Closes #21009

---

## Checklist


authentik/providers/saml/api/providers.py
```diff
@@ -335,10 +335,7 @@ class SAMLProviderViewSet(UsedByMixin, ModelViewSet):
                 name="force_binding",
                 location=OpenApiParameter.QUERY,
                 type=OpenApiTypes.STR,
-                enum=[
-                    SAML_BINDING_REDIRECT,
-                    SAML_BINDING_POST,
-                ],
+                enum=list(SAML_BINDINGS_SUPPORTED),
```
BeryJu:
```suggestion
                enum=SAML_BINDINGS_SUPPORTED,
```

should be fine without the cast

BeryJu (approved):
small nit

PR #24718, opened 2026-08-04
providers/oauth2: send back-channel logout requests when a user is deactivated

PeshekDotDev:
### What does this PR change?

This sends an oidc backchannel logout signal when a user is deactivated.

Without it, when an admin deactivates a user, they will be kept logged in to services configured for backchannel oidc logout. This could lead to a user getting access to information that they should no longer be allowed access to.

### How was this tested?

Both through unit tests and manual testing through the browser

### Linked issues

Closes #20420

---

## Checklist


authentik/providers/oauth2/signals.py
```diff
@@ -94,31 +115,30 @@ def user_session_deleted_oauth_backchannel_logout_and_tokens_removal(
...
+    """Send back-channel logout requests and remove user tokens when deactivated"""
     if instance.is_active:
         return
-    AccessToken.objects.including_expired().filter(user=instance).delete()
+
+    access_tokens = (
+        AccessToken.objects.including_expired().select_related("provider").filter(user=instance)
+    )
+
+    revocations = backchannel_logout_revocations(access_tokens)
```
rissson: I don't think this is necessary since we delete the user's sessions when deactivating them, no?

PeshekDotDev: Well, in theory, but not always. We don't have an official signal listening for user post save and checking for is_active = false to delete sessions. So if someone is deactivated via scim, ldap, kerberos, blueprint, or scheduled offboarding has revoke sessions disabled, or someone uses a put request instead of patch to update a user, then the user will be deactivated but their session will not be killed until they visit authentik again via `authentik/core/middleware.py:63`. and by then, the logic that existed before in this `user_deactivated` signal would have already deleted their tokens, so no ability so send the backchannel logout

If we had a global signal that listened for user post save and deleted all sessions if is_active = false, then i could just remove the `access_tokens.delete()` from here and everything would work as expected. Maybe that's the real issue?

But in it's current state, we would either have to:

Add backchannel logout into this section and keep the token deletion. Because this lets us guarantee that we are logging out and deleting access tokens if a bug ever causes someone to be deactivated but not delete their session

or

We guarantee that deactivation always deletes all sessions, and then we remove access_tokens.delete() from this function.

I think if we do want to go the second route, we should still merge this and then make an issue for that. I figured someone else would have more context as to why we delete tokens in here etc and why this is the way it is. probably just being careful before and there wasn't a thought about backchannel logout

rissson: Yeah I don't think we should add custom logic here. We should instead fix the fact that sessions are not always deleted when a user is deactivated. I don't want custom code here for oauth, 'cause then we need to duplicate that for every provider/other things that rely on knowning when a user is deactivated via session removal.

PeshekDotDev: Yeah that does make more sense. We can skip deleting access tokens on deactivation so that the regular session delete can do backchannel logout, and I can look into the edgecases later

PeshekDotDev:
#25088 actually fixes this issue, so we can just merge in the tests now

rissson (approved):

PR #24734, opened 2026-08-04
web/admin: add user app entitelments screen

### LINKED ISSUE #23194: Easier method to find application entitlements that a user has
At the moment this requires going to each application or several API calls. Would be better if this could be checked directly on a user's account via the user directory and/or an API call that references the user.

atereshkin:
### What does this PR change?

This adds a "user application entitlements" screen in the admin UI where a single user's entitlements can be observed across all applications. The "Applications" tab in the user view is now split into two "subtabs" - "Applications" and "Application entitlements", the former preserving the old Applications view and the latter showing entitlements. 

The entitlements API endpoint now accepts `for_user` query parameter that allows filtering by individual user.

To simplify reviewing a user's entitlements, see #23194

### How was this tested?

Manual test + a few new tests in the suite for the new API endpoint.

### Linked issues

closes #23194
---

## Checklist

### DIFF
--- a/authentik/core/models.py
+++ b/authentik/core/models.py
@@ -483,21 +483,28 @@ class User(SerializerModel, AttributesMixin, AbstractUser):
     def app_entitlements(self, app: Application | None) -> QuerySet[ApplicationEntitlement]:
         """Get all entitlements this user has for `app`."""
         if not app:
-            return []
+            return ApplicationEntitlement.objects.none()
+        return self.all_app_entitlements().filter(app=app)
+
+    def all_app_entitlements(self) -> QuerySet[ApplicationEntitlement]:
+        """Get all entitlements this user has, across all applications."""
         all_groups = self.all_groups()
-        qs = app.applicationentitlement_set.filter(
-            Q(
-                Q(bindings__user=self) | Q(bindings__group__in=all_groups),
-                bindings__negate=False,
+        return (
+            ApplicationEntitlement.objects.filter(
+                Q(
+                    Q(bindings__user=self) | Q(bindings__group__in=all_groups),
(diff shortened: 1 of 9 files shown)


rissson (approved):

authentik/core/models.py
```diff
@@ -483,21 +483,28 @@ def group_attributes(self, request: HttpRequest | None = None) -> dict[str, Any]
...
+                    Q(bindings__user=self) | Q(bindings__group__in=all_groups),
+                    bindings__negate=False,
+                )
+                | Q(
+                    Q(~Q(bindings__user=self), bindings__user__isnull=False)
+                    | Q(~Q(bindings__group__in=all_groups), bindings__group__isnull=False),
+                    bindings__negate=True,
+                ),
+                bindings__enabled=True,
             )
```
BeryJu: Kinda unrelated to this but we can use `FilterPolicyEngine` for this now so we don't need to repeat this query

BeryJu (changes requested):
Kinda torn between having a tab with all entitlements and making the user-app table expandable and showing the entitlements scoped in there

atereshkin:
> Kinda torn between having a tab with all entitlements and making the user-app table expandable and showing the entitlements scoped in there

The latter option is what I considered first in fact, but it seems that would only go half way towards addressing the original need - seeing all the entitlements  at once. It would still require clicking (and lots of it if you have more than a few apps) to do so. 
I don't know how much sense entitlements make outside of the context of a specific app, but it seems to be the premise of this entire feature that they do.

authentik/core/models.py
atereshkin: Done.

PR #24770, opened 2026-08-04
web/elements: show certs that can't be selected as grayed out

### LINKED ISSUE #24738: Show filtered certificates as grayed out instead of not showing
### Describe the bug

At the current moment, you can add a certificate that isn't supported by saml, and when you select a saml cert, the cert doesn't show. What we should do instead is show it as grayed out so the user knows the cert isn't missing, but also knows they can't select it for that purpose

### How to reproduce

Add a PQ cert
Go to saml provider
Click encryption cert selector
Cert does not show

### Expected behavior

You should see the cert underneath all valid certs and it should be grayed out. highlighting it should give detail that the cert cannot be used for this purpose

### Screenshots

_No response_

### Additional context

_No response_

### Deployment Method

Other (please specify)

### Version

2026.8

### Relevant log output

```shell

```

PeshekDotDev:
### Linked issues

Closes #24738

---

## Checklist


PeshekDotDev:
Screenshot 2026-08-04 at 6 08 02 PM

web/src/admin/common/ak-crypto-certificate-search.ts
```diff
@@ -104,26 +105,69 @@ export class AkCryptoCertificateSearch extends CustomListenerElement(AKElement)
...
-        const certificates = await aki(CryptoApi).cryptoCertificatekeypairsList(args);
-        return certificates.results;
+
+        return [...usable, ...unusable];
     };
 
+    optionDisabled = (item: CertificateKeyPair): boolean => this.#unusableReason(item) !== null;
+
+    renderDescription = (item: CertificateKeyPair): SlottedTemplateResult =>
+        this.#unusableReason(item) ?? nothing;
```
kensternberg-authentik: Line 368 of `SearchSelect.ts` shows that `renderDescription()` should return a `null`, rather than `nothing`.  `nothing` isn't nullish, so `renderMenuItems` will still render the description block, it'll just be empty.  That takes up space on the page and makes scrolling through the valid options take longer.  Return `null` here for consistency.

web/src/admin/common/ak-crypto-certificate-search.ts
```diff
@@ -104,26 +105,69 @@ export class AkCryptoCertificateSearch extends CustomListenerElement(AKElement)
...
         }
-        if (this.allowedKeyTypes?.length) {
-            args.keyType = this.allowedKeyTypes;
+        const { results } = await aki(CryptoApi).cryptoCertificatekeypairsList(args);
+
+        const usable: CertificateKeyPair[] = [];
+        const unusable: CertificateKeyPair[] = [];
+
+        for (const item of results) {
+            (this.#unusableReason(item) ? unusable : usable).push(item);
```
kensternberg-authentik: CryptoAPI is paginated, and clamps to 20 by default.  SearchSelect can only handle a single API page of selectable items, which is its own unsolved problem.  If you have 20 verification certificates and three signing ones, it's possible you'll get a SearchSelect of 20 unusable items and not be able to pick a certificate you want.

web/src/admin/common/ak-crypto-certificate-search.ts
PeshekDotDev: Always love situations like that lol. Fixed

web/src/admin/common/ak-crypto-certificate-search.ts
PeshekDotDev: I made a filtered call and then a second call after that fills up the rest if usable is under 20. Let me know what you think

web/src/admin/common/ak-crypto-certificate-search.ts
```diff
@@ -104,26 +105,85 @@ export class AkCryptoCertificateSearch extends CustomListenerElement(AKElement)
...
+        }
+
+        if (
+            this.allowedKeyTypes?.length &&
+            (!item.keyType || !this.allowedKeyTypes.includes(item.keyType))
+        ) {
+            return msg("This certificate is not a valid option here: unsupported key type.", {
+                id: "crypto.certificate-search.unsupported-key-type.description",
+            });
+        }
```
kensternberg-authentik, line 129: If I'm reading this right, this _re-implements_ the code in `crypto/api.py:CertificateKeyPairFilter`.  I don't suppose there's a way to move all this logic entirely into the server and deliver a richer data set from `cryptoCertificatekeypairsList()`?

web/src/admin/common/ak-crypto-certificate-search.ts
```diff
@@ -104,26 +105,85 @@ export class AkCryptoCertificateSearch extends CustomListenerElement(AKElement)
...
             args.search = query;
         }
+
+        const restrictions: CryptoCertificatekeypairsListRequest = {};
+        if (!this.noKey) {
+            restrictions.hasKey = true;
+        }
         if (this.allowedKeyTypes?.length) {
-            args.keyType = this.allowedKeyTypes;
+            restrictions.keyType = this.allowedKeyTypes;
```
kensternberg-authentik, line 147: CryptoAPI is paginated, and clamps to 20 by default.  SearchSelect can only handle a single API page of selectable items, which is its own unsolved problem.  If you have 20 verification certificates and three signing ones, it's possible you'll get a SearchSelect of 20 unusable items and not be able to pick a certificate you want.

kensternberg-authentik: As far as I can tell, this issue has not been addressed.

web/src/admin/common/ak-crypto-certificate-search.ts
```diff
@@ -104,26 +105,85 @@ export class AkCryptoCertificateSearch extends CustomListenerElement(AKElement)
...
+        if (Object.keys(restrictions).length === 0) {
+            const { results } = await api.cryptoCertificatekeypairsList(args);
+            return results;
         }
-        const certificates = await aki(CryptoApi).cryptoCertificatekeypairsList(args);
-        return certificates.results;
+
+        // The API only returns one page of results, so ask it for the usable keypairs directly.
+        // Otherwise, a page full of unusable ones could push the usable ones out of the list.
+        const [{ results: usable }, { results: all }] = await Promise.all([
```
kensternberg-authentik, line 159: `Promise.all` is a code smell these days. If one of these requests succeeds and the other fails (unlikely, I know, but weirder things have happened), this will crash silently, leaving the page in a weird state.  Consider using [_AllSettled_](https://www.greatfrontend.com/questions/quiz/how-is-promiseall-different-from-promiseallsettled) and unpacking the result by hand instead.

kensternberg-authentik: This _might_ be a good place to put a `{ results: unusable }` with the adjustments wanted on line 129 instead.  I know enough about DRF to know that you can invert the filter _and_ provide extra fields in the serializer (like your "reason" fields), although I don't know how OpenAPI feels about that ... the last time I tried things got weird.

web/src/elements/ak-list-select/ak-list-select.ts
```diff
@@ -320,16 +350,26 @@ export class ListSelect extends AKElement implements IListSelect {
...
-            ([value, label, desc]: SelectOption) => html`
+        return options.map(([value, label, desc]: SelectOption) => {
+            const disabled = this.#disabledOptions.has(value);
+
+            return html`
                 <li
                     role="option"
                     value=${value}
                     class="ak-select-item"
+                    aria-disabled=${ifDefined(disabled || undefined)}
```
kensternberg-authentik, line 361: Put the `aria-disabled` expression on the button in addition to having it here. The button gets focus, not the list item. Having it here tells the listener what to skip; having it on the button will let someone using accessibility aids hear that the object is disabled.

web/src/elements/ak-list-select/ak-list-select.ts
kensternberg-authentik, line 235: This doesn't check if the item is disabled. Disabled rows can still be accessed via the keyboard (which is good, because you want a disabled user to be able to access the row and see/hear the reason for why it's disabled), but now the row can be selected via the keyboard.  Don't let the code set `aria-selected` when  the certificate isn't selectable.

kensternberg-authentik, line 381: Consider moving this _outside_ and below the button object, then linking it to the button by [`aria-describedby`](https://accessiblyapp.com/blog/aria-describedby/).

kensternberg-authentik (approved):
On the server, there is a difference between `key_data__exact=""` (this certificate has no key) and `key_data__exact=None`, the latter of which is a rare case where the database has been able to correlate the algorithm with its known set. While rare, this seems wrong; maybe you need a third error type for this distinction?

web/src/admin/common/ak-crypto-certificate-search.ts
PeshekDotDev: It has been. I verified twice in the browser with manual testing

web/src/admin/common/ak-crypto-certificate-search.ts
PeshekDotDev: I altered this to use promise.allSettled

web/src/elements/ak-list-select/ak-list-select.ts
PeshekDotDev: fixed

web/src/elements/ak-list-select/ak-list-select.ts
PeshekDotDev: fixed

PeshekDotDev: fixed

mo74m3ed (commented):
[]()

PR #24792, opened 2026-08-05
tasks: aggregate status from logs instead of legacy field

melizeche:
`Task.aggregated_status` is set by a trigger that reads the `_messages` field. `0005_tasklog` moved logs into the `TaskLog` table but left the trigger untouched, and nothing writes `_messages` anymore, so `aggregated_status` can only ever equal the task's state, and `info`/`warning`/`error` are unreachable.

As a result, tasks that log errors display as **Done**, the "exclude successful" task view matches nothing, and the status summary always reports zero for those levels. 13 non-test `self.error()` / `self.warning()` call sites are invisible in the UI. Tasks that raise are unaffected, `rejected` is a state, not log-derived.

## How

- Trigger aggregates over `TaskLog`, as explicit plpgsql `IF/ELSE` so the lookup only runs on the transition to `done`, hitting the existing `(task_id, previous)` index.
- Batched backfill for tasks left on `done` despite warning/error logs; it reuses the trigger, so the query only selects affected rows.

## Example

before with a broken blueprint: all good
image

after:

image

---

## Checklist


authentik/tasks/models.py
```diff
@@ -76,19 +76,23 @@ class Meta(TaskBase.Meta):
                 name="update_aggregated_status",
                 operation=pgtrigger.Insert | pgtrigger.Update,
                 when=pgtrigger.Before,
+                # Logs live in TaskLog (see 0005_tasklog), so the status of a finished task
+                # is aggregated from its log rows. The table is referenced unqualified so
+                # that it resolves through the search path to the current tenant's schema.
```
rissson: This seems AI generated. Let's remove it.

rissson (approved):
approving provided the added tests pass

authentik/tasks/migrations/0008_remove_task_update_aggregated_status_and_more.py
```diff
@@ -0,0 +1,72 @@
...
+
+import pgtrigger.compiler
+import pgtrigger.migrations
+from django.db import migrations
+
+# Number of tasks to re-aggregate per statement
+BATCH_SIZE = 1000
+
+
+def backfill_aggregated_status(apps, schema_editor):
```
rissson: Thinking about this, I'd rather not backfill. Old tasks will have a wrong status, but I think it's fine.

rissson (approved):
approving pending one change

authentik/tasks/migrations/0008_remove_task_update_aggregated_status_and_more.py
melizeche: I think there's value in backfilling tasks, admins could get info they missed

PR #24794, opened 2026-08-05
stages/password: lock password login after failed attempts

dominic-r:


authentik/stages/password/api.py
```diff
@@ -1,21 +1,31 @@
 """PasswordStage API Views"""
 
+from django.utils.translation import gettext_lazy as _
+from rest_framework.exceptions import ValidationError
 from rest_framework.viewsets import ModelViewSet
 
 from authentik.core.api.used_by import UsedByMixin
+from authentik.enterprise.license import LicenseKey
```
BeryJu: can't import enterprise from oss

authentik/stages/password/api.py
```diff
@@ -1,21 +1,31 @@
...
 from authentik.flows.api.stages import StageSerializer
 from authentik.stages.password.models import PasswordStage
 
 
 class PasswordStageSerializer(StageSerializer):
     """PasswordStage Serializer"""
 
+    def validate_failed_attempts_before_lockout(self, limit: int) -> int:
```
dominic-r: use conditional inheretance

BeryJu (commented):
Would also imo be nicer if it was more agnostic but that requires big open-ended changes from the base PR

Also need to make sure PasswordDevice.objects.select_for_update

dominic-r:
> Would also imo be nicer if it was more agnostic but that requires big open-ended changes from the base PR

Call the strings like authenticator instrad then filter by device or similar

PR #24795, opened 2026-08-05
core: add composite user status

dominic-r:
### Linked issues

---

## Checklist


authentik/core/tests/test_users.py
```diff
@@ -173,3 +174,37 @@ def test_password_hash_ignored_outside_blueprint_context(self):
 
         self.assertTrue(serializer.is_valid(), serializer.errors)
         self.assertNotIn("password_hash", serializer.validated_data)
+
+
+class TestUserStatus(TestCase):
```
BeryJu, line 179: nit: separate test case should be in separate file

dominic-r:
Superseded: composite status was dropped; the user API now exposes password_locked in #24794.

PR #24838, opened 2026-08-06
web/bugfix: Repair or replace bad color definitions in the UI

kensternberg-authentik:
web/bugfix: Repair or replace bad color definitions in the UI

These are six different cases where a survey of the colors turned up… nothing. In three cases, there are typos: the CSS custom property simply didn’t exist in Patternfly. In one, the color was originally defined by a non-existent property (`danger-color--50`), then re-assigned to `palette--red-100`. I’ve pulled the re-assignment upward. And in `Sidebar.css`, these two colors are referenced, but do not exist in Patternfly. I’ve extrapolated from the existing palette to provide those definitions, and they look pretty good.

To review these changes, I needed to update Storybook so that it now had an _explicit_ collection of valid paths, including the fonts path.  The fonts path resolves to a path outside the `./web` folder, so it needed to be added, and then the project itself needed a new explicit permission since using the `server.fs.allow` field is no longer allowed to default.

## Checklist

-   [🌿] The project has been linted, built, and tested (`make all`)
-   [🌿] The documentation has been updated and formatted (`make docs`)


web/src/elements/ak-dual-select/components/ak-dual-select-controls.ts
```diff
@@ -37,7 +37,7 @@ export class AkDualSelectControls extends CustomEmitterElement<DualSelectEventTy
             }
 
             .pf-c-button {
-                --pf-c-button--m-plain--Color: var(--pf-global--Color-300) !important;
+                --pf-c-button--m-plain--Color: var(--pf-global--Color--300) !important;
```
kensternberg-authentik, line 40: So important we didn't even check to see if it was defined.  _Sigh_.

GirlBossRush (approved):

PR #24839, opened 2026-08-06
internal/outpost/ldap: type switching for uidNumber/gidNumber

tacerus:
### What does this PR change?

Treat both integer and string uidNumber/gidNumber attribute values equally.

When the uidNumber attribute was set as an integer without an explicit gidNumber, the system would generate a gidNumber instead of mapping the uidNumber. The mapping only worked if uidNumber was set as a string.

As uidNumber/gidNumber are commonly integers, avoid the need for faking strings in the user attributes by matching the behavior.

This also reduces implementation confusion, as the documentation states

> The gidNumber attribute of each virtual group is equal to the uidNumber of the user.

without imposing any limitations on the data type.

### How was this tested?

Define a user with a custom `uidNumber` attribute. Set the attribute to a number: first a string, then an integer one. Observe the result when querying the uidNumber/gidNumber attributes of the user and the users virtual group.

### Linked issues

---

## Checklist

### DIFF
--- a/internal/outpost/ldap/utils.go
+++ b/internal/outpost/ldap/utils.go
@@ -7,6 +7,22 @@ import (
 	api "goauthentik.io/packages/client-go"
 )
 
+func getAttributeString(attributes map[string]any, key string) (string, bool) {
+	val, ok := attributes[key]
+	if !ok || val == nil {
+		return "", false
+	}
+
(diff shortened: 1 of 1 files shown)


tacerus:
CI failure appears not related to my change, but let me know if I missed something.

PR #24844, opened 2026-08-06
enterprise/requests: integrate agents with requests

BeryJu:
currently, an agent user always inherits full perms from its parent user, which is great for ease of use but not great for lease privilege.

This wires the agents system into requests, and adds APIs that can be used by the future MCP server to request access for different apps/entitlements on an agent user.

If the owner already has access to the application, the owner can approve access directly, if the owner doesn't have access yet then the owner needs to approve it _and_ it goes through the normal requests flow


kensternberg-authentik (approved):

GirlBossRush (approved):

rissson (approved):
There are places that could be optimised in terms of number of DB queries, but we can always loop back on that later

PR #24845, opened 2026-08-06
flows: allow splicing a planned flow into an active plan

dominic-r:
Adds `FlowPlan.insert_plan()` to insert another planned flow's bindings and markers into an active plan, and a `check_authentication` toggle on `FlowPlanner` so callers that have already established the user's identity mid-flow can plan a flow for a pending user. Groundwork for running required action flows inside a login, used by the next PR in this stack.


dominic-r:
@gergosimonyi 's infinite wisdom has permitted me to close this PR!

dominic-r:
(We discussed a better implementation that would not require this change)

PR #24850, opened 2026-08-06
web/admin: add next actions card on the user page

dominic-r:
Adds a Next actions card to the user overview for viewing, adding, and removing required flows without editing user attributes directly. The card remains available whenever an Enterprise license is installed, including after it expires.

### DIFF
--- a/website/docs/users-sources/user/next_actions.mdx
+++ b/website/docs/users-sources/user/next_actions.mdx
@@ -22,7 +22,11 @@ If a listed flow doesn't exist or its policies deny the user, the session remain
 
 ## Require an action for a user
 
-Set the `goauthentik.io/user/next-actions` attribute on the user, through the user API, a [blueprint](../../customize/blueprints/index.mdx), or from within a flow, for example with a [User Write stage](../../add-secure-apps/flows-stages/stages/user_write/index.mdx) in an enrollment flow:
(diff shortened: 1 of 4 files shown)


website/docs/users-sources/user/next_actions.mdx
```diff
@@ -22,7 +22,11 @@ If a listed flow doesn't exist or its policies deny the user, the login fails un
 
 ## Require an action for a user
 
-Set the `goauthentik.io/user/next-actions` attribute on the user, through the user API, a [blueprint](../../customize/blueprints/index.mdx), or from within a flow, for example with a [User Write stage](../../add-secure-apps/flows-stages/stages/user_write/index.mdx) in an enrollment flow:
+1. Log in to authentik as an administrator and open the authentik Admin interface.
+2. Navigate to **Directory** > **Users** and click the name of the user.
+3. In the **Next actions on login** card, select a flow and click **Add**. Pending actions are listed in the same card and can be removed there.
+
+Because the actions are stored in the `goauthentik.io/user/next-actions` attribute, they can also be set through the user API, a [blueprint](../../customize/blueprints/index.mdx), or from within a flow, for example with a [User Write stage](../../add-secure-apps/flows-stages/stages/user_write/index.mdx) in an enrollment flow:
```
dewi-tik, line 29: They could also be set via the usual YAML codebox.

PR #24877, opened 2026-08-07
events: use absolute URLs (constructed using base_url from the tenant) when sending notifications outside (e.g. via email)

atereshkin:
### What does this PR change?

This makes notifications containing a hyperlink use absolute URLs when they are going outside of the system (i.e. via email, Slack or webhook). 

We've added an ability to include a hyperlink in notifications earlier but it was useless for external notifications because those hyperlinks were relative URLs. 

### How was this tested?

Added a few new tests specifically for this + manually by triggering an access review notification.

---

## Checklist


rissson (approved):

authentik/events/models.py
```diff
@@ -657,6 +661,14 @@ class Notification(SerializerModel):
     seen = models.BooleanField(default=False)
     user = models.ForeignKey(User, on_delete=models.CASCADE)
 
+    @property
+    def hyperlink_absolute(self) -> str | None:
+        """The hyperlink resolved against the tenant's configured base URL, for use
+        outside of the authentik UI where relative URLs do not resolve"""
+        if not self.hyperlink:
+            return self.hyperlink
```
```suggestion
            return None
```

authentik/events/models.py
```diff
@@ -594,14 +596,16 @@ def send_email(self, notification: Notification) -> list[str]:
...
                 "label": notification.hyperlink_label,
             }
         if notification.event:
             context["title"] += notification.event.action
             for key, value in notification.event.context.items():
                 if not isinstance(value, str):
                     continue
+                if key == "hyperlink":
+                    value = build_absolute_url(value)  # noqa: PLW2901
                 context["key_value"][key] = value
```
```suggestion
                context["key_value"][key] = value
                if key == "hyperlink":
                    context["key_value"][key] = value = build_absolute_url(value) 
```

I'd prefer this instead of disabling errors

authentik/tenants/utils.py
```diff
@@ -31,3 +33,18 @@ def get_unique_identifier() -> str:
...
+    are returned unchanged."""
+    if not url:
+        return url
+    parsed = urlsplit(url)
+    if parsed.scheme or parsed.netloc:
+        return url
+    base_url = get_current_tenant(only=["base_url"]).base_url
+    if not base_url:
+        return url
+    return f"{base_url}/{url.lstrip('/')}"
```
BeryJu: should use urljoin

BeryJu: also needs to respect CONFIG.get("web.path")

website/docs/install-config/configuration/configuration.mdx
```diff
@@ -837,8 +837,6 @@ Set this to the scheme and host only, without a path. All of authentik is served
 
 Defaults to an empty string.
 
-This setting is currently unused until it can be marked as required, starting from authentik version 2026.11.
-
```
BeryJu, line 841: if this PR is getting backported this needs to stay

emilburzo (approved):

BeryJu (changes requested):

authentik/tenants/utils.py
atereshkin: Are you sure? The only current callers of this already pass URLs that already account for CONFIG.get("web.path"). I cannot think of a good case where we'd need to pass an unprefixed URL here. Maybe the function just needs a better name?

BeryJu: fair point, yeah I forgot that that config is used when constructing a URL, in that case maybe just a better name

website/docs/install-config/configuration/configuration.mdx
atereshkin: Reinstated, but I've changed the wording because it's actually used by object lifecycles rather than being 100% unused.

BeryJu (approved):

PR #24887, opened 2026-08-07
root: add optional pyroscope for profiling

BeryJu:
we've experimented with this in the past but pyroscope used to not support memory profiling in pythin. We already have this in go, with the same env var. Mainly a test build for me to use tbh

### DIFF
--- a/authentik/lib/debug.py
+++ b/authentik/lib/debug.py
@@ -1,3 +1,5 @@
+from os import environ
+
 from structlog.stdlib import get_logger
 
 from authentik.lib.config import CONFIG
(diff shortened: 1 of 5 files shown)


authentik/lib/debug.py
```diff
@@ -29,3 +33,31 @@ def start_debug_server(port_offset: int = 0, **kwargs) -> bool:
         return False
     LOGGER.debug("Starting debug server", host=host, port=port)
     return True
+
+
+def start_pyroscope(component: str, **tags: str) -> bool:
+    """Attempt to start the pyroscope profiler in the current process.
+    Must be called *after* forking, the profiler's agent threads do not survive fork().
+    Returns true if the profiler was started, otherwise false"""
+    server = environ.get("AUTHENTIK_PYROSCOPE_HOST")
```
rissson, line 42: I want to start a new top-level config key called `observability` (or something of that nature), in which we can stuff the different implementations we support. So for example:
```yaml
observability:
  pyroscope:
    enable: false
    host: "http://pyroscope"
  sentry: # this would only be for traces and performance stuff, error reporting stays separate I think
    enabled: true
    traces_sample_rate: 0.1
  otlp: # or whatever else
    enabled: true
```

Open to changes on this, I don't have this fully refined in my head yet

rissson (approved):

authentik/lib/debug.py
BeryJu: I agree, but for the time being just a raw env var is fine I think, similar to the go implementation, which is also un-documented

rissson: fair enough!

PR #24890, opened 2026-08-07
providers/oauth2: cache ProviderInfoView get_claims

BeryJu:
optimisation sweep

### DIFF
--- a/authentik/providers/oauth2/views/provider.py
+++ b/authentik/providers/oauth2/views/provider.py
@@ -3,6 +3,7 @@
 from typing import Any
 
 from django.apps import apps
+from django.core.cache import cache
 from django.http import HttpRequest, HttpResponse, JsonResponse
 from django.shortcuts import get_object_or_404, reverse
 from django.views import View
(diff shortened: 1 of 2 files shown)


authentik/providers/oauth2/views/provider.py
```diff
@@ -138,6 +139,16 @@ def get_info(self, provider: OAuth2Provider) -> dict[str, Any]:
             pass
         return config
 
+    def get_claims_cached(self, provider: OAuth2Provider) -> list[str]:
+        """Same as self.get_claims but cached to avoid re-evaluating property-mappings"""
+        key = f"authentik/providers/oauth2/provider_info/claims/{provider.pk}"
```
rissson, line 144: Would it be worth caching on the hash of something like `ScopeMapping.objects.filter(provider=provider).order_by("scope_name").values_list("pk", flat=True)` instead? So we get cache invalidation "for free"

BeryJu: Fair point, I was gonna add a signal on ScopeMapping post_save that prunes the cache for each of the providers it uses

PR #24894, opened 2026-08-07
lib: use msgspec json decoder for requests Session 

BeryJu:
the last one

### DIFF
--- a/authentik/lib/tests/test_utils_http.py
+++ b/authentik/lib/tests/test_utils_http.py
@@ -1,0 +1,63 @@
+"""Test http utils"""
+
+from decimal import Decimal
+
+from django.test import TestCase
+from requests.exceptions import JSONDecodeError
+
+from authentik.lib.config import CONFIG
+from authentik.lib.utils.http import (
+    DebugSession,
+    MsgspecHTTPAdapter,
+    MsgspecResponse,
+    get_http_session,
+)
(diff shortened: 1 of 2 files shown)


authentik/lib/tests/test_utils_http.py
```diff
@@ -0,0 +1,63 @@
...
+        self.assertIsInstance(session.get_adapter("https://goauthentik.io"), MsgspecHTTPAdapter)
+        self.assertIsInstance(session.get_adapter("http://goauthentik.io"), MsgspecHTTPAdapter)
+
+    @CONFIG.patch("log_level", "trace")
+    def test_session_adapters_debug(self):
+        """Test that the debug session also uses the msgspec adapter"""
+        session = get_http_session()
+        self.assertIsInstance(session, DebugSession)
+        self.assertIsInstance(session.get_adapter("https://goauthentik.io"), MsgspecHTTPAdapter)
+        self.assertIsInstance(session.get_adapter("http://goauthentik.io"), MsgspecHTTPAdapter)
```
rissson, line 63: Any way we could add a test to ensure that the msgspec decoder is used by default? So that in future requests versions, if they start passing args to the `json` method, we can catch that?

rissson (approved):
Just a small testing thing, otherwise LGTM

PR #24899, opened 2026-08-08
lifecycle/container: test jemalloc

BeryJu:


rissson (approved):
I'd also setup https://github.com/gnzlbg/jemallocator for the rust part. Having it directly in the code might allow us to provide metrics about it in the future

PR #24902, opened 2026-08-09
core: return the intended status code from error views for all request methods

### LINKED ISSUE #24230: /providers/scim/{id}/sync/object/ returns 405 when invalid ID is used or task errors for other reason
The `/providers/scim/{id}/sync/object/` endpoint returns a 405, rather than a descriptive error, when an invalid ID is used or the SCIM task fails for any other reason e.g,. the SCIM tasks fails to enqueue.

Souptik96:
### What does this PR change?

The four views wired up as django's `handler400/403/404/500` in `authentik/root/urls.py`
subclassed `TemplateView` directly. `TemplateView` only implements `get`, so its
`dispatch()` answers a failing `POST`/`PUT`/`PATCH`/`DELETE` with `405 Method Not
Allowed` and an `Allow: GET, HEAD, OPTIONS` header, discarding the status code the
handler stands for.

- Adds a shared `ErrorView` base in `authentik/core/views/error.py` that renders the
  error page for every request method, so `BadRequestView`/`ForbiddenView`/
  `NotFoundView`/`ServerErrorView` return 400/403/404/500 respectively regardless of
  method. This also removes the duplicated `template_name` and the
  `ServerErrorView.dispatch` `# pragma: no cover` wrapper it made redundant.
- Widens the dramatiq result handling in `OutgoingSyncProviderStatusMixin.sync_object`
  (`authentik/lib/sync/outgoing/api.py`) from `ResultFailure` to `ResultError`, so
  `ResultTimeout`/`ResultMissing` — a task that is queued but has not reported yet —
  returns the task's messages instead of escaping as an unhandled 500.
- Adds `authentik/core/tests/test_error_views.py` covering both the handler status
  codes per method and the end-to-end unresolvable-API-path case.

Reported in #24230: `POST /api/v3/providers/scim/{id}/sync/object/` with an invalid ID
returns `405 Method Not Allowed` instead of a descriptive error. The endpoint is not
actually at fault — it is the error handlers.

Two things combine:

1. The DRF router matches a detail route's id with `(?P<pk>[^/.]+)`, so an id
   containing a dot never resolves and django raises `Http404` at the resolver, before
   any view runs. `handler404` then answers the `POST` with 405 rather than 404.
2. If the sync task is queued but has not reported a result yet, `get_result()` raises
   `ResultTimeout`/`ResultMissing`, which was not caught. That unhandled exception
   reaches `handler500`, which likewise answers the `POST` with 405 instead of 500.

The defect is repo-wide, not SCIM-specific: any non-GET request that reaches an error
handler on any endpoint was reported as 405. `schema.yml` declares only 200/400/403 for
this operation, so a 405 is not a documented outcome and the web UI's
`SyncObjectForm.ts` has no way to surface it usefully.

### How was this tested?

`make all` could **not** be run in my environment, and I want to be upfront about that:
`uv sync --frozen` fails locally while building `dumb-init` (`'Compiler' object has no
attribute 'linker_exe'`) and then `psycopg-c` (`couldn't run 'pg_config'
--includedir`), and I have no docker/postgres available. So the django test suite and
the lint matrix could not execute against the real project. **Please rely on CI for
`make all`.**

What I did instead, to avoid submitting an unverified guess:

- Built a standalone django + DRF harness that imports the real `error.py` by path and
  drives the handlers through `RequestFactory`, plus a DRF `APIClient` reproducing the
  router-level 404 and the unhandled-exception paths.
- Confirmed the bug on unmodified source: each handler returns its intended code for
  `GET` but `405` for `POST`/`PUT`/`PATCH`/`DELETE`; a pk containing a dot and an
  unhandled exception both surface as 405.
- Confirmed the fix: all methods return the handler's own status code, and the
  `Allow` header is gone.
- Negative control: with the new test file kept and only `error.py` reverted, the test
  fails with 16 assertion failures, the first being `AssertionError: 405 != 500`.
- `ruff check` and `black --check` pass on the three touched files.

No serializers or viewsets changed shape, so no `make gen` output changes are expected.

### Linked issues

closes #24230

## Checklist

- [ ] The project has been linted, built, and tested (`make all`) — see "How was this
      tested?"; blocked locally by the `uv sync` build failures described above.
- [x] The documentation has been updated (n/a — no user-facing documentation change)
- [ ] The documentation has been formatted (`make docs`) — n/a, no docs touched

---

Disclosure: yes, I used agentic AI assistance while preparing this change. The
diagnosis, the repros, and the negative control described above are real and were run;
please review accordingly, and let me know if you would rather I close this.

### DIFF
--- a/authentik/lib/sync/outgoing/api.py
+++ b/authentik/lib/sync/outgoing/api.py
@@ -1,6 +1,6 @@
 from django.db.models import Model
 from dramatiq.actor import Actor
-from dramatiq.results.errors import ResultFailure
+from dramatiq.results.errors import ResultError
 from drf_spectacular.utils import extend_schema
 from rest_framework.decorators import action
 from rest_framework.fields import BooleanField, CharField, ChoiceField
@@ -119,7 +119,11 @@ class OutgoingSyncProviderStatusMixin:
         )
         try:
             msg.get_result(block=True)
-        except ResultFailure:
+        # ResultFailure: the task ran and raised; its logs are the useful output.
+        # ResultTimeout/ResultMissing: the task is queued but hasn't reported yet.
+        # Neither is a server fault, so report whatever the task has logged so far
+        # rather than letting the error escape as a 500.
+        except ResultError:
             pass
         task: Task = msg.options["task"]
         task.refresh_from_db()
--- a/authentik/lib/sync/outgoing/api.py
+++ b/authentik/lib/sync/outgoing/api.py
@@ -119,10 +119,6 @@ class OutgoingSyncProviderStatusMixin:
         )
         try:
             msg.get_result(block=True)
-        # ResultFailure: the task ran and raised; its logs are the useful output.
-        # ResultTimeout/ResultMissing: the task is queued but hasn't reported yet.
-        # Neither is a server fault, so report whatever the task has logged so far
-        # rather than letting the error escape as a 500.
         except ResultError:
             pass
         task: Task = msg.options["task"]
--- a/authentik/core/tests/test_error_views.py
+++ b/authentik/core/tests/test_error_views.py
@@ -1,0 +1,69 @@
+"""Test the views wired up as django's handler400/403/404/500"""
+
+from django.test import RequestFactory, TestCase
+
+from authentik.core.tests.utils import create_test_brand
+from authentik.core.views.error import (
+    BadRequestView,
+    ForbiddenView,
+    NotFoundView,
+    ServerErrorView,
+)
+
+HANDLERS = (
+    (BadRequestView, 400),
+    (ForbiddenView, 403),
+    (NotFoundView, 404),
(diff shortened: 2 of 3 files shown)


authentik/lib/sync/outgoing/api.py
```diff
@@ -119,7 +119,11 @@ def sync_object(self, request: Request, body: SyncObjectSerializer, pk: int) ->
         )
         try:
             msg.get_result(block=True)
-        except ResultFailure:
+        # ResultFailure: the task ran and raised; its logs are the useful output.
+        # ResultTimeout/ResultMissing: the task is queued but hasn't reported yet.
+        # Neither is a server fault, so report whatever the task has logged so far
+        # rather than letting the error escape as a 500.
```
rissson:
```suggestion
```

BeryJu (approved):

PR #24913, opened 2026-08-09
root: test with python 3.15.0rc2

BeryJu:
don't merge

current issues:
- pyroscope removed (doesn't compile)


rissson:
marking as draft to avoid unintentional merges

PR #24963, opened 2026-08-10
enterprise/stages/source: configurable failure action

BeryJu:
based off of https://github.com/goauthentik/authentik/pull/24333

reworked to use dataclasses instead of dict and cleaned up code


authentik/core/sources/matcher.py
```diff
@@ -85,6 +103,11 @@ def get_action(
             property = matchable_property.property
             if matching_mode in [matchable_property.link_mode, matchable_property.deny_mode]:
                 if not properties.get(property, None):
+                    self.failure = MatchFailure(
+                        reason=MatchFailureReason.MISSING_PROPERTY,
+                        property=property,
+                        source_slug=self.source.slug,
+                    )
```
BeryJu: Turn this into an exception

rissson (approved):

authentik/core/sources/matcher.py
BeryJu: I'll leave this as-is for now as changing this into an exception requires quite a few other changes

Issue #22891, opened 2026-06-05
Support custom CA for DUO servers


runleveldev:
### Is your feature request related to a problem?

I'm trying to use DUO behind a firewall which performs packet inspection. This breaks the functionality since the DUO API package is using it's bundled certificates without an override. I can work around this by mounting my certificate bundle over top of DUO's, but it'd be nice to have a built-in solution.

### Describe the solution you'd like

Allow selecting a certificate in the DUO setup stage which is passed to the DUO API client constructor as the certificate/bundle to verify with.

### Describe alternatives that you've considered

Adding this mount to the docker compose works, but just seems hacky.

```
    - /etc/ssl/certs/ca-certificates.crt:/ak-root/.venv/lib/python3.14/site-packages/duo_client/ca_certs.pem
```

### Additional context

See https://github.com/duosecurity/duo_client_python/blob/f2fb544037567e6b7ecf78adeacfe991b55ee2bc/duo_client/client.py#L216

dewi-tik:
At the moment your workaround is what we'd suggest. Fixing the underlying issue is significantly more difficult than you'd expect. However, we're happy to receive any contributor PRs related to this.

dominic-r:
We're closing this in favor of this parent issue: #25295

PR #24965, opened 2026-08-10
stages/authenticator_duo: return a descriptive error when the Duo API is unreachable

### LINKED ISSUE #22896: 405 Method Not Allowed when DUO API certificate doesn't verify
### Describe the bug

I'm using Authentik behind a packet inspecting firewall which re-encypts using a organization CA. I filed #22891 as a feature enhancement involving this. That said, it was hard to track this down as the root cause because, when I was attempting to import authenticators, POST /api/v3/stages/authenticator/duo/\<uuid\>/import_devices_automatic/ returned 405 Method Not Allowed instead of a descriptive error such as certificate not valid.

### How to reproduce

1. Setup a proxy, a reverse proxy is probably easier to test with, just make sure it's got a cert not signed by one of the CAs the DUO vendored ca_certs.pem file trusts.
2. Point the DUO Authenticator setup stage at it.
3. Attempt to import users
4. Observe 405 Method Not Allowed in the console.

### Expected behavior

Error 400 with a JSON error describing the SSL problem.

### Screenshots

_No response_

### Additional context

_No response_

### Deployment Method

Docker

### Version

2026.5.2

### Relevant log output

```shell
{"auth_via": "session", "domain_url": null, "event": "dropping exception", "exc": "SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1081)')", "host": "<my duo url>", "level": "debug", "logger": "authentik.lib.sentry", "pid": 46379, "request_id": "c8ea38a92a6046ecb125631c8d5e859c", "schema_name": "public", "timestamp": "2026-06-05T13:49:10.757643"}
{"auth_via": "session", "domain_url": "<my duo url>", "event": "/api/v3/stages/authenticator/duo/0334e04c-af0d-49fc-bc49-f37a222b2e5c/import_devices_automatic/", "host": "<my duo url>", "level": "info", "logger":"authentik.asgi", "method": "POST", "pid": 46379, "remote": "64.186.39.4", "request_id": "c8ea38a92a6046ecb125631c8d5e859c", "runtime": 1753, "schema_name": "public", "scheme": "https", "status": 405, "timestamp": "2026-06-05T13:49:10.764531", "user": "akadmin", "user_agent": "Mozilla/5.0 (X11; Linux x86_64; rv:140.0) Gecko/20100101 Firefox/140.0"}
```

### LINKED ISSUE #25295: Duo API uses bundled CAs so org or MITM CAs fail with a misleading 405
## What

: Duo API uses bundled CAs, so org/MITM CA breaks imports and surfaces as a misleading 405. #22896 points at #22891.

## Want

- : Duo API uses bundled CAs, so org/MITM CA breaks imports
- Surfaces as a misleading 405
- #22896 points at #22891

## Ideas this came from

- [#22891 Support custom CA for DUO servers](https://github.com/goauthentik/authentik/issues/22891) — Is your feature request related to a problem? I'm trying to use DUO behind a firewall which performs packet inspection. This breaks the functionality since the 
- [#22896 405 Method Not Allowed when DUO API certificate doesn't verify](https://github.com/goauthentik/authentik/issues/22896) — I'm using Authentik behind a packet inspecting firewall which re-encypts using a organization CA. I filed #22891 as a feature enhancement involving this. That s

abhisheksharma2411:
### What does this PR change?

`_duo_import_devices` caught only `RuntimeError`, but `duo_client` surfaces transport failures as the underlying socket/TLS error. `SSLCertVerificationError` is an `OSError`, not a `RuntimeError`, so it escaped the handler entirely.

This catches transport failures explicitly and maps them to the documented `400` with a descriptive message, ordered narrowest-first since `SSLCertVerificationError` < `SSLError` < `OSError`:

| Failure | Message |
|---|---|
| Certificate verification failed | `Failed to connect to Duo: TLS certificate verification failed.` |
| Any other TLS failure | `Failed to connect to Duo: TLS error.` |
| Any other connection failure | `Failed to connect to Duo.` |

`RuntimeError` handling is unchanged.

Closes #22896. A Duo endpoint behind an untrusted certificate (a packet-inspecting firewall re-encrypting with an org CA, in the reporter's case) produced a bare **405 Method Not Allowed** with no error body, so nothing indicated a TLS problem.

It compounds: `OSError` is also in `authentik.lib.sentry.ignored_classes`, so the exception was dropped from error reporting as well. The reporter's only clue was a `"dropping exception"` debug log.

The messages are deliberately categorical and do not echo exception text, preserving the existing behaviour of not leaking internals to the API caller.

### How was this tested?

The defect was reproduced with failing tests **before** the fix. The 405 reproduces locally with no proxy involved, which confirms it is authentik's own behaviour rather than the reporter's firewall:

```
authentik.asgi: {'method': 'POST', 'status': 405,
  'event': '/api/v3/stages/authenticator/duo/.../import_devices_automatic/'}

FAILED test_api_import_automatic_tls_failure
  - ssl.SSLCertVerificationError: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed
FAILED test_api_import_automatic_connection_failure
  - OSError: connection refused
```

Three tests added, following the existing `test_api_import_automatic_invalid` pattern of patching `duo_client.admin.Admin.get_users_iterator`:

1. `test_api_import_automatic_tls_failure` — `SSLCertVerificationError` → 400 + certificate message
2. `test_api_import_automatic_tls_error` — generic `SSLError` → 400 + TLS message, so a handshake failure is not misreported as a certificate problem
3. `test_api_import_automatic_connection_failure` — plain `OSError` → 400 + connection message

After the fix:

```
python manage.py test authentik.stages.authenticator_duo.tests.AuthenticatorDuoStageTests
  → 13 passed, 0 failed

coverage report --include="authentik/stages/authenticator_duo/api.py"
  → 96.88%   (the 4 uncovered lines are pre-existing; every added line is covered)

ruff check   → All checks passed!
ruff format  → 2 files already formatted
```

### Linked issues

closes #22896

closes #25295
Possibly related: #22891 requests custom CA support for Duo, which would let this configuration work rather than just fail clearly. This PR only makes the failure diagnosable.

---

## Checklist

    Ran the affected suite plus `ruff check` / `ruff format --check` on the changed files, and coverage on `api.py`. I did not run the full `make all` locally (full web build plus complete suite); happy to if you'd like, otherwise CI covers it.

    No documentation change — this only replaces an unhandled exception with the error response the endpoint already documents (`400: Bad request`). Let me know if the error strings should be documented anywhere.

---

Supersedes #24903, which I had to abandon: I force-pushed that branch to correct a commit trailer and did it from a shallow clone, which orphaned the history and caused GitHub to auto-close the PR and show the whole repository as added. This branch is rebuilt on current `main` with the identical two-file change. Apologies for the noise.

### DIFF
--- a/authentik/stages/authenticator_duo/api.py
+++ b/authentik/stages/authenticator_duo/api.py
@@ -4,6 +4,7 @@ from ssl import SSLCertVerificationError, SSLError
 from typing import Any
 
 from django.http import Http404
+from django.utils.translation import gettext_lazy as _
 from drf_spectacular.types import OpenApiTypes
 from drf_spectacular.utils import OpenApiResponse, extend_schema, inline_serializer
 from guardian.shortcuts import get_objects_for_user
@@ -217,25 +218,25 @@ class AuthenticatorDuoStageViewSet(UsedByMixin, ModelViewSet):
         except SSLCertVerificationError as exc:
             LOGGER.warning("failed to verify duo api certificate", exc=exc)
             return {
-                "error": "Failed to connect to Duo: TLS certificate verification failed.",
+                "error": _("Failed to connect to Duo: TLS certificate verification failed."),
                 "count": created,
             }
         except SSLError as exc:
             LOGGER.warning("tls error connecting to duo", exc=exc)
             return {
-                "error": "Failed to connect to Duo: TLS error.",
+                "error": _("Failed to connect to Duo: TLS error."),
                 "count": created,
             }
         except OSError as exc:
             LOGGER.warning("failed to connect to duo", exc=exc)
             return {
-                "error": "Failed to connect to Duo.",
+                "error": _("Failed to connect to Duo."),
                 "count": created,
             }
         except RuntimeError as exc:
             LOGGER.warning("failed to get users from duo", exc=exc)
             return {
-                "error": "An internal error occurred while importing devices.",
+                "error": _("An internal error occurred while importing devices."),
                 "count": created,
             }
 
--- a/authentik/stages/authenticator_duo/api.py
+++ b/authentik/stages/authenticator_duo/api.py
@@ -1,5 +1,6 @@
 """AuthenticatorDuoStage API Views"""
 
+from ssl import SSLCertVerificationError, SSLError
 from typing import Any
 
 from django.http import Http404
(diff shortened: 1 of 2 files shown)


abhisheksharma2411:
Five days in with no review, so a status note rather than a nudge — the checks list on this PR reads worse than the actual state, and I'd rather correct that than have it be the reason nobody starts.

**Every check on this head is green.** Codecov posts a new check run per coverage upload, so the API returns ~100 rows for ~83 distinct checks and the intermediate ones are partial uploads. Reducing to the latest run per check name:

```
distinct checks: 83
failing (latest per name): 0
```

The earlier reds in the list are superseded rows, not current results. Nothing here is waiting on me.

Happy to rebase onto current `main` if that helps, or to split it if the shape is the problem — just say which.

BeryJu (approved):

PR #24991, opened 2026-08-11
website/docs: preserve host and port in proxy provider nginx redirects

### LINKED ISSUE #24951: Proxy Provider nginx config uses $http_host in redirects and headers (HTTP/3 + security concerns)
### Describe the bug

The nginx configuration generated/suggested by the Proxy Provider uses `$http_host` in several places (e.g. redirects and `X-Original-URL`), while using `$host` elsewhere.

Using `$http_host` is problematic because it reflects the raw client-supplied `Host` header, which is untrusted, may include ports, and may be missing or inconsistent in HTTP/3 environments.

This can lead to incorrect redirect URLs, authentication issues, or inconsistent behavior.

### How to reproduce

1. Use the Proxy Provider nginx configuration as provided in the documentation/UI
2. Deploy behind nginx with HTTP/2 or HTTP/3 enabled
3. Trigger an authentik login flow (e.g. accessing a protected upstream)
4. Observe redirects using `$scheme://$http_host$request_uri`

Depending on the client/request:

* redirects may contain incorrect or unexpected host values causing 500 errors

### Expected behavior

The nginx configuration suggested by the Proxy Provider should use $host instead of $http_host when constructing the X-Original-URL header and authentication redirect URLs.

For example:

proxy_set_header X-Original-URL $scheme://$host$request_uri;
return 302 /outpost.goauthentik.io/start?rd=$scheme://$host$request_uri;

This would provide consistent behavior with the existing proxy_set_header Host $host; configuration and avoid relying on the raw client-supplied Host header.

### Screenshots

_No response_

### Additional context

The config already uses `$host` for `proxy_set_header Host`, but `$http_host` is still used in:

```nginx
proxy_set_header X-Original-URL $scheme://$http_host$request_uri;
return 302 /outpost.goauthentik.io/start?rd=$scheme://$http_host$request_uri;
```

This inconsistency can be avoided by using `$host` everywhere:

```nginx
proxy_set_header X-Original-URL $scheme://$host$request_uri;
return 302 /outpost.goauthentik.io/start?rd=$scheme://$host$request_uri;
```

More details: https://bsod.wtf/guide/nginx-http-host-vs-host-http3-config-fix/
https://www.getpagespeed.com/server-setup/nginx/nginx-http3-http_host-fix

### Deployment Method

Docker

### Version

2026.8.0-rc4

### Relevant log output

```shell

```

rowkav09:
### What does this PR change?
Uses `$host` instead of `$http_host` when building `X-Original-URL` and the sign-in redirect (`rd=`) in the Proxy Provider nginx recipes and e2e test config. `$http_host` is the raw client-supplied Host header; `$host` is nginx's validated value, already used for the actual `Host` header two lines above in the same files.

authentik's outpost parses `X-Original-URL` directly (`getNginxForwardUrl` in `mode_common.go`) with no validation, so the unvalidated client Host header flows straight into authentik's own redirect construction.

### How was this tested?
Config-only change (variable substitution, no structural changes); verified via `git grep` that no other `$http_host` usage was missed and that `X-Forwarded-Host` in the ingress snippet (a distinct, legitimate use) is untouched. Existing e2e proxy-forward-auth test already exercises this config.

### Linked issues

closes #24951

### DIFF
--- a/tests/e2e/proxy_forward_auth/nginx_single/nginx.conf
+++ b/tests/e2e/proxy_forward_auth/nginx_single/nginx.conf
@@ -42,7 +42,7 @@ server {
         # ensure the host of this vserver matches your external URL you've configured
         # in authentik
         proxy_set_header    Host $host;
-        proxy_set_header    X-Original-URL $scheme://$http_host$request_uri;
+        proxy_set_header    X-Original-URL $scheme://$host$request_uri;
         add_header          Set-Cookie $auth_cookie;
         auth_request_set    $auth_cookie $upstream_http_set_cookie;
         proxy_pass_request_body off;
@@ -56,6 +56,6 @@ server {
         add_header Set-Cookie $auth_cookie;
         return 302 /outpost.goauthentik.io/start?rd=$request_uri;
         # For domain level, use the below error_page to redirect to your authentik server with the full redirect path
-        # return 302 https://localhost/outpost.goauthentik.io/start?rd=$scheme://$http_host$request_uri;
+        # return 302 https://localhost/outpost.goauthentik.io/start?rd=$scheme://$host$request_uri;
     }
 }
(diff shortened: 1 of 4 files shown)


BeryJu (approved):

dewi-tik:
Thanks for the contribution. Applied some changes so that non-standard ports are handled and a few other tweaks

dominic-r (approved):

dewi-tik (approved):

PR #24999, opened 2026-08-11
flows: send messages with the responses

rissson:
As discussed.

@BeryJu I'll probably need a hand with this.

Does not remove the messages from the websocket connection for now.


kensternberg-authentik (approved):

BeryJu (commented):
I think it would be better to have the challenge be

```
"component": "xyz",
"meta": {
"messages": [],
}
```

rissson:
@BeryJu I put it under the `flow_info` key. Did you want to create a new `meta` key?

BeryJu:
> @BeryJu I put it under the `flow_info` key. Did you want to create a new `meta` key?

flow_info is probably fine too tbh

authentik/flows/stage.py
```diff
@@ -295,7 +301,7 @@ def get_challenge(self, *args, **kwargs) -> RedirectChallenge:
         )
 
     def challenge_valid(self, response: ChallengeResponse) -> HttpResponse:
-        return HttpChallengeResponse(self.get_challenge())
+        return HttpChallengeResponse(self.get_challenge(), self.request)
```
BeryJu:
```suggestion
        return HttpChallengeResponse(self.get_challenge())
```

BeryJu (approved):

PR #25008, opened 2026-08-11
web: Flip the admin interface to path-based routing

GirlBossRush:
## Details

This PR flips `/if/admin/` to path-based routing, converts the shared runtime-stateful components to real search params, and deletes the legacy hash router.

New router-core `search-params` accessors back the state conversion: `ak-tabs` selection becomes `?page=`, `Table` search becomes `?q=` (overridable per table, with namespaced pagination), and the notification drawers follow. That also heals the tab deep links the user flip deferred.

On the routing side, `admin/Routes.ts` becomes 55 `URLPattern` routes — literal-before-param ordering reserves a future `/core/applications/new` — and `<ak-router-view>` replaces the outlet. The command palette, create wizard, and table filters move over with it, along with 60 in-admin `#/` links across 46 files and the shared nav. The legacy `src/elements/router/` modules are deleted; `Router404` stays, since the core outlet uses it.

Stacked on #25007.

## Testing

`120-admin-routing` e2e 7/7, `121` base-path smoke skip-guarded, and `110-user-routing` regression 6/6 with tab deep links now selecting natively. 142 unit and browser tests, a clean full build, and `lint-check` green.


rissson:
(╯°□°)╯︵ ┻━┻ flip it!

BeryJu (approved):

PR #25010, opened 2026-08-11
core: Redirect source linking to the path-based user settings URL

GirlBossRush:
When an authenticated user links a social/OAuth source, `SourceFlowManager.handle_existing_link` redirects them to the user settings "Connected services" tab. Updates that redirect from the legacy hash form to the path-based `?page` form the flipped user interface understands:

`…/if/user/#/settings;page-sources` → `…/if/user/settings?page=page-sources`

## Tests updated
- Unit: `test_source_flow_manager.py::test_authenticated_link`.
- Selenium e2e (`test_source_oauth_oauth2.py`): the post-link URL assertion and `find_settings_tab_panel` now use the path + `?page=` form instead of the `#…;{json}` hash suffix.

## Base
**Stacked on #25008** (the admin flip), not `main` — the redirect's tab selection depends on `ak-tabs` reading `?page=` search params, which lands with the admin flip. Merges after the flip stack.

### DIFF
--- a/authentik/core/sources/flow_manager.py
+++ b/authentik/core/sources/flow_manager.py
@@ -412,7 +412,7 @@ class SourceFlowManager:
             reverse(
                 "authentik_core:if-user",
             )
-            + "#/settings;page-sources"
+            + "settings?page=page-sources"
         )
 
     def handle_enroll(
--- a/authentik/core/tests/test_source_flow_manager.py
+++ b/authentik/core/tests/test_source_flow_manager.py
@@ -86,7 +86,7 @@ class TestSourceFlowManager(TestCase):
         self.assertEqual(response.status_code, 302)
         self.assertEqual(
             response.url,
-            reverse("authentik_core:if-user") + "#/settings;page-sources",
+            reverse("authentik_core:if-user") + "settings?page=page-sources",
         )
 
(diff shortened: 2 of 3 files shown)


BeryJu (changes requested):
Just as a reminder for myself @BeryJu replace with urlencode

GirlBossRush:
Superseded by #25108. That PR's top commit (`web,core: point the sources-tab redirects at the path segment`) rewrites the same three files — `authentik/core/sources/flow_manager.py`, `authentik/core/tests/test_source_flow_manager.py`, and `tests/e2e/test_source_oauth_oauth2.py` — targeting `settings/sources` instead of `?page=page-sources`, which path-routed tabs ignore. #25382 now bases on #25108 directly.

GirlBossRush:
Superseded by #25108 — that PR's top commit (`web,core: point the sources-tab redirects at the path segment`) rewrites the same three files (`authentik/core/sources/flow_manager.py`, `authentik/core/tests/test_source_flow_manager.py`, `tests/e2e/test_source_oauth_oauth2.py`) and targets `settings/sources` instead of `?page=page-sources`, which path-routed tabs ignore. #25382 now bases on #25108 directly.

PR #25024, opened 2026-08-12
providers/scim: fix group membership removals

### LINKED ISSUE #25003: vSphere SCIM does not sync groups properly
### Describe the bug

vSphere's SCIM sync can sync groups properly when a user is added, but not when a user is removed

### How to reproduce

1. Set up vSphere & Authentik & SCIM
2. Give a group access to a folder
3. Add a user to the group
4. This works, the user can now see the folder
5. Remove the user from the group
6. The user can still see the folder, even after manually triggering a SCIM sync

### Expected behavior

vSphere should be removing people from groups, or else the SCIM sync is out-of-sync

### Screenshots

_No response_

### Additional context

_No response_

### Deployment Method

Docker

### Version

2026.5.6

### Relevant log output

```shell

```

dominic-r:
### What

Removing users from a SCIM group was never written to servers without PATCH support (such as vCenter) whenever the change was invisible to `diff()`: `to_schema` only attached `members` to the payload when the list was non-empty, `diff()` compared payloads with a list-union merge that cannot see a member disappear, and the state recorded on the connection was whatever the server last echoed, which vCenter-like servers don't echo at all (`204 No Content`, or responses without a member list). The write was then skipped as "nothing changed", leaving only `patch_compare_users`, which sends PATCH requests to a server that authentik itself has configured as having no PATCH support.

### Changes

- Group payloads now always state the member list when PATCH is unsupported, including when it is empty — that is the only way to remove the last member without PATCH.
- `diff()` compares members by ID rather than merging lists, and no longer trusts a recorded state that has no member list: it cannot prove the members are unchanged, so the group is written once and the recorded state converges. This is also what heals deployments whose connections were recorded by earlier versions.
- After every group write, the connection records what was written — the sent payload overlaid with the response — instead of the response alone. Recording only the response silently reverted the recorded member list on servers that answer `204` or omit `members`, which made the next removal look like a no-op.
- `create()` no longer sends a member PATCH to providers that report no PATCH support; the members are already part of the create request, and the extra request failed with task retries against strict servers.
- Both member-removal paths now build their operations through one pair of helpers using the RFC 7644 filter path `members[value eq "<id>"]`. Previously the reconcile path used a `{"op": "remove", "path": "members", "value": [...]}` shape, so a server implementing only the filter form silently lost removals.
- `ServiceProviderConfiguration.default()` actually marks itself as a fallback. pydantic silently drops underscore-prefixed constructor kwargs, so `is_fallback` was always `False` and the PUT-failed-so-try-PATCH fallback in `update_group` was dead code.

Note: when `create()` links a group that already exists remotely on a provider without PATCH support, membership now waits for the next sync rather than sending a PATCH the server would reject.

Closes #25003

### DIFF
--- a/authentik/providers/scim/clients/groups.py
+++ b/authentik/providers/scim/clients/groups.py
@@ -76,16 +76,12 @@ class SCIMGroupClient(SCIMClient[Group, SCIMProviderGroup, SCIMGroupSchema]):
 
         if not self._config.patch.supported:
             users = list(obj.users.order_by("id").values_list("id", flat=True))
-            connections = SCIMProviderUser.objects.filter(
+            member_ids = SCIMProviderUser.objects.filter(
                 provider=self.provider, user__pk__in=users
-            )
-            members = []
-            for user in connections:
-                members.append(
-                    self._create_group_member(user.scim_id),
-                )
-            if members:
-                scim_group.members = members
+            ).values_list("scim_id", flat=True)
+            # The member list is also sent when it is empty, as without PATCH that is the only
+            # way to remove the last member of a group
+            scim_group.members = [self._create_group_member(x) for x in member_ids]
         else:
             del scim_group.members
         return scim_group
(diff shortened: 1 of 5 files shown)


KingHenry6666:
Just independently tested, the original issue is fixed. I'll do a bit more testing of edge cases and weird scenarios and see if it holds up.

KingHenry6666:
Just finished testing quite a few edge cases, including if a full SCIM sync works (In case the vCenter powers off or loses connectivity), mass removing people, adding/removing them from scim's provisoined group list, etc. Everything I tested works.

BeryJu (approved):

PR #25041, opened 2026-08-13
stages/authenticator_duo: allow verifying the Duo API against a custom CA chain

### LINKED ISSUE #22891: Support custom CA for DUO servers

### LINKED ISSUE #25295: Duo API uses bundled CAs so org or MITM CAs fail with a misleading 405

abhisheksharma2411:
### What does this PR change?

Adds an optional **CA chain** to the Duo authenticator stage, so the Duo API can be verified against an internal CA instead of the bundle vendored inside `duo_client`.

- New `ca_chain` FK to `CertificateKeyPair` on `AuthenticatorDuoStage`, mirroring `LDAPSource.peer_certificate`
- Passed to both `Auth` and `Admin` clients
- Omitted entirely when unset, so `duo_client` keeps its own roots and existing installations are unaffected

Closes #22891. Duo behind a TLS-inspecting proxy currently cannot be reached at all — `duo_client` validates against `duo_client/ca_certs.pem` with no override. The only workaround is mounting a host bundle over that file inside the container, which the reporter fairly described as hacky.

@dewi-tik noted on the issue that this is *"significantly more difficult than you'd expect"*, and the reason is real: `duo_client` accepts `ca_certs` **only as a filesystem path**, while authentik keeps certificates in the database.

```python
def __init__(self, ikey, skey, host, ca_certs=DEFAULT_CA_CERTS, ...):
    """ca_certs - Path to CA pem file."""
```

authentik already solves exactly this shape in `LDAPSource.server()`, which materialises `client_certificate` to a `NamedTemporaryFile` for `ldap3`. This follows that precedent rather than inventing a new one.

On writing certificate material to disk: the file holds only the **CA chain**, which is public certificate data, never a private key. It is still written `0600` into a private `mkdtemp()` directory, so other users on the host cannot read it.

### How was this tested?

```
manage.py test authentik.stages.authenticator_duo.tests.AuthenticatorDuoStageTests
  → 15 passed  (5 new)

ruff check   → All checks passed!
ruff format  → 7 files already formatted
```

The five new tests pin the behaviour that matters:

1. **No chain configured** → `ca_certs` is not passed at all
2. **Chain configured** → the file exists, contains exactly the configured `certificate_data`, and is mode `0600`
3. **The kwarg reaches the `Auth` client** — asserted by reading back `client.ca_certs`
4. **The kwarg reaches the `Admin` client**
5. **Without a chain, `client.ca_certs == DEFAULT_CA_CERTS`** — so the fallback is the bundled roots rather than nothing

That third and fourth case matter because the value has to survive all the way into the constructor; asserting only on `_duo_client_kwargs()` would pass even if the wiring were dropped.

### Linked issues

closes #22891

closes #25295
Related: #24965 (open) makes the *failure* diagnosable when the CA is wrong — it currently surfaces as a bare 405. This makes the configuration actually work. They're complementary and independent.

---

## Checklist

    Ran the affected suite plus `ruff check` / `ruff format --check`. I did not run the full `make all` locally; CI covers it.

    **One thing I could not do and would rather flag than hide:** `schema.yml` needs regenerating, since this adds a serializer field. `make gen-build` failed in my environment — with `AUTHENTIK_TENANTS__ENABLED=true` the tenant migrations error out locally (`UnicodeEncodeError: 'idna' codec ... label too long`), and without it the generated schema *drops* the six `/tenants/` endpoints, which would be a damaging diff. I deliberately left `schema.yml` untouched rather than commit an incomplete one. Happy to add the regenerated schema if someone can point me at what my local setup is missing, or if a maintainer regenerates it.

    Not updated — happy to add a note to the Duo stage docs describing when the CA chain is needed if you'd like one; I wasn't sure whether a new optional field warrants it.


abhisheksharma2411:
Updated — the two red gates are addressed, and the schema caveat in the description no longer applies.

**`test-gen`.** I noted originally that I couldn't regenerate `schema.yml` locally. I got the toolchain running (`make migrate` via `lifecycle.migrate`, then `gen-build`, then the client-ts generator), so the generated artifacts are now in the branch:

```
blueprints/schema.json                                       |  6 ++
schema.yml                                                   | 24 ++++
packages/client-ts/src/models/AuthenticatorDuoStage.ts       | 13 +++
packages/client-ts/src/models/AuthenticatorDuoStageRequest.ts| 13 +++
.../models/PatchedAuthenticatorDuoStageRequest.ts            | 13 +++
```

That is exactly the file set the failing job diffed, and the change is purely the additive `ca_chain` property — no reordering or collateral edits. `client-go` and `client-rust` came back unchanged, matching what the CI diff reported.

Worth flagging one thing I hit, in case it saves someone else the trouble: running the client-ts generator without `pnpm install` at the repo root leaves prettier unable to resolve `@goauthentik/prettier-config`, and every file in `packages/client-ts` then shows as modified on formatting alone. Installing the workspace first reduces it to the three real files.

**`lint (black, python)`.** The migration was the one file failing `ci-lint-black`. Formatted with `black==26.5.1` to match CI; `black --check authentik packages tests scripts lifecycle` is clean across all 2,232 files.

**The earlier `test-e2e` failures were unrelated and have cleared.** They were a `authentik_tasks_tasklog` FK violation in task teardown and a Selenium `DetachedShadowRootException` — neither touching the Duo stage. Rebasing onto current `main` (the branch was 105 commits behind) cleared both; all 11 e2e suites now pass.

Squashed into the single commit rather than stacking fixups, so the branch stays one logical change. Happy to split the generated artifacts into their own commit if you'd rather review them separately.

abhisheksharma2411:
The red on this branch is a flake, and the matrix makes that unusually easy to show: `test-migrations-from-stable` runs 20 identical shards and **19 passed**. Only PostgreSQL 18 / Run 3 failed.

```
PostgreSQL 14-alpine  Runs 1–10   ✅ ×10
PostgreSQL 18-alpine  Runs 1,2,4–10 ✅ ×9
PostgreSQL 18-alpine  Run 3        ❌
```

Same commit, same job definition, same migrations — so nothing in the diff distinguishes run 3 from the nineteen that passed. `ci-core-mark` is red only because it aggregates that one shard.

Everything else is green: 96 successful checks, and the `test-gen` artifacts from my earlier comment are still in place.

I don't have permission to re-run jobs on this repo, so I've left it rather than force-pushing a no-op commit to shake it loose — happy to rebase if you'd prefer a clean board before review.

abhisheksharma2411:
Same note as on #24965, plus the one red on this head, which I'm confident is a flake and can show rather than assert.

**Latest run per check name: 2 failing** — `ci-core-mark`, which is the aggregate, and one `test-migrations-from-stable` shard. The rest of the ~78 distinct checks are green; the other reds in the list are superseded codecov rows from partial uploads.

**The migration failure is a flake, and this matrix makes it provable.** `test-migrations-from-stable` runs the same commit across shards — identical inputs, so any disagreement is the environment, not the diff:

```
PostgreSQL 14-alpine   5 shards   5 success
PostgreSQL 18-alpine   7 shards   6 success, 1 failure
                                  -----------------
                                  11 success, 1 failure
```

Eleven of twelve shards of the same commit pass. That's two different outcomes from identical inputs, which is the only kind of flake evidence I'd act on — a cross-PR comparison would prove nothing here.

`ci-core-mark` fails as the aggregate of that one shard, so it clears when the shard is re-run.

Happy to rebase for a fresh run if that's easier than a re-run.

abhisheksharma2411:
Rebased onto current `main` (`b2a50acf`) — this had gone to `mergeable_state: dirty`. New head `5fbbe14b2`, and GitHub now reports `mergeable: true`.

The conflict was with #24965, which has since merged: it added `from ssl import SSLCertVerificationError, SSLError` to `tests.py` in the same import block this branch touches. Resolved by keeping both, in isort order (`pathlib` < `ssl` < `stat`). Nothing else in the two changes overlaps, and `0009_authenticatorduostage_ca_chain` is still the only 0009 in the app, so there's no duplicate migration number to renumber.

The two are worth reading together, though, because they're the same failure from opposite ends. #24965 makes a Duo endpoint behind an untrusted certificate report a descriptive error instead of a bare 405. This one gives the operator a way to make that endpoint *trusted* — which is the actual fix when Duo is reached through a TLS-inspecting proxy that re-signs with an internal CA. The error message #24965 added is what points at this field.

Verified after the rebase rather than trusting that it applied cleanly:

```
python manage.py test --keepdb authentik/stages/authenticator_duo/tests.py
  18 passed in 6.10s
```

That includes #24965's `test_api_import_automatic_tls_error` and `test_api_import_automatic_connection_failure` alongside this branch's `test_ca_chain_reaches_the_duo_admin_client`, which is the combination the merge conflict actually put at risk.

```
ruff check      All checks passed
ruff format     7 files already formatted
black --check   17 files unchanged
```

The generated artifacts survived the rebase intact — `git diff main...HEAD` on them is still only the additive `ca_chain` property, in the three schema components and the blueprint:

```
schema.yml            AuthenticatorDuoStage, AuthenticatorDuoStageRequest,
                      PatchedAuthenticatorDuoStageRequest
blueprints/schema.json  authentik_stages_authenticator_duo.authenticatorduostage
packages/client-ts    3 models, +13 each
```

No reordering or collateral edits, which was the thing to check — `schema.yml` auto-merged and a bad merge there would show up as `test-gen` red rather than as a conflict.

Board is otherwise unchanged from my last note: this needs a reviewer more than it needs anything from me. Happy to split the generated artifacts into their own commit if that reads more easily.

PR #25075, opened 2026-08-13
website/docs: document debugging the Rust components

dominic-r:
Found while auditing 2026.8 docs for staleness ahead of the release — see the [Slack thread](https://authentiksecurity.slack.com/archives/C08C0SJ4V6D/p1786634134866959).

### `debugging.md`

The page's second and final section was headed "authentik Server / Outposts (Golang)" and opened:

> Outposts, as well as some auxiliary code of the authentik server, are written in Go.

After 2026.8 the server and proxy outpost are Rust; LDAP, RADIUS, and RAC are still Go (their Dockerfiles still run `go build`). So the heading was wrong about the server, right about three of four outposts, and there was no Rust guidance at all — for what is now the whole front of the request path.

Split into a Rust section and a Go section. The Rust section covers:

- debugger tooling and how to run a single component (`cargo run -- server`, `cargo run -- proxy`)
- **tokio-console**, which is already wired up — `packages/ak-common/src/tracing.rs` starts a `console_subscriber::ConsoleLayer` when `listen.debug_tokio` is set. This was undocumented, and `debug_tokio` was commented out in `default.yml` during 2026.8, so it is off unless explicitly set.
- the per-crate `AUTHENTIK_LOG__RUST_LOG__*` levels, also undocumented

The Go section now also points at `AUTHENTIK_LISTEN__DEBUG`, which is the listener those outposts actually serve.

### `full-dev-environment.mdx`

> authentik is primarily a Django application running under gunicorn, proxied by a **Go** application that serves static files.

gunicorn is still accurate; the front proxy is not. `lifecycle/ak` routes `server`, `worker`, and `allinone` to the `authentik` Rust binary, implemented in `src/server/`. The page's own prerequisites list (line 22) already requires Rust, so it contradicted itself — and this is the first architectural sentence a new contributor reads.

Note: the `AUTHENTIK_LISTEN__DEBUG` scope correction referenced by the new Go section is in #25074.

### DIFF
--- a/website/docs/developer-docs/setup/debugging.md
+++ b/website/docs/developer-docs/setup/debugging.md
@@ -58,6 +58,24 @@ After re-creating the containers with `AUTHENTIK_DEBUGGER` set to `true` and the
 
 If the authentik instance is running on a remote server, the `.vscode/launch.json` file needs to be adjusted to point to the IP of the remote server. Alternatively, you can forward the debug port via an SSH tunnel, using `-L 9901:127.0.0.1:9901`.
 
-## authentik Server / Outposts (Golang)
+## authentik Server / Proxy outpost (Rust)
 
-Outposts, as well as some auxiliary code of the authentik server, are written in Go. These components can be debugged using standard Golang tooling, such as [Delve](https://github.com/go-delve/delve).
+The authentik server entrypoint and the proxy outpost are written in Rust. The server accepts incoming requests, serves static assets, and forwards the rest to the Python core; the Python core itself is covered in the section above.
+
+These components can be debugged with standard Rust tooling, such as [`rust-gdb`, `rust-lldb`](https://doc.rust-lang.org/book/appendix-04-useful-development-tools.html), or the [CodeLLDB](https://marketplace.visualstudio.com/items?itemName=vadimcn.vscode-lldb) extension for Visual Studio Code. Build a debug binary with `cargo build` and run the component you want to inspect, for example `cargo run -- server` or `cargo run -- proxy`.
+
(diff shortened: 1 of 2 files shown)


website/docs/developer-docs/setup/debugging.md
```diff
@@ -58,6 +58,24 @@ After re-creating the containers with `AUTHENTIK_DEBUGGER` set to `true` and the
...
 
-## authentik Server / Outposts (Golang)
+## authentik Server / Proxy outpost (Rust)
 
-Outposts, as well as some auxiliary code of the authentik server, are written in Go. These components can be debugged using standard Golang tooling, such as [Delve](https://github.com/go-delve/delve).
+The authentik server entrypoint and the proxy outpost are written in Rust. The server accepts incoming requests, serves static assets, and forwards the rest to the Python core; the Python core itself is covered in the section above.
+
+These components can be debugged with standard Rust tooling, such as [`rust-gdb`, `rust-lldb`](https://doc.rust-lang.org/book/appendix-04-useful-development-tools.html), or the [CodeLLDB](https://marketplace.visualstudio.com/items?itemName=vadimcn.vscode-lldb) extension for Visual Studio Code. Build a debug binary with `cargo build` and run the component you want to inspect, for example `cargo run -- server` or `cargo run -- proxy`.
```
rissson, line 65: sry but this is just llm generated garbage.

dominic-r: absolutely, i let the agent do its thing for the entire artifact now I have to fix this thing

PR #25079, opened 2026-08-13
website/docs: document ForceAuthn and typed AKQL JSON values

dominic-r:
Found while auditing 2026.8 docs for staleness ahead of the release — see the [Slack thread](https://authentiksecurity.slack.com/archives/C08C0SJ4V6D/p1786634134866959).

Two 2026.8 behavior changes appear only in the release notes and nowhere in the docs.

### SAML `ForceAuthn`

`authentik/providers/saml/views/sso.py` now re-authenticates when the SP requests it:

```python
# If the SP requested ForceAuthn, we must re-authenticate the user regardless of
# any existing session, mirroring the OIDC provider's prompt=login handling.
if auth_n_request and auth_n_request.force_authn:
    reauth_response = self.check_force_authn(request)
```

`ForceAuthn` appeared nowhere in `website/docs/` outside the release notes. This matters more than a typical new feature: it needs no configuration, applies to any SP that sets the attribute, and changes login behavior for existing applications the moment an instance is upgraded. The new section says so explicitly, so anyone investigating "users are being asked to sign in more often since 2026.8" lands on the answer.

### AKQL typed JSON values

#24028 gave `JSONSearchField` `value_types = [str, bool, int, float, Decimal]`, so JSON fields now match Booleans and numbers rather than only strings. The **JSON and nested fields** section showed only string comparisons, so nothing was wrong — it just did not say the capability exists. Adds a note and two examples.

### DIFF
--- a/website/docs/add-secure-apps/providers/saml/index.mdx
+++ b/website/docs/add-secure-apps/providers/saml/index.mdx
@@ -153,6 +153,14 @@ In authentik, it's also possible to configure the default SAML NameID policy use
 By default, users are free to change their email addresses. Therefore, it is recommended to either: avoid using a user's email address as the NameID attribute or, if possible, disallow changing email addresses.
 :::
 
+## ForceAuthn:ak-version[2026.8]
+
+An SP can set the `ForceAuthn` attribute on its `AuthnRequest` to require that the user authenticates again, even when they already have a valid authentik session.
+
+authentik honors this attribute. When an incoming `AuthnRequest` sets `ForceAuthn="true"`, the user is sent through the provider's authentication flow before the SSO flow continues. If the provider does not have an authentication flow configured, authentik uses the authentication flow configured for the associated brand. This behaves similarly to `prompt=login` for OAuth 2.0 and OpenID Connect providers. Once the user has re-authenticated, the SSO flow proceeds normally.
+
+This requires no configuration on the provider, and it applies to any SP that sets the attribute. If users of an existing application are prompted to sign in again more often after upgrading to 2026.8, check whether that SP sets `ForceAuthn`.
+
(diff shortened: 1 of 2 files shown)


website/docs/add-secure-apps/providers/saml/index.md
```diff
@@ -153,6 +153,14 @@ In authentik, it's also possible to configure the default SAML NameID policy use
 By default, users are free to change their email addresses. Therefore, it is recommended to either: avoid using a user's email address as the NameID attribute or, if possible, disallow changing email addresses.
 :::
 
+## ForceAuthn:ak-version[2026.8]
+
+An SP can set the `ForceAuthn` attribute on its `AuthnRequest` to require that the user authenticates again, even when they already have a valid authentik session.
+
+authentik honors this attribute. When an incoming `AuthnRequest` sets `ForceAuthn="true"`, the user is sent through the provider's authentication flow before the SSO flow continues, in the same way that `prompt=login` behaves for OAuth 2.0 and OpenID Connect providers. Once the user has re-authenticated, the SSO flow proceeds normally.
```
dewi-tik:
```suggestion
authentik honors this attribute. When an incoming `AuthnRequest` sets `ForceAuthn="true"`, the user is sent through the provider's authentication flow before the SSO flow continues. If the provider does not have an authentication flow configured, authentik uses the authentication flow configured for the associated brand. This behaves similarly to `prompt=login` for OAuth 2.0 and OpenID Connect providers. Once the user has re-authenticated, the SSO flow proceeds normally.
```

I assume?

dewi-tik (changes requested):

dewi-tik (approved):

PR #25093, opened 2026-08-14
core: scope user path_startswith filter to the path subtree

### LINKED ISSUE #23212: Clicking a user tree node shows users from sibling paths
### Describe the bug

The user path tree in the sidebar renders correctly (e.g., `users/group1` and `users/group11` appear as separate siblings under `users`). However, clicking `group1` in the tree also displays users whose path is `users/group11` in the content area. The tree navigation is correct, but the filtered user list is wrong.

### How to reproduce

1. Create a user with path `users/group1`
2. Create another user with path `users/group11`
3. In the sidebar tree, expand `users` — see `group1` and `group11` as correct siblings
4. Click `group1`
5. Observe that the user list shows both the user in `group1` and the user in `group11`

### Expected behavior

Clicking `group1` should only show users whose path is `users/group1` or a sub-path like `users/group1/sub`. Users under `users/group11` should not appear.

### Screenshots

_No response_

### Additional context

_No response_

### Deployment Method

Docker

### Version

2026.5.3

### Relevant log output

```shell

```

NoiceHax:
`path_startswith` on the users API used a plain `startswith` lookup, so it matched any path with the same text prefix. Filtering on `users/group1` also returned users under `users/group11`. In the admin UI that means clicking a node in the user path tree shows users from sibling nodes.

I swapped it for a small filter method that matches the path exactly or as a prefix followed by a slash. Trailing slashes get stripped off the value first, and an empty value returns the queryset unfiltered, same as the old lookup did.

The test is `test_path_startswith` in authentik/core/tests/test_users_api.py, next to `test_paths`. It creates users at `<root>/group1`, `<root>/group1/sub` and `<root>/group11`, then checks the sibling never comes back and that filtering on `<root>` still returns all three. I could not run the suite locally, so CI is the real check.

Closes #23212


abhisheksharma2411 (commented):
This looks right to me, and it's a nice catch — `foo/bar` matching `foo/bar2` is the classic prefix-vs-subtree slip, and the `Q(path=value) | Q(path__startswith=f"{value}/")` shape is the standard fix. The test picking `group1` / `group11` as the sibling pair is exactly the case I'd have wanted to see.

I poked at it a bit rather than just reading:

- `Q` is already imported at `authentik/core/api/users.py:10`, so no new import needed 👍
- `git grep 'lookup_expr="startswith"'` turns up only this one, so there's no sibling instance of the same bug lurking elsewhere
- the concrete beneficiary is the admin UI — `web/src/admin/users/UserListPage.ts` passes `pathStartswith: this.activePath` from the path tree, so selecting `foo/group1` there was quietly listing `foo/group11`'s users too. Might be worth a line in the PR description, since it makes the user-visible impact obvious

Two small things, neither blocking:

**1. It's a semantics change on a public API filter.** Anyone using `path_startswith` as a loose prefix *search* (`foo/gr` to find `foo/group1`) gets nothing back now. That's the intended fix and the right call, but it might deserve a release-note line so it doesn't look like a regression from the outside.

**2. Is the empty guard reachable?** I might have this wrong, but I think django-filter short-circuits on `EMPTY_VALUES` before it ever calls a `method=`, so `path_startswith=` never reaches `filter_path_startswith` — which would leave `if not value` catching only the all-slashes case like `/`. Harmless either way, and defensive is fine; I mostly mention it so a future cleanup doesn't delete it as dead code without noticing it's the `/` guard.

On that: `"/"` used to match nothing (paths have no leading slash — `path = models.TextField(default="users")`) and now returns everything unfiltered. Genuinely trivial, just noting it since it's the one input where the change widens rather than narrows.

Nice change 🙂

NoiceHax:
Thanks for the review, and for checking for sibling instances of the same bug.

On the admin UI: the first paragraph of the description already calls this out. Selecting foo/group1 in the user path tree was quietly showing foo/group11's users too, which is the visible impact.

On the release note: the per-release changelog is generated from commit subjects (make gen-changelog), so this will land under minor changes/fixes as the fix commit on its own. No manual entry needed.

On the empty guard: you're right that django-filter short-circuits on empty values, so if not value effectively guards the all-slash / case after the rstrip. I kept it for that reason, so a future cleanup doesn't mistake it for dead code.

BeryJu (approved):

PR #25102, opened 2026-08-14
packages/django-dramatiq-postgres/broker: chunked purge queryset

### LINKED ISSUE #22581: 2026.5.0 Used all my RAM
### Describe the bug

I upgraded and then immediately ran out of RAM (Authentik Python used it up)
I then increased the SWAP file and it used that up too.

### How to reproduce

Upgrade to 2026.5.0 using Docker

### Expected behavior

Worked like before

### Screenshots

_No response_

### Additional context

_No response_

### Deployment Method

Docker

### Version

2026.5.0

### Relevant log output

```shell

```

rissson:
probably closes https://github.com/goauthentik/authentik/issues/22581

Fix the usual memory issue in django when deleting many objects


authentik/lib/utils/db.py
```diff
@@ -1,39 +1,5 @@
 """authentik database utilities"""
 
-import gc
-from collections.abc import Generator
+from django_dramatiq_postgres.utils import chunked_queryset
```
BeryJu: Instead of just importing it here to make it available i'd rather update all imports

rissson: fair enough

rissson: Actually, it seems we import from authentik already in `packages/django-postgres-cache/django_postgres_cache/tasks.py`, so might as well do it like that too.

BeryJu (approved):

PR #25148, opened 2026-08-17
stages/prompt: preserve empty initial values

tungyhardDevOps:
### What does this PR change?
Preserves empty initial values as defaults when creating read-only prompt fields. This allows optional read-only text and text-area prompts whose initial value expression returns an empty string to be submitted successfully.

Adds a regression test covering a non-required read-only prompt with an empty initial value.
Read-only prompt fields already set `allow_blank=True`, but the empty initial value was dropped because defaults were only assigned when they were truthy.As a result, an empty string was not retained as the read-only field's default and user settings flows could reject the field with "This field may not be
blank."
### How was this tested?
- Added `test_read_only_blank_value` to verify that an optional read-only prompt
  preserves an empty default and validates an empty submission.
- Ran Python syntax compilation for the changed files.
- Ran `git diff --check`.

The focused Django test suite was not run locally because the available Python
runtime did not include the project test dependencies.
### Linked issues
closes #18899

---

## Checklist


rissson:
hiya, there is no need to keep merging main here. We'll get to reviewing this when we can.

tungyhardDevOps:
> hiya, there is no need to keep merging main here. We'll get to reviewing this when we can.

alright thanks for the feedback

rissson (approved):

BeryJu (approved):

Issue #18179, opened 2025-11-17
Wild language mix


Skeeve:
### Describe the bug

The issue is described here: https://github.com/goauthentik/authentik/issues/12516
and here: https://github.com/goauthentik/authentik/issues/9433

Both of them are closed.

### How to reproduce

1. Go to Settings (for the user)
2. Change from "Automatische Erkennung…" to EN - English
3. Save
4. Notice that there will be a wild mix between (in my case) German and English.

### Expected behavior

The UI should respect the settings.

### Screenshots

Example for flows:

<img width="1181" height="310" alt="Image" src="https://github.com/user-attachments/assets/0f585699-a1fb-463c-839b-581562adec54" />

Clicking "Create" gives me for example "Benutzer Schreib Stage" instead of "User Write Stage"

<img width="581" height="991" alt="Image" src="https://github.com/user-attachments/assets/303a73a6-8030-4b1a-8e9c-d76f9cd53e16" />

### Additional context

Also note: My Browser (Vivaldi) is set to english.

Also my Mac OS is set to english.

### Deployment Method

Docker

### Version

2025.10.1

### Relevant log output

```shell

```

BeryJu:
This is the expected behaviour when a certain string is not translated. Our translations are community-maintained here, and anyone can check here if they want to help out:  https://app.transifex.com/authentik/authentik/dashboard/

Skeeve:
> This is the expected behaviour when a certain string is not translated. Our translations are community-maintained here, and anyone can check here if they want to help out: https://app.transifex.com/authentik/authentik/dashboard/

So German is the default language? I highly doubt it.

BeryJu:
I mis-read your issue, yeah this is odd indeed

dewi-tik:
Can you please re-test this once 2025.12 is released? It includes significant changes to locale handling.

GirlBossRush:
This issue occurs because stage descriptions are stored in the model docstring. Our message extraction currently can't access this information—this is a known limitation we're actively addressing.

Skeeve:
I beg to disagree. It must be something related to the browser I guess.

Snapshots from Vivaldi:

<img width="419" height="185" alt="Image" src="https://github.com/user-attachments/assets/cc5abf15-7ec7-4087-a488-0114b6c424a0" />

<img width="757" height="726" alt="Image" src="https://github.com/user-attachments/assets/4ce6c089-85b3-49c4-828a-d2e953460d28" />

---

[...]

PR #25156, opened 2026-08-17
web/i18n, core: Persist locale in a cookie and reload to switch

GirlBossRush:
## Details

Locale becomes fixed for the lifetime of a page. Instead of swapping the language live in JS (the source of mixed client/server language state — #18179 / #22954), every locale-change path now persists the preference to the Django language cookie (`authentik_language`, which `LocaleMiddleware` already reads) and reloads, so the **server** re-renders every string — including server-rendered flow challenges — in the new locale.

### Change
- `persist.ts`: `persistLocale` / `applyLocaleChange` write the `authentik_language` cookie + reload.
- Switch the flow locale selector, the user-settings prompt stage, and `SessionContextController`'s post-render locale application off live `setLocale`.
- Delete the `sessionStorage["authentik:locale"]` path and drop it from `autoDetectLanguage`; the server-resolved hint (`window.authentik.locale`) is now authoritative.
- `InterfaceView.dispatch` honors `?locale=` server-side, validated via `check_for_language` (forces `TemplateResponse.render()` inside the `override`, since it renders lazily).

### Tests
- Colocated web unit test for the cookie helper.
- Python test for the server-side `?locale=` override (applied when valid, ignored when not).

### DIFF
--- a/authentik/core/middleware.py
+++ b/authentik/core/middleware.py
@@ -11,7 +11,7 @@ from django.core.exceptions import ImproperlyConfigured
 from django.http import HttpRequest, HttpResponse, HttpResponseBadRequest
 from django.utils.deprecation import MiddlewareMixin
 from django.utils.functional import SimpleLazyObject
-from django.utils.translation import override
+from django.utils.translation import check_for_language, override
 from structlog.contextvars import STRUCTLOG_KEY_PREFIX
 
 from authentik.lib.tracing import active_tracer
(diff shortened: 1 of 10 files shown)


authentik/core/views/interface.py
```diff
@@ -54,6 +55,21 @@ def dispatch(self, request: HttpRequest, *args: Any, **kwargs: Any) -> HttpRespo
...
+        locale = request.GET.get("locale")
+        if locale and check_for_language(locale):
+            with override(locale):
+                response = super().dispatch(request, *args, **kwargs)
+                # TemplateResponse renders lazily, after this context exits; force it
+                # now so the shell is rendered while the override language is active.
+                if hasattr(response, "render") and callable(response.render):
+                    response.render()
+                return response
+        return super().dispatch(request, *args, **kwargs)
```
BeryJu: I dont think this should be view specific, this should be a middleware

PR #25160, opened 2026-08-17
core: distroless server image

emilburzo:
### TODO

- [ ] stacked PRs attempt: docker vs lifecycle
- [ ] debug: python/rust dev dependencies + curl, wget

---

Distroless style server image. 

The final stage starts `FROM scratch` and has only the files belonging to an explicit list of packages + a list of extra paths. This means there's nothing in the image that isn't actively used (like wget, adduser, bash, etc).

### Changes

- A prune script assembles the runtime filesystem by asking dpkg which files belong to the packages in the explicit list, adds the paths no package owns, and copies the result to a staging directory. It also writes a dpkg status directory so SBOM tooling still sees the package list in an image with no dpkg database.

- Three text files define the packages to keep, extra paths to keep, paths to remove again. The build fails if an entry matches nothing, so the lists cannot go stale.

- lifecycle/ak was removed because there is no shell to run it. It was replaced by the authentik binary as a boot subcommand, and /lifecycle/ak ships as asymlink to that binary.

- ak.py, the console script used in dev checkouts, now `exec`s instead of shelling out, and dispatches to the compiled binary if there is one, `cargo` if not, and the config loader for `dump_config`.

- The blueprint shell treats `readline` as optional, since the image no longer ships `libreadline`. The command still runs, just without history and line editing.

- A `server-debug` image is built and published: the same build but with all the useful tooling included: a shell, apt, the dev dependency group. Needed for `ak test-all`, `AUTHENTIK_DEBUGGER` and `docker exec ... bash`.


lifecycle/container/Dockerfile
```diff
@@ -60,7 +60,8 @@ RUN --mount=type=secret,id=GEOIPUPDATE_ACCOUNT_ID \
     /bin/sh -c "GEOIPUPDATE_LICENSE_KEY_FILE=/run/secrets/GEOIPUPDATE_LICENSE_KEY /usr/bin/entry.sh || echo 'Failed to get GeoIP database, disabling'; exit 0"
 
 # Stage: download Rust toolchain
-FROM ghcr.io/goauthentik/fips-debian:trixie-slim-fips@sha256:ed780d7d3d789d7056855e8c7e3c37d21368ea42f39305fe431c19ae2e08d6a3 AS rust-toolchain
+# TEMPORARY: pinned to goauthentik/fips#178 (remove wget/curl). Revert before merge.
+FROM ghcr.io/goauthentik/fips-debian:trixie-slim-fips-pr-178@sha256:e048842e22989f2c0eaa7029d8a3d662a816504b0a9275917b5832a02e8cecda AS rust-toolchain
```
emilburzo:
```suggestion
# TEMPORARY: pinned to goauthentik/fips#178 (remove wget/curl). Revert before merge.
FROM ghcr.io/goauthentik/fips-debian:trixie-slim-fips-pr-178@sha256:cd66513ddc1a2735f120db27add7dffa409f601fec3ec813a69bba9b126e845e AS rust-toolchain

File: lifecycle/container/Dockerfile [hunk 2]
```diff
@@ -82,7 +84,8 @@ RUN --mount=type=bind,target=rust-toolchain.toml,src=rust-toolchain.toml \
 # Stage: Download uv
 FROM ghcr.io/astral-sh/uv:0.12.7@sha256:95f2aa1fe59274951cfe9b0cbc7972e879ff1004bc8945d130a32eb0dbd85945 AS uv
 # Stage: Base python image
-FROM ghcr.io/goauthentik/fips-python:3.14.7-slim-trixie-fips@sha256:61a327d4e98787225872b21d03305a77de4cc179d76c9f0f4b44b3cbe089ae59 AS python-base
+# TEMPORARY: pinned to goauthentik/fips#178 (remove wget/curl). Revert before merge.
+FROM ghcr.io/goauthentik/fips-python:3.14.7-slim-trixie-fips-pr-178@sha256:9f7cb71b77da3d38b3c7b77bb1cd913e9db90270137321771e97d25b9d07e8e8 AS python-base
```
emilburzo:
```suggestion
# TEMPORARY: pinned to goauthentik/fips#178 (remove wget/curl). Revert before merge.
FROM ghcr.io/goauthentik/fips-python:3.14.7-slim-trixie-fips-pr-178@sha256:395b057183f7e567f3cef6b06e7dffd5e10b7e8bedb94866b613449e1ca5ab28 AS python-base
```

emilburzo:
Closing in favor of stacked PR: https://github.com/goauthentik/authentik/pull/26320

PR #25182, opened 2026-08-18
test(outpost/proxy): gracefully skip postgres store tests when database is unreachable

aoright:
### Summary
- In `internal/outpost/proxyv2/postgresstore/postgresstore_test.go`, `internal/outpost/proxyv2/postgresstore/connpool_test.go`, and `internal/outpost/proxyv2/application/session_postgres_test.go`, gracefully skip unit tests using `t.Skipf` when PostgreSQL connection fails in developer environments.
- Guard `CleanupTestDB` from potential nil pointer dereferences.


rissson:
That code is not used anymore and will be removed soon.

Issue #25191, opened 2026-08-18
OpenID Connect source: HTTP 500 when the IdP's `groups` claim contains objects instead of strings


dunningdan:
### Describe the bug

An `openidconnect` OAuth source crashes with an unhandled `TypeError` (HTTP 500) on the callback if the IdP returns `groups` as a list of objects. Login through the source becomes impossible.

[`sources/oauth/types/oidc.py#L89`](https://github.com/goauthentik/authentik/blob/main/authentik/sources/oauth/types/oidc.py#L89) copies the claim through unchanged:

  ```python
  "groups": info.get("groups", []),
  ```

[`core/sources/flow_manager.py#L133-L143`](https://github.com/goauthentik/authentik/blob/main/authentik/core/sources/flow_manager.py#L133-L143) then uses each element as a dict key:

  ```python
  self.groups_properties = {
      group_id: self.mapper.build_object_properties(
          object_type=Group,
          request=request,
          user=None,
          group_id=group_id,
          **self.user_info,
      )
      for group_id in self.user_properties.setdefault("groups", [])
  }
  ```

Dicts aren't hashable, so this raises `TypeError: cannot use 'dict' as a dict key`.

**Example**: Oracle Identity Domains returns this from `/oauth2/v1/userinfo` when the `groups` scope is requested:

  ```json
  [
    {"id": "0db0…", "name": "pixee-admins",   "$ref": "https://…/admin/v1/Groups/0db0…"},
    {"id": "28b5…", "name": "Administrators", "$ref": "https://…/admin/v1/Groups/28b5…"}
  ]
  ```

### How to reproduce

1. Create an OAuth source, provider type **OpenID Connect**, against an IdP that returns object-shaped `groups`.
2. Make sure the user belongs to at least one group at the IdP.
3. Sign in through the source → callback returns **HTTP 500**, `system_exception` event recorded.

A user in *no* groups signs in fine (empty claim, so no dict is used as a key).

Reproducible without an IdP, in `ak shell`, against any existing `openidconnect` source:

  ```python
  from authentik.core.models import User
  from authentik.core.sources.mapper import SourceMapper
  from authentik.sources.oauth.models import OAuthSource

  source = OAuthSource.objects.get(slug="<your-openidconnect-source>")
  info = {"email": "user@example.com",
          "groups": [{"id": "0db0", "name": "team-a", "$ref": "https://idp.example.com/Groups/0db0"}]}

  props = SourceMapper(source).build_object_properties(
[...]

PR #25195, opened 2026-08-18
sources/oauth: cover object group property mappings

rigel08:
## What does this PR do?

Adds a test for OAuth groups returned as objects.

This makes sure we have coverage for the object-based group format and catches the current `TypeError` behavior if it regresses.

## Testing

Ran:

`uv run pytest --reuse-db authentik/sources/oauth/tests/test_property_mappings.py -q`

All 5 tests pass.


abhisheksharma2411 (commented):
Thanks for adding coverage here — object-shaped `groups` genuinely is an untested path, and it's a good thing to have pinned. One thing I think needs rethinking before this lands, and it's about what the test asserts rather than how it's written.

**The test locks in a crash as expected behaviour.**

```python
with self.assertRaises(TypeError):
    OAuthSourceFlowManager(self.source, request, IDENTIFIER, {"info": info}, {})
```

That `TypeError` is #25191 — an open, triaged `bug` filed two days ago. It's unhandled, so it surfaces as **HTTP 500** and makes login through the source impossible for any user who belongs to at least one group at an IdP that returns objects (Oracle Identity Domains is the reported case). The PR body describes it as *"catches the current `TypeError` behavior if it regresses"*, but a crash going away isn't a regression — it's the fix. Whoever fixes #25191 will find this test red and have to work out whether they broke something.

I've walked into this exact trap twice recently in my own work, so I sympathise — a test written around current behaviour reads as a safety net right up until the behaviour turns out to be the bug.

Would you be up for turning it into the assertion the fix will need?

```python
def test_group_property_mappings_with_object_groups(self):
    """Object-shaped `groups` claims must not crash the flow manager (#25191)."""
    info = deepcopy(INFO)
    info["groups"] = [{"id": "group-1", "name": "Admins"}]
    request = self.request_factory.get("/", user=AnonymousUser())

    flow_manager = OAuthSourceFlowManager(self.source, request, IDENTIFIER, {"info": info}, {})
    self.assertIn("group-1", flow_manager.groups_properties)
```

That fails today for the right reason and passes the moment #25191 is fixed — which makes this PR the regression test for it, rather than something the fix has to undo. If you'd rather keep the scope to "characterise today's behaviour", marking it `@expectedFailure` with a `# TODO(#25191)` would at least keep the intent legible to whoever picks the bug up.

**Two smaller things, both mechanical:**

**1. The unrelated reformatting will fail `ci-lint-black`.** This repo sets `line-length = 100` (`pyproject.toml` L153) and the original one-liner is 99 characters, so black actively wants it collapsed back. I ran it rather than guessed:

```
$ black --check --diff authentik/sources/oauth/tests/test_property_mappings.py
-        flow_manager = OAuthSourceFlowManager(
-            self.source, request, IDENTIFIER, {"info": INFO}, {}
-        )
+        flow_manager = OAuthSourceFlowManager(self.source, request, IDENTIFIER, {"info": INFO}, {})
would reformat ...
```

Reverting those three hunks keeps the diff to just the new test, which is easier to review anyway. Your PR currently shows only 9 checks, so the lint job probably hasn't run yet — first-time contributor PRs wait on a maintainer to release the full run, which is why this hasn't shown up as red.

**2. Missing newline at end of file** — the last hunk ends with `No newline at end of file`.

The instinct to cover this path is the right one; it's just worth pointing it at the behaviour we want rather than the one we currently have 🙂

abhisheksharma2411:
Heads up so this doesn't blindside you — I've opened #25208 with a fix for #25191, and it touches this same file.

I'd rather that not read as stepping on your PR, because it isn't meant to. You found a genuinely untested path, and the coverage gap you spotted is the reason the crash sat there. The fix turned out to need two layers, which was more than a repointed assertion could carry:

- `SourceFlowManager` skips identifiers it can't use as a dict key and records a Configuration error event, so no source can 500 on this — including a user property mapping that merges something odd into `groups`.
- `OpenIDConnectType` reduces object entries to the `id`/`value`/`sub` they carry, so the Oracle case in the issue actually gets its groups rather than merely not crashing.

Your test becomes `test_object_group_property_mappings` in mine, asserting the mapping result instead of the `TypeError`. If a maintainer would rather land yours first, say so and I'll drop my equivalent test and rebase on top of it — the discovery is yours either way, and I've said as much in the PR description.

The two mechanical notes from my earlier review still apply if you keep this open: the reformatted hunks will fail `ci-lint-black`, and there's a missing trailing newline 🙂

rigel08:
Thanks for the heads up, no worries at all. I’m happy for #25208 to carry the actual fix. I’ll clean up my PR based on your feedback and let the maintainers decide which one they’d prefer to land first.

Appreciate the explanation and the credit for finding the untested path!

rissson:
I think to give credit to both of you for working on this, we'll merge this one first, and the other ones then, if everyone is happy with that.

rissson (approved):

PR #25204, opened 2026-08-18
core/sources: forward kwargs when updating an existing connection

### LINKED ISSUE #25115: SourceFlowManager.get_action() drops **kwargs when re-linking an existing connection, breaking source login for already-authenticated users (Plex: NOT NULL violation on plex_token)
**authentik version:** 2026.5.6 (also present on `main` as of 2026-08-13)
**Deployment:** helm / Kubernetes
**Source type:** Plex (but the defect is in shared `core` code, so any source whose `update_user_connection()` reads from `kwargs` is affected)

## Description

`authentik/core/sources/flow_manager.py`, `SourceFlowManager.get_action()` has three calls to
`update_user_connection()`. Two forward `**kwargs`; the middle one does not:

```python
def get_action(self, **kwargs) -> tuple[Action, UserSourceConnection | None]:
    # When request is authenticated, always link
    if self.request.user.is_authenticated:
        new_connection = self.user_connection_type(
            source=self.source, identifier=self.identifier
        )
        new_connection.user = self.request.user
        new_connection = self.update_user_connection(new_connection, **kwargs)   # ok
        if existing := self.user_connection_type.objects.filter(
            source=self.source, identifier=self.identifier
        ).first():
            existing = self.update_user_connection(existing)                     # <-- kwargs dropped
            return Action.AUTH, existing
        return Action.LINK, new_connection

    action, connection = self.matcher.get_user_action(self.identifier, self.user_properties)
    if connection:
        connection = self.update_user_connection(connection, **kwargs)           # ok
    return action, connection
```

Subclasses rely on `kwargs` to carry the credential. `PlexSourceFlowManager`
(`authentik/sources/plex/plex.py`):

```python
def update_user_connection(self, connection, **kwargs) -> UserPlexSourceConnection:
    """Set the access_token on the connection"""
    connection.plex_token = kwargs.get("plex_token")
    return connection
```

With no `kwargs`, `kwargs.get("plex_token")` returns `None`, the previously-valid token on the
existing row is overwritten with `NULL`, and `PostSourceStage` fails when it saves.

## Reproduction

1. Configure a Plex source and link a user (connection row now exists with a valid `plex_token`).
2. Sign in to authentik so the browser holds a session.
3. While still signed in, start a source login for that same user — e.g. an OIDC client sending
   `prompt=login`, which forces the authentication flow to re-run against a live session — and
   click "Sign in with Plex".

Requires **both** an authenticated session **and** a pre-existing connection for
`(source, identifier)`; that is the only path reaching the faulty line. Anonymous logins are
unaffected, which makes this look intermittent — it reproduces reliably in a normal browser and
never in a private window.

## Result

```
django.db.utils.IntegrityError: null value in column "plex_token" of relation
"authentik_sources_plex_userplexsourceconnection" violates not-null constraint
DETAIL:  Failing row contains (1, null).

  File "/authentik/flows/views/executor.py", line 297, in get
    stage_response = self.current_stage_view.dispatch(request)
  File "/authentik/core/sources/stage.py", line 25, in dispatch
    connection.save()
```

## Expected

Re-authenticating an existing source connection preserves (or refreshes) the credential and
succeeds, as the anonymous path already does.

## Suggested fix

```diff
-            existing = self.update_user_connection(existing)
+            existing = self.update_user_connection(existing, **kwargs)
```

This also makes the branch consistent with the two sibling calls, and lets the token be
*refreshed* on re-auth rather than merely preserved — probably the intended behaviour, since a
Plex token can be revoked or rotated independently of authentik.

## A second failure on the same branch (not yet root-caused)

After working around the NOT NULL violation (`ALTER TABLE
authentik_sources_plex_userplexsourceconnection ALTER COLUMN plex_token DROP NOT NULL`), the
same authenticated path still fails, now further along:

```
"event": "No Pending user to login.", "level": "warning",
"logger": "authentik.flows.stage",
"stage": "default-source-authentication-login",
"stage_view": "authentik.stages.user_login.stage.UserLoginStageView"
```

The user is shown "Request has been denied. / Unknown error" — the empty message being
consistent with `stage_invalid()` being called without one.

This is reproducible under exactly the same preconditions as above (authenticated session +
pre-existing connection), and never on the anonymous path. Same-session comparison from one
browser, 12 seconds apart:

| `redeem_token` ran as | outcome                                     |
| --------------------- | ------------------------------------------- |
| unauthenticated       | `login` event created — succeeds            |
| session               | `No Pending user to login.` — access denied |

I have not traced which line drops `pending_user`: `get_action()` returns `Action.AUTH` with a
connection whose `.user` is a valid FK, and `handle_auth()` passes
`PLAN_CONTEXT_PENDING_USER: connection.user` into `_prepare_flow()`, so on a reading of the code
it should be present by the time `UserLoginStageView` runs. Reporting it as an observation in
case it is the same defect seen further downstream, or a second one in the same branch.

## Impact / severity

Any deployment where users hold an authentik session and re-authenticate through a source hits
this. It is a hard 500 with no user-facing explanation, and it is data-destructive in intent —
it attempts to null a stored credential; only the NOT NULL constraint prevents silent corruption.
A source model with a nullable credential column would corrupt the row instead of erroring.

## Workaround

Insert a `user_logout` stage as the first stage of the authentication flow used for
re-authentication, so the source stage runs anonymous and takes the working branch. Safe because
the executor keeps the flow plan in memory and `stage_ok()` re-persists it after the session
flush (`flows/views/executor.py:422`). Cost: the user is signed out of other applications in
that browser.

### LINKED ISSUE #25325: Plex source token becomes invalid or drops kwargs on re-auth
## What

Plex source token: “Plex token invalid, please re-authenticate” and SourceFlowManager.get_action() dropping kwargs (NOT NULL plex_token).

## Want

- Plex source token: “Plex token invalid, please re-authenticate”
- SourceFlowManager.get_action() dropping kwargs (NOT NULL plex_token)

## Ideas this came from

- [#17089 Plex token invalid, please re-authenticate source.](https://github.com/goauthentik/authentik/issues/17089) — Since I updated to 2025.8.3 version I'm getting spam by the Plex token invalid, please re-authenticate source. error. I even tried to remove Plex from the sourc
- [#25115 SourceFlowManager.get_action() drops **kwargs when re-linking an existing connection, breaking source login for already-authenticated users (Plex: NOT NULL violation on plex_token)](https://github.com/goauthentik/authentik/issues/25115) — authentik version: 2026.5.6 (also present on main as of 2026-08-13) Deployment: helm / Kubernetes Source type: Plex (but the defect is in shared core code, so a

abhisheksharma2411:
Closes #25115.

`SourceFlowManager.get_action()` calls `update_user_connection()` three times. Two forward `**kwargs`; the branch taken when an already-authenticated user re-links an existing connection does not:

```python
new_connection = self.update_user_connection(new_connection, **kwargs)   # ok
if existing := self.user_connection_type.objects.filter(...).first():
    existing = self.update_user_connection(existing)                     # kwargs dropped
    return Action.AUTH, existing
```

Sources carry their credential in those kwargs. `PlexSourceFlowManager.update_user_connection()` reads `kwargs.get("plex_token")`, so on this path it gets `None`, the previously valid token on the existing row is overwritten with NULL, and `PostSourceStage` fails on save with a NOT NULL violation.

Plex is where it surfaces loudly, but the defect is in shared `core` code — any source whose `update_user_connection()` reads from kwargs is affected. Thanks to the reporter for localising it to the line.

## Tests

Two: the existing connection receives the kwargs, and the LINK path — already correct — keeps doing so.

**One thing worth flagging, because it nearly cost me a useless test.** My first version subclassed the flow manager and accumulated kwargs into a single dict:

```python
def update_user_connection(self, connection, **kwargs):
    seen.update(kwargs)      # <- passes with or without the fix
```

That test passed against the *unfixed* code. `get_action()` updates a fresh unsaved connection before it reaches the existing row, and that first call always forwarded kwargs — so the accumulated dict was already populated by the time the buggy call ran, and the assertion proved nothing.

The version here records `(connection.pk, kwargs)` per call and asserts on the call for the *saved* connection. With the fix reverted it fails with:

```
AssertionError: None != 'a-credential' : get_action() dropped kwargs when
updating the existing connection
```

which is the check I'd want on a fix this small.

## Verification

```
authentik/core/tests/ + authentik/sources/plex/   305 passed
black --check authentik packages tests scripts lifecycle   2231 files unchanged
ruff check authentik                                       All checks passed!
```

No schema or client regeneration needed — no model or serializer change.

Closes #25325


abhisheksharma2411:
Flagging the one red check so it doesn't cost a reviewer time: `test-openid-conformance (oidc_backchannel)` is a webdriver startup failure, not a behaviour change.

```
TestOpenIDConformanceBackchannel.test_oidcc_backchannel_logout_certification_test_plan
ValueError: Webdriver failed after 3.
```

The driver never came up, so no conformance assertion was reached. The same job **passed on attempt 1 of this exact commit** (`d4158b04d`) and failed on attempt 2 — same code, both outcomes:

| attempt | `oidc_backchannel` |
|---|---|
| 1 | ✅ success |
| 2 | ❌ `Webdriver failed after 3` |

For what it's worth, this also isn't a repo-wide flare — it's green on `main` and on the other open PRs I checked, which is why I went and compared the two attempts rather than assuming.

On the earlier attempt: that run was cancelled by CI after 86 jobs had already passed, which left `ci-core-mark` red through no fault of the branch. I don't have permission to re-run jobs, so I closed and reopened to re-trigger — apologies for the notification noise, it was the only lever available to me.

The change itself is unrelated to authentik as an OIDC provider: it only affects `SourceFlowManager.get_action()` forwarding `**kwargs` when re-linking an existing source connection. Happy to rebase if you'd rather see a clean run.

authentik/core/sources/flow_manager.py
```diff
@@ -154,7 +154,12 @@ def get_action(self, **kwargs) -> tuple[Action, UserSourceConnection | None]:  #
             if existing := self.user_connection_type.objects.filter(
                 source=self.source, identifier=self.identifier
             ).first():
-                existing = self.update_user_connection(existing)
+                # `**kwargs` carries the credential for sources that store one
+                # (Plex's `plex_token`, for example). Dropping it here made
+                # `update_user_connection()` overwrite a valid stored token with
+                # None on every re-link, so the AUTH path is the one that must
+                # forward it — the LINK path below already did.
```
rissson:
```suggestion
```

rissson (changes requested):
otherwise lgtm

rissson:
also, please read https://github.com/goauthentik/authentik/blob/main/AI_POLICY.md

abhisheksharma2411:
Thanks for applying the suggestion directly and for merging main in — and sorry you had to do that rather than me.

## Disclosure

Per `AI_POLICY.md`, which you pointed me at: this contribution was AI-assisted. I used Claude Code for the investigation, the patch, and the test, and I reviewed and directed the work throughout. On this one that specifically included throwing away a first version of the regression test, which recorded kwargs into a single accumulated dict and therefore **passed against the unfixed code** — `get_action()` calls `update_user_connection()` twice on that path and the first call always forwarded kwargs, so the assertion proved nothing. The test now keys on the saved connection's pk, and I confirmed it fails on unpatched code before trusting it.

I should have disclosed up front instead of waiting to be asked, and I'm sorry for that. Happy to answer questions on any part of the change.

I take your point on the comment, too — it was doing the job a commit message should do. I've trimmed the equivalent one on #25208 for the same reason.

rissson (approved):

PR #25205, opened 2026-08-18
web/flow: fix remember-me being cleared by CSRF token rotation

so-jelly:
### Details

The identification stage's "Remember me on this device" toggle never is cleared on login, the next visit the username is not prefilled and the toggle is unchecked.

`RememberMeController` derived a "session ID" from the first 8 characters of the `authentik_csrf` cookie, stored it alongside the username, and deleted the stored username whenever the two no longer matched. Django rotates the CSRF token on login, so that value can never match on the user's next visit.

**Reproduction (no login required)**

1. Open a flow page with an identification stage that has `enable_remember_me` on.
2. In the browser console: `localStorage.setItem("authentik-remember-me-user", "user@example.com"); localStorage.setItem("authentik-remember-me-session", "OLDTOKEN")`.
3. Reload — `authentik-remember-me-user` is deleted, the field is empty, the toggle unchecked.

**Fix**

The stored username is now cleared only by an explicit user action: switching the toggle off, or following the "Not you?" link, which already calls `RememberMeStorage.reset()` from `FormStatic`. The legacy `authentik-remember-me-session` key is deleted so state written by earlier versions is cleaned up.

Adds `RememberMeController.unit.test.ts` covering the behavior, including a regression case asserting the username survives a rotated CSRF cookie (fails on `main`, passes here).

### Checklist

- [x] Local tests pass (`cd web && pnpm test`)

### DIFF
--- a/web/src/flow/stages/identification/controllers/RememberMeController.ts
+++ b/web/src/flow/stages/identification/controllers/RememberMeController.ts
@@ -67,6 +67,12 @@ export class RememberMeController implements ReactiveController {
     protected logger = ConsoleLogger.prefix("controller/remember-me");
     protected autoSubmitAttempts = 0;
 
+    /**
+     * Whether the feature is currently on, seeded from {@linkcode defaultChecked} so that a
+     * returning user records keystrokes without having to re-toggle the switch.
+     */
+    protected rememberMeEnabled: boolean;
+
     constructor(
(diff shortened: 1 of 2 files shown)


so-jelly:
Hello, there appears to be an approval to run the remaining CI workflows.

PR #25208, opened 2026-08-19
sources: don't 500 on an object-shaped `groups` claim

### LINKED ISSUE #25191: OpenID Connect source: HTTP 500 when the IdP's `groups` claim contains objects instead of strings

abhisheksharma2411:
Closes #25191

## What

An `openidconnect` source whose IdP returns `groups` as a list of objects made the callback raise `TypeError: cannot use 'dict' as a dict key` out of `SourceFlowManager.__init__`, which surfaced as **HTTP 500**. Every user belonging to at least one group at such an IdP was locked out of the source entirely; a user in no groups signed in fine.

Two changes, layered, either of which stands on its own.

### 1. The flow manager no longer aborts on an identifier it cannot key on

`groups_properties` is a dict keyed by group identifier. A source is expected to hand over identifiers, but nothing enforces that — and it isn't only IdPs: a user property mapping can merge anything at all into `groups`, since `MERGE_LIST_UNIQUE` doesn't police the shape. Either way the exception escaped `__init__`, before any mapping had run, so an operator had no way to intervene.

Unusable entries are now skipped and a `CONFIGURATION_ERROR` event records how many were dropped. Recording rather than silently discarding matters here: a user arriving with fewer groups than the IdP granted is a difference an admin needs to be able to see, and it's the one thing worse than a 500 in an SSO product.

### 2. OpenID Connect sources understand the common object shapes

Not crashing would still leave the reporter without groups, so `OpenIDConnectType` now reduces an object entry to the identifier it carries — `id`, `value` or `sub` — and recovers the group name from `name`, `display` or `displayName`. That covers SCIM's `value`/`display` ([RFC 7643 §2.4](https://datatracker.ietf.org/doc/html/rfc7643#section-2.4)) and the `id`/`name` that Oracle Identity Domains' userinfo endpoint returns.

Entries with no recognised identifier are passed through untouched, so the flow manager reports them — an unsupported shape yields one diagnostic instead of two.

The name is looked up from the original claim rather than stashed on `info` (the approach `EntraIDType` takes with `raw_groups`). That keeps the change inert for a claim of plain strings: `info` reaches property mappings exactly as it does today, with no new key.

## Compatibility

A claim of plain strings behaves identically — `test_group_base_properties` and `test_grup_property_mappings` are untouched and still pass. The only inputs whose behaviour changes are the ones that returned HTTP 500.

## Testing

Five new tests. I confirmed each fails against unpatched code before the fix — the two headline ones with the exact `TypeError` from the issue's traceback:

```
FAILED ...test_property_mappings.py::test_object_group_property_mappings
  TypeError: cannot use 'dict' as a dict key (unhashable type: 'dict')
FAILED ...test_source_flow_manager.py::test_unusable_group_identifier_does_not_abort
  TypeError: cannot use 'dict' as a dict key (unhashable type: 'dict')
FAILED ...test_property_mappings.py::test_object_group_base_properties
  AssertionError: Lists differ: [{'id': '0db0', 'name': 'pixee-admins', ...}] != ['0db0', '28b5']
FAILED ...test_property_mappings.py::test_object_group_base_properties_without_name
  AssertionError: Lists differ: [{'id': '0db0'}] != ['0db0']
```

`test_unrecognised_object_group_is_passed_through` passes both before and after by design — it pins the pass-through that keeps the two layers from duplicating a diagnostic.

`authentik/sources/oauth`, `authentik/core/tests` and `authentik/core/sources` are green (352 passed), as are `ruff check authentik/` and `black --check`.

## Docs

Added a paragraph to the group synchronization section of the source property mappings page: identifiers have to be simple values, unusable ones are skipped with a Configuration error event, and the shapes OIDC sources handle without a mapping.

---

A note on overlap: #25195 adds a test to this same file that asserts `assertRaises(TypeError)` for this input. I left a comment there suggesting it be repointed at the fixed behaviour — if that lands first I'm happy to drop my equivalent test and rebase on it, and it should keep the credit for finding the gap in coverage.


authentik/sources/oauth/types/oidc.py
```diff
@@ -81,10 +81,47 @@ class OpenIDConnectType(SourceType):
...
     urls_customizable = True
 
+    # Members an object-shaped `groups` entry may use for the group's stable
+    # identifier and for its human-readable name, in preference order. SCIM
+    # names them `value` and `display` (RFC 7643 section 2.4); Oracle Identity
+    # Domains' userinfo endpoint uses `id` and `name`.
+    group_id_claims = ("id", "value", "sub")
+    group_name_claims = ("name", "display", "displayName")
+
+    def get_group_id(self, group: dict[str, Any]) -> str | None:
```
rissson: Since this is technically a breaking change, I'd refrain from making this change in this PR so that we can backport the fix in core/sources/flow_manager.py.

Can you move this to another PR once this one is merged?

rissson (changes requested):

abhisheksharma2411:
Done — pushed in 85f9434. This PR is now only `core/sources/flow_manager.py` plus its test and the matching docs line, so it should backport cleanly.

Agreed on the reasoning, and it's a better split than what I had: the OIDC change alters what a working source produces (an object-shaped claim starts yielding groups instead of a 500), while the flow manager change only turns a crash into a skip. Those belong on different release tracks.

I've kept the OIDC half on a branch and will open it once this merges, as you asked. For reference it's `OpenIDConnectType` reducing an object entry to its `id`/`value`/`sub` and recovering the name from `name`/`display`/`displayName`, plus four tests and the docs sentence I removed here.

One thing I'd flag for that follow-up, since it affects how you'd want to scope it: the two halves are genuinely independent — I re-ran the suite after stripping the OIDC changes and the flow manager test passes unchanged (15 passed). So the backport doesn't depend on the follow-up landing at all.

I also cut the docstring on `_keyable_group_ids` down to the essentials, taking the hint from the comment you removed on #25204.

## Disclosure

Per `AI_POLICY.md`: this contribution was AI-assisted. I used Claude Code for the investigation, the patch, and the test, and I reviewed and directed the work throughout — including rejecting a first version of the regression test that passed against unfixed code, and re-running the suite with the fix reverted to confirm the test actually goes red.

I should have disclosed that up front on this PR and on #25204 rather than after you asked — that's on me, and I'm sorry for the extra round trip. I'm happy to walk through any part of the change or answer questions on it directly.

rissson:
@abhisheksharma2411 since #25195 is merged, could you rebase this?

abhisheksharma2411:
Rebase wasn't needed in the end — you'd already merged `main` in yourself in 7c0276b, so I've added a commit on top of that rather than force-pushing a rebase over your merge.

Worth flagging *why* it was red, because it isn't the branch being stale:

**#25195 landed asserting the `TypeError` this PR removes.** Its test does

```python
with self.assertRaises(TypeError):
    OAuthSourceFlowManager(self.source, request, IDENTIFIER, {"info": info}, {})
```

so once both are on the same branch it fails with `AssertionError: TypeError not raised`. I reproduced that locally on your merge commit before changing anything. 48ffe3f repoints it at the behaviour the fix produces:

```python
flow_manager = OAuthSourceFlowManager(self.source, request, IDENTIFIER, {"info": info}, {})
self.assertEqual(flow_manager.groups_properties, {})
```

Empty rather than populated is deliberate and worth being precise about: with only the flow manager guard, `{"id": "group-1"}` is still not usable as a dict key, so it is skipped. It becomes a *mapped* group in the OIDC follow-up, and I'll tighten this assertion to `assertIn("group-1", ...)` there — so the same test tracks the behaviour across both PRs instead of being rewritten twice.

19 tests pass across `sources/oauth` and `core/tests`; `black` and `ruff` clean.

I did review #25195 before it merged and suggested exactly this change to the author, so I should have anticipated the collision and said something here when it landed rather than leaving you to hit a red board. Sorry about that.

rissson (approved):

PR #25238, opened 2026-08-19
stages/identification: only join Source subtypes that can render a login button

ViktorBarzin:
### What does this PR change?

On the identification stage, `select_subclasses()` is called with no arguments:

```python
current_stage.sources.filter(enabled=True).order_by("name").select_subclasses()
```

With no arguments, django-model-utils' `InheritanceManager` LEFT JOINs **every**
`Source` subtype table into a single wide row. The loop underneath only ever uses
the result to call `source.ui_login_button(request)`, and `Source.ui_login_button`
returns `None` — so subtypes that don't override it can't contribute a button and
don't need to be joined.

This narrows the join to the subclasses that do override it. The list is derived,
not hardcoded, so a new login-capable source type is picked up automatically:

```python
def login_capable_source_subclasses() -> list[type[Source]]:
    subclasses = []
    for relation in Source._meta.related_objects:
        if not getattr(relation.field, "parent_link", False):
            continue
        model = relation.related_model
        if model.ui_login_button is not Source.ui_login_button:
            subclasses.append(model)
    return subclasses
```

On current `main` that resolves to `OAuthSource`, `SAMLSource`, `PlexSource`,
`TelegramSource` and `KerberosSource` — the LDAP and SCIM sources are sync-only and
are the ones dropped from the join.

I walked `Source._meta.related_objects` rather than `all_subclasses(Source)` (which
is the more idiomatic helper here, and is what `core/tests/test_models.py` uses)
specifically because the parent-link walk can only ever yield concrete subclasses
that have a table to join against. Happy to switch it to `all_subclasses` with an
`_meta.abstract` guard if you'd prefer the consistency.

This query dominates first-load latency of the login flow. On our instance
(self-hosted, 3 OAuth sources configured) `GET
/api/v3/flows/executor/default-authentication-flow/` sat at ~1.4–1.8s TTFB, and
this one query accounted for essentially all of it:

| query | time |
|---|---|
| `select_subclasses()` (current behaviour) | 1527 ms |
| `select_subclasses(<login-capable only>)` | ~14 ms |
| `.all()` with no polymorphic join, for reference | 5 ms |

The cost is DB-side row width, not Python and not round-trips: cProfile put the time
in `psycopg` `wait()`, and the timing was identical when pointed straight at the
Postgres primary instead of through PgBouncer. The absolute numbers will vary with
how many subtype tables an install has and how wide they are, but the shape of the
fix doesn't.

### How was this tested?

Being upfront about what I did and didn't run:

- **Verified in production.** We've run this narrowing as a small patch over the
  official server image since 2026-06-28, and it is currently live across an
  upgrade chain of 2026.2.4 → 2026.2.6 → 2026.5.6 → 2026.8.0. Login (password, MFA,
  and the OAuth source buttons) works normally on all four.
- **Output equivalence.** The set of concrete source types returned and the rendered
  login buttons are identical before and after — that was checked when we first
  wrote the patch, and the flow's rendered source buttons have been unchanged in
  production since.
- **Subclass coverage.** Searched the repo for `def ui_login_button`: 6 hits — the
  base in `core/models.py` plus exactly those 5 overrides. Nothing in
  `authentik/enterprise/` overrides it.
- **Lint/format.** `ruff check` and `ruff format --check` on the changed file report
  the same result before and after this patch (clean, modulo one import-sorting
  finding that also fires on unmodified `main` when the file is linted outside the
  package).
- **I have not run `make all`.** I don't have an authentik dev environment set up, so
  the full suite and the web build are unverified by me. The existing
  `test_source_<Model>` tests generated in `core/tests/test_models.py` already cover
  `ui_login_button` per subclass. If you'd like a regression test that pins the
  narrowed source set or asserts the challenge output is unchanged, tell me the shape
  you want and I'll add it.

### Linked issues

None that I could find — I searched the tracker for `select_subclasses` and found
nothing covering this path.

---

## Checklist

-   [ ] The project has been linted, built, and tested (`make all`) — see above; ruff
        clean on the changed file, but I could not run the full suite
-   [x] The documentation has been updated and formatted (`make docs`) — no
        user-facing behaviour change, so no docs change


authentik/stages/identification/stage.py
```diff
@@ -385,9 +403,12 @@ def get_challenge(self) -> Challenge:
...
         ui_sources = []
-        sources: list[Source] = (
-            current_stage.sources.filter(enabled=True).order_by("name").select_subclasses()
-        )
+        sources: list[Source] = current_stage.sources.filter(enabled=True).order_by("name")
+        # Narrow the polymorphic join: a bare select_subclasses() LEFT JOINs every
+        # Source subtype table, but only subclasses overriding ui_login_button can
+        # contribute a button below.
+        if login_capable := login_capable_source_subclasses():
+            sources = sources.select_subclasses(*login_capable)
```
rissson:
```suggestion
        sources: list[Source] = current_stage.sources.filter(enabled=True).order_by("name").select_subclasses(*login_capable_source_subclasses())
```

the explanation already exists in the function above

authentik/stages/identification/stage.py
```diff
@@ -69,6 +69,24 @@ def get_login_serializers():
     return mapping
 
 
+def login_capable_source_subclasses() -> list[type[Source]]:
```
rissson, line 74: maybe worth an lru_cache, or even make it a static list evaluated at import

rissson (changes requested):
if you haven't already read https://github.com/goauthentik/authentik/blob/main/AI_POLICY.md

rissson (approved):

ViktorBarzin:
I should have disclosed this up front. The patch and the PR description were both written with Claude Code.

What is not AI generated is the reason it exists. We profiled this query on our own instance in June, and the narrowed version has run in production here ever since, currently on 2026.8.0. The 1527ms and 14ms figures are measurements from that instance, not estimates.

Both review points are pushed:

Folded the narrowing into the queryset chain and dropped the duplicated comment. I wrapped it across lines instead of taking your suggestion verbatim because the single-line form is 145 characters and ruff allows 100.

Added `@cache` rather than a list built at import. `stage.py` gets imported while apps are still loading, and if the lookup ran too early and came back empty, `select_subclasses()` would join every subtype again, which is the bug this fixes. Say the word if that ordering worry is unfounded and I will make it a module-level constant.

authentik/stages/identification/stage.py
```diff
@@ -69,6 +70,25 @@ def get_login_serializers():
...
+    """Concrete Source subclasses that can render a UI login button.
+
+    ``Source.ui_login_button`` returns None, so a source can only appear on the
+    identification stage if its subclass overrides it. Walking the parent links
+    rather than ``__subclasses__`` keeps this to subclasses that actually have a
+    table to join against.
+    """
+    subclasses = []
+    for relation in Source._meta.related_objects:
```
BeryJu:
```suggestion
    for source_type in all_subclasses(Source):
```

ViktorBarzin:
Switched to `all_subclasses(Source)`. I kept a `_meta.abstract` guard because on 2026.8.0 it returns 24 subclasses, 17 of them abstract, and 16 of those abstract ones do override `ui_login_button`. Those are the per-provider OAuth proxies like `GoogleOAuthSource`. Without the guard, `select_subclasses()` receives models that have no table.

Verified by running the exact code in this push against a live 2026.8.0 instance:

- derived list is `KerberosSource`, `OAuthSource`, `PlexSource`, `SAMLSource`, `TelegramSource`
- same pks and same concrete types as the unnarrowed query, on a stage with 3 enabled OAuth sources
- 1245ms unnarrowed against 14ms narrowed, with the helper itself at 0.08ms uncached

authentik/stages/identification/stage.py
```diff
@@ -69,6 +70,22 @@ def get_login_serializers():
...
+    ``Source.ui_login_button`` returns None, so a source only reaches the
+    challenge below if its subclass overrides it. Abstract subclasses are skipped
+    because they have no table to join against.
+    """
+    return [
+        source_type
+        for source_type in all_subclasses(Source)
+        if not source_type._meta.abstract
+        and source_type.ui_login_button is not Source.ui_login_button
+    ]
```
rissson, line 86:
```suggestion
LOGIN_CAPABLE_SOURCE_SUBCLASSES: list[type[Source]] = [
    source_type
    for source_type in all_subclasses(Source)
    if not source_type._meta.abstract
    and source_type.ui_login_button is not Source.ui_login_button
]
```

following up on my previous comment, would that work as well? cc. @BeryJu

BeryJu: I suppose yeah

rissson (approved):

ViktorBarzin:
I checked the import-time version before it lands and it comes out incomplete. At the point `stage.py` is imported, the constant holds three of the five types.

```
CONSTANT : ['KerberosSource', 'OAuthSource', 'SAMLSource']
TRUTH    : ['KerberosSource', 'OAuthSource', 'PlexSource', 'SAMLSource', 'TelegramSource']
MISSING  : ['PlexSource', 'TelegramSource']
```

`stage.py` is imported during `django.setup()`, and of the source model modules only `authentik.sources.oauth.models` and `authentik.sources.saml.models` are imported before it. Plex and Telegram are not registered yet when the comprehension runs. Kerberos happens to already be there.

I measured it by copying the package, putting 036f895's `stage.py` in place, adding one line at module scope that writes the constant to stderr, then running `django.setup()` with that copy on `PYTHONPATH`.

The failure is quiet rather than loud, which is what makes it awkward. A type missing from the narrowed list comes back as a base `Source`, and `Source.ui_login_button()` returns None, so the button simply is not rendered. I reproduced that part on my own instance by omitting a type I do have:

```
complete list  -> ['OAuthSource', 'OAuthSource', 'OAuthSource']   buttons: ['Facebook', 'Github', 'Google']
type omitted    -> ['Source', 'Source', 'Source']                 buttons: []
```

So on current main this would drop the Plex and Telegram login buttons with nothing in the logs. I noticed the backport label, so it seemed better to raise it now than after.

Anything evaluated once app loading has finished is fine. The cached function managed that by running on first request rather than by design. `AppConfig.ready()` would also work if you would rather keep a module-level name. I am happy to push whichever you prefer.

PR #25396, opened 2026-08-21
website/docs: show how to handle an object-shaped `groups` claim

abhisheksharma2411:
### What does this PR change?

Documents how to handle an object-shaped `groups` claim, using property mappings. No behaviour change.

Originally this PR added built-in normalization to `OpenIDConnectType`. @rissson's objection was right and I've replaced it — see the comment below for the evidence that changed my mind.

#25208 turned an unusable group identifier from an HTTP 500 into a skip, and documented that such entries are dropped. It stops there: an affected operator is told what they lost and not what to do about it. This adds the remedy, which is the extension point authentik already has.

Two snippets: reducing object entries to identifiers in a **user** property mapping, which runs before the identifiers are read; and recovering the provider's own group names in a **group** property mapping, which receives `group_id` alongside the original claim.

### How was this tested?

Both snippets were run against a source with an object-shaped claim before being written down, rather than composed on the page:

```
input:  [{"id": "g1", "name": "Admins"}, {"value": "g2", "display": "Devs"}, "plain"]
result: {"g1": {"name": "Admins"}, "g2": {"name": "Devs"}, "plain": {"name": "plain"}}
```

That covers the `id`/`name` shape, the SCIM `value`/`display` shape (RFC 7643 §2.4), and a plain string entry passing through untouched.

`cspell` and `prettier --check` both clean on the changed file.

### Linked issues

refs #25191, refs #25208

---

## AI usage disclosure

Per [AI_POLICY.md](https://github.com/goauthentik/authentik/blob/main/AI_POLICY.md): written with AI assistance (Claude), used for drafting and for exploring the codebase. I ran the snippets myself against a real source before documenting them, and can explain and defend any part of this without assistance.

---

## Checklist

-   [ ] The project has been linted, built, and tested (`make all`) — *docs-only; ran `cspell` and `prettier --check` on the changed file*


abhisheksharma2411:
Pushed `11d002ef` — one of the CI failures was mine, and I want to be precise about which, because the rest of the red is not what it first looks like.

## Mine: `lint (spellcheck, node)`

`recognised` is a forbidden word in `cspell.config.jsonc`. I reproduced it with the repo's own command rather than reading it out of the CI log:

```
authentik/sources/oauth/tests/test_property_mappings.py:177:27
  - Forbidden word (recognised) fix: (recognized)
```

Fixed, along with two `unrecognised` spellings cspell did *not* flag — it matches the standalone word, so the prefixed form slipped past the check while still being inconsistent with the surrounding code. `test_unrecognised_object_group_is_passed_through` is renamed accordingly. cspell now reports `Issues found: 0`; 22 tests still pass and `ruff check` / `ruff format --check` are clean.

## Not mine, but I can't call it a flake either

`test-e2e (proxy)` failed, and `ci-core-mark` fails as its aggregate. The four failures are all the same thing:

```
TestProviderProxyForward.test_traefik    ValueError: Webdriver failed after 3
TestProviderProxyForward.test_caddy      ValueError: Webdriver failed after 3
TestProviderProxyForward.test_envoy      ValueError: Webdriver failed after 3
TestProviderProxy.test_proxy_basic_auth  ValueError: Webdriver failed after 3
```

Browser startup exhausting its retries, in proxy-outpost tests. This PR is three files: `oidc.py`, its test module, and one docs line — nothing on the proxy path.

**But I want to flag the limit of that claim rather than assert a flake.** I checked `test-e2e (proxy)` across the last 14 runs of `ci-main.yml`: 11 success, 1 failure (mine), 2 where the job didn't run at all. The catch is that **all 11 successes are in-repo runs, and both of the other fork runs are the ones where the job was absent** — so I have no fork-PR baseline to compare against, and a fork-vs-in-repo comparison is exactly the confound that would make a real environmental difference look like a flake. I got caught by that same confound comparing against dependabot runs on an earlier PR, so I'd rather say "unexplained, and not on my diff's path" than "known flake".

If a maintainer re-runs just that job on the same commit, that would settle it either way — two outcomes from identical inputs is the only evidence I'd trust here. Happy to be told it's a known one.

## And a correction to a reading I nearly reported

`codecov/project` and `codecov/patch` looked red — the API showed 15 non-success check runs including `codecov/patch: 21.42% of diff hit`. That was me reading stale rows. Codecov posts a new check run per upload, so the sequence is progressive as shards land:

```
22:08  project  40.07%
22:10  patch    19.04%   project  45.08%
22:14  patch    21.42%   project  64.59%
22:16  patch   100.00%   project  91.08%
22:34  patch   100.00%   project  91.54%  (-0.04%)   <- current
```

Current state is **`codecov/patch` success at 100% of diff hit** and **`codecov/project` success at 91.54%**. Every new line is covered. The intermediate reds were partial uploads, not a coverage regression — noting it because it would be an easy thing for either of us to misread on a glance at the checks list.

`test-migrations-from-stable - PostgreSQL 18-alpine` shows `cancelled`, which I read as collateral from the e2e failure rather than an independent result.

rissson:
So this one I'm not so sure about. Since there are no standard for what the `groups` claim should look like, I don't think we should assume anything in there, and instead leave it up to the user to write property mappings to handle whatever their source gives them.

abhisheksharma2411:
You're right, and I've replaced the PR rather than argue for it.

I went to check whether the property-mapping route actually works today, because "leave it to the user" only holds if the user can in fact do it. It does — and more completely than I expected. `user_properties` is built by the mapper *before* `_keyable_group_ids` reads it, so a user property mapping gets first look at the claim:

```
without any mapping:              {}
with a user property mapping:     {'g1': {...}, 'g2': {...}}
```

And the names survive too. Running both halves against a source with a mixed claim:

```
input:  [{"id": "g1", "name": "Admins"}, {"value": "g2", "display": "Devs"}, "plain"]
result: {"g1": {"name": "Admins"}, "g2": {"name": "Devs"}, "plain": {"name": "plain"}}
```

So the extension point already covers every shape I was proposing to hard-code, including ones I hadn't thought of — which is the argument against hard-coding, not for it. My premise was that "no standard" means an IdP returning objects isn't misconfigured, so authentik should cope. Yours is that "no standard" is exactly why authentik shouldn't guess. Yours is the better reading: mine would have baked five claim names into the source type and made every other shape a second-class citizen that needs a mapping anyway.

**What I've replaced it with is docs-only.** #25208 tells an affected operator what they lost — entries dropped, Configuration error event — and doesn't tell them what to do instead. That gap is worth closing, and closing it with your answer rather than mine. The PR is now the two snippets above, both run before being written down rather than composed on the page.

Retitled and the description rewritten. If you'd rather not carry the recipe in the docs at all, say so and I'll close this — it's a small thing and I'd rather it match how you want the surface to read than get merged on my say-so.

One thing I should flag since it's adjacent: the earlier draft of #25208's docstring claimed operators had "no way to intervene, since mappings only run after this point". That's wrong for *user* mappings, as the above shows. It didn't ship — you'd trimmed the docstring during review and I took your version on the rebase — but I'd been carrying the wrong mental model since, and it's what made this PR look necessary to me. Worth saying out loud in case it's written down anywhere else.

website/docs/users-sources/sources/property-mappings/index.md
```diff
@@ -77,3 +77,77 @@ return {
 The `groups` attribute is a special attribute that must contain group identifiers. By default, those identifiers are also used as the group name. Each identifier is then given to group property mappings as the `group_id` variable, if extra processing needs to happen.
 
 An identifier has to be a simple value such as a string. Entries that are not, such as the objects some identity providers return in an OpenID Connect `groups` claim, are skipped, and a **Configuration error** event records how many were dropped.
+
+#### Object-shaped OpenID Connect group claims
+
+OpenID Connect does not define a standard `groups` claim or its value shape. Some identity providers return group objects instead of simple identifiers.
```
dominic-r:
```suggestion
The OpenID Connect standard does not define a standard `groups` claim or its value shape. Some identity providers return group objects instead of simple identifiers.
```

dominic-r (approved):

PR #25397, opened 2026-08-21
packages/ak-axum: extract the rightmost untrusted IP from X-Forwarded-For

### LINKED ISSUE #25393: Client IP extraction regression on 2026.8.0 when behind multiple proxies
### Describe the bug

When the Authentik instance is behind multiple reverse proxies `X-Forwarded-For` HTTP header carries a list of IPs of all hops but the last, i.e. `1.2.3.4, 172.17.0.1`.

Authentik 2026.5.6 extracted `1.2.3.4`, which was the expected outcome - the real client ip (equal to `X-Real-Ip`).
Authentik 2026.8.0 extracts `172.17.0.1`, which is the IP of the first reverse proxy on the way.

The issue is that 2026.5.6 used the leftmost address from the header and 2026.8.0 uses the rightmost ([`packages/ak-axum/src/extract/client_ip.rs:`](https://github.com/goauthentik/authentik/blob/0aeab0f1b7cfa20ac82af169712f73c6a389c2b7/packages/ak-axum/src/extract/client_ip.rs#L46) `client_ip::rightmost_x_forwarded_for`), even though it can be trusted by `trusted_proxy_cidrs`.

So in this setup all clients appear to connect from the same IP.

I guess the correct behaviour should be to extract the rightmost **untrusted** IP.

### How to reproduce

```bash
curl -s -o /dev/null -H 'X-Forwarded-For: 1.2.3.4, 172.17.0.1' http://localhost:9000/
docker compose logs --tail 1 server
```

### Expected behavior

**Expected:** `"remote": "1.2.3.4"`
**Actual:**   `"remote": "172.17.0.1"`
`172.17.0.1` is covered by `172.16.0.0/12`, which is in defaults for `trusted_proxy_cidrs`.

### Screenshots

_No response_

### Additional context

_No response_

### Deployment Method

Docker

### Version

2026.8.0

### Relevant log output

```shell

```

yzxcj797:
Fixes #25393.

## Summary

`extract_client_ip` resolved the client IP via `client_ip::rightmost_x_forwarded_for` — the plain **rightmost** `X-Forwarded-For` entry. Behind multiple reverse proxies that is the adjacent proxy, not the client: with `X-Forwarded-For: 1.2.3.4, 172.17.0.1` and the default `trusted_proxy_cidrs` (which cover `172.16.0.0/12`), **every** client of the instance appeared as `172.17.0.1` — the reporter's exact reproduction.

The extractor now walks the chain from the right and skips addresses matched by `trusted_proxy_cidrs` (`ip_addr_trusted`, the same predicate `TrustedProxy` uses), returning the **first untrusted address** — the originating client:

- `1.2.3.4, 172.17.0.1` → `1.2.3.4` (the report's case);
- a spoofed leftmost entry still loses to the untrusted address the proxy appended (rightmost-untrusted, not leftmost, keeps spoofing ineffective);
- when every entry is trusted, the leftmost (oldest) entry is the best remaining guess at the client.

The `client-ip` dependency becomes unused and is dropped from `ak-axum` and the workspace.

## Testing

Added to `packages/ak-axum/src/extract/client_ip.rs`:

- `rightmost_untrusted_x_forwarded_for` — the report's exact header resolves to `1.2.3.4`;
- `all_forwarded_for_entries_trusted_falls_back_to_leftmost`.

Existing tests pin the untrusted-chain behavior (rightmost of `192.0.2.51, 192.0.2.42`, multiple XFF, IPv6) and pass unchanged.

**Verification note:** I could not run `cargo test` on this Windows machine — building the workspace requires the aws-lc toolchain (nasm + cmake + libclang); nasm and cmake were installed but libclang is not obtainable in this environment. The patch was verified with `rustfmt --edition 2024 --check` (clean — the files parse and are formatted) and a full desk-check against the existing code; CI runs the complete suite.


packages/ak-axum/src/extract/client_ip.rs
```diff
@@ -33,6 +36,38 @@ where
...
+/// Get the rightmost IP from the `X-Forwarded-For` chain that is not itself a
+/// trusted proxy.
+///
+/// Each proxy appends the address it received the request from, so entries
+/// added by our own trusted proxies are skipped from the right and the first
+/// untrusted address is the originating client. Using the rightmost untrusted
+/// entry (instead of the plain leftmost) keeps client-supplied spoofed values
+/// from being trusted: a client can only append to the left of its own
+/// address. When every entry is trusted, the leftmost (oldest) entry is the
+/// best remaining guess at the client.
```
rissson:
```suggestion
```

rissson (approved):

PR #25412, opened 2026-08-23
root: tracing refactor

BeryJu:
~~this is not our final otel implementation, in fact this probably won't ever get merged. This is to explore how otel works, and see what kind of performance overhead it has when directly instrumenting python. This is a relatively minimal implementation replacing sentry with otel + claude fixing the forking stuff and adding tracing for django middlewares~~

~~This will be merged, however purposefully stay undocumented as we rework the implementation and eventually migrate this to rust~~

This is a refactor of the current tracing in preparation for OTEL


scripts/compose.tempo.yml
```diff
@@ -0,0 +1,64 @@
+## Local OTLP trace testing.
```
rissson: maybe `compose.tracing.yml` as the file name?

rissson (approved):

PR #25414, opened 2026-08-23
lib/expression: use literal fallback in ak_obj_attr

### LINKED ISSUE #25381: ak_obj_attr fallback ambiguous working
### Describe the bug

The [docs](https://docs.goauthentik.io/sys-mgmt/object-attributes/#use-object-attributes-in-property-mappings) for the new object attributes feature state that it's possible to set a value as a fallback option when using the function ak_obj_attr with property mappings. In my view the surrounding text implies that it's possible to set a fallback value that will be returned as is when the preceding object attribute is not set or its value is empty.
>Always use ak_obj_attr() with a fallback (e.g., "en-US"). An object attribute definition does not guarantee that every object has a value for that key.

However the fallback value is also treated as an object attribute. This can create an exception when both attributes are not defined. I checked the [code](https://github.com/goauthentik/authentik/blob/version-2026.8/authentik/lib/expression/evaluator.py#L177) and tried this with:
```python
return {
    "quota": ak_obj_attr(request.user, "nextcloud_quota", "1 GB"),
}
```

### How to reproduce

1. Create a new scope mapping which uses ak_obj_attr with a fallback option. 
2. Assign this scope mapping to an OAuth2 provider.
3. Preview the provider output with a user that doesn't have the object attribute set.

### Expected behavior

If treating the fallback option as an object attribute is the intended behaviour, then I suggest clarifying the docs to reflect this. Otherwise I suggest an update to the code that returns the fallback value as is. This makes this feature also more useful for my use case.

### Screenshots

_No response_

### Additional context

_No response_

### Deployment Method

Docker

### Version

2026.8.0

### Relevant log output

```shell
Context
{
    "asn": {
        "asn": x,
        "as_org": "x",
        "network": "x"
    },
    "geo": {
        "lat": x,
        "city": "",
        "long": x,
        "country": "NL",
        "continent": "EU"
    },
    "message": "Failed to execute property mapping",
    "exception": [
        {
            "frames": [
                {
                    "name": "<module>",
                    "lineno": 5,
                    "filename": "/authentik default OAuth Mapping: OpenID 'Nextcloud'"
                },
                {
                    "name": "handler",
                    "lineno": 3,
                    "filename": "/authentik default OAuth Mapping: OpenID 'Nextcloud'"
                },
                {
                    "name": "expr_obj_attr",
                    "lineno": 178,
                    "filename": "/authentik/lib/expression/evaluator.py"
                },
                {
                    "name": "inner",
                    "lineno": 253,
                    "filename": "/ak-root/.venv/lib/python3.14/site-packages/django/utils/functional.py"
                }
            ],
            "exc_type": "AttributeError",
            "is_cause": false,
            "is_group": false,
            "exc_notes": [],
            "exc_value": "'User' object has no attribute '1 GB'",
            "exceptions": [],
            "syntax_error": null
        }
    ],
    "expression": "return {\n    \"quota\": ak_obj_attr(request.user, \"nextcloud_quota\", \"1 GB\"),\n}",
    "http_request": {
        "args": {},
        "path": "/api/v3/providers/oauth2/26/preview_user/",
        "method": "GET",
        "request_id": "980636f0a4e942e6bc32a700024edcb2",
        "user_agent": "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/151.0.0.0 Safari/537.36"
    }
}
User
{
    "pk": 4,
    "email": "x",
    "username": "x"
}
```

ace2016:
### What does this PR change?
Changes `ak_obj_attr` so its fallback argument is returned as a literal value when the requested attribute is missing.

The fallback was incorrectly treated as an attribute name on the object. This made literal fallback values ambiguous and could return an unrelated object attribute instead of the supplied fallback.

### How was this tested?
- Added focused regression coverage for existing attributes, literal fallbacks, empty-string fallbacks, and `None`.
- Ran `make test authentik/lib/tests/test_evaluator.py`.
- `git diff --check` passed.

### Linked issues
closes #25381

---

## Checklist

### DIFF
--- a/authentik/lib/expression/evaluator.py
+++ b/authentik/lib/expression/evaluator.py
@@ -174,9 +174,7 @@ class BaseEvaluator:
         return fallback value."""
         attrs = getattr(obj, "attributes", {})
         value = get_path_from_dict(attrs, attr_key)
-        if value is None and fallback:
-            return getattr(obj, fallback)
-        return value
+        return fallback if value is None else value
 
     def expr_event_create(self, action: str, **kwargs):
         """Create event with supplied data and try to extract as much relevant data
--- a/authentik/lib/expression/evaluator.py
+++ b/authentik/lib/expression/evaluator.py
@@ -174,7 +174,7 @@ class BaseEvaluator:
         return fallback value."""
         attrs = getattr(obj, "attributes", {})
         value = get_path_from_dict(attrs, attr_key)
-        if value is None and fallback:
+        if value is None and fallback is not None:
             return getattr(obj, fallback, fallback)
         return value
 
--- a/authentik/lib/expression/evaluator.py
+++ b/authentik/lib/expression/evaluator.py
@@ -174,7 +174,9 @@ class BaseEvaluator:
         return fallback value."""
         attrs = getattr(obj, "attributes", {})
         value = get_path_from_dict(attrs, attr_key)
-        return fallback if value is None else value
+        if value is None and fallback:
+            return getattr(obj, fallback, fallback)
+        return value
 
     def expr_event_create(self, action: str, **kwargs):
         """Create event with supplied data and try to extract as much relevant data
--- a/authentik/lib/tests/test_evaluator.py
+++ b/authentik/lib/tests/test_evaluator.py
@@ -39,6 +39,16 @@ class TestEvaluator(TestCase):
         """Test expr_is_group_member"""
         self.assertFalse(BaseEvaluator.expr_is_group_member(create_test_admin_user(), name="test"))
 
+    def test_expr_obj_attr(self):
+        """Test expr_obj_attr"""
+        user = create_test_user()
+        user.attributes = {"locale": "en-US"}
+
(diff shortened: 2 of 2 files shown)


authentik/lib/expression/evaluator.py
```diff
@@ -174,9 +174,7 @@ def expr_obj_attr(obj: Model, attr_key: str, fallback: str | None = None) -> Any
         return fallback value."""
         attrs = getattr(obj, "attributes", {})
         value = get_path_from_dict(attrs, attr_key)
-        if value is None and fallback:
-            return getattr(obj, fallback)
-        return value
+        return fallback if value is None else value
```
BeryJu: I meant to fix this before .8 shipped but since .8 shipped with the old implementation we can't just change it;

I think we can do 

```python
if value is None and fallback:
    return getattr(obj, fallback, fallback)
return value
```

which will keep the current behaviour of allowing fallback to be an attribute name but will still fallback on using it as a value

BeryJu (changes requested):

BeryJu (approved):

PR #25415, opened 2026-08-23
crypto: add managed secrets API

### LINKED ISSUE #24961: Introduce a "Secret" object

dominic-r:


rissson:
we should have a longer discussion about this before actually implementing anything.

My thinking is as follows, following along the lines of #24961 :

- have a new `Secret` model, which is referenced from all the places where we store a secret (like oauth2 client_secret, notification transport credentials, etc.)
- we then have our usual permission system for that secret
- we also can configure rotation policies on that secret (how often, if it should be rotated at all)
- the admin interface should allow creating a secret "on the fly" from the modals where we need one

dominic-r:
Gonna push up changes in a few min which implement what you suggested, just to give an idea and try this out + marking as draft

dominic-r:
Sounds like a good idea, hadn't seen that issue before

dominic-r:
Marking as draft for now, will also discuss with G

authentik/secrets/models.py
```diff
@@ -0,0 +1,112 @@
...
+        if not isinstance(data, dict):
+            raise ValueError("Credential must be a JSON or YAML object")
+        return data
+
+    def replace_value(self, value: str, request: Request | None = None) -> None:
+        """Replace and audit the value, then signal consumers."""
+        if value == self.value:
+            return
+        from authentik.events.middleware import audit_ignore
+        from authentik.events.models import Event, EventAction
```
BeryJu, line 76: Top-level imports if possible

authentik/secrets/models.py
```diff
@@ -0,0 +1,112 @@
...
+                event.save()
+            secret_value_changed.send(sender=Secret, secret=self)
+
+    def rotate(self, request: Request | None = None) -> str:
+        """Generate and store a new text value."""
+        if self.type != SecretType.TEXT:
+            raise ValueError("Only text secrets can be rotated.")
+        value = generate_secret_value()
+        self.replace_value(value, request)
+        return value
```
BeryJu, line 95: as long as this stays manual rotation i'm fine with it, automated rotation shouldn't be done by authentik

BeryJu (commented):
- Smaller PR pls
- First PR that adds the system
- Second PR that adds the usage
- Have to remove the migrations that remove the old fields, those need to wait until 27.2
- Move secret model next to crypto ....maybe

authentik/crypto/secrets/apps.py
```diff
@@ -0,0 +1,12 @@
+"""Managed secrets app configuration."""
+
+from authentik.blueprints.apps import ManagedAppConfig
+
+
+class AuthentikSecretsConfig(ManagedAppConfig):
+    """Managed secrets app configuration."""
+
+    name = "authentik.crypto.secrets"
+    label = "authentik_secrets"
```
BeryJu, line 10:
```suggestion
    label = "authentik_crypto_secrets"
```

if this stays a sub-app in crypto it should have the matching label

BeryJu:
The idea with adding this to `crypto` is to rename `crypto` to `secrets` and then this will make more sense. TODO for me @BeryJu

authentik/crypto/secrets/models.py
```diff
@@ -0,0 +1,127 @@
...
+if TYPE_CHECKING:
+    from rest_framework.request import Request
+    from rest_framework.serializers import Serializer
+
+
+class SecretType(models.TextChoices):
+    """How a secret value is entered and displayed."""
+
+    TEXT = "text", _("Text")
+    MULTILINE = "multiline", _("Multi-line text")
```
rissson, line 34: Why are we making a distinction between those two?

authentik/crypto/secrets/models.py
```diff
@@ -0,0 +1,127 @@
...
+    FILE = "file", _("File")
+
+
+secret_value_changed = Signal()
```
rissson: should be in signals.py

authentik/crypto/secrets/models.py
```diff
@@ -0,0 +1,127 @@
...
+
+
+def generate_secret_value() -> str:
+    """Generate a value safe for HTTP Basic authentication and similar protocols."""
+    return generate_id(128)
```
rissson, line 40: Why not use the `default_token_length` system setting? As an admin I'd like to be able to configure the length system-wide here. If `default_token_length` doesn't fit, then let's create a new setting.
Also, 128 seems a bit extreme, let's reduce this to like 64

authentik/crypto/secrets/models.py
```diff
@@ -0,0 +1,127 @@
...
+
+class Secret(SerializerModel, ManagedModel, CreatedUpdatedModel):
+    """A named value that can be shared by configuration objects."""
+
+    secret_uuid = models.UUIDField(primary_key=True, editable=False, default=uuid4)
+    name = models.TextField(unique=True)
+    type = models.TextField(choices=SecretType.choices, default=SecretType.TEXT)
+    value = models.TextField(default=generate_secret_value)
+
+    def get_json(self) -> dict:
```
rissson, line 63: Should there be a `SecretType.JSON` then?

authentik/crypto/secrets/api.py
```diff
@@ -0,0 +1,191 @@
...
+        secret = self.get_object()
+        if secret.type != SecretType.TEXT:
+            raise ValidationError({"non_field_errors": [_("Only text secrets can be rotated.")]})
+        try:
+            value = secret.rotate(request)
+        except DjangoValidationError as exc:
+            raise ValidationError({"non_field_errors": exc.messages}) from exc
+        can_view = request.user.has_perm("authentik_crypto_secrets.view_secret_value") or (
+            request.user.has_perm("authentik_crypto_secrets.view_secret_value", secret)
+        )
```
rissson:
```suggestion
        can_view = request.user.has_perm("authentik_crypto_secrets.view_secret_value", secret)
```
doesn't this also check global permissions?

authentik/crypto/secrets/models.py
dominic-r: I could create a setting for sure. 128 was done because it matches the OAuth client secret generator.  https://github.com/goauthentik/authentik/blob/7b025acae2626f3a6ec70750ca8370f3e722a15f/authentik/providers/oauth2/models.py#L75-L77

melizeche:
First pass review:
Add a disclaimer/notice in the docs and the UI that secrets are not encrypted at rest. Maybe overkill, but it helps ensure people don't think this feature replaces HashiCorp Vault or similar products. In the docs you can mention that this doesn't change anything about how it was stored before this feature

dominic-r:
also: 
* tighten the rotate secret message to be more explicit that it doesn't update external sources
* improve release notes for user advantages (e.g. better rbac granularity)

PR #25438, opened 2026-08-24
Renew certificates in outposts

### LINKED ISSUE #25437: Outposts serve expired certificates
### Describe the bug

The LDAP outpost caches the certificates on the first start and then keeps serving that certificate forever from cache.

Especially with short lived ACME certificates this leads to incidents when the outposts are not restarted before the certificates expire although the server already has new certificate data.

### How to reproduce

1. Create a short lived certificate via the disk import functionality
2. Start an ldap outpost with TLS support enabled
3. Renew the certificate on disk. It will be imported in authentik-server automatically
4. Wait for the original certificate to expire. The renewed one should still be valid.
5. Connections to LDAP outpost will now fail with certificate errors because the certificate is expired.

### Expected behavior

Outposts should fetch the current certificate data from time to time.

### Screenshots

_No response_

### Additional context

_No response_

### Deployment Method

Kubernetes

### Version

2025.12

### Relevant log output

```shell

```

flx5:
## Details

The outposts have to expire the certificates in the cache and fetch updated information. Otherwise, they continue to serve expired certificates even when the server already has newer ones.

### What does this PR change?

Add a TTL for certificate data and fetch new certificates after configured time.

Not updating certificates leads to incidents because of expired certificates.
Manually restarting the outposts is not reasonable since ACME certificates are short-lived and should auto-update.

### How was this tested?
Manual testing.

### Linked issues

Closes https://github.com/goauthentik/authentik/issues/25437

---

## Checklist

### DIFF
--- a/internal/outpost/ak/crypto.go
+++ b/internal/outpost/ak/crypto.go
@@ -5,40 +5,59 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"encoding/pem"
+	"time"
 
 	log "github.com/sirupsen/logrus"
+	"goauthentik.io/internal/config"
+
 	api "goauthentik.io/packages/client-go"
 )
+
+type CertCacheEntry struct {
+	certificate *tls.Certificate
+	fingerprint string
+	refreshAt   time.Time
+}
 
 type CryptoStore struct {
 	api *api.CryptoAPIService
 
 	log *log.Entry
 
-	fingerprints map[string]string
-	certificates map[string]*tls.Certificate
+	certificates map[string]CertCacheEntry
+
+	cacheTTL time.Duration
 }
 
 func NewCryptoStore(cryptoApi *api.CryptoAPIService) *CryptoStore {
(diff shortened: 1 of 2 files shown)


internal/outpost/ak/crypto.go
```diff
@@ -90,19 +110,24 @@ func (cs *CryptoStore) Fetch(uuid string) error {
 			Leaf:        x509cert,
 		}
 	}
-	cs.certificates[uuid] = &tcert
-	cs.fingerprints[uuid] = cfp
+
+	cs.certificates[uuid] = CertCacheEntry{
+		certificate: &tcert,
+		fingerprint: cfp,
+		refreshAt:   time.Now().Add(time.Hour * 24),
```
tacerus: Is it intentional that this sets 24 hours instead of `cs.cacheTTL`?

flx5: Nope, missed that one.

PR #25439, opened 2026-08-24
outpost/proxy: avoid breaking header parsing in non-compliant backends

### LINKED ISSUE #25410: Proxy outpost sends lowercase HTTP/1.1 header names upstream, breaking backends that match `Content-Length` case-sensitively (regression in 2026.8.0)
### Describe the bug

Since 2026.8.0 the proxy outpost writes all header field names to the upstream in lowercase, including `content-length` and `host`. Backends that compare field names case-sensitively — common among embedded HTTP servers — then don't see a body length, treat the request as malformed and return `400 Bad Request`.
 
Requests without a body are unaffected, so `GET` traffic keeps working and only requests with a body fail.
 
### Cause
 
The Rust proxy outpost (`src/outpost/proxy/reverse_proxy.rs`, new in 2026.8.0; the Go `internal/outpost/proxyv2/` implementation shipped up to and including 2026.5.6) forwards via `hyper_util::client::legacy::Client`. `http::HeaderName` is normalised to lowercase on construction (by design, matching HTTP/2) and hyper writes names as stored, so the HTTP/1.1 wire format ends up all-lowercase. Go's `net/http` canonicalised field names, so every request the old outpost produced used the conventional title-case form.
 
hyper has an opt-in for exactly this case: `http1_title_case_headers` on the client builder — *"write header names as title case at the socket level [...] does not affect HTTP/2"*. For the outpost that is one line in `src/outpost/proxy/upstream.rs::build_client()`:
 
```rust
Ok(Client::builder(TokioExecutor::new())
    .set_host(false)
    .http1_title_case_headers(true)   // restores the pre-2026.8 wire format
    .build(connector))
```
 
HTTP/2 upstreams are unaffected by that flag, so they keep the lowercase form RFC 9113 §8.2.1 requires.

### How to reproduce

1. Run an upstream that looks up the header case-sensitively, as several embedded HTTP servers do:
```python
   import socket
 
   srv = socket.create_server(("0.0.0.0", 8099))
   while True:
       conn, _ = srv.accept()
       data = b""
       while b"\r\n\r\n" not in data:
           chunk = conn.recv(65536)
           if not chunk:
               break
           data += chunk
       head = data.split(b"\r\n\r\n", 1)[0]
       ok = any(l.startswith(b"Content-Length:") for l in head.split(b"\r\n")[1:])
       body = b"OK" if ok else b"Bad HTTP request"
       status = b"200 OK" if ok else b"400 Bad Request"
       conn.sendall(
           b"HTTP/1.1 " + status
           + b"\r\nContent-Length: " + str(len(body)).encode()
           + b"\r\nConnection: close\r\n\r\n" + body
       )
       conn.close()
```
 
2. Create a proxy provider with:
   - internal host `http://<host>:8099`
   - external host `https://repro.example.com`
   - `skip_path_regex` set to `.*`, so the test needs no authentication
   Create an application and assign the provider to an outpost.
3. Send a request with a body:
```
   curl -X POST --data-binary 'x' https://repro.example.com/   ->  400 Bad Request
```
 
   A `GET`, or a `POST` with an empty body, returns `200`.
 
4. Point the same upstream at a standalone outpost from `ghcr.io/goauthentik/proxy:2026.5.6` (Go implementation) → `200 OK`.

### Expected behavior

Conventional title-case field names when serialising HTTP/1.1 requests to the upstream, as the outpost did up to 2026.5.6.
 
If that is unwelcome as a default, an opt-in would be sufficient — comparable to HAProxy's `h1-case-adjust` / `option h1-case-adjust-bogus-server`, which exist for exactly this class of backend.

### Screenshots

_No response_

### Additional context

Field names are case-insensitive per RFC 9110 §5.1, so hyper's output is legal and these upstreams are, strictly speaking, the buggy party. In practice many embedded web servers do a literal `strcmp` and only accept `Content-Length`, which is why title-case is what essentially everything on the HTTP/1.1 leg emits: browsers, curl, nginx, Apache `mod_proxy`, Squid, Varnish, HAProxy (which ships `h1-case-adjust` specifically for these backends) — and authentik's own outpost up to 2026.5.6. Emitting title-case costs nothing for compliant upstreams, since it is equally valid there.

### Deployment Method

Docker

### Version

2026.8.0

### Relevant log output

```shell

```

rissson:
### Linked issues

closes #25410

---

## Checklist

### DIFF
--- a/src/outpost/proxy/upstream.rs
+++ b/src/outpost/proxy/upstream.rs
@@ -28,13 +28,13 @@ pub(super) fn build_client(insecure: bool) -> Result<UpstreamClient> {
         builder.with_native_roots()?
     }
     .https_or_http()
-    .enable_http1()
-    .enable_http2()
+    .enable_all_versions()
     .build();
     // Forward the request's own `Host` upstream instead of deriving it from the
     // (internal) upstream URI authority. The proxy sets `Host` explicitly.
     Ok(Client::builder(TokioExecutor::new())
         .set_host(false)
+        .http1_title_case_headers(true)
         .build(connector))
 }
 


BeryJu (approved):
should we add a test for this too?

rissson:
> should we add a test for this too?

I don't think we need to bother

PR #25450, opened 2026-08-24
web: Fix Sentry Initialization

GirlBossRush:
This PR makes `errorReporting.enabled` actually control whether Sentry reports errors, and moves Sentry initialization out of Lit element constructors so reporting starts earlier.  Production error reporting now follows the admin's `errorReporting.enabled` setting, while debug/Spotlight behavior remains available independently. Sentry also initializes early enough to capture errors that occur during application startup.

## Details

Today, the administrator setting does not affect whether the Sentry SDK runs.

Sentry is initialized with:

```ts
init({ enabled: process.env.NODE_ENV !== "production" })
```

This has the opposite effect from what we want: Sentry is disabled in production, including the builds shipped in our containers. The container's web stage sets `NODE_ENV=production` (`lifecycle/container/Dockerfile:12`), and the bundler inlines that value (`bundler/utils/node.js:37`).

As a result, enabling error reporting in authentik does not cause production errors to be reported.

`browserTracingIntegration` had the opposite, registered only in production, where the SDK itself was disabled. This had development therefore had Sentry enabled without the tracing integration.

Both conditions were introduced in #19504 and have been present since 2026.2.0.

### What changed

`common/sentry/index.ts` has been split by responsibility:

* **`utils.ts`** contains the Sentry enablement policy as the pure function `sentryEnabled({ errorReporting, debug, search, production })`.

  * `errorReporting.enabled` controls normal reporting.
  * `CanDebug` can independently enable Sentry, as required for Spotlight.
  * Development supports `?disable-sentry` to disable Sentry for a single page load without rebuilding.

* **`apply.ts`** performs Sentry initialization as an entrypoint side effect. Each application entrypoint imports it before loading its Lit elements.

  Previously, `configureSentry()` ran from the constructors of `AdminInterface`, `UserInterface`, and `FlowExecutor`. Errors thrown while their modules were loading or custom elements were being registered therefore occurred before Sentry was initialized.

  `browserTracingIntegration` is now registered whenever Sentry is enabled. Its automatic instrumentation remains disabled because authentik's router creates tracing spans itself.

* **`error.ts`** contains `SentryIgnoredError` without importing the Sentry runtime.

  Six form and API helpers only need this error class. Importing it through the old `index.ts` unnecessarily pulled in `globalAK` and, transitively, the generated API client.


kensternberg-authentik (approved):
Nothing syntactically wrong, and it builds.  Been so long since I looked at Sentry I'll have to take your word on the configuration being right.  :-)

PR #25461, opened 2026-08-25
website/docs: document ak_call_policy gotchas

severin:
## Details

This PR updates the documentation of the `ak_call_policy` function with some gotchas that I came across when looking deeper into how the function works.

### What does this PR change?

It adds a note about error handling:

- Exceptions from the called policy do not bubble up
- Specifying an invalid name raises an exception

~And a note about keywords potentially "tainting" the existing context~

### How was this tested?

By looking at the code and playing around with `ak_call_policy`. 

Disclaimer: I also let an AI agent review the change against the codebase

---

## Checklist


website/docs/expressions/reference/_functions.mdx
```diff
@@ -48,6 +48,25 @@ result = ak_call_policy("test-policy-2", foo="bar")
...
+Keyword arguments are written directly into the context of the current `PolicyRequest`, any key you pass can
+remain visible to:
+
+- The rest of the caller, after `ak_call_policy` returns.
+- Other policies bound to the same flow, stage binding, application, or source.
+- Any further `ak_call_policy` calls made during the same evaluation.
+
+Passing a key that already exists in the context (such as `pending_user` or `prompt_data`) overwrites
+it for the remainder of the evaluation.
+:::
```
severin: @rissson @BeryJu is this actually intended behaviour or rather a bug?

BeryJu: that is definitely a bug, I think it's missing either a copy or deepcopy call there

BeryJu: https://github.com/goauthentik/authentik/pull/25462

severin: @BeryJu thanks for the quick fix 🚀 

I updated the PR to only include the info about error handling...

BeryJu (approved):

website/docs/expressions/reference/_functions.mdx
dewi-tik: Thanks @severin

PR #25464, opened 2026-08-25
blueprints: handle absent blueprint references

### LINKED ISSUE #25329: Blueprint delete of stage bindings is unsafe when KeyOf is still referenced
## What

blueprint delete of stage bindings is unsafe (!KeyOf still referenced; deleting a missing binding fails the whole apply).

## Want

- Blueprint delete of stage bindings is unsafe (!KeyOf still referenced
- Deleting a missing binding fails the whole apply)

## Ideas this came from

- [#20309 Blueprints fail to delete stage when it references another stage using !KeyOf](https://github.com/goauthentik/authentik/issues/20309) — I am trying to add a delete option for a blueprint, so it can be cleanly uninstalled with "one click". However, even though the blueprint execution is "successf
- [#20456 Deleting non-existent flow stage bindings via blueprints fails](https://github.com/goauthentik/authentik/issues/20456) — Trying to delete a flow stage binding from a flow fails if the binding is already deleted. Consider the following blueprint Applying this blueprting when the fl

ace2016:
### What does this PR change?
Stores the resolved instance state for blueprint entries marked as `absent` before deleting the instance.

Adds regression coverage for:

- Deleting entries that reference another absent entry through `!KeyOf`.
- Reapplying deletion of an already-missing flow stage binding.

`!KeyOf` resolves references using the state stored on another blueprint entry. Entries marked as `absent` did not store their resolved instance before deletion, causing later entries referencing them to fail during blueprint application.

Recording the entry state before deletion keeps reference resolution consistent with entries in other states and allows absent blueprints to be applied idempotently.

### Linked issues
closes #25329

---

## Checklist

### DIFF
--- a/authentik/blueprints/tests/test_v1_state.py
+++ b/authentik/blueprints/tests/test_v1_state.py
@@ -3,9 +3,12 @@
 from django.test import TransactionTestCase
 
 from authentik.blueprints.v1.importer import Importer
-from authentik.flows.models import Flow
+from authentik.flows.models import Flow, FlowDesignation, FlowStageBinding
 from authentik.lib.generators import generate_id
 from authentik.lib.tests.utils import load_fixture
+from authentik.stages.identification.models import IdentificationStage
+from authentik.stages.password.models import PasswordStage
+from authentik.stages.user_login.models import UserLoginStage
 
 
 class TestBlueprintsV1State(TransactionTestCase):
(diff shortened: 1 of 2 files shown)


ace2016:
Hello @BeryJu could you please give this a review when you are free? thank you

BeryJu (approved):

Issue #22742, opened 2026-05-28
Huge number of Certificate Discovery tasks being scheduled after Update


pascal260303:
### Describe the bug

After doing the update to 2026.5.0 a huge amount of Certificate Discovery Tasks were being scheduled. I didn't notice immediately so this was the number of tasks being scheduled after some days:
```
>>> from authentik.tasks.models import Task
>>> count = Task.objects.filter(state="queued").count()
>>> print(f"Wartende Tasks: {count}")
Wartende Tasks: 249308
```
Each single task finishes successfully and a single task would only run a few seconds, but there where multiple tasks per second being newly scheduled so this accumulated.
I stopped the tasks with this
```
Task.objects.filter(state="queued").update(state="done")
```
and paused the Certificate Discovery task, because I saw in the webui that it was this task being scheduled. After that the number of tasks being queued went back to zero and stayed there.
I have 14 directories in the certs folder and each has 2 Files "fullchain.pem" and "privkey.pem". 
For now pausing the task works, but I need it, since all the certificates are acme certificates and might change pretty soon.
I did not change the scheduling of the task, it was at all times "11 * * * *".

I just updated to 2026.5.2 but nothing changed. I initially updated from 2026.2.1 to 2026.5.0 and now to 2026.5.2

### How to reproduce

1. have multiple Directories  with Certificates in the certs/ Folder
2. Let the Certificate Discovery Task run normally
3. See the tasks accumulating and your DB suffering.

### Expected behavior

Certificate Discovery Task running once per hour, like configured.

### Deployment Method

Docker

### Version

2026.5.2

### Relevant log output

```shell

```

pyro-tec:
Having the same issue, in addition pausing the task doesn't help as it restarts itself every few hours

pascal260303:
Yes I noticed the task restarting itself too. Tried to find the issue but failed so I wrote a bash script that stops all scheduled tasks and pauses it again every 30 mins if the number of scheduled tasks is over 100. That is only a workaround.

For me what might be related is that the certs dir was previously owned by a user with id 1000, but that user doesn't exist anymore and there is no new user with id 1000.
I tried setting the owner to the new supposed user, but it keeps getting overwritten. I don't know where I can change that.

Mase3206:
[...]

PR #25483, opened 2026-08-25
crypto: don't dispatch discovery if one is queued

BeryJu:
https://github.com/goauthentik/authentik/issues/22742

### DIFF
--- a/authentik/crypto/tasks.py
+++ b/authentik/crypto/tasks.py
@@ -91,7 +91,7 @@ class CertificateEventHandler(FileSystemEventHandler):
         for tenant in Tenant.objects.filter(ready=True):
             with tenant:
                 if Task.objects.filter(
-                    actor_name=certificate_discovery.actor_name, state=TaskState.QUEUED
+                    tenant=tenant, actor_name=certificate_discovery.actor_name, state=TaskState.QUEUED
                 ).exists():
                     continue
                 Schedule.dispatch_by_actor(certificate_discovery)
--- a/authentik/crypto/tasks.py
+++ b/authentik/crypto/tasks.py
@@ -91,7 +91,9 @@ class CertificateEventHandler(FileSystemEventHandler):
         for tenant in Tenant.objects.filter(ready=True):
             with tenant:
                 if Task.objects.filter(
-                    tenant=tenant, actor_name=certificate_discovery.actor_name, state=TaskState.QUEUED
+                    tenant=tenant,
+                    actor_name=certificate_discovery.actor_name,
+                    state=TaskState.QUEUED,
                 ).exists():
                     continue
                 Schedule.dispatch_by_actor(certificate_discovery)
--- a/authentik/crypto/tasks.py
+++ b/authentik/crypto/tasks.py
@@ -9,6 +9,7 @@ from cryptography.hazmat.primitives.serialization import load_pem_private_key
 from cryptography.x509.base import load_pem_x509_certificate
 from django.conf import settings
 from django.utils.translation import gettext_lazy as _
+from django_dramatiq_postgres.models import TaskState
 from dramatiq.actor import actor
 from dramatiq.middleware import Middleware
 from structlog.stdlib import get_logger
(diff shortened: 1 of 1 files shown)


abhisheksharma2411 (commented):
The dedup is the right idea and the `run_tasks` extraction removes a real copy-paste pair. Two things about the predicate, one of which I think is a genuine gap.

## `QUEUED` alone leaves a window open

`TaskState` runs `WAITING_FOR_DEPENDENCIES → QUEUED → CONSUMED → PREPROCESS → RUNNING → POSTPROCESS → DONE`. The guard matches only `QUEUED`, so a discovery task that a worker has already claimed but not yet started — sitting in `CONSUMED` or `PREPROCESS` — does not suppress a new dispatch. A burst of certificate writes that straddles that transition still enqueues the duplicate this PR is meant to prevent.

`WAITING_FOR_DEPENDENCIES` has the same shape: the task exists and has not scanned anything yet.

So the set that means "a discovery is already pending and has not looked at the filesystem yet" is closer to:

```python
state__in=(
    TaskState.WAITING_FOR_DEPENDENCIES,
    TaskState.QUEUED,
    TaskState.CONSUMED,
    TaskState.PREPROCESS,
)
```

**And to be explicit about the other half: excluding `RUNNING` is correct and I would not change it.** A certificate written *while* a scan is in progress may already have been passed over, so that event does need to enqueue a fresh run. If the predicate is widened, it should stop short of `RUNNING` deliberately rather than by omission — worth a comment saying so, because the next person to read this will wonder.

## Check-then-dispatch is not atomic

`.exists()` and `Schedule.dispatch_by_actor()` are separate statements, so two events that interleave between them both dispatch. This is not hypothetical for watchdog specifically: writing a new certificate commonly raises `IN_CREATE` **and** `IN_CLOSE_WRITE`/`IN_MODIFY`, so `on_created` and `on_modified` both fire for one file and both call `run_tasks()`.

That still leaves this PR strictly better than the status quo, and a duplicate discovery run is idempotent rather than harmful — so I would not hold the PR for it. But the guard is best-effort, and it reads like a guarantee. If you wanted it to be one, the usual shape is a conditional insert (or an advisory lock keyed on the actor name) rather than a read followed by a write.

## Small note on the query, probably a non-issue

Nothing in `TaskBase.Meta.indexes` leads with `actor_name` — the set is `(queue_name)`, `(queue_name, state)`, `(queue_name, state, eta)`, `(message_id, …)` variants, and `(state, mtime, result_expiry)`; authentik adds `(rel_obj_content_type, rel_obj_id)`. So this filter has no covering index and leans on `state` being selective.

In practice `QUEUED` rows drain quickly and the count stays small, so I expect this to be fine and I am **not** claiming a performance problem — I checked because the query now runs on every certificate file event, per ready tenant, and that is the kind of thing worth knowing before it is on a hot path rather than after.

---

Verified the dispatch side does what the guard assumes: `dispatch_by_actor` → `schedule.send()` → `send_with_options` enqueues through the Postgres broker synchronously, so a second event genuinely does observe the first task's row rather than racing an async write.

*AI usage disclosure, per [AI_POLICY.md](https://github.com/goauthentik/authentik/blob/main/AI_POLICY.md): this review was written with AI assistance (Claude). I read the state machine, the index definitions and the dispatch path myself and can defend each point above without it.*

authentik/crypto/tasks.py
```diff
@@ -85,25 +87,30 @@ def dispatch(self, event: FileSystemEvent) -> None:
             return None
         return super().dispatch(event)
 
+    def run_tasks(self):
+        for tenant in Tenant.objects.filter(ready=True):
+            with tenant:
+                if Task.objects.filter(
+                    actor_name=certificate_discovery.actor_name, state=TaskState.QUEUED
```
rissson:
```suggestion
                    tenant=tenant, actor_name=certificate_discovery.actor_name, state=TaskState.QUEUED
```

rissson (approved):

PR #25486, opened 2026-08-25
Add RepoCloud to Installation section

cosark:
Adds RepoCloud as an additional deployment option in the Installation section of the README, matching the existing text link format used by other providers.


dominic-r:
Thank you for your contribution, however this is not something we're considering for the time being

PR #25515, opened 2026-08-26
Fix/obj attr literal fallback

### LINKED ISSUE #25390: Enterprise Audit Diff: For attributes, show nested fields in attributes
Currently we show the full diff of the `attributes` field even if only one field within that has changed

Also fields in attributes are not sorted

https://github.com/goauthentik/internal-customer-ref/issues/2

ace2016:
### What does this PR change?
Displays changes to nested object attributes as individual audit diff entries instead of displaying the complete `attributes` object. Nested attribute paths use the existing `attributes.<path>` notation and are sorted to provide deterministic output.
Adds regression coverage for changed, added, removed, unchanged, nested, and sorted attribute fields.
Enterprise audit events currently compare model fields only at the top level. When one nested attribute changes, the audit diff displays the complete previous and new `attributes` objects, making the actual change difficult to identify.

Showing each changed attribute separately makes audit events clearer and avoids including unrelated unchanged attributes.
### How was this tested?
- Added focused regression coverage for nested attribute diffs.
- Ran Ruff against the changed Python files.
- Ran Python syntax compilation for the changed files.
- Ran `git diff --check`.
- Ran the Docker-backed project test workflow.
### Linked issues
closes #25390

---

## Checklist

### DIFF
--- a/authentik/enterprise/audit/middleware.py
+++ b/authentik/enterprise/audit/middleware.py
@@ -69,11 +69,37 @@ class EnterpriseAuditMiddleware(AuditMiddleware):
     def diff(self, before: dict, after: dict, update_fields: list[str] | None = None) -> dict:
         """Generate diff between dicts"""
         diff = {}
+
+        def diff_attributes(
+            previous: dict[str, Any], current: dict[str, Any], path: str = "attributes"
+        ) -> dict:
+            attribute_diff = {}
+            for key in sorted(previous.keys() | current.keys()):
+                previous_value = previous.get(key)
+                current_value = current.get(key)
+                if isinstance(previous_value, dict) and isinstance(current_value, dict):
+                    attribute_diff.update(
+                        diff_attributes(previous_value, current_value, f"{path}.{key}")
+                    )
+                elif previous_value != current_value:
+                    attribute_diff[f"{path}.{key}"] = {
+                        "previous_value": previous_value,
+                        "new_value": current_value,
+                    }
+            return attribute_diff
+
         for key, value in before.items():
             if update_fields and key not in update_fields:
                 continue
             if after.get(key) != value:
-                diff[key] = {"previous_value": value, "new_value": after.get(key)}
+                if (
+                    key == "attributes"
+                    and isinstance(value, dict)
+                    and isinstance(after.get(key), dict)
+                ):
+                    diff.update(diff_attributes(value, after[key]))
+                else:
+                    diff[key] = {"previous_value": value, "new_value": after.get(key)}
         for key, value in after.items():
             if update_fields and key not in update_fields:
                 continue
--- a/authentik/enterprise/audit/tests.py
+++ b/authentik/enterprise/audit/tests.py
@@ -288,3 +288,42 @@ class TestEnterpriseAudit(APITestCase):
             update_fields=["is_active"],
         )
         self.assertEqual(diff, {"is_active": {"new_value": True, "previous_value": False}})
+
+    def test_diff_attributes(self):
+        """Test nested attribute diff"""
+        diff = EnterpriseAuditMiddleware(None).diff(
+            {
+                "attributes": {
+                    "department": "Engineering",
+                    "name": "Alice",
+                    "settings": {"theme": "light", "timezone": "UTC"},
+                },
+            },
+            {
+                "attributes": {
+                    "department": "Security",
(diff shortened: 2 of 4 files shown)


ace2016:
Hi @BeryJu, could you please share the link or instructions for signing the required CLA?

abhisheksharma2411 (commented):
The audit-diff half reads well — flattening nested attribute changes into individual `attributes.<path>` entries is a real improvement over dumping the whole object, and sorting for deterministic output is the right instinct for something that ends up in an audit record.

I'd separate the `expr_obj_attr` change out, though, and I think it needs rethinking rather than just re-homing.

**It isn't mentioned in the PR description.** The Details section describes the audit diff only, but `authentik/lib/expression/evaluator.py` is a different subsystem — `expr_obj_attr` is exposed to policy expressions and property mappings, so a semantic change there reaches SAML/OIDC claims and group mappings. Two unrelated changes under one title makes the second one easy to merge unread.

**And the semantics it introduces are ambiguous in a way that can silently substitute user data for an intended constant.**

Today `fallback` means exactly one thing — *the name of another attribute to read*:

```
expr_obj_attr(user, 'missing', <fallback>)   -- current main
  fallback='name'       -> 'Ada Lovelace'
  fallback='email'      -> 'ada@example.test'
  fallback='pk'         -> 1099
  fallback='en-GB'      -> raises AttributeError: 'User' object has no attribute 'en-GB'
  fallback='Anonymous'  -> raises AttributeError: 'User' object has no attribute 'Anonymous'
```

Unambiguous, and a non-attribute fails loudly. With `getattr(obj, fallback, fallback)` it becomes *"an attribute name if one happens to exist, otherwise a literal"*:

```
expr_obj_attr(user, 'missing', <fallback>)   -- this PR
  fallback='en-GB'      -> 'en-GB'                    literal (as intended)
  fallback='Anonymous'  -> 'Anonymous'                literal (as intended)
  fallback='name'       -> 'Ada Lovelace'             <-- leaked a user field
  fallback='email'      -> 'ada@example.test'         <-- leaked a user field
  fallback='username'   -> 'vzixK3FdJffptDgfrIIj'     <-- leaked a user field
  fallback='pk'         -> 1099                       <-- leaked a user field
  fallback='is_active'  -> True                       <-- leaked a user field
  fallback='path'       -> 'users'                    <-- leaked a user field
```

(Run against a real `create_test_user()` on both versions, swapping only `evaluator.py` and confirming the swap took effect first.)

Your own tests pin both meanings side by side, which is what surfaced it for me:

```python
self.assertEqual(BaseEvaluator.expr_obj_attr(user, "missing", "username"), user.username)  # attribute
self.assertEqual(BaseEvaluator.expr_obj_attr(user, "missing", "en-GB"), "en-GB")           # literal
```

So which branch runs is decided by whether the operator's chosen default collides with a field on the model — invisibly, with no error either way. Someone writing `expr_obj_attr(user, "display_name", "name")` meaning the literal string `"name"` gets `Ada Lovelace` in the claim instead. On `User` the colliding namespace is large and mostly inherited from Django, so this isn't an exotic case: `name`, `email`, `path`, `pk`, `is_active`, `type`, `uuid` are all live.

It's also fragile across upgrades. Adding a field to `User` in a future release changes the meaning of expressions already deployed in the field, with no migration and no warning — the expression keeps working and starts returning something else.

**Suggestions, in order of how much I'd argue for them:**

1. **Split it out.** The audit-diff change stands on its own and is easy to review; this one deserves its own PR and its own discussion. That alone would unblock the good half.
2. **Make the two meanings explicit rather than inferred** — e.g. keep `fallback` as attribute-name-only and add a separate `default=` for literals, so neither can shadow the other. That also preserves the loud `AttributeError` for a typo'd attribute name, which is currently the only thing that catches one.
3. If a literal default really is the desired behaviour and the attribute-name reading is the accident, then say so and **remove** the attribute lookup rather than layering the literal on top — the danger is specifically the two coexisting.

The `if fallback:` → `if fallback is not None:` change is fine and worth keeping either way; `""` should be a usable default rather than falling through to `None`. Worth calling that out separately since it's an independent, uncontroversial fix.

Happy to be wrong about the intent here — if `expr_obj_attr`'s `fallback` was always documented as a literal somewhere I haven't found, that changes the analysis. The docstring (*"return fallback value"*) is genuinely ambiguous, which is probably how we got here.

PR #25517, opened 2026-08-26
root: deprecate non-UTF8 database encoding

### LINKED ISSUE #25495: UTF8 encoding being required
Change warning to error and add mention in docs

> Thank you for the quick response!
> 
> Technically, a server encoding of UTF8 is [required by authentik since `2025.8`](https://docs.goauthentik.io/releases/2025.8#database-encoding-requirements), however, this is nowhere else present in the docs. We should also probably change [this warning](https://github.com/goauthentik/authentik/blob/version/2026.8.0/authentik/root/db/base.py#L32-L38) to an error if we're serious about that, enough time has passed.
> 
> I'll leave this issue open until one or both of those happen. 

 _Originally posted by @gergosimonyi in [#25445](https://github.com/goauthentik/authentik/issues/25445#issuecomment-5422247808)_

dominic-r:
authentik has required a UTF8 database since 2025.8 but only warned about it, letting installs run until something downstream failed on bytes it could not decode. The system check now fails startup and points at the docs.

Closes: #25495


dewi-tik (approved):
Docs LGTM

gergosimonyi (commented):
Two thoughts

1. We should create deprecation events for a release or two before just failing.
2. The rest of `_check_encoding` can be removed after this, right? Django client encoding [defaults to `UTF8`](https://github.com/django/django/blob/5.2.17/django/db/backends/postgresql/base.py#L267) anyway.

rissson:
>     1. We should create deprecation events for a release or two before just failing.

technically we announced it already https://docs.goauthentik.io/releases/2025.8/#database-encoding-requirements

website/docs/releases/2026/v2026.11.mdx
```diff
@@ -40,6 +40,14 @@ helm repo update
 helm upgrade authentik authentik/authentik -f values.yaml --version ^xxxx.x
 ```
 
+## Deprecations
+
+### Databases that are not `UTF8` are deprecated
```
dominic-r:
```suggestion
### Non-`UTF8` database server encoding is deprecated
```

website/docs/install-config/configuration/configuration.mdx
```diff
@@ -67,6 +67,8 @@ To check if your config has been applied correctly, you can run the following co
 
 authentik requires PostgreSQL for application data, configuration, sessions, and background task coordination. authentik supports PostgreSQL 14 - 18.
 
+The database should use the `UTF8` encoding, which is PostgreSQL's default. Unless a different encoding was explicitly chosen when the database was created, no change is needed. Run `SHOW server_encoding;` to check; authentik logs a warning on any other encoding, and support for them will be removed in an upcoming version. Converting an existing database means dumping it, re-creating it with `ENCODING 'UTF8'`, and restoring the dump.
```
dominic-r:
```suggestion
The database must use the `UTF8` encoding, which is PostgreSQL's default. Unless a different encoding was explicitly chosen when the database was created, no change is needed. Run `SHOW server_encoding;` to check; authentik logs a warning on any other encoding, and support for them will be removed in an upcoming version. Converting an existing database means dumping it, re-creating it with `ENCODING 'UTF8'`, and restoring the dump.
```

website/docs/install-config/configuration/configuration.mdx
```diff
@@ -67,6 +67,8 @@ To check if your config has been applied correctly, you can run the following co
 
 authentik requires PostgreSQL for application data, configuration, sessions, and background task coordination. authentik supports PostgreSQL 14 - 18.
 
+The database must use the `UTF8` encoding, which is PostgreSQL's default. Unless a different encoding was explicitly chosen when the database was created, no change is needed. Run `SHOW server_encoding;` to check; authentik logs a warning on any other encoding, and support for them will be removed in an upcoming version. Converting an existing database means dumping it, re-creating it with `ENCODING 'UTF8'`, and restoring the dump.
```
gergosimonyi:
```suggestion
The database must use the `UTF8` server encoding, which is PostgreSQL's default. Unless a different encoding was explicitly chosen when the database was created, no change is needed. Run `SHOW server_encoding;` to check; authentik logs a warning on any other encoding, and support for them will be removed in an upcoming version. Converting an existing database means dumping it, re-creating it with `ENCODING 'UTF8'`, and restoring the dump.
```

gergosimonyi (approved):

PR #25547, opened 2026-08-27
website/integrations: add Nutanix PE & clean up Nutanix

KingHenry6666:
# Why Restructure?
Nutanix AHV is the name of the Hypervisor, not the web management interface. AHV itself does not support SSO. The elements that we as engineers work with are known as Prism Element and Prism Central.

# Why add Nutanix PE?
Prism Central is for managing multiple clusters, and may be unnecessary for smaller deployments and homelabs. Additionally, there may be a need to enable SSO on each PE appliance for auditing or work needs.


KingHenry6666:
NPE Documentation is finished. I wrote it based on a working configuration I set up. I will review NPC in a bit.

dominic-r (approved):

PR #25560, opened 2026-08-28
tenants: fix base URL validation rejecting internal hostnames

### LINKED ISSUE #25546: base url invalid
### Describe the bug

Hi, using authentik 2026.8.0

On initial setup there's issues with "base url" being overly strict.

The documentation suggests I simply specify it in the desired format:
https://docs.goauthentik.io/sys-mgmt/settings/#base-url

The akadmin user creation prompts for this value, but my server for example was trying:
https://auth.svr001
...and this was failing.

I then tried wiping all volumes and setting it via the environment variable as here:
https://docs.goauthentik.io/install-config/configuration/#authentik_web__base_url
...which did set it in the container's environment variables, and the config, tested via:
docker compose exec worker ak dump_config | grep -i -A2 -B2 base_url
...but still it prompted for me to enter it again anyway on first setup

I followed the advice here to bypass this:
https://docs.goauthentik.io/troubleshooting/login/?utm_source=chatgpt.com#cant-access-initial-setup-flow-during-installation-steps

but it then complains that the user isn't recognised regardless.

I then attempted using this other alternative method to get in via recovery session:
docker compose exec server ak create_recovery_key 10 akadmin

...which let me in, and i see the base url field populated from the environment variable in the ui inside system >> settings.
however if i try to save those options in any way it continues to fail to save anything because it considers the base url that is already set, to be invalid.

Is there a way to relax this validation or what's the resolution for this?

### How to reproduce

- create new authentik install via docker compose
- as base url, use something non-conventional, like auth.svr001
- not accepted by the ui validation, for base url on both initial setup and when saving system settings

### Expected behavior

That it would be accepted

### Screenshots

_No response_

### Additional context

_No response_

### Deployment Method

Docker

### Version

2026.8.0

### Relevant log output

```shell

```

emilburzo:
Fixes https://github.com/goauthentik/authentik/issues/25546


authentik/tenants/api/settings.py
```diff
@@ -76,6 +76,14 @@ def map_serializer_field(self, auto_schema, direction):
...
     """Settings Serializer"""
 
+    # the model field is a URLField, and URLValidator rejects internal hostnames.
+    base_url = CharField(
+        required=False,
+        allow_blank=True,
+        max_length=200,
+        help_text=Tenant._meta.get_field("base_url").help_text,
+        validators=[BASE_URL_VALIDATOR],
+    )
```
rissson: The validator should be set on the model field instead, which allows you to remove this, and a bunch more of the validation code that you added on this PR

rissson (changes requested):

authentik/tenants/api/settings.py
emilburzo: I did try that, but because the model field is `models.URLField`, I could not get rid of the `URLValidator`. it seems `validators = [...]` only appends to the list of validators, it doesn't replace it

rissson: Ah yeah, I forgot about that. In that case, change the model field to become:
```
base_url = models.TextField(validators=[DomainlessURLValidator(...)])
```

emilburzo: that introduces a migration though, right? do we want that? since this needs to be backported to .8

rissson: That's fine for this. It will change the column type from `varchar(200)` to `text` which is painless in postgres. Especially for a table with 1 row

emilburzo: I went with `models.CharField` and max length 200 instead, so that there's no DDL change at all. 

Why: concerned about upgrades/downgrades and any surprises with the data during conversions.

rissson: Got it. cc. @gergosimonyi when moving from tenants to settings, let's change this to a textfield

rissson (approved):

PR #25564, opened 2026-08-28
web/admin: split flow diagram logic to generate a graph in JSON on the backend, render the graph in the client

kensternberg-authentik:
This commit introduces a new FlowDiagram workflow.

# WHY

To make it possible to add “edit stage,” “edit policy,” and “edit binding” controls directly onto the Mermaid diagram.

The existing FlowDiagram built the Mermaid graph entirely on the server, transmitting the Markdown to the client and feeding that directly into the client-side Mermaid parser. While the resulting graph showed the slugs for every policy, flow, and stage, those slugs are not always the correct UUID or PK for a stage or policy, and it wasn’t easy to reverse-engineer them. By sending the graph in a more traditional, JSON-based format of Nodes-and-Edges, it was fairly simple to move the “generate the language Mermaid consumes” onto the client.

# NOTE

This commit *does not change the user experience*. This is laying the groundwork for the controls, it does not add the controls themselves. It is strictly the minimum change necessary to support the next step.  The diagram is unchanged:

image

# DETAILS

The biggest change is simple: the back-end flows_diagram.py has been revised so that, instead of generating the stringly-based `node -> edge -> node` format that Mermaid wants, it stores `Node[]` and `Edge(from, to)[]` arrays derived using the same logic @BeryJu originally provided, and then sends those to the client. The logic is the same. The key type he provided I moved into enums, and rather than do the whole “are we into the loop proper and did the last stage have a source?” lookback dance that’s been replaced with just keeping a record of what should be looked back on via a `previous_policies` field that just gets updated with each iteration.

---

## Checklist

-   [☘️] Tests for the new feature have been written.
-   [☘️] The project has been linted, built, and tested (`make all`)
-   [☘️] The documentation has been updated and formatted (`make docs`)
-   [☘️] I have read the [AI usage policy](https://github.com/goauthentik/authentik/blob/main/AI_POLICY.md).


authentik/flows/api/flows_diagram.py
```diff
@@ -1,233 +1,254 @@
...
+    target: str
+    type: DiagramEdgeTypes = DiagramEdgeTypes.PROCEED
+
+
+@dataclass
+class DiagramGraph:
+    nodes: list[DiagramNode] = field(default_factory=list)
+    edges: list[DiagramEdge] = field(default_factory=list)
+
+
```
kensternberg-authentik: You have no idea how much it irritates me that these next two constructors are almost exactly the same, but I could not for the life of me figure out how to abstract them. FlowStageBinding and PolicyBinding have different `Meta` declarations so the trick of creating an 'intersection type' for them wasn't possible. Django's magic can be really annoying.

BeryJu (approved):
lgtm for backend

PR #25569, opened 2026-08-28
providers/oauth2: scope the end-session early return to this provider's flow

### LINKED ISSUE #25545: Blank page when triggering OIDC logout while already logged out
### Describe the bug

Not sure if this is actually a bug or rather "desired" (but unexpected) behavior, but I observe the following phenomenon when using Authentik as an OIDC provider in web apps.

When logging out of an OIDC application while the user session in Authentik has already ended (for example, after previously logging out in another tab), the browser gets stuck on a completely blank white page with no content, error message, or redirection.

### How to reproduce

1. Log in to Authentik.
2. Open two separate browser tabs:
    - Tab A: app 1 (e.g. custom internal app connected via OIDC)
    - Tab B: app 2 (e.g. some third-party app such as Discourse, connected via OIDC)
3. In tab A, trigger logout -> Authentik post-logout screen is shown.
4. On the post-logout screen, click to log out completely -> redirected to the Authentik login screen.
5. Switch to tab B and trigger logout in app 2
6. A blank white page is rendered (HTTP `200 OK` with an empty response body). No redirect happens and no UI is displayed.

### Expected behavior

The user should either be redirected to the configured `post_logout_redirect_uri` / application or to the Authentik login screen.

### Screenshots

_No response_

### Additional context

_No response_

### Deployment Method

Docker

### Version

2026.8.0

### Relevant log output

```shell

```

abhisheksharma2411:
## Details

Closes #25545.

`EndSessionView.dispatch` returned an empty `200` whenever *any* flow plan was present in the session:

```python
if SESSION_KEY_PLAN in request.session:
    return HttpResponse(status=200)
```

The guard is there so a front-channel logout iframe cannot replace an in-flight invalidation plan, but it matches every plan rather than that one.

The reporter's sequence produces a foreign plan: logging out in tab A leaves the user on the login screen, which stores an **authentication** flow plan in `SESSION_KEY_PLAN`. Logging out from a second application in tab B then hits the early return and renders a blank page — `200 OK`, empty body, no redirect — instead of running the invalidation flow.

This compares the stored plan's `flow_pk` against the flow the request would actually run, so an unrelated plan falls through to the normal logout path. `self.flow` is resolved in `resolve_provider_application()`, which `PolicyAccessView.dispatch` only calls after this override, so the override resolves it first and defers to `super()` for the resolution failures `PolicyAccessView` already handles.

Identity rather than `FlowDesignation` is the discriminator on purpose: a provider's `invalidation_flow` is not required to carry `FlowDesignation.INVALIDATION`, and the existing `test_active_flow_plan_returns_early` uses one that does not.

## Checklist

- [x] Local tests pass (`ak test authentik/`)

- [ ] The API schema has been updated (`make gen-build`)

- [ ] The code has been formatted (`make web`)

If applicable

- [ ] The documentation has been updated
- [ ] The documentation has been formatted (`make website`)

## Testing

`authentik/providers/oauth2/tests/` — **303 passed**, run against a local Postgres 18 + Redis.

The new `test_unrelated_flow_plan_does_not_return_blank` seeds an authentication-flow plan and asserts the request redirects into the invalidation flow. It was written first and failed on `main` with `AssertionError: 200 != 302` — the reporter's blank page, reproduced.

Both directions are covered, verified by reverting each half:

| mutation | result |
|---|---|
| drop the early return entirely | 2 fail — the iframe tests that motivated the guard |
| restore the original unscoped `if SESSION_KEY_PLAN in request.session` | 1 fail — the new test |

So the guard still does its original job, and the scoping is what the new test is pinning.

## AI assistance disclosure

Per `AI_POLICY.md`: I used Claude Code while working on this. It ran the searches and edits under my direction, and I reviewed every line before committing.

Two places where that mattered, since the policy asks for the extent rather than the fact:

My first assertion in the new test was `assertNotEqual(response.content, b"")`, which failed even with the fix applied — a 302 has an empty body by definition, so the assertion was testing the wrong thing. I replaced it with an assertion on the redirect target and on the plan actually being replaced.

I also nearly filed a second, unrelated report against `ExpiringModel` — `is_expired` does `now() > self.expires` while both field defaults are `expires=None, expiring=True`, which reads like a guaranteed `TypeError`. I wrote the reproduction before reporting it and it disproved me: `expiring_model_pre_save` in `authentik/core/signals.py` fills `expires` on save, and nothing in the codebase writes those models through `bulk_create` or a queryset `.update()`, so the state is unreachable. That one is not in this PR because it is not a bug.

I can explain this change, why the flow-identity comparison is the correct discriminator, and what it affects, without assistance.

### DIFF
--- a/authentik/providers/oauth2/tests/test_end_session.py
+++ b/authentik/providers/oauth2/tests/test_end_session.py
@@ -287,6 +287,30 @@ class TestEndSessionView(OAuthTestCase):
             self.client.session[SESSION_KEY_PLAN].flow_pk, self.invalidation_flow.pk.hex
         )
 
+    def test_unresolvable_application_with_a_plan_falls_through(self):
+        """A plan present plus an unresolvable application must not return a blank 200.
+
+        The early return resolves the provider itself to learn which flow this request
+        would run. When that resolution fails, PolicyAccessView has to be the one to
+        build the response, so the request falls through instead of being answered with
+        an empty body.
+        """
+        self._brand_authentication_flow()
+        plan = FlowPlan(flow_pk=create_test_flow(FlowDesignation.AUTHENTICATION).pk.hex)
+        session = self.client.session
+        session[SESSION_KEY_PLAN] = plan
+        session.save()
+
+        response = self.client.get(
+            reverse(
+                "authentik_providers_oauth2:end-session",
+                kwargs={"application_slug": "does-not-exist"},
+            ),
+            HTTP_HOST=self.brand.domain,
+        )
+
+        self.assertEqual(response.status_code, 404)
+
     def test_frontchannel_iframe_callback_preserves_injected_stages(self):
         """Stages injected into the logout plan survive the iframe's end-session hit.
 
--- a/authentik/providers/oauth2/tests/test_end_session.py
+++ b/authentik/providers/oauth2/tests/test_end_session.py
@@ -255,6 +255,38 @@ class TestEndSessionView(OAuthTestCase):
         self.assertEqual(response.status_code, 200)
         self.assertEqual(self.client.session[SESSION_KEY_PLAN].flow_pk, plan.flow_pk)
 
+    def test_unrelated_flow_plan_does_not_return_blank(self):
+        """A plan for a different flow must not short-circuit end-session (#25545).
+
+        After logging out in another tab the user lands on the login screen, which
+        stores an *authentication* flow plan in SESSION_KEY_PLAN. A logout triggered
+        from a second application then hit the early return meant for front-channel
+        logout iframes and rendered an empty 200 — a blank white page — instead of
+        running the invalidation flow.
+        """
+        self._brand_authentication_flow()
+        other_flow = create_test_flow(FlowDesignation.AUTHENTICATION)
+        plan = FlowPlan(flow_pk=other_flow.pk.hex)
+        session = self.client.session
+        session[SESSION_KEY_PLAN] = plan
+        session.save()
+
+        response = self.client.get(
+            reverse(
+                "authentik_providers_oauth2:end-session",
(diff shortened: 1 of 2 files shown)


abhisheksharma2411:
Pushed `f9f93a6` covering the branch codecov flagged, plus a note on the other three reds.

**codecov/patch** — the 2 uncovered lines were the resolution-failure path in `dispatch`. Added `test_unresolvable_application_with_a_plan_falls_through`: a plan in the session plus a slug that resolves to nothing must fall through to `PolicyAccessView` rather than return an empty body.

Being straight about what that test is worth: for the reachable case it does not *distinguish*, it only covers. `resolve_provider_application` raises `Http404` for a missing application, and `Http404` produces a 404 whether I catch it and defer to `super().dispatch()` or let it propagate. The `except` earns its place for the other two — `PolicyAccessView.dispatch` catches `Application.DoesNotExist` and `Provider.DoesNotExist` and turns them into an access-denied response, so calling `resolve_provider_application()` myself without mirroring that would convert those into 500s. Both are hard to reach deliberately (they need a FK pointing at a deleted row), which is why the test exercises the `Http404` route. Happy to drop the `except` and let everything propagate if you would rather not carry a defensive branch for unreachable states.

**test-rust** is failing on `main` independently of this PR — the last five `main` runs are all `failure`. This branch touches only `authentik/providers/oauth2/`, no Rust.

**test-e2e (rac)** and **ci-core-mark** — RAC e2e is unrelated to the OIDC logout path, and `ci-core-mark` is the aggregate that goes red when any upstream job does. Happy to rebase if a green run is wanted once `main` recovers.

`authentik/providers/oauth2/` is **303 passed** locally against Postgres 18 + Redis, and the end-session file is 14/14.

PR #25571, opened 2026-08-29
policies/reputation: reset negative score to 0 on successful login

### LINKED ISSUE #20148: Option to automatically set reputation score to 0 upon successful login

manantlerio:
### What does this PR change?

On a successful login, the reputation entry for the matching `(identifier, IP)` pair is now updated as follows, instead of being unconditionally incremented by 1:

- a **negative** score is reset to 0
- any **other** score is raised by 1

Both branches are a single `CASE` expression inside the existing `INSERT ... ON CONFLICT DO UPDATE`, so it stays one statement and there is no read-then-write race between concurrent logins.

- Only the exact `(identifier, IP)` pair is touched. Other usernames behind the same IP, and the same username from other IPs, are unaffected.
- A login with no existing entry still creates one at 1, matching the previous behaviour.
- The tenant's `reputation_lower_limit` / `reputation_upper_limit` clamping is unchanged.
- Expired entries are not touched, and expiry is not refreshed on update, both consistent with `update_score`.

No model, migration, API or settings change, so nothing needs configuring on upgrade.

Per the discussion on #20175: unconditionally incrementing means an attacker who shares an IP with a legitimate user earns extra retries per successful login. @gergosimonyi asked whether this should just be the new default rather than another flag, and @dewi-tik confirmed after speaking with @BeryJu that changing the default is preferred over adding a tenant setting.

The behaviour implemented here is the one confirmed in https://github.com/goauthentik/authentik/issues/20148#issuecomment-5441847794, corrected per @gergosimonyi's review on this PR to keep the increment for non-negative scores.

### How was this tested?

Eight tests in `authentik/policies/reputation/tests.py` covering both branches and the surrounding semantics: negative resets to 0, zero increments to 1, positive increments by 1, no existing entry creates one at 1, the upper limit still bounds the increment, another identifier on the same IP is unaffected, the same identifier on another IP is unaffected, and a failed-then-successful login ends at 0.

`ruff check` and `ruff format` are clean. I do not have PostgreSQL/Redis available locally, so I have not run the Django suite myself - the previous revision passed `test-unittest` in CI, and I am relying on CI again for this revision.

### Linked issues

Closes #20148

---

## AI usage disclosure

Per [AI_POLICY.md](https://github.com/goauthentik/authentik/blob/main/AI_POLICY.md): I used Claude Code to navigate the codebase, draft this change, and draft this description. I reviewed and edited all of it, and I understand and take responsibility for what is submitted here. Happy to answer review questions directly.

---

## Checklist

-   [ ] The project has been linted, built, and tested (`make all`) - `ruff` clean locally; I cannot run the Django suite on this machine, so I am relying on CI


gergosimonyi (commented):
This doesn't look quite right. The intention is for a successful login to (with a single SQL query)

- reset the reputation to 0 if it was negative
- increase the reputation by 1 if it was non-negative

The second part is removed here.

manantlerio:
Good catch - I dropped the increment instead of making it conditional. Fixed in b7632029: both branches now in one statement (`CASE WHEN score < 0 THEN 0 ELSE score + 1`) inside the existing `ON CONFLICT DO UPDATE`. Also restored the missing-row case (old `update_score(..., 1)` created a row at 1; my first pass created none).

website/docs/customize/policies/types/reputation.mdx
```diff
@@ -13,7 +13,7 @@ Use a Reputation policy when you want authentik to react to repeated failed auth
 authentik tracks a reputation score over time:
 
 - failed logins decrease the score
-- successful logins increase the score
+- a successful login resets a negative score back to 0, and raises any other score by 1, for that username and client IP pair
```
gergosimonyi:
```suggestion
- successful logins increase the score (with negative scores being reset to 0)
```

gergosimonyi (approved):

Issue #22196, opened 2026-05-10
Bug: Runtime values are used as structured log event names or field keys


angrychow:
### Describe the bug

# Bug: Runtime values are used as structured log event names or field keys

## Anti-pattern

This report identifies logging calls where dynamic runtime data is used as the log event name or as a structured field key. In structured logging, event names and field keys should be stable schema elements. Runtime-specific data such as IDs, URLs, object names, errors, query labels, or caller-controlled keys should be emitted as field values under stable keys instead of changing the event or key surface.

## Impact on observability platforms

When event names or field keys vary with runtime data, observability backends can see unnecessary cardinality growth in log streams, indexes, facets, labels, and dashboards. This makes queries harder to write, splits one logical event across many names or fields, increases storage and indexing cost, and can reduce the usefulness of alerts and aggregations. When opaque identifiers or shifted key/value pairs become field names, the resulting records are also harder for humans to scan and interpret.

## Affected entrypoints

Commit: `c6b5869b488b07ddd598d1f8066d3d054ae806c0`

Found 10 affected entrypoint(s): 7 message-name entrypoint(s), 3 structured-key entrypoint(s).

- `key` `authentik/sources/kerberos/sync.py:74`
- `key` `authentik/sources/ldap/sync/groups.py:127`
- `key` `authentik/sources/ldap/sync/users.py:88`
- `msg` `authentik/blueprints/v1/importer.py:372`
- `msg` `authentik/events/logs.py:40`
- `msg` `authentik/providers/saml/processors/authn_request_parser.py:71`
- `msg` `authentik/tasks/models.py:143`
- `msg` `authentik/tasks/models.py:146`
- `msg` `authentik/tasks/models.py:149`
- `msg` `authentik/tasks/schedules/scheduler.py:26`

### How to reproduce

Trigger any affected entrypoint listed below and inspect the emitted structured log record.

### Expected behavior

Log event names and structured field keys should remain stable across executions. Runtime-specific data should be emitted as structured field values under fixed, descriptive keys.

### Deployment Method

Docker

### Version

c6b5869b488b07ddd598d1f8066d3d054ae806c0

### Relevant log output

```shell

```

angrychow:
> take

Thank you!

CodeBhavyaG:
I have raised a pull request please review #25121

PR #25573, opened 2026-08-30
*: use stable structured log event names and field keys

manantlerio:
### What does this PR change?

Runtime values were being used as structured log event names and, in the LDAP and Kerberos synchronizers, as structured field keys. Runtime data now goes into fields under fixed keys.

- `sources/kerberos`, `sources/ldap`: the property-mapping output was spread as kwargs (`**defaults`), so mapping-controlled attribute names became log field keys. It is now logged under a single `attributes` key, which matches how `attributes` is already used elsewhere in these synchronizers. This also removes a crash: a mapping that produced a key colliding with a structlog argument (`event`) raised `TypeError`.
- `blueprints`: `f"Entry invalid: {exc}"` becomes `"Entry invalid"`. The exception was already attached as the `error` field.
- `providers/saml`: the ACS URL mismatch warning interpolated both URLs into the event name; they are now `request_acs_url` and `provider_acs_url` fields. The `CannotHandleAssertion` message is unchanged.
- `tasks`: exception text is no longer used as the task log event name. This branch only runs when an `Exception` is passed as the message, so the event name was previously the exception's own `str()`. It is now `"Task has encountered an error"`, or `"Task has encountered an error and will be retried"` for `Retry`. Full exception details are still recorded in the `exception` attribute via `exception_to_dict`.
- `tasks/schedules`: `f"Sent {count} scheduled tasks"` becomes `"Sent scheduled tasks"` with a `count` field, matching the base `Scheduler` in `packages/django-dramatiq-postgres`.

`authentik/events/logs.py:40` was listed in the issue but is intentionally left unchanged. `LogEvent.log()` replays an already-captured log record when a task log cannot be written to the database, so its event name has to stay the one from the original call site. The sources feeding it are fixed above.

Event names and field keys are schema, not data. When they vary with runtime values, one logical event splits across many names and the field surface changes per deployment, which inflates cardinality in log indexes and facets and makes the records harder to query and read. The LDAP and Kerberos case is sharper than that: because the keys came from property mappings, a directory attribute named `event` collided with structlog's own argument and raised `TypeError` during sync.

### How was this tested?

Not run locally. `make all` needs a Postgres and Redis dev environment I have not stood up, so I would rather rely on CI here than claim a run I did not do.

What I did check by hand, against `main`:

- `attributes=` is already used as a log field key in `authentik/sources/ldap/sync/users.py`, so the new key name is consistent with the surrounding code.
- The base `Scheduler` in `packages/django-dramatiq-postgres/django_dramatiq_postgres/scheduler.py` already logs `"Sent scheduled tasks", count=count`, so the subclass now matches it.
- `exception_to_dict` uses structlog's `ExceptionDictTransformer`, which retains the exception type, message and traceback. Nothing is lost by dropping the exception text from the event name.
- No behaviour changes outside logging. The `CannotHandleAssertion` message is byte-identical to before, and the `tasks/models.py` branch only touches a path where the message argument was already an exception object.

Happy to add a regression test for the `event` key collision if you would like one.

### Linked issues

refs #22196

---

## Checklist

-   [ ] The project has been linted, built, and tested (`make all`) - not run locally, see "How was this tested?" above
-   [ ] The documentation has been updated and formatted (`make docs`) - no documentation changes required

## AI usage disclosure

Required by [AI_POLICY.md](https://github.com/goauthentik/authentik/blob/main/AI_POLICY.md).

Tool: Claude Code (Anthropic, Claude Opus). Extent: the AI drafted the patch, the commit message and the text of this PR body from the entrypoint list in #22196.

I have reviewed every hunk against the surrounding code, checked the claims above about existing behaviour rather than taking them on trust, and I understand what this change does and why it is correct. I take responsibility for what is submitted here, and I am happy to answer review questions on any part of it. The commit carries a `Co-Authored-By` trailer accordingly.

### DIFF
--- a/authentik/blueprints/v1/importer.py
+++ b/authentik/blueprints/v1/importer.py
@@ -349,7 +349,7 @@ class Importer:
             try:
                 retry_serializer = self._validate_single(entry)
             except EntryInvalidError as exc:
-                self.logger.warning("Entry invalid on retry", entry=entry, error=exc)
+                self.logger.warning(f"Entry invalid on retry: {exc}", entry=entry, error=exc)
                 if raise_errors:
                     raise exc
                 return None
@@ -407,7 +407,7 @@ class Importer:
                 if entry.get_state(self._import) == BlueprintEntryDesiredState.ABSENT:
                     serializer = exc.serializer
                 else:
-                    self.logger.warning("Entry invalid", entry=entry, error=exc)
+                    self.logger.warning(f"Entry invalid: {exc}", entry=entry, error=exc)
                     if raise_errors:
                         raise exc
                     return False
--- a/authentik/blueprints/v1/importer.py
+++ b/authentik/blueprints/v1/importer.py
@@ -349,7 +349,7 @@ class Importer:
             try:
                 retry_serializer = self._validate_single(entry)
             except EntryInvalidError as exc:
-                self.logger.warning(f"Entry invalid on retry: {exc}", entry=entry, error=exc)
+                self.logger.warning("Entry invalid on retry", entry=entry, error=exc)
                 if raise_errors:
                     raise exc
                 return None
@@ -407,7 +407,7 @@ class Importer:
                 if entry.get_state(self._import) == BlueprintEntryDesiredState.ABSENT:
                     serializer = exc.serializer
                 else:
-                    self.logger.warning(f"Entry invalid: {exc}", entry=entry, error=exc)
+                    self.logger.warning("Entry invalid", entry=entry, error=exc)
                     if raise_errors:
                         raise exc
                     return False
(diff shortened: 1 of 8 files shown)


authentik/blueprints/v1/importer.py
```diff
@@ -407,7 +407,7 @@ def _apply_models(self, raise_errors=False) -> bool:
                 if entry.get_state(self._import) == BlueprintEntryDesiredState.ABSENT:
                     serializer = exc.serializer
                 else:
-                    self.logger.warning(f"Entry invalid: {exc}", entry=entry, error=exc)
+                    self.logger.warning("Entry invalid", entry=entry, error=exc)
                     if raise_errors:
```
BeryJu, line 411: These log messages contain the error message on purpose as they are returned by the API so having them included in the string makes them easier to read

BeryJu (changes requested):
lgtm aside from the one note

manantlerio:
Reverted, thanks - kept the error in the message string. Did the same for the `Entry invalid on retry` line just above it, since it is the same case. `importer.py` is back to original; the rest of the PR is unchanged.

BeryJu (approved):

PR #25577, opened 2026-08-30
docs(sources): fix group-sync disable snippet and document override behavior

chrisblech:
## What

The 2024.8 release notes say to disable source group synchronization with
`return {"groups": []}`. That's a no-op — the provider's groups are still
imported. This corrects it and documents the surrounding merge/override
behavior.

## Why `[]` doesn't work

Group claims are written into the base user properties before any mapping runs
(OAuth/OIDC: `get_base_user_properties` → `"groups": info.get("groups", [])`;
SAML: the group attribute). `SourceMapper.build_object_properties` then merges
each mapping result into those base properties with `MERGE_LIST_UNIQUE`
(deepmerge `append_unique`), so appending `[]` changes nothing. The merge config
and base behavior are identical in `version/2024.8.0` and current — the snippet
was never correct. `None` is the only value that clears the attribute.

## Changes

- Correct the release-note snippet to `{"groups": None}`.
- Add to the property mappings page:
  - merging (not replacing) is the default;
  - disable with `{"groups": None}` (not `[]`);
  - exact override via two name-ordered mappings (`aa …` → `None`, then
    `ab …` → target list; relies on `PropertyMappingManager.order_by("name")`),
    or by assigning to the `properties` accumulator and returning `None`;
  - groups a source creates/links are reconciled per (group, source) on every
    login, which the group view doesn't surface (matters with Name link
    matching).

## Files

- `website/docs/releases/2024/v2024.8.mdx`
- `website/docs/users-sources/sources/property-mappings/index.mdx`

## Notes for reviewers

- Both override techniques rely on current behavior — the two-mapping one on
  execution order (`order_by("name")`), the single-mapping one on `properties`
  being the live accumulator. Flagging in case you'd prefer a dedicated replace
  semantic.
- The historical release note is corrected here; happy to instead leave it and
  only fix the living page.

## AI usage disclosure

Prepared with AI assistance (Anthropic's Claude): cross-checking behavior against
the authentik source, reproducing the list-merge locally with `deepmerge`,
drafting the docs, and running Prettier. I verified in a running 2026.8 instance
that `{"groups": []}` doesn't disable group sync, `{"groups": None}` does, and
the two-mapping approach works. I've reviewed all changes, understand them, and
can explain them without AI assistance.

## Checklist

### DIFF
--- a/website/docs/releases/2024/v2024.8.mdx
+++ b/website/docs/releases/2024/v2024.8.mdx
@@ -43,7 +43,7 @@ slug: "/releases/2024.8"
 
     ```python
     return {
-        "groups": [],
+        "groups": None,
     }
     ```
 
--- a/website/docs/users-sources/sources/property-mappings/index.mdx
+++ b/website/docs/users-sources/sources/property-mappings/index.mdx
@@ -78,6 +78,83 @@ The `groups` attribute is a special attribute that must contain group identifier
 
 An identifier has to be a simple value such as a string. Entries that are not, such as the objects some identity providers return in an OpenID Connect `groups` claim, are skipped, and a **Configuration error** event records how many were dropped.
 
+#### Disabling group synchronization
+
+Returning an empty list does **not** disable group synchronization. The
+`groups` a source collects (from the OpenID Connect `groups` claim, or the
+SAML group attribute) are placed into the base user properties before any
+property mapping runs, and mapping results are merged into them with a
+unique-append strategy. `return {"groups": []}` therefore appends nothing and
+the provider's groups are still imported.
+
+To discard the collected groups, return `None` for the attribute. As described
+in [Expression semantics](#expression-semantics), an attribute set to `None` is
+dropped entirely:
+
+```python
+return {"groups": None}
+```
+
+#### Overriding the collected groups
+
+Because results are merged rather than substituted, returning a list _adds to_
+the provider's groups instead of replacing them. There are two ways to end up
+with an exact set of groups and drop whatever the source collected.
+
+**Two ordered mappings (recommended).** Property mappings run in alphabetical
+order by name. Setting an attribute to `None` drops it (see
+[Expression semantics](#expression-semantics)); a later mapping can then set it
+to the value you want. Assign both as user property mappings on the source:
+
+- `aa - discard source groups`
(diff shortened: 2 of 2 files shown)


dominic-r:
Thank you for your contribution, however, the 2024 release has long passed, and as a result, I don't see the need to modify release notes so far back.

chrisblech:
Hey @dominic-r, thanks for taking the time to review this — and fair point on not wanting to
churn a 2024 release note. I understand the reluctance to touch history that far
back.

Could I ask you to reconsider, or at least split the decision?

1. Most of this PR isn't the release note. The bulk is additions to the living
   property mappings page (`users-sources/sources/property-mappings/index.mdx`):
   how group sync merges rather than replaces, how to actually disable it
   (`{"groups": None}`) and override it, and how groups a source creates or links
   are reconciled per (group, source) — which isn't visible in the group view.
   That content stands on its own and isn't tied to the old release note. If the
   release-note change is the blocker, I'm happy to drop it and keep just the docs
   additions.

2. On the release note itself, my concern isn't historical accuracy — it's that
   the snippet is a factual error that currently misleads users. That note is the
   only place in the docs documenting how to disable source group sync, so it's
   what search returns (and what AI assistants quote as authoritative) when
   someone researches this. `{"groups": []}` is a silent no-op: users apply it,
   believe groups are no longer imported, and they still are. Since group
   membership commonly feeds authorization, that's the kind of silent
   misconfiguration that can have security consequences — which seems worth
   correcting regardless of the release's age.

If a full edit to the note feels wrong, I'm glad to do whichever is least
intrusive: just the one-line snippet fix, or leaving the note as-is and adding a
short correction that points to the property mappings page. Whatever you prefer.

Either way, thanks for maintaining authentik.

dominic-r:
The PR is a meat proxy. 78 lines of unreviewed AI slop. I'll take the spirit of it and write the None vs [] note on the mappings page myself.

dominic-r:
https://github.com/goauthentik/authentik/pull/25578

dominic-r:
https://github.com/goauthentik/authentik/pull/25578

dominic-r:
I've credited you in the PR. Thanks for reporting!

PR #25597, opened 2026-08-31
root: skip database-calling middlewares for liveness probe

severin:
This PR is a "fix"/performance improvement of the liveness probe endpoint.

As per Kubernetes best practices liveness probes should not hit any dependencies (like databases):

> When your app has a strict dependency on back-end services, you can implement both a liveness and a readiness probe. The liveness probe passes when the app itself is healthy, but the readiness probe additionally checks that each required back-end service is available. This helps you avoid directing traffic to Pods that can only respond with error messages.
(from https://kubernetes.io/docs/concepts/workloads/pods/probes)

Currently the liveness probe implementation triggers a few database queries (for tenant and brand; from middlewares).

This PR fixes this by skipping (irrelevant) middlewares that do db queries for the liveness probe endpoint. I'll also open another PR with a suggestion for a more "radical" approach (edit: see https://github.com/goauthentik/authentik/pull/25598).

### What does this PR change?

It adds "skips" to middlewares that trigger database queries when processing a liveness probe endpoint request.

To remove any (accidental) external dependencies for the liveness probe endpoint.

### How was this tested?

By adding a unit test and manually verifying that the endpoint returns the desired status.

### Disclaimer

I used an AI agent (opencode with Claude Opus and Gemini Flash) to help with the implementation. All the changes were reviewed and tested manually.

## Checklist

### DIFF
--- a/authentik/brands/middleware.py
+++ b/authentik/brands/middleware.py
@@ -18,6 +18,10 @@ class BrandMiddleware:
         self.get_response = get_response
 
     def __call__(self, request: HttpRequest) -> HttpResponse:
+        # Skip for liveness probe
+        if request.path == "/-/health/live/":
+            return self.get_response(request)
+
         locale_to_set = None
         if not hasattr(request, "brand"):
             brand = get_brand_for_request(request)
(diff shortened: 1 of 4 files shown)


BeryJu:
closed in favour of #25598

PR #25598, opened 2026-08-31
root: implement liveness endpoint in a middleware

severin:
This is a suggestion on top of https://github.com/goauthentik/authentik/pull/25597. I personally find it the better approach because there's no easy way for regressions (i.e. new middlewares that add database queries and would have to be remembered to skip). Also, it's more efficient (and therefore probably more performant, though I'm not sure how quantifiable the impact is) since almost the whole Django request processing can be skipped.

---

As per Kubernetes best practices liveness probes should not hit any dependencies (like databases):

> When your app has a strict dependency on back-end services, you can implement both a liveness and a readiness probe. The liveness probe passes when the app itself is healthy, but the readiness probe additionally checks that each required back-end service is available. This helps you avoid directing traffic to Pods that can only respond with error messages.
(from https://kubernetes.io/docs/concepts/workloads/pods/probes)

Currently the liveness probe implementation triggers a few database queries (for tenant and brand; from middlewares).

This PR proposes the following way to address this: It implements the liveness endpoint in a middleware that runs before anything else, therefore any side-effects from other middlewares are skipped.

Note: This means that liveness endpoint calls no longer are treated as "normal" Django requests and won't be reported in e.g. Prometheus metrics. I added a mention of this in the docs but it should probably also be called out in the release notes.

Since this is quite a big change to how things work currently I'm also proposing a less intrusive (but more manual) approach in https://github.com/goauthentik/authentik/pull/25597

### What does this PR change?

It implements the liveness endpoint in a middleware that runs before anything else instead of a Django view. Therefore any side-effects from other middlewares are skipped.

To remove any (accidental) external dependencies for the liveness probe endpoint

### How was this tested?

By adding a unit test and manually verifying that the endpoint returns the desired status.

### Disclaimer

I used an AI agent (opencode with Claude Opus and Gemini Flash) to help with the implementation. But all the changes were reviewed and tested manually.

## Checklist


authentik/root/urls.py
```diff
@@ -47,6 +47,5 @@
 urlpatterns = [
     path(CONFIG.get("web.path", "/")[1:], include(_urlpatterns)),
     path("-/metrics/", MetricsView.as_view(), name="metrics"),
-    path("-/health/live/", LiveView.as_view(), name="health-live"),
```
BeryJu, line 50: I'd prefer keeping this as a placeholder view (doesn't have to have any logic), just so we can be consistent with using `reverse` for the path

severin: @BeryJu I'm not 100% sure whether I understand you correctly: you would like to have the view around AND the middleware? The view would be "dead code" then (because it is never rendered because the middleware intercepts calls to its route), but it would more consistently indicate that the path/route exists?

BeryJu: yeah basically just

```python
from django.views import View

...

# Dummy view for URL source of truth, handled in middleware
path("-/health/live/", View.as_view(), name="health-live"),
```

severin: @BeryJu pushed this in https://github.com/goauthentik/authentik/pull/25598/commits/488ae23f8e1c3b00fca345480025f2d609498718

dewi-tik (approved):

BeryJu (approved):

authentik/root/urls.py
severin: @BeryJu thanks for cleaning this up. Sorry for not properly looking at the test 🙈

BeryJu: no worries, all good

PR #25609, opened 2026-08-31
web: show pre-selected value in paginated search-selects

### LINKED ISSUE #25406: When editing binding in stage-binding within flow, it shows no object
### Describe the bug

It is rather difficult to describe... So here's a quick screen-capture of said odd behavior. When editing, it does not list the correct, assigned policy:
Image

It is only hidden in the list when clicking in the "Select an object..." field. Then, only when the list lazy-loads and would show the policy in the list, it auto-completes. But this way, it makes changing and saving the binding rather frustrating.

### How to reproduce

1. Go to a flow
2. Go to Stage Bindings
3. Edit a existing stage binding
4. It will sometimes (!) not show the linked policy. I suspect it is related to which lazy-loaded, limited policies it initially lists.

### Expected behavior

1) Being able to see policy and save straight away.
2) Being able to see all policies or at least indicate that entering something to filter is needed.

### Screenshots

_No response_

### Additional context

_No response_

### Deployment Method

Docker

### Version

2026.8.0

### Relevant log output

```shell

```

GirlBossRush:
This PR fixes an issue where editing a policy or stage binding would show an empty selector instead of the currently bound object.

`ak-search-select` only auto-selects an object that appears in the fetched API page. Those lists are paginated (~20/page), so when the bound policy/stage sorts onto a later page, `.selected` never matched it and the field rendered blank — the value only appeared after the user typed enough to pull it into the list.

The fix merges the binding instance's embedded object (`policyObj` / `stageObj` / `groupObj` / `userObj`) into the fetched results when the current page doesn't already include it, so the selection renders immediately. This mirrors the existing pattern in `TokenForm`. Applied to the policy, group, and user selectors in `PolicyBindingForm` and the stage selector in `StageBindingForm`.

Fixes #25406

### DIFF
--- a/web/src/admin/flows/StageBindingForm.ts
+++ b/web/src/admin/flows/StageBindingForm.ts
@@ -140,9 +140,6 @@ export class StageBindingForm extends ModelForm<FlowStageBinding, string> {
                             args.search = query;
                         }
                         const stages = await aki(StagesApi).stagesAllList(args);
-                        // The list is paginated, so the currently-bound stage may not appear on
-                        // the first page. Merge it in so it renders as selected without the user
-                        // having to search for it first.
                         const selectedStage = this.instance?.stageObj;
                         if (
                             selectedStage &&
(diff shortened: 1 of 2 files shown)


kensternberg-authentik (approved):
It's a good hack for the time being.

PR #25615, opened 2026-08-31
core: load and serialize users once in `GroupViewSet.list`

gergosimonyi:
Alternative to #25366


rissson (approved):

BeryJu (approved):
if this is needed I'm fine with this, but I'd rather not have it

PR #25640, opened 2026-09-01
website/docs: flows-stages: update section

dewi-tik:
### What does this PR change?

Add descriptions to files, reorder sidebar, change some wording, remove example directory, add redirects for moved files

### Linked issues

---

## Checklist


website/docs/add-secure-apps/flows-stages/flow/snippets.mdx
```diff
@@ -12,12 +18,10 @@ plan.redirect("https://foo.bar")
 return False
 ```
 
-This policy should be bound to the stage after your redirect should happen. For example, if you have an identification and a password stage, and you want to redirect after identification, bind the policy to the password stage. Make sure the stage binding's option _Evaluate when stage is run_ is enabled.
+## Deny flow when user is authenticated
 
-### Deny flow when user is authenticated
+When used with authentik 2022.7 or later, set the flow _Denied action_ to _CONTINUE_. This will redirect already authenticated users to the default interface if they try to use the respective flow.
```
dominic-r: would remove the 2022 date, it's too old

dominic-r (approved):

PR #25641, opened 2026-09-01
root, web, docs: pnpm 12.4.0 upgrade. Remove Corepack.

GirlBossRush:
## Details

This PR upgrades pnpm to 12.4.0 across the root, `web`, and `website` workspaces, and removes Corepack. The pnpm team no-longer recommends Corepack as part of their setup. Given the gradual migration away from Corepack in the Node community, it looks like Corepack may be deprecated entirely, and instead allowing each package manager to enforce its own compatibility matrix.

Corepack was the only thing verifying the `+sha512.` suffix on `packageManager` — pnpm's own version switcher reads the version and ignores the hash — so two lint checks take that job over:

- `lint-runtime` resolves the pinned version's `dist.integrity` from the registry, re-encodes it as hex, and compares. On a mismatch it prints the corrected line to paste in. It also checks the pin satisfies `engines.pnpm`, which had drifted. An unreachable registry warns and skips rather than failing, so offline installs still work.
- `lint-catalogs` now also covers the seven places that pin pnpm: five `packageManager` fields and the two pnpm image tags. The Dockerfiles turned out to be three patches behind the manifests.

While in here, the `scripts/node` helpers and `packages/logger-js` moved to TypeScript. Both run directly under Node's type stripping, so there's no build step, and `tsc -b` now covers them — the root project had `files: []` and was type-checking nothing.

### Notes on the two package changes

`packages/logger-js` — `types` resolved to the browser implementation while the `node` condition ran the Node one, so importing it by name on Node described the wrong value. Each condition now carries its own declarations. `publishConfig.exports` points published consumers at the emitted JS, since Node refuses to strip types for files whose real path is under `node_modules`.

`packages/esbuild-plugin-live-reload` — `@goauthentik/logger-js` was an optional peer that the plugin imported unconditionally, ranged `^1.0.1` against a package sitting at 2.0.2, and referenced from the emitted `.d.ts`. The package already ships a console-backed `createLogger` that the client entry uses; the plugin entry uses it now too, and the dependency is gone.

### How was this tested?

- `pnpm lint` — type-check, runtime pins, catalog drift, spellcheck
- every commit checked out individually, confirming the scripts run and `tsc -b` passes at each point in the history
- `make node-preinstall` in a worktree with no `node_modules` and nothing built, since it runs before `pnpm install`
- `packages/logger-js` packed and installed as a plain npm dependency, confirming published consumers get the emitted JS


pnpm-workspace.yaml
```diff
@@ -19,7 +19,7 @@ catalog:
   "@typescript-eslint/eslint-plugin": "^8.65.0"
   "@typescript-eslint/parser": "^8.65.0"
   "@typescript-eslint/utils": "^8.65.0"
-  "@types/node": "^26.1.2"
+  "@types/node": "^26.4.0"
```
GirlBossRush, line 22: Bumping this intentionally to ensure that the lock files aren't stale after the pnpm upgrade.

website/pnpm-workspace.yaml
```diff
@@ -36,7 +36,7 @@ catalog:
   "@typescript-eslint/eslint-plugin": "^8.65.0"
   "@typescript-eslint/parser": "^8.65.0"
   "@typescript-eslint/utils": "^8.65.0"
-  "@types/node": "^26.1.2"
+  "@types/node": "^26.4.0"
```
GirlBossRush, line 39: Bumping this intentionally to ensure that the lock files aren't stale after the pnpm upgrade.

BeryJu (approved):

PR #25646, opened 2026-09-01
web/elements: fix ak-diagram race condition

kensternberg-authentik:
web/element: fix race condition in ak-diagram

## Why

While working on a different feature that introduces a small delay in ak-diagram, I noticed that once in awhile it wouldn’t render at all. Diagram relies on network transactions for its data, then on the component rendering, and *then* on Mermaid finishing its own internal render pass… the last of which Lit’s scheduler cannot see and for which it has no telemetry.

## Details

This commit re-structures the schedule of events inside ak-diagram so that the diagram renders the *last* Mermaid diagram passed to it heedless of the order in which the involved renderers complete.

For my own purposes (that feature mentioned above), I have also provided an override-or-callback that you can call to further manipulate the diagram before it’s been rendered.

## Checklist

-   [♨️] The project has been linted, built, and tested (`make all`)
-   [♨️] The documentation has been updated and formatted (`make docs`)
-   [♨️] I have read the [AI usage policy](https://github.com/goauthentik/authentik/blob/main/AI_POLICY.md).


web/src/elements/Diagram/ak-diagram.ts
```diff
@@ -54,30 +61,63 @@ export class Diagram extends AKElement {
...
+    protected override updated(changedProperties: PropertyValues<this>): void {
+        super.updated(changedProperties);
+        if (changedProperties.has("diagram") || changedProperties.has("activeTheme")) {
+            requestAnimationFrame(() =>
+                this.#renderMermaid().catch((error: unknown) => {
+                    console.warn("Could not render diagram:", error);
+                }),
+            );
+        }
+    }
```
GirlBossRush, line 76:
```suggestion
   protected mermaidRenderFrameID = -1;

    protected override updated(changedProperties: PropertyValues<this>): void {
        super.updated(changedProperties);
        
        if (changedProperties.has("diagram") || changedProperties.has("activeTheme")) {
            cancelAnimationFrame(this.mermaidRenderFrameID);
            
            this.mermaidRenderFrameID = requestAnimationFrame(() =>
                this.#renderMermaid().catch((error: unknown) => {
                    console.warn("Could not render diagram:", error);
                }),
            );
        }
    }
```

web/src/elements/Diagram/ak-diagram.ts
```diff
@@ -54,30 +61,63 @@ export class Diagram extends AKElement {
...
+    // class, or pass in a callback. Meant to let clients fill in details using the SVG
+    // ForeignObject protocol. The diagram is completely torn down and replaced every time, so
+    // whatever you put here must fully render and never assume anything from a previous pass is
+    // still present.
+    //
+    protected diagramUpdated(): void {
+        if (this.diagramUpdatedCallback) {
+            this.diagramUpdatedCallback.call(this.diagramUpdatedCallback, this);
+        }
+    }
```
GirlBossRush:
```suggestion
    /**
     * Callback to be called when the diagram has been updated. This is called after the diagram has
     * been rendered, and can be used to fill in details using the SVG ForeignObject protocol.
     *
     * Meant to let clients fill in details using the SVG ForeignObject protocol. 
     * The diagram is completely torn down and replaced every time.
     * Whatever you put here must fully render and never assume anything from a previous pass is still present.
     */
    protected diagramUpdated(): void {
        this.diagramUpdatedCallback?.(this)
    }
```

GirlBossRush (approved):

PR #25648, opened 2026-09-01
website/docs: document password reset behavior for inactive users

### LINKED ISSUE #18391: Forgot password when a user is deactivated
### Describe the bug

An user which is deactivated can use the forgot password feature: it receives an e-mail and is able to change its password (of course it's not able to authenticate).

### How to reproduce

1. As administrator:
- create a recovery flow
- create a new user (set an e-mail address and set a password)
- deactivate this user

2. In a private window, go on the authentication flow:
- start the login as user
- use the Forgot Password feature
-> you should receive an e-mail and be able to change the password even if your user is deactivated

### Expected behavior

- to let the flow being executed until the password is sent
- to inform the user an e-mail was sent: this allows us to avoid showing which users exist and which users do not.
- but to not really send the e-mail + to not authorize to continue the recovery procedure

### Screenshots

_No response_

### Additional context

Of course we can implement a Deny Stage in the recovery flow to manage this problem.
But from a certain point of view it can be considered as a bug or like something to improve (for example: having a parameter somewhere to decide if we accept this behavior or not)

### Deployment Method

Docker

### Version

2025.10.2

### Relevant log output

```shell

```

dewi-tik:
### What does this PR change?

Adds notes to docs about inactive users being reactivated by recovery flow/password reset depending on settings.

### Linked issues

Closes #18391

---

## Checklist

### DIFF
--- a/website/docs/add-secure-apps/flows-stages/stages/deny/index.mdx
+++ b/website/docs/add-secure-apps/flows-stages/stages/deny/index.mdx
@@ -43,3 +43,31 @@ return ak_client_ip in ip_network("203.0.113.0/24")
 3. Bind that policy to the Deny stage in the flow.
 
 When the policy passes, the Deny stage runs and the flow stops immediately.
+
+### Example: Block password recovery for inactive users
+
+Account deactivation prevents authentication, but it does not automatically prevent recovery stages from sending email or changing a password. To stop inactive users before those actions occur:
+
+1. Add a Deny stage to the recovery flow after the Identification stage and before the Email stage.
+2. Create an [Expression policy](../../../../customize/policies/types/expression/index.mdx) with the following expression:
+
+    ```python
+    pending_user = request.context.get("pending_user")
+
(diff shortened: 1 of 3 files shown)


website/docs/add-secure-apps/flows-stages/stages/deny/index.mdx
```diff
@@ -43,3 +43,31 @@ return ak_client_ip in ip_network("203.0.113.0/24")
 3. Bind that policy to the Deny stage in the flow.
 
 When the policy passes, the Deny stage runs and the flow stops immediately.
+
+### Example: Block password recovery for inactive users
```
dominic-r, line 47: why can password activity occur for inactive users?

dewi-tik: See https://github.com/goauthentik/authentik/issues/18391

dominic-r (approved):

PR #25719, opened 2026-09-03
source/saml: properly verify audience in response matches entity id

PeshekDotDev:
### Linked issues

---

## Checklist

### DIFF
--- a/authentik/sources/saml/processors/response.py
+++ b/authentik/sources/saml/processors/response.py
@@ -40,6 +40,7 @@ from authentik.lib.utils.time import timedelta_from_string
 from authentik.sources.saml.exceptions import (
     InvalidEncryption,
     InvalidSignature,
+    MismatchedAudience,
     MismatchedRequestID,
     MissingSAMLResponse,
     SAMLException,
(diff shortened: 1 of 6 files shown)


authentik/sources/saml/tests/test_response.py
```diff
@@ -85,6 +90,22 @@ def test_success(self):
...
+                "SAMLResponse": b64encode(
+                    load_fixture("fixtures/response_success.xml").encode()
+                ).decode()
+            },
+        )
+
+        parser = ResponseProcessor(self.source, request)
+        with self.assertRaises(MismatchedAudience):
+            parser.parse()
+
```
emilburzo, line 136: would it make sense to add more tests here? e.g.:

- no `AudienceRestriction` at all
- multiple `Audience` elements in one `AudienceRestriction`, one of them "us" (ok - OR)
- multiple `AudienceRestriction` with one `Audience`, one of them "us" (fail - AND)

emilburzo (commented):
should we also already drop the audience override in this PR?

authentik/sources/saml/processors/response.py
```diff
@@ -143,6 +144,24 @@ def _verify_conditions(self):
...
+                get_element_text(audience).strip()
+                for audience in restriction.findall(f"{{{NS_SAML_ASSERTION}}}Audience")
+            ]
+            if entity_id not in audiences:
+                LOGGER.warning(
+                    "Assertion audience does not match source entity ID",
+                    entity_id=entity_id,
+                    audiences=audiences,
+                )
+                raise MismatchedAudience()
```
BeryJu, line 179: Should probably only throw if we're not in any of the audience restriction elements?

PeshekDotDev:
@emilburzo yea, dropping the audience_override in this PR is a good idea, I removed it

authentik/sources/saml/processors/response.py
PeshekDotDev: The spec is odd. The following is allowed in a SAML response:

```
<saml:Conditions>
  <saml:AudienceRestriction>
    <saml:Audience>https://sp-a.example.com</saml:Audience>
    <saml:Audience>https://sp-b.example.com</saml:Audience>
    <saml:Audience>https://sp-c.example.com</saml:Audience>
    <saml:Audience>https://sp-d.example.com</saml:Audience>
  </saml:AudienceRestriction>
  <saml:AudienceRestriction>
    <saml:Audience>https://sp-a.example.com</saml:Audience>
  </saml:AudienceRestriction>
</saml:Conditions>
```

In this example, only sp-a is allowed to consume this because it is the only sp that exists in both (all) audience restrictions

https://docs.oasis-open.org/security/saml/v2.0/saml-core-2.0-os.pdf#page=23 -- Line 922

If desired, we could do something like throw an error if we don't exist in any, and then throw a different error if we are in one audience restriction but not all

authentik/sources/saml/tests/test_response.py
PeshekDotDev: Good calls. Those are added, and I also added one that makes sure we pass when we are present in both `AudienceRestriction` elements

emilburzo (approved):

PR #25734, opened 2026-09-04
web: Close command palette when opening About

### LINKED ISSUE #25726: Clicking About authentik in command palette doesn't close the palette
It does close when you exit the About authentik, but it's kept open in background

dominic-r:
Opening About returned a promise that resolved only after the dialog closed, leaving the command palette visible behind it.

Closes: #25726


kensternberg-authentik (approved):
Eewwwwwwww.... relying on the _side effect_.  Gross.

Issue #16776, opened 2025-09-15
Allow non-http(s) urls for applications


PrivatePuffin:
**Is your feature request related to a problem? Please describe.**
We host multiple game servers which allow steam links `steam://` to open them from the browser directly.
Currently the Application system does not allow us to use said links, which is a shame.

**Describe the solution you'd like**
Allow non https links for applications, so users can be offered non-webpage links as well.

**Describe alternatives you've considered**
Having to host some sort of forwarding, which is convoluted at best

PR #25801, opened 2026-09-05
core, web: add configurable additional links to application cards

### LINKED ISSUE #25800: Allow additional links on application cards
### The problem

The user library gives every application exactly one action: open it in a browser.

Plenty of self-hosted services also ship a native client, and the card says nothing about it — users
go hunting elsewhere and install the wrong one. Those clients then ask for a server address, which
authentik already knows from the launch URL but never offers in a form you can copy.

### What I'd like

An optional row under the application card, configured per application:

- **Links** — label, URL, optional icon. Native clients, documentation, status page, support.
  Nothing is derived automatically.
- **A copy button** — puts the application's address on the clipboard, shows it in a tooltip, and
  confirms when it lands. The address is no longer printed on the card: it took a whole line to say
  what a button says better.
- **A heading** for the row, optional, aligned left, centre or right.

Two switches in the application's **UI settings** turn each half on, next to *Open in new tab* and
*Hide from Application Dashboard*. Both default to off, and the link editor stays usable whether the
switch is on or not.

The row renders in the list view as well as the grid. That is not a detail: below 768px the library
forces list mode, so a card-only feature is invisible on every phone.

### Related: #16776

#16776 asks for custom schemes such as `steam://` as the application URL. The two are complementary
rather than competing: that one changes what the primary launch does, this one keeps the web launch
primary and adds the rest beside it.

### How it is built

- **Storage** — one `JSONField` on `Application`, following the `Tenant.footer_links` precedent. The
  whole block (switches, heading, alignment, links) lives in that single object, so the display
  settings cost no extra column and no second migration. The earlier bare-list shape is still read,
  so nothing already stored is lost.
- **Icons** — the existing `Application.meta_icon` / `ak-app-icon` convention.
- **Admin UI** — the existing `ak-array-input`.
- **Security** — URLs are validated server-side and anything that is not `http(s)` is rejected. A URL
  typed without a scheme is completed to `https://`, but a scheme we refuse is never rewritten into
  one we accept.
- **API** — no new endpoint, and no extra request on the dashboard.

I did consider a dedicated `ApplicationLink` model with a foreign key and explicit ordering. I went
with the JSON field because it is the smaller change and follows an existing pattern — happy to
convert it if you would rather have the model.

### Alternatives considered

- **Brand custom CSS** — can restyle, cannot create links.
- **`meta_description`** — not shown on the card, HTML escaped, and mixes configuration with prose.
- **One application per platform** — works, but splits one service across several cards.
- **An external landing page** — works, adds a hop and duplicates what authentik already holds.
- **Homepage, Homarr and friends** — replacing the portal defeats the point of a single door behind
  SSO.
- **Custom JavaScript** — far too much surface for something this small.

### Screenshots

before : 
Image
Image
after : 
Capture d&#39;écran 2026-09-06 185357
Capture d&#39;écran 2026-09-06 185445
Capture d&#39;écran 2026-09-06 190600

### Two things I noticed on the way

Both independent of this proposal, and happy to split them out:

- `ak-file-search-input` appears to fire two identical requests to `/api/v3/admin/file/` per
  instance.
- `ak-app-icon` only constrains `max-height`, so an SVG without intrinsic dimensions can overflow
  its row.

### Disclosure

I built this with an AI assistant, working under my direction, and I would rather say so upfront.
The design decisions are mine, and claims about existing authentik behaviour — `Tenant.footer_links`,
the library, the icon conventions, the components reused — were checked against the codebase rather
than assumed. If the project has a policy on AI-assisted contributions, I will follow it.

The implementation is open as #25801.

---

PloxPlox:
Hello,

## What does this PR change?

Today an application card in the user dashboard offers one thing: a link that opens the service
in a browser. This adds an optional row underneath — small icons linking wherever the
administrator points them, and a button that copies the application's address.

The usual case is a native client. Jellyfin, Immich, Nextcloud and plenty of others ship mobile
and desktop apps, and right now nothing on the card tells a user they exist. The same row works
just as well for documentation, a status page, or a support channel. A native client also asks
for a server address before it will connect; that address is already on the card as the launch
URL, but only as somewhere to go, never as text you can copy.

Nothing is guessed. The administrator types a label, a URL, and picks an icon. authentik does not
infer the store from the hostname and does not ship a set of platform logos.

**Everything is configured per application**, in the application form, beside the switches that
already govern the card. An application that turns neither half on is unchanged — same markup,
same card dimensions, to the pixel.

### Additional context

before : 
Image
Image
after : 
Capture d&#39;écran 2026-09-06 185357
Capture d&#39;écran 2026-09-06 185445
Capture d&#39;écran 2026-09-06 190600

### What it touches, and what it cannot

I have tried to keep the blast radius small and easy to check.

| | |
|---|---|
| New model field | one, `Application.application_links`, a `JSONField` defaulting to `[]` |
| New migration | one, additive — no column is altered or dropped |
| New API endpoint | none |
| New dependency | none |
| `UIConfig` / `enabledFeatures` | untouched |
| Existing fields | untouched |
| Flows, providers, policies, outposts, sources | untouched |
| Behaviour with nothing configured | identical to today |
| How to undo it | clear the field, or drop it |

One shared component is touched: `ak-radio` gains an `inline` option, which selects
`pf-m-inline` instead of `pf-m-stack`. PatternFly already ships the modifier. It defaults to the
previous class, so every other radio group in the codebase renders exactly as before — the diff
there is one conditional and a property.

The library list cache is unaffected. It stores *which* applications a user may see; the rows
themselves are re-fetched from the database by `_expand_applications` before serialization, so an
edited link shows up immediately rather than after the 24-hour timeout.

Translations are left alone. `msg()` ids and translator notes are in place, and the nightly
`Translation - Extract and compile` job will pick the new strings up in its own PR rather than
having them ride along with this one.

## Why is this change needed?

When a self-hosted service has a mobile app, the administrator ends up being the distribution
channel: sending store links by hand, writing a wiki page, answering the same question every few
weeks. Users who go looking themselves often install the wrong thing, because searching "Element"
or "Jellyfin" in a store returns several unrelated results.

Refs #16776, *Allow non-http(s) urls for applications* (open, `enhancement/confirmed`). That issue
asks for `steam://` launch URLs so a native application can be started from the dashboard. It is
the same need approached from the other side, and the two fit together — see the note on URL
schemes below.

## On the storage question from #25800

The issue asked whether these should live in a `JSONField` on `Application` or in a dedicated
`ApplicationLink` model with a foreign key. This PR answers with the former, and the block has
since grown: it carries two switches, an optional heading and its alignment alongside the links.

Those settings are only ever read and written together, and nothing queries or orders by them, so
one additive column carries the lot rather than four columns and four migrations. If you would
rather see a related model, say so — the serializer is the only place that knows the shape, and I
will convert it.

## How was this tested?

Twelve tests in `authentik/core/tests/test_applications_api.py` cover the default value, a full
round-trip, an incomplete block being stored complete, the optional icon, a label being required
on `PATCH` as well as `PUT`, the rejected URL schemes and their case-insensitivity, a scheme-less
URL being completed to HTTPS, that completion declining to rescue a scheme we refuse, the rejected
alignment values, clearing the block, and reading a value stored in the earlier shape.

Writing them found two bugs, both now fixed, and both the same mistake at different depths. A
nested serializer takes `partial` from the **root** serializer, so during a `PATCH` every key
absent from the payload is skipped rather than defaulted. At the link level that accepted a link
with no label, stored it, and blew up on the next read. At the block level a payload of
`{"enabled": true}` was stored exactly like that — no links, no alignment, no address flag. Both
the block and each link are now validated by a serializer instantiated on the spot, which is its
own root, keeping `PUT` and `PATCH` under one set of rules.

`ModelSerializer` also refuses a writable nested serializer outright, whatever field it targets,
so the block is set aside before delegating to `super()` and written back afterwards.

The full Python suite runs green on this branch: **4147 passed, 26 skipped, 346 subtests**. The
nine errors it also reports are all in `authentik/sources/kerberos/`, and are the environment
rather than the branch: `k5test` shells out to `/usr/sbin/kdb5_util` while building its test
realm, and the `krb5-kdc` package that provides it is not installed on my machine, so each of the
nine fails with `FileNotFoundError` during setup, before any test body runs. Nothing in this
branch touches `authentik/sources/`.

Checked by hand on a local instance with six applications: with nothing configured, with links
only, with the copy button only, and with both. Also checked in the list view, below and above
the 768px breakpoint, and in a right-to-left locale.

The screenshots are light-theme only, but the new styles declare **no colour of their own** —
every value comes from tokens you already redefine per theme (`--pf-c-card--BackgroundColor`,
`--pf-global--BorderColor--100`, `--pf-global--Color--200`, `--ak-accent`,
`--pf-global--success-color--100`). Grepping the touched style blocks for hex, `rgb()` and
`hsl()` returns nothing, so dark mode follows by construction.

### The list view

`LibraryPage` forces `ViewMode.List` below 768px and ignores the stored preference:

```ts
this.#gridModeMatcher = window.matchMedia("(width > 768px)");

protected synchronizeViewModeWithMediaQuery(matches = this.#gridModeMatcher.matches) {
    if (!matches) {
        this.viewMode = ViewMode.List;
        return;
    }
    this.viewMode = LibraryPage.ViewModeStorage.read(ViewMode.Grid);
}
```

A feature drawn only on cards is therefore invisible on every phone and tablet, with no way for
the user to switch. The row is rendered in both views: in the list it takes a second grid line
under the application name, so the kebab menu keeps its column.

### The copy button

Safari only honours a clipboard write while it is still processing the gesture that triggered it,
so the address is resolved during render and never awaited before `writeText`. The fallback uses
a `contentEditable` range rather than a hidden `textarea`, which is the shape iOS accepts —
`select()` on a non-editable field copies nothing there. If both paths fail the tooltip keeps
showing the address, selectable, since nothing else on the card writes it out.

Hover styling is confined to `@media (hover: hover)`: on iOS a first tap on a hover-styled element
is sometimes spent simulating the hover, which would cost the user a second tap.

### A note on URL schemes

Link URLs are restricted to `http` and `https`, checked in the serializer rather than only in the
component. An administrator-supplied URL could otherwise carry `javascript:`, which on an identity
provider is stored XSS.

This does not stand in the way of #16776. The risk is `javascript:` and `data:`, not application
schemes — `steam://` cannot execute anything in the page. The check is one named constant,
`ALLOWED_LINK_SCHEMES`, so widening it is a one-line change. If anything a link row is a better
home for `steam://` than the launch URL, since it is explicitly secondary and the web launch stays
the primary action.

I could not find an equivalent check on `Tenant.footer_links`, so this is an addition rather than
a pattern I am copying. Tell me if you would rather it lived somewhere else, or covered footer
links too.

## How this was written

This was built with an AI assistant, working under my direction. I would rather say so plainly
than have you wonder halfway through the review.

What that means concretely:

- **The scope was deliberately kept small.** One field, one additive migration, no new endpoint,
  no new dependency. I wanted a change you can evaluate in one sitting and revert by clearing a
  field — not one that reaches into flows or providers.
- **Nothing in this description is assumed.** Every number above comes from a command that was
  actually run, and the output is quoted rather than summarised. Where something could not be
  verified on my machine, it says so instead of glossing over it.
- **The design calls are mine.** Refusing to derive labels from URLs, moving the configuration
  onto the application rather than onto the viewer, keeping the links centred while only the
  heading moves — those were my decisions, and in at least one case taken against the assistant's
  advice.

If your project would rather not take AI-assisted contributions, tell me and I will close this
without argument. I would rather ask than have it discovered later.

## Linked issues

closes #25800
refs #16776

---

## Checklist

- [ ] `make all` — see the note below
- [x] `make docs`

I ran the parts of `make all` rather than the target itself, so here is exactly what passed:

| Step | Result |
| --- | --- |
| `ruff check` | all checks passed |
| `bandit -iii` | clean |
| `mypy --strict` | no issues in 2241 source files |
| `ak makemigrations --check` | no changes detected |
| `gen` (schema + clients) | regenerated, `git diff --exit-code` clean on all five `test-gen` paths |
| `web` (prettier, eslint, `lint:types`) | clean |
| `test` | 4147 passed, 26 skipped, 346 subtests |

The box stays unticked for two reasons, both environmental. `cargo-deny` and `cargo-machete` need
a Rust toolchain I do not have installed; this branch changes no `Cargo.toml`, no `Cargo.lock` and
no `.rs` file, so I do not expect them to have anything to say. And the nine Kerberos errors above
come from the same kind of gap. CI is better placed than I am to confirm both.

---


dewi-tik:
Thank you for this - we're reviewing and considering options. But we like the idea. Thanks

PloxPlox:
Thanks! And no worries if it's easier for you to just push changes directly on the branch rather than going back and forth — whatever's simplest on your end.

PR #25802, opened 2026-09-05
core: remove legacy user and group attribute filters

dominic-r:
Filtering users or groups by a JSON attribute, such as `department=engineering`, loads every matching object and its related objects before pagination. It then throws that data away and runs another query for the requested page. Asking for 100 users can therefore load all 10,000 matches first.

This replaces `len(queryset.filter(...))` with `.exists()` in the user and group attribute filters. The preliminary check becomes `SELECT 1 … LIMIT 1`, avoiding loading those objects and their relations.

**The endpoint still returns the full requested page and all its fields.** The boolean from `.exists()` is discarded. The next line still returns the same filtered queryset, and the existing permissions and pagination still apply. Requests without an attribute filter are unaffected.

The preliminary check remains because some invalid lookups only raise `ValueError` when Django compiles the query. For example, a non-boolean `department__isnull` value must still reach the existing handler, which returns the original queryset. A regression test covers that behavior as well as malformed JSON, non-object JSON, matching results, and no matches.

### Before and after

Local fixture: 10,000 users, 10,000 groups, and 10,000 memberships, with one user per group and approximately 1 KB of attributes per object. All 10,000 objects match the filter. Times are medians of three warmed Django API-client requests against dev PostgreSQL, including response rendering but excluding network/browser time. The tenant's maximum page size was temporarily raised to 1,000 for the larger requests.

| Requested page | Before | After | Speedup |
|---|---:|---:|---:|
| 100 users | 28.40 s | 416 ms | 68× |
| 1,000 users | 30.62 s | 4.12 s | 7.4× |
| 100 groups | 1.54 s | 93 ms | 17× |
| 1,000 groups | 2.02 s | 541 ms | 3.7× |

These are measurements for this fixture, not a guarantee for every deployment. Original and changed 100-object API responses matched byte-for-byte for both users and groups on the same fixture. The fixture and page-size setting were rolled back afterward.

### Validation

- The query-loading regression fails on the original code for both filters.
- `make test` for the user API, group API, and attribute filters: 68 tests and 16 subtests pass. Used a fresh `COVERAGE_FILE` to exclude stale local coverage data.
- Black, Ruff, and the pending-migration check pass.

Made with GPT 6 Astra in Codex, which investigated the bottleneck, wrote the code and tests, ran the benchmarks and checks, and drafted this description under maintainer direction.

### DIFF
--- a/authentik/core/api/groups.py
+++ b/authentik/core/api/groups.py
@@ -341,7 +341,7 @@ class GroupFilter(FilterSet):
         for key, _value in value.items():
             qs[f"attributes__{key}"] = _value
         try:
-            _ = len(queryset.filter(**qs))
+            queryset.filter(**qs).exists()
             return queryset.filter(**qs)
         except ValueError:
             return queryset
--- a/authentik/core/tests/test_attribute_filters.py
+++ b/authentik/core/tests/test_attribute_filters.py
@@ -1,0 +1,49 @@
+"""Attribute filtering must not load all matches before pagination."""
+
+from django.test import TestCase
+from rest_framework.exceptions import ValidationError
+
+from authentik.core.api.groups import GroupFilter
+from authentik.core.api.users import UsersFilter
+from authentik.core.models import Group, User
+from authentik.core.tests.utils import create_test_user
+from authentik.lib.generators import generate_id
+
+
+class TestAttributeFilters(TestCase):
+    """Check filtering without materializing matching objects or their relations."""
+
+    def test_filter_before_pagination(self):
+        user = create_test_user(attributes={"department": "engineering"})
+        group = Group.objects.create(name=generate_id(), attributes={"department": "engineering"})
+        group.users.add(user)
+        for filter_class, model, relation, expected in (
+            (UsersFilter, User, "groups", user),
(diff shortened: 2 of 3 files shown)


authentik/core/tests/test_attribute_filters.py
```diff
@@ -0,0 +1,49 @@
+"""Attribute filtering must not load all matches before pagination."""
```
gergosimonyi, line 1: What reason is there to test these?

BeryJu (commented):
dont like fixed query count assertions in tests, thresholds are much better (and tests can maybe be a bit more minimal), change itself is good

authentik/core/api/groups.py
```diff
@@ -341,7 +341,7 @@ def filter_attributes(self, queryset, name, value):
         for key, _value in value.items():
             qs[f"attributes__{key}"] = _value
         try:
-            _ = len(queryset.filter(**qs))
+            queryset.filter(**qs).exists()
             return queryset.filter(**qs)
         except ValueError:
             return queryset
```
BeryJu, line 347: actually maybe just remove this, AKQL is OSS which has much better filtering support for attributes on users & groups

PR #25808, opened 2026-09-05
flows: count cache keys in the database

dominic-r:
Flow cache statistics currently download every matching cache key and then count the strings in Python. This asks PostgreSQL for the count directly, so the response from the database is one number regardless of how many keys match.

The cache backend owns the matching rules. A shared private queryset method keeps `keys()` and the new `count_keys()` consistent; the flow serializer and cache-info endpoint use the latter. Cache clearing still uses the existing key-list API. Exact keys, prefix wildcards, complex patterns, key versions, and inclusion of expired entries keep their existing behavior. No flow execution or cache expiration policy changes.

This is a modest optimization, not a dramatic endpoint speedup. Local timings measure one count operation against PostgreSQL, excluding HTTP and browser rendering. Both versions make one query; the improvement is avoiding key transfer and Python list creation. Counts matched at every size. Figures are medians of three runs after warmup through 1,000 keys; 10,000 is one stress run per version. Fixtures were rolled back.

| Rows | Before | After | Queries before → after |
|---:|---:|---:|---:|
| 50 | 0.72 ms | 0.58 ms | 1 → 1 |
| 100 | 0.68 ms | 0.65 ms | 1 → 1 |
| 500 | 1.26 ms | 1.24 ms | 1 → 1 |
| 1,000 | 1.35 ms | 1.13 ms | 1 → 1 |
| 10,000 | 4.93 ms | 2.16 ms | 1 → 1 |

Validation: 32 focused cache, flow API, and planner tests passed, plus four subtests. The new backend test covers exact/prefix/complex/missing matches, versions, and expired entries against the database. An existing flow API test now checks the returned cache count. No query-budget assertions were added. Black, Ruff, and `make gen` passed, with no generated API changes. Broader CI remains to run.

Made with GPT 6 Astra, which assisted with investigation, implementation, tests, benchmarks, and this description.

### DIFF
--- a/authentik/flows/api/flows.py
+++ b/authentik/flows/api/flows.py
@@ -57,7 +57,7 @@ class FlowSerializer(ModelSerializer):
 
     def get_cache_count(self, flow: Flow) -> int:
         """Get count of cached flows"""
-        return len(cache.keys(f"{cache_key(flow)}*"))
+        return cache.count_keys(f"{cache_key(flow)}*")
 
     def get_export_url(self, flow: Flow) -> str:
         """Get export URL for flow"""
@@ -123,7 +123,7 @@ class FlowViewSet(UsedByMixin, ModelViewSet):
     @action(detail=False, pagination_class=None, filter_backends=[])
     def cache_info(self, request: Request) -> Response:
         """Info about cached flows"""
-        return Response(data={"count": len(cache.keys(f"{CACHE_PREFIX}*"))})
+        return Response(data={"count": cache.count_keys(f"{CACHE_PREFIX}*")})
 
     @permission_required(None, ["authentik_flows.clear_flow_cache"])
     @extend_schema(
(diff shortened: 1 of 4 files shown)


BeryJu (changes requested):
Makes sense

the only I dont like is adding custom methods to the cache backend since `django.core.cache`'s `cache` object is not typed at all

authentik/flows/api/flows.py
```diff
@@ -57,7 +58,9 @@ class FlowSerializer(ModelSerializer):
 
     def get_cache_count(self, flow: Flow) -> int:
         """Get count of cached flows"""
-        return len(cache.keys(f"{cache_key(flow)}*"))
```
rissson, line 60: I'd rather add a `keys_count` method on the cache backend.

rissson (changes requested):

authentik/flows/api/flows.py
BeryJu: thats what it was and I what I voted against since the `django.core.cache.cache` object is un-typed I dont like adding custom methods to it, but I also don't like directly calling the ORM here

rissson: I mean, `keys` already isn't standard

PR #25858, opened 2026-09-08
feat(ldap-outpost): Render map[string]any as proper ldap attrs

### LINKED ISSUE #16954: Nested user attributes not exposed as valid LDAP attributes in LDAP Outpost
**Describe the bug**
When exposing custom attributes via the LDAP outpost, nested attributes are not returned as individual LDAP attributes. Instead, they appear in a serialized map format that does not seem to comply with LDAP.

**To Reproduce**

1. Configure a user in Authentik with a custom attribute containing nested values, e.g. settings.locale = en.
2. Query the user via the LDAP outpost using ldapsearch.
3. Inspect the LDAP response.

See that the attribute is returned as a serialized map string rather than as a proper multi-valued LDAP attribute.

**Expected behavior**
I would expect the LDAP outpost to flatten or map the nested attribute into standard LDAP attributes, for example:

```
settingsLocale: en
```

or

```
settings.locale: en
```

instead of:

```
settings: map[locale:en]
```

-   authentik version: 2025.8.3
-   Deployment: Docker

**Additional context**

josegomezr:
## Details

When finding map[string]any values, instead of generating an ldap value with the string representation of the go map value it generates ldap keys in the form of:

```
origKey--subKey: value
```

These new "virtual" keys are also searchable as the LDAP search filters are executed after the entry generation phase.

### What does this PR change?

TL;DR:

Old:

```
address: map[country:DE locality:foo postalCode:123456 state:bar street:baz]
```

New:
```
address--country: DE
address--locality: foo
address--postalCode: 123456
address--state: bar
address--street: baz
```

### How was this tested?

Unit tests are included in the PR

### Linked issues

Closes: goauthentik/authentik#16954

---

## Checklist

No AI was used in this changeset.

### DIFF
--- a/internal/outpost/ldap/utils/utils.go
+++ b/internal/outpost/ldap/utils/utils.go
@@ -56,6 +56,18 @@ func AttributesToLDAP(
 	for attrKey, attrValue := range attrs {
 		entry := &ldap.EntryAttribute{Name: keyFormatter(attrKey)}
 		switch t := attrValue.(type) {
+		case map[string]any:
+			// When finding a nested map: generate prefixed entries in the form of
+			// key--subkey.
+			prefixedFormatter := func(key string) string {
+				return keyFormatter(attrKey + "--" + key)
+			}
+			nestedAttrs := AttributesToLDAP(t, prefixedFormatter, valueFormatter)
+			for _, attr := range nestedAttrs {
+				if len(attr.Values) > 0 {
+					attrList = append(attrList, attr)
+				}
+			}
 		case []string:
 			entry.Values = valueFormatter(t)
 		case *[]string:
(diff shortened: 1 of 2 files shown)


josegomezr:
Grafted from SUSE/authentik#89

rissson (approved):
I'm happy with this, but I'm unsure about the `--` separator. cc. @BeryJu

josegomezr:
> I'm happy with this, but I'm unsure about the `--` separator. cc. @BeryJu

If it's worth, there's some method to the madness. The reason for `--` is to obviously differentiate what attributes are faked from the outpost.

It's more likely than an ldap attribute is camelcased than having dashes, let alone double dashes.

Issue #18792, opened 2025-12-12
Dry-run policiesd


rissson:
### Is your feature request related to a problem?

Ability to dry-run policies to try out changes/new policies

### Describe the solution you'd like

Store the policy result data into the flow context (even if the policy is not in dry-run)

Allows inspection of the data from the inspector _after_ the policy ran, and not only of the policy result.
Allows custom expression policies from that data.

### Describe alternatives that you've considered

`ak_call_policy` from an expression policy. Does not give access to the policy data.

### Additional context

ref https://github.com/goauthentik/internal-customer-ref/issues/2

jsachs:
## Proposal
Add binding-level dry-run support for policies. Dry-run policies execute normally but are excluded from the effective ALL/ANY decision and from user-facing aggregate messages.

Individual results from all policy bindings, including non-dry-run bindings, are stored in flow context under `policy_results`, exposing `passing`, `messages`, and `raw_result` to later expression policies and the flow inspector. The implementation covers all three policy engines, flow-plan caching and re-evaluation, API/client generation, Admin UI configuration, documentation, and cache invalidation.

PR #25989, opened 2026-09-09
policies: fix policybinding cache invalidation

jsachs:
### What does this PR change?
Invalidate cached policy results when either a policy or a policy binding changes. Previously, updated or deleted bindings would not invalidate the policy cache.

This prevents stale `negate` and `failure_result` behavior. It is necessary to avoid creating stale `dry_run` behavior when implemented.

### How was this tested?
Signal and engine tests updated.

You can also test manually by doing the following in your local dev environment:
1. Create a basic application
2. Create an Expression Policy binding with `return True` for the new application.
3. Go to http://localhost:9000/if/user/#/library in another tab. You should see the new application. Now the cache is warm.
4. Return to the policy binding and enable Negate Result.
5. Immediately refresh the other tab. The application should immediately disappear.
6. You can test the same behavior by deleting the policy binding instead of negating it.

### Linked issues
https://github.com/goauthentik/authentik/issues/18792

---

## Checklist

  - Most of this work was done in OpenCode with GPT 5.6 Sol.
  - Code was reviewed line-by-line to confirm my understanding of my own changes and each AI-assisted change.


authentik/policies/signals.py
```diff
@@ -41,14 +44,15 @@ def invalidate_policy_cache(sender, instance, update_fields=None, **_):
     if sender == User and update_fields and set(update_fields) <= {"last_login"}:
         return
 
-    if sender == Policy:
+    if sender in (Policy, PolicyBinding):
+        bindings = PolicyBinding.objects.filter(policy=instance) if sender == Policy else [instance]
```
BeryJu, line 48: this is probably the main reason this isn't already done, trying to avoid queries in signals (not too important for this, its not really in the hot path)

authentik/policies/signals.py
```diff
@@ -41,14 +44,15 @@ def invalidate_policy_cache(sender, instance, update_fields=None, **_):
...
         total = 0
-        for binding in PolicyBinding.objects.filter(policy=instance):
-            prefix = f"{CACHE_PREFIX}{binding.policy_binding_uuid.hex}_{binding.policy.pk.hex}*"
-            keys = cache.keys(prefix)
+        for binding in bindings:
+            prefix = f"{CACHE_PREFIX}{UUID(str(binding.policy_binding_uuid)).hex}_*"
```
BeryJu, line 51: `binding.policy_binding_uuid` should already be of the UUID type, shouldn't need to cast it to str and then to UUID again to get hex

authentik/policies/signals.py
jsachs: Sorry if I'm missing something, can you explain to me how this is adding a performance-sensitive query?

`PolicyBinding.objects.filter(policy=instance)` was previously present at line 46.

authentik/policies/signals.py
jsachs: I did this because there was a test failure in `test_export_validate_import_policies` when I was running my tests locally. I asked Codex to help me with this one since I was pretty unclear why it was happening.

Since a `PolicyBinding` save now uses the existing `instance` directly, we'd get the following behavior:
1. blueprint assigns UUID (`string`) to the model instance
2. django writes it to the database
3. `post_save` gets the same instance, which has yet to be reloaded/coerced
4. `.hex` is called on this UUID which is not yet an actual `UUID`

authentik/policies/signals.py
BeryJu: ah ok then I misread the diff, my bad

authentik/policies/signals.py
BeryJu: ok valid, I'd rather fix this in the blueprint importer than have a special case for it here

jsachs: Ack, updated.

BeryJu (approved):

PR #25997, opened 2026-09-10
lib/sync/outgoing: prevent duplicate group creation in outgoing providers

ryanpesek:
### Summary
Prevent concurrent outgoing provisioning tasks from creating duplicate remote objects and duplicate local provisioned-object records.
Rapid consecutive changes, such as creating a group and immediately adding members, can schedule overlapping synchronization tasks. Both tasks could observe that no provider connection existed and independently create the same object remotely.

### Changes
- Add a tenant-, provider-, and object-scoped PostgreSQL advisory lock around outgoing writes.
- Hold the lock through remote creation/update, local connection persistence, and initial group membership synchronization.
- Use the same lock for subsequent membership changes.
- Retry tasks when the 30-second lock acquisition timeout is reached.
- Preserve Entra and Google Workspace post-write membership behavior inside the lock.
- Reject unknown membership task actions.
- Add a concurrent SCIM group regression test that verifies:
- Only one create operation occurs.
- The second writer waits for the first.
- Both writers converge on one local connection.
 
### Scope
The locking is implemented in the shared outgoing synchronization client, so it applies to SCIM, Microsoft Entra, and Google Workspace users and groups.
This change does not modify or clean up existing duplicate records.

### Testing
- New SCIM concurrency tests: 2 passed
- SCIM, Microsoft Entra, and Google Workspace suites: 101 passed
- Black and Ruff checks passed

---

## Checklist

### DIFF
--- a/authentik/enterprise/providers/google_workspace/clients/groups.py
+++ b/authentik/enterprise/providers/google_workspace/clients/groups.py
@@ -97,8 +97,8 @@ class GoogleWorkspaceGroupClient(
             # Resource missing is handled by self.write, which will re-create the group
             raise
 
-    def write(self, obj: Group):
-        google_group, created = super().write(obj)
+    def _write(self, obj: Group):
+        google_group, created = super()._write(obj)
         self.create_sync_members(obj, google_group)
         return google_group, created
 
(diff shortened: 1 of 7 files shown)


BeryJu (commented):
looks good, I'll probably refactor this a bit so it doesn't need to touch enterprise code

PR #26016, opened 2026-09-10
policies: expose policy results in flow context

jsachs:
### What does this PR change?
- expose policy binding results to later flow policies through `context["policy_results"]`
- preserve results from planning and runtime policy evaluation for flow inspection
- snapshot policy output before retaining it in flow context
- document the policy result context contract for expression policies

- in order for dry runs to provide meaningful information, policy binding results must be stored somewhere

### How was this tested?
- flow planner and inspector tests
- policy process tests
- TODO: manual testing?

### Known review topics
- make planning-stage results consistent between cold and warm flow-plan cache paths
- bound the traversal and retained size of `raw_result`
- avoid repeatedly copying accumulated results into flow history
- clarify skipped terminal-stage representation in the inspector

### Linked issues
https://github.com/goauthentik/authentik/issues/18792

---

## Checklist


jsachs:
@BeryJu a few design questions I'm bumping into here:
1. It sounds like we need `policy_results` to become a reliable input to later policies. In addition to `dry_run` support, we're also trying to support expression policies consuming this data. This means the current cache behavior won't work. It sounds like we might need to re-evaluate the planning stage policies if there is a cache it, but this means we effectively rebuild the plan and remove a lot of the value of having the cache in the first place.
2. We need to decide on some sort of consistent way to bound the `raw_result`, otherwise the cache could potentially get very large. This also leads to a general concern of these results enlarging a session.
3. I think we want a new `policy_results` for each newly planned execution? I believe right now existing entries would be merged, leaving behind deleted/disabled bindings.
4. What do we want to do about failed flow-entry attempts? Right now the snapshot comes before the `result.passing` check. Maybe we should move it after this check?

@dewi-tik may also have some ideas?

PR #25976, opened 2026-09-09
website/docs: improve radius provider docs

dewi-tik:
### Linked issues

Closes https://github.com/goauthentik/authentik/issues/25324

---

## Checklist

-   [x] The project has been linted, built, and tested (`make all`)
-   [x] The documentation has been updated and formatted (`make docs`)
-   [x] I have read the [AI usage policy](https://github.com/goauthentik/authentik/blob/main/AI_POLICY.md).

PR #26029, opened 2026-09-10
providers/radius: allow returning additional properties, continue on failed mapping

BeryJu:
Currently in radius property mappings you have to do `packet["attr"] = "value"` which is a bit clunky and doesn't match any of the other property mappings. With this PR you can also now do `return {"attr": "value"}`.

Also currently radius stops when a single mapping fails, to align this with oauth/saml, it now continues and just logs the failure

ref https://github.com/goauthentik/authentik/pull/25976


dewi-tik (approved):
Docs LGTM

PR #26030, opened 2026-09-10
web/test: fix two e2e locators that assume a settled UI state

GirlBossRush:
## Details

This PR fixes two failures in the browser suite where a locator assumes the UI has settled into one particular state.


abhisheksharma2411 (commented):
Both diagnoses check out, and the second one especially — I traced it rather than take the comment's word.

## The `New Provider` duplication is real

`Table` renders the create control from two places:

```
Table.ts:607  renderEmpty()      → :614  <div slot="primary">${this.renderObjectCreate()}</div>
Table.ts:912  renderToolbar()    → :913  ${this.renderObjectCreate()}
                                   :959  <div class="pf-c-toolbar__content" part="toolbar-secondary">
                                   :961      ${this.renderToolbar()} …
```

and `renderEmpty()` is what's on screen while rows load (`:690`). So during the loading window there genuinely are two buttons with the accessible name `New Provider`, and Playwright strict mode fails on the ambiguity — which is exactly why it only bites when the request is slow, and why it would look like flake rather than a locator bug.

Scoping to `[part="toolbar-secondary"]` is the right fix: it's the part attribute on the container that wraps `renderToolbar()`, so it selects the persistent control rather than the transient one. A `.first()` would also have made the error go away and would have silently picked whichever rendered first — worse, because the empty-state button is the one that disappears.

## The remember-me change

The insight that remember-me *submits* the identification stage rather than only pre-filling it is the part that makes the old code wrong rather than merely fragile — a wait for `$identificationStage` after a flow has already advanced past it can only ever time out.

Restructuring to branch on `isVisible()` and converge on the password stage reads correctly to me, and the `ak-stage-identification, ak-stage-password` first-match in `SessionFixture` is the same shape applied to the fixture.

## One question

Both fixes are timing-dependent by nature, so: did you get a failing run before and a passing run after, or is this reasoned from the code? Not a challenge — the reasoning stands on its own and I'd take it either way. But for a flake fix the before/after is the evidence that matters most, and if you have the failing output it's worth pasting, because the next person to see these locators will want to know whether they were *observed* to fail or *predicted* to.

LGTM.

kensternberg-authentik (approved):

PR #26052, opened 2026-09-11
lifecycle/charts: init

BeryJu:
move charts back into monorepo

todo:
- test release CI
- create S3 bucket
- migrate existing repo's charts to s3


.github/workflows/release-publish.yml
```diff
@@ -152,6 +152,51 @@ jobs:
...
+      - name: Download release container artifact
+        run: |
+          gh release download "helm-build-${{ matrix.name }}" --pattern "${{ matrix.name }}.tgz"
+          gh attestation verify "${{ matrix.name }}.tgz" --owner goauthentik
+      - id: publish-s3
+        name: Push to S3
+        run: |
+          helm plugin install https://github.com/hypnoglow/helm-s3.git
+          helm repo add goauthentik s3://TODO-NAME
+          helm s3 push ./${{ matrix.name }}*.tgz goauthentik
```
rissson: Does this only update the index.yaml? It should do so, and set the URL to the tgz to the release artifact

Makefile
```diff
@@ -325,6 +331,26 @@ docker:  ## Build a docker image of the current source tree
 test-docker:
 	BUILD=true ${PWD}/scripts/test_docker.sh
 
+helm-docs:
+	docker run \
+		--rm \
+		-v "${PWD}:/helm-docs" \
+		-u $(UID) \
+		docker.io/jnorwood/helm-docs:v1.12.0 \
```
rissson: Do we want to put those in a compose.yml file? So we get dependabot updates

.github/workflows/release-publish.yml
BeryJu: I thought we were gonna put the tgz in the bucket too?

rissson: I wouldn't? It's extra cost for us, when we can have github serve it for us

BeryJu: How would we backfill old releases?

rissson: We don't. Old releases already link to helm chart repo github releases for the tarballs, they're not in the gh-pages branch

BeryJu: fair, but yeah this plugin will push the tgz and update the index file

rissson: ah yeah, and there is no way to setup a custom URL apparently

PR #26066, opened 2026-09-11
web/user: fix enterprise check on user library

### LINKED ISSUE #25815: Agents tab buttons show briefly in the user interface before the enterprise license check hides them.
### Describe the bug

On unlicensed installations in 2026.8.1, the Agents tab button on the user interface displayed briefly, then hides, during the webUI load.

It looks related to this recently implemented PR: web/user: fix missing enterprise check on agents tab in user interface - #25371

### How to reproduce

1. Run an unlicensed version
2. Load the user interface and watch the top nav bar

I tested using Safari on MacOS.

### Expected behavior

The agent buttons should not display at all, even briefly, on unlicensed installations.

### Screenshots

_No response_

### Additional context

_No response_

### Deployment Method

Docker

### Version

2026.8.1

### Relevant log output

```shell

```

kensternberg-authentik:
web/bug/25815: fix display of capabilities tabs on the User Applications page

## What

Fixes a visible race condition on the User Applications page

The test for license status would allow the “Agents” tab to be shown briefly when the `licenseSummary.status` was undefined. This looks like a race condition; sometimes the licensing information hadn’t arrived when the display rendered. This commit extracts the predicate into a function and makes it positive (“You *can* do this”) rather than negative.

Also: separates copy/pasta machinery from the intent. 🗣🔥🔥🔥

Closes #25815

## Checklist

-   [🐜] The project has been linted, built, and tested (`make all`)
-   [🐜] The documentation has been updated and formatted (`make docs`)
-   [🐜] I have read the [AI usage policy](https://github.com/goauthentik/authentik/blob/main/AI_POLICY.md).


web/src/user/ak-interface-user.ts
```diff
@@ -123,6 +133,23 @@ class UserInterface extends WithLicenseSummary(
         });
     }
 
+    protected canUseNavItem = ({ feature, capability }: UserNavItem) => {
```
kensternberg-authentik, line 136: Arrowed to prevent having to write `(navItem) => this.canUseNavItem(navItem)` in the `render()` pass. We're gonna enclose `this` somewhere for this feature, it may as well be here. I think it makes the `filter()` expression read more cleanly, anyway.

web/src/user/ak-interface-user.ts
```diff
@@ -142,25 +169,27 @@ class UserInterface extends WithLicenseSummary(
...
-            navItems.push({ label: msg("Agents"), link: toUserInterface("agents") });
-        }
+
+        const CE = CapabilitiesEnum;
+        const navItems: UserNavItem[] = [
+            { label: msg("Applications"), link: toUserInterface("library") },
+            {
+                label: msg("Discover"),
+                link: toUserInterface("requests"),
+                feature: "requests" as Feature,
```
kensternberg-authentik, line 183: If you hover over `feature:`, VSCode gives you a list of valid strings, and "requests" and "agents" are there, but for some reason Typescript _still_ needs to be told these are the feature keys.

abhisheksharma2411 (commented):
The race diagnosis is right, and the mechanism is worth naming precisely because it's the kind that survives review: the old guard was a **negative** test.

```ts
this.licenseSummary?.status !== LicenseSummaryStatusEnum.Unlicensed
```

`undefined !== Unlicensed` is `true`, so an unresolved license summary passed the check and the tab rendered until the real status arrived. Replacing "not the bad one" with "one of the good ones" is what actually fixes it — any allowlist would have been immune here, and no amount of extra `?.` would have been.

## This also fixes a second case, which the description doesn't claim

The old condition admitted **four** statuses — `Valid`, `ExpirySoon`, `Expired`, and `undefined`. The new one admits two. So `Expired` stops showing the enterprise tabs as well, not just the transient undefined.

That's a fix rather than a side effect, and I'd say so in the description, because it's the change most likely to be read as a regression by whoever sees an expired-license instance lose its Agents tab. The new allowlist is **exactly** the backend's own definition:

```python
# authentik/enterprise/models.py:78-80
@property
def is_valid(self) -> bool:
    """Quickly check if a license is valid"""
    return self in [LicenseUsageStatus.VALID, LicenseUsageStatus.EXPIRY_SOON]
```

Frontend and backend now agree on what "licensed" means, where before the UI was more permissive than the API it's gating. Worth a line in the PR body — "aligns with `LicenseUsageStatus.is_valid`" is a stronger justification than "fixes a race", and it tells the next person where the pair lives.

## Two small things

**`canUseNavItem` returns `boolean | undefined`.** The expression starts with `status &&`, so when `licenseSummary` is unresolved it evaluates to `undefined`, not `false`. `Array.prototype.filter` treats that as falsy so the behaviour is right, but the predicate's type is wider than its name implies. `Boolean(status) && …` or `status !== undefined && …` keeps it a true predicate if that matters under your TS settings.

**The `as Feature` casts look unnecessary.** `feature` is typed `Feature` on `UserNavItem` and `"requests"` / `"agents"` are literal keys of `enabledFeatures`, so the annotation on `navItems` should already narrow them. If they're there to defeat widening, fine — but if they're vestigial, dropping them means a typo in a feature name is caught by the compiler instead of asserted past.

Neither blocks. Nice cleanup — turning three parallel `if (…) navItems.push(…)` blocks into data plus one predicate is the change that makes the next nav item a one-line addition instead of a fourth copy of the condition.

BeryJu (approved):

Issue #13792, opened 2025-04-07
Unable to bind LDAP connection when using TLS certificate as identity


anshulkamath:
**Describe the bug**

When we do not provide a `user` and `password` to our LDAP source, we get errors when invoking the LDAP protocol. 

**To Reproduce**

1. Find an LDAP source that binds using SASL
2. Add the LDAP source to Authentik
3. Do not add a user or password (because we are using the TLS connection as the identity)
4. Set up the source as normal otherwise
5. Try and sync the source to authentik

**Expected behavior**

Authentik should know which type of LDAP connection to make to the LDAP server based on the provided credentials (or absence thereof)

**Screenshots**

**Logs**
Output of docker-compose logs or kubectl logs respectively

**Version and Deployment (please complete the following information):**

-   authentik version: 2025.2.2
-   Deployment: helm

**Additional context**

anshulkamath:
Possibly fixed by https://github.com/goauthentik/authentik/pull/13791

PR #26067, opened 2026-09-11
sources/ldap: support SASL external

jsachs:
## Details

SASL was previously explored here: https://github.com/goauthentik/authentik/issues/13792

### What does this PR change?
Add `SASL EXTERNAL` as an LDAP source service bind method. This means that sync and writeback operatinos can be performed using TLS client cert verification.

User verification using a password still depends on `SIMPLE` binds.

This is a standard LDAP auth mechanism, which is useful for sync/writeback operations as the existing TLS connection can be reused, rather than needing to rely on a user-level bind with an admin password.

### How was this tested?
- unit tests added
- e2e testing with an openldap container added
- screenshot for new admin interface:
image

## Checklist


jsachs:
@dewi-tik are there any other e2e tests you'd like me to consider in addition to the openldap container?

Also, I'm not sure if I made the correct UI decision here. I could see this also being a toggle for now.

tests/e2e/test_source_ldap_sasl_external.py
```diff
@@ -0,0 +1,241 @@
...
+    LDAPSourcePropertyMapping,
+)
+from authentik.sources.ldap.password import LDAPPasswordChanger
+from authentik.sources.ldap.sync.users import UserLDAPSynchronizer
+from authentik.tasks.models import Task
+from tests.live import E2ETestCase
+
+OPENLDAP_IMAGE = (
+    "docker.io/osixia/openldap:2.6.10-alpha@"
+    "sha256:80a577d7d4471c4db662195111e5709665b6fcbd6679094d05402b8c620e2607"
```
BeryJu: small nit, prefer ghcr images if possible

tests/e2e/test_source_ldap_sasl_external.py
```diff
@@ -0,0 +1,241 @@
...
+        )
+        self.client_keypair = CertificateKeyPair.objects.create(
+            name=generate_id(),
+            certificate_data=certificate_files["client.crt"].decode(),
+            key_data=certificate_files["client.key"].decode(),
+        )
+
+        fixtures = Path(__file__).parent / "fixtures" / "openldap-sasl"
+        self.openldap = self.run_container(
+            image=OPENLDAP_IMAGE,
```
BeryJu: this was changed after you started this PR, but there's `self.pinned_image` now, instead of defining the image in the code add it to the e2e compose file instead, so we can get dependabot to update it.

BeryJu (commented):
other than those two things, lgtm

tests/e2e/test_source_ldap_sasl_external.py
jsachs: Ack. I think the only well-supported GHCR image for openldap is this one, and I don't think it has the configuration required for this use case: https://github.com/rroemhild/docker-test-openldap/pkgs/container/docker-test-openldap

BeryJu (approved):

tests/e2e/test_source_ldap_sasl_external.py
BeryJu: fair

PR #26081, opened 2026-09-12
packages/ak-common/config: fix list values from env variables parsing (#7836)

### LINKED ISSUE #7836: Make https listener (9443) optional
If you are deploying authentik using a reverse proxy, there is no point in having a https listener running ad the revers epoxy is taking care of all https related things.

Two posible solutions:
- Add an option to disable the https layer.
- Make the https listen address nullable. 

Just ignoring the additional port. But thats not that clean.

chilikk:
## Details

Fixed the issue stated in [#7836](https://github.com/goauthentik/authentik/issues/7836#issuecomment-5382201554) by dropping empty list values when parsing a comma-separated list from an environment variable.

### What does this PR change?

Empty values are dropped from the list before attempting to parse each value.

Fixes a regression in 2026.8 where the HTTPS listener can no longer be disabled by setting the env variable to the empty value. This worked find in 2026.5.6

### How was this tested?

Uncommented an already existing test case in ak-common/config.

### Linked issues

closes #7836 

---

## Checklist

### DIFF
--- a/website/docs/install-config/configuration/configuration.mdx
+++ b/website/docs/install-config/configuration/configuration.mdx
@@ -399,6 +399,8 @@ List of comma-separated `address:port` values for HTTPS.
 
 Applies to the Server and Proxy outposts.
 
+Set to an empty value to disable the HTTPS listener.
+
 Defaults to `[::]:9443`.
 
 ##### `AUTHENTIK_LISTEN__LDAP`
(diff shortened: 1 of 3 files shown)


website/docs/install-config/configuration/configuration.mdx
```diff
@@ -399,6 +399,8 @@ List of comma-separated `address:port` values for HTTPS.
 
 Applies to the Server and Proxy outposts.
 
+Set to an empty value to disable the HTTPS listener.
```
rissson, line 404: Please add this comment for the HTTP and metrics config options as well.

rissson (changes requested):
a small docs nit, but otherwise LGTM

rissson (approved):

PR #26095, opened 2026-09-14
providers/oauth2: use managed secrets for OAuth and proxy providers

dominic-r:


authentik/crypto/secrets/migrations/0002_preserve_role_permissions.py
```diff
@@ -0,0 +1,129 @@
...
+    permissions = {}
+    for codename, name in [
+        ("view_secret", "Can view Secret"),
+        ("change_secret", "Can change Secret"),
+        ("rotate_secret", "Rotate secret's value"),
+        ("view_secret_value", "View secret's value"),
+    ]:
+        permissions[codename], _ = Permission.objects.using(alias).get_or_create(
+            content_type=secret_type, codename=codename, defaults={"name": name}
+        )
```
BeryJu, line 56: there's a command to force permission creation, `from django.contrib.auth.management import create_permissions`, see `authentik/providers/ldap/migrations/0004_alter_ldapprovider_options_and_more.py`

authentik/crypto/secrets/migrations/0002_preserve_role_permissions.py
```diff
@@ -0,0 +1,129 @@
...
+        secret_fields = [
+            field.attname for field in Model._meta.fields if field.related_model is Secret
+        ]
+        content_type = (
+            ContentType.objects.using(alias).filter(app_label=app_label, model=model_name).first()
+        )
+        if not content_type:
+            continue
+        for action in ["view", "change"]:
+            grants = {"view_secret"}
```
BeryJu, line 68: I think for migration permissions if you have change perms they should get migrated to view_secret, nothing should migrate to change_secret/rotate_secret

cc @gergosimonyi

authentik/crypto/secrets/models.py
```diff
@@ -78,6 +78,12 @@ def validate_value(self, value: str) -> None:
                 b64decode(value, validate=True)
             except (BinasciiError, ValueError) as exc:
                 raise ValidationError(_("Value must be base64-encoded.")) from exc
+        if self._state.adding:
+            return
+        if self.oauth2_providers.exists():
+            from authentik.providers.oauth2.utils import validate_client_secret
+
+            validate_client_secret(value)
```
BeryJu, line 86: I think this should be a signal

authentik/providers/oauth2/models.py
```diff
@@ -236,11 +246,14 @@ class OAuth2Provider(WebfingerProvider, Provider):
         verbose_name=_("Client ID"),
         default=generate_id,
     )
-    client_secret = models.CharField(
-        max_length=255,
-        blank=True,
+    secret = models.ForeignKey(
```
BeryJu, line 249: `client_secret_ref` imo

authentik/providers/proxy/models.py
```diff
@@ -137,9 +141,25 @@ class ProxyProvider(OutpostModel, OAuth2Provider):
         blank=True,
     )
 
-    cookie_secret = models.TextField(default=get_cookie_secret)
+    cookie_secret_ref = models.ForeignKey(
```
BeryJu, line 144: Changing this secret will invalidate all sessions
Either
- Show warning in UI when rotating
- Have M2M with only one secret being marked active, which is then used for signing
- Add secrets versioning

authentik/providers/proxy/models.py
```diff
@@ -74,6 +75,9 @@ class ProxyProvider(OutpostModel, OAuth2Provider):
     """Protect applications that don't support any of the other
     Protocols by using a Reverse-Proxy."""
 
+    # Remove the legacy credential columns in 2027.2.
+    _cookie_secret = models.TextField(default="")
```
BeryJu, line 79: don't rename

authentik/providers/oauth2/models.py
```diff
@@ -219,6 +220,15 @@ class Meta:
 class OAuth2Provider(WebfingerProvider, Provider):
     """OAuth2 Provider for generic OAuth and OpenID Connect Applications."""
 
+    # Remove the legacy credential columns in 2027.2.
+    _client_secret = models.CharField(
```
BeryJu, line 224: don't rename, will break instances during upgrade

BeryJu (changes requested):

web/src/elements/buttons/IconRotateSecretButton.ts
```diff
@@ -0,0 +1,151 @@
...
+            } finally {
+                modal.inert = false;
+                dialog.closedBy = closedBy;
+            }
+        };
+
+        return renderModal(
+            html`<p>
+                    ${msg(
+                        "This replaces the value for every object using this secret. Update any external systems that use it.",
```
melizeche: We could improve the wording here so it can be a little more on the nose for the user that this rotates only for authentik and doesn't rotate external secrets

PR #26135, opened 2026-09-14
sources/ldap: respect accountExpires attributes

### LINKED ISSUE #22035: Take the expiration date in LDAP/AD into account

I would like to use the `accountExpires` attribute from Active Directory to automatically deactivate accounts during synchronization.

It would be nice to have an extension of the standard mapping templates.

I tried to create my own mapping rule, but it doesn't work.

```
from datetime import datetime, timezone, timedelta

expires = list_flatten(ldap.get("accountExpires"))
now = datetime.now(timezone.utc)

return {
  "is_active": not (expires and expires < now),
  "attributes": {
    "accountExpires": list_flatten(ldap.get("accountExpires")),
    "countryCode": list_flatten(ldap.get("countryCode")),
    "notes": list_flatten(ldap.get("description"))
  }
}
```

Image

### Additional context

_No response_

jsachs:
### What does this PR change?
Adds support for Active Directory’s `accountExpires` attribute during LDAP user synchronization.
- Deactivates expired accounts and removes their authentik sessions.
- Supports raw LDAP timestamps and schema-decoded datetimes, including both AD “never expires” values.
- Combines expiration, AD disabled/locked status, and existing FreeIPA lock status before saving the user, preventing one check from overriding another.
- Preserves AD’s existing authority over `is_active` when `userAccountControl` is available. Extending or clearing expiration allows reactivation when directory restrictions permit it.
- Reports invalid expiration values before updating the affected user.
- Documents synchronization timing and reactivation behavior.

authentik currently synchronizes AD disabled/locked status but does not account for account expiration. An expired AD account can therefore remain active in authentik.
A custom property mapping is not sufficient to address this reliably: the existing AD handler can overwrite the mapping’s `is_active` result based on userAccountControl. AD’s timestamp encoding and “never expires” sentinels also require special handling.
This change evaluates directory restrictions together so that expired accounts are consistently deactivated.

### How was this tested?
Ran the LDAP source test suite, which includes new coverage.

### Linked issues
Closes #22035

## Checklist


authentik/sources/ldap/sync/vendor/ms_ad.py
```diff
@@ -5,9 +5,48 @@
...
 from authentik.core.models import User
+from authentik.lib.sync.outgoing.exceptions import StopSync
+from authentik.sources.ldap.models import flatten
 from authentik.sources.ldap.sync.base import BaseLDAPSynchronizer
 
+FILETIME_EPOCH = datetime(1601, 1, 1, tzinfo=UTC)
+FILETIME_NEVER = 9223372036854775807
+
+
+def account_expired(value: Any) -> bool:
```
BeryJu: prefer helpers on `MicrosoftActiveDirectory` instead of bare functions

BeryJu (approved):

PR #23658, opened 2026-07-03
docs: add Zenith to installation options

odpay:
| Q                                      | A
| -------------------------------------- | ---
| Bug fix? (use the a.b branch)          | ❌
| New feature/enhancement? (use the a.x branch)      | ❌
| Deprecations?                          | ❌
| BC breaks? (use the c.x branch)        | ❌
| Automated tests included?              | ❌
| Related user documentation PR URL      |
| Related developer documentation PR URL |
| Issue(s) addressed                     |

## Description

hey, i work on Zenith Hosting, a startup doing one-click hosting for open source apps. we added Mautic to our marketplace, so this PR adds a small managed-hosting note in the self-host section (links to https://zenith.hosting/host/mautic?ref=gh).

we share a cut of revenue with the projects we host. happy to explain if you're curious: hello@zenith.hosting

No LLMs were involved.

---
### 📋 Steps to test this PR:

1. view the rendered README self-host section

PR #26139, opened 2026-09-14
docs: add Easypanel installation guide

Ahson-Shaikh:
Adds an Easypanel install guide alongside Docker Compose/Kubernetes/AWS. authentik has an official one-click template on Easypanel (https://easypanel.io/templates/authentik) that provisions PostgreSQL and Redis automatically.


BeryJu:
just like https://github.com/goauthentik/authentik/pull/23658, we're not adding new installation methods to the docs.

PR #26143, opened 2026-09-14
web/flows: render CAPTCHA widgets without an iframe wrapper

GirlBossRush:
## Details

This PR removes the iframe wrapper the CAPTCHA stage generated around every interactive widget. Providers are resolved by the challenge's `js_url` instead of by sniffing `window` globals, and widgets render directly in the flow document via each vendor's explicit render API. This let's us drop the `postMessage` bridge, generated document template, `document.write` fallback, blob URLs, and _some_ of the height-sync observers.

- Fixed an issue where a second CAPTCHA stage never initializes. The executor reuses one `<ak-stage-captcha>` and only swaps `challenge`, so stage two kept showing stage one's solved widget and the flow stalled.
- Fixed reCAPTCHA Enterprise using the classic API script 
- Fixed race condition where `FormFixture.selectSearchValue` not awaiting its own `fill`
- Fixed a11y for the "Provider Type", "Order" and "Stage" controls.


web/src/flow/stages/captcha/CaptchaStage.ts
```diff
@@ -232,370 +272,218 @@ export class CaptchaStage
...
-        }
 
-        const { globalName } = Controller;
-        const controller = new Controller(this);
-
-        try {
-            await this.#run(controller);
-            this.#logger.debug(`[${globalName}]: handler succeeded`);
+            if (this.challenge?.interactive) {
+                await controller.mount(this.container);
```
kensternberg-authentik: Up on line 409, you check to see if the captcha load is stale.  This is an async event, which means other things are happening while you're `await`ing; you surrender the execution thread to other tasks in the queue.  Is it possible the load went stale due to one of those tasks?  Should you check again here?

web/src/flow/stages/captcha/CaptchaStage.ts
```diff
@@ -232,370 +272,218 @@ export class CaptchaStage
...
-    #refreshControllers() {
-        if (!this.challenge) {
-            this.#logger.debug("No challenge, skipping controller refresh.");
+        // A flow can present several CAPTCHA stages in a row, and the executor reuses this
+        // element for each of them rather than replacing it — only `challenge` changes. The
+        // widget therefore has to be rebuilt here, not just when the element first renders,
+        // or the second stage keeps showing the first stage's solved widget.
+        if (
+            changedProperties.has("challenge") &&
+            this.#challengeSignature !== this.#loadedSignature
```
kensternberg-authentik: AIUI, the signature is sometimes just the URL of the Captcha object. Is it possible to have two captcha stages in a row from the same URL? Even by accident? What happens?

web/src/flow/stages/captcha/CaptchaStage.ts
```diff
@@ -232,370 +272,218 @@ export class CaptchaStage
...
-
-        this.#run(this.activeController);
-    };
+        // The container lives outside this element's subtree, so dropping the stage does
+        // not take it with it.
+        this.#frameObserver?.disconnect();
+        this.#frameObserver = undefined;
+        this.#container?.remove();
+        this.#container = undefined;
+        this.widgetLoaded = false;
```
kensternberg-authentik: Add `this.#loadedSignature = null;` here so the component knows to reload the captcha on disconnect/reconnect.  I know, I know, bloody unlikely, but weirder things happen.

kensternberg-authentik (commented):
There's a lot of state machine in here.

Suggestion: Have a paged with a *solved* Captcha on it, but before leaving the page, change the theme or locale.  See what happens.  :-)

kensternberg-authentik (approved):
Sorry, meant to approve.  There are some edge cases that I think should be looked at, though.

PR #26150, opened 2026-09-14
lib: support display names in ak_send_email

### LINKED ISSUE #25054: ak_send_email address limitation
### Describe the bug

The ak_send_email function works great for Template based emails through an Expression Policy.

I use this for Activation emails, for "Admin Approval" required type registration, when I flip a new account to Active, it triggers a nice templated email to the user.

The problem here, is that there's a bit of an inconsistency in the "To:" field when compared to using an Email Stage.

The field shows ONLY the email address in the email, whereas when an email is sent through an Email Stage, it has a proper "Name <email@address.com>" value.

While this is a "small" thing, it's also a matter of end user presentation and inconsistency.

I feel this could be easily resolved by simply allowing the address field to accept both possible address types, specific address (as it already does) or the properly formatted "Name <email@address.com>".  

Another option would be to simply add another field to the allowed parameters of ak_send_email, rfc_address perhaps, to allow properly formatted RFC To: values.

The third option, if you so choose, you could also potentially do a backend lookup on the database to find the corresponding user that corresponds to that email address and populate the "To" header appropriately before sending the email.  This method, however, would likely have problems with duplicate email addresses in your database, so the first two options are likely cleaner and easier to implement.

### How to reproduce

Create an Expression Policy that uses ak_send_email.

Example below:
> 
>     ak_send_email(
>         address=[str(target_user.email)],
>         subject="Account Activated",
>         template="email/account_activated.html",
>         context={
>             "user": target_user,
>             "url": login_url,
>             "authentik_name": company_name
>         }
>     )
> 

Setup either a flow or Notification Rule to consume it.

Trigger said flow/rule to process the Policy and send the email.

Open the email and notice the "To" field has only the email address and not a "proper" Name <email@address.com>" styled header.

### Expected behavior

I would expect this to have clean and consistent header values, including the "To" field, when any email comes from this platform.

### Screenshots

Email Stage example: (ignore the "From" field being inconsistent here, this screenshot is before I fixed my From setting in my .env)
Image

ak_email_send example:
Image

### Additional context

_No response_

### Deployment Method

Docker

### Version

2026.2.2

### Relevant log output

```shell

```

melizeche:
## Details

Allow `ak_send_email` recipients to include display names via (name, email) tuples or Name <email> strings. 
This aligns expression emails with the Email stage and adds documentation and test coverage.

### Linked issues

closes #25054


dewi-tik (approved):

authentik/lib/utils/email.py
```diff
@@ -19,12 +43,14 @@ def normalize_addresses(addr: str | list[str] | None) -> list[tuple[str, str]] |
     if addr is None:
         return None
     if isinstance(addr, str):
-        return [("", addr)]
+        return [_normalize_single_address(addr)]
+    if isinstance(addr, tuple):
```
BeryJu, line 47: if isinstance(addr, str | tuple):

BeryJu (approved):
1 nit

PR #26152, opened 2026-09-15
packages/django-dramatiq-postgres: reconcile pending tasks during continuous notifications

ryanpesek:
### What does this PR change?

Periodically reconcile runnable tasks from PostgreSQL into the consumer's pending set, even while PostgreSQL notifications continue arriving.

Reconciliation uses the existing `consumer_listen_timeout` as its interval and does not run while the consumer has reached its prefetch limit.

A regression test covers continuous stale notifications preventing a runnable task from being discovered.

PostgreSQL `NOTIFY` messages are delivered to every worker. In deployments where workers have different database latency, a lower-latency worker can claim and complete tasks before a higher-latency worker processes the corresponding notifications.

The higher-latency worker then spends its time processing stale notification IDs. Because the broker only queried the task table when no notification was available, a continuous notification stream could prevent that fallback query indefinitely, leaving runnable tasks undiscovered by the worker.

This was reproduced with workers in two regions sharing a PostgreSQL database. Restarting the affected worker caused its startup scan to find the runnable backlog and immediately resume processing at full concurrency.

---

## Checklist


BeryJu (approved):

rissson (approved):

rissson:
@ryanpesek do you mind resolving the conflicts? and then we can get this in

ryanpesek:
> @ryanpesek do you mind resolving the conflicts? and then we can get this in

Done!

PR #26166, opened 2026-09-15
tenants: fix syntax error

gergosimonyi:
This is python 2 syntax.

May be worth investigating how this was not discovered earlier.


gergosimonyi:
`ruff` complains because of this. Weird. I'm gonna close this and take care of it in the tenants removal PR with no backporting.

PR #26169, opened 2026-09-15
web/admin: add copy button to policies and property mappings pages

dewi-tik:
### What does this PR change?

Before:
image

After:
image

### Linked issues

---

## Checklist


BeryJu (commented):
I'd rather have this as a broadly usable API for anything, but with other objects like flows you get into the situation of how much should be copied

Might need to do a preview API for copying that would then show a tree of checkboxes of what you want to copy

PR #26171, opened 2026-09-15
tenants: remove

gergosimonyi:
Works on my machine 🤷


authentik/admin/migrations/0001_initial.py
```diff
@@ -0,0 +1,167 @@
...
+            ],
+            options={
+                "verbose_name": "Version history",
+                "verbose_name_plural": "Version history",
+                "db_table": "authentik_version_history",
+                "ordering": ("-timestamp",),
+                "managed": False,
+                "default_permissions": [],
+            },
+        ),
```
BeryJu, line 32: This shouldn't be in here, that model is unmanaged and used by system migrations

authentik/admin/migrations/0002_system_settings.py
```diff
@@ -0,0 +1,69 @@
+from json import loads
```
BeryJu, line 1: potentially nice as a single migration but not required

BeryJu (approved):
I spot-checked files and they seemed fine

authentik/admin/models.py
```diff
@@ -1,7 +1,132 @@
...
+    """Global settings for this authentik instance"""
+
+    id = models.BooleanField(primary_key=True, default=True)
+
+    avatars = models.TextField(
+        help_text=_("Configure how authentik should show avatars for users."),
+        default="gravatar,initials",
+    )
+    base_url = models.CharField(
+        max_length=200,
```
emilburzo, line 31: there was a wish from marc to change this to a textfield

https://github.com/goauthentik/authentik/pull/25560#discussion_r3882080248

authentik/brands/api.py
```diff
@@ -96,7 +96,7 @@ class Themes(models.TextChoices):
 
 def get_default_ui_footer_links():
     """Get default UI footer links based on current tenant settings"""
```
emilburzo, line 98:
```suggestion
    """Get default UI footer links based on current system settings"""
```

authentik/core/tests/test_users_avatars.py
```diff
@@ -21,7 +21,7 @@ def setUp(self) -> None:
 
     def set_avatar_mode(self, mode: str):
         """Set the avatar mode on the current tenant."""
```
emilburzo, line 23:
```suggestion
        """Set the avatar mode on the current system settings."""
```

authentik/events/consumer.py
```diff
@@ -5,13 +5,12 @@
...
 from channels.exceptions import DenyConnection
 from channels.generic.websocket import JsonWebsocketConsumer
-from django.db import connection
 
 from authentik.core.models import User
 
 
 def build_user_group(user: User):
-    return sha256(f"{connection.schema_name}/group_client_user_{user.uuid}".encode()).hexdigest()
+    return sha256(f"group_client_user_{user.uuid}".encode()).hexdigest()
```
emilburzo, line 13: shouldn't this be `CONFIG.get("postgresql.default_schema")`? 

or is there nothing using the old name after a restart?

authentik/flows/signals.py
```diff
@@ -23,9 +22,7 @@ def delete_cache_prefix(prefix: str) -> int:
 @receiver(monitoring_set)
 def monitoring_set_flows(sender, **kwargs):
     """set flow gauges"""
-    GAUGE_FLOWS_CACHED.labels(tenant=connection.schema_name).set(
-        len(cache.keys(f"{CACHE_PREFIX}*") or [])
-    )
+    GAUGE_FLOWS_CACHED.set(len(cache.keys(f"{CACHE_PREFIX}*") or []))
```
emilburzo, line 25: this drops labels I think?

authentik/lib/sync/outgoing/models.py
```diff
@@ -96,7 +96,7 @@ def get_sync_time_limit_ms(self) -> int:
     def sync_lock(self) -> pglock.advisory:
         """Postgres lock for syncing to prevent multiple parallel syncs happening"""
         return pglock.advisory(
-            lock_id=f"goauthentik.io/{connection.schema_name}/providers/outgoing-sync/{str(self.pk)}",
+            lock_id=f"goauthentik.io/providers/outgoing-sync/{str(self.pk)}",
```
emilburzo, line 99: any risk if the server/worker is upgraded/restarted in a delayed fashion?

authentik/outposts/consumer.py
```diff
@@ -22,13 +21,11 @@
 
 
 def build_outpost_group(outpost_pk: str | UUID) -> str:
-    return sha256(f"{connection.schema_name}/group_outpost_{str(outpost_pk)}".encode()).hexdigest()
+    return sha256(f"group_outpost_{str(outpost_pk)}".encode()).hexdigest()
 
 
 def build_outpost_group_instance(outpost_pk: str | UUID, instance: str) -> str:
```
emilburzo, line 27: starting to sound like a broken record, but same q if the different string affects anything

authentik/sources/scim/views/v2/base.py
```diff
@@ -109,7 +110,9 @@ def filter_query(self, request: Request, query: QuerySet) -> QuerySet:
             raise SCIMInvalidFilterError("Unsupported filter.") from exc
 
     def paginate_query(self, query: QuerySet) -> Page:
-        per_page = int(self.request.tenant.pagination_default_page_size)
+        per_page = int(
+            get_system_settings(["pagination_default_page_size"]).pagination_default_page_size
+        )
```
emilburzo, line 115: any option for not introducing an extra db call per page?

authentik/admin/files/backends/file.py
```diff
@@ -84,7 +83,7 @@ def file_url(
 
         def _file_url(name: str, request: HttpRequest | None) -> str:
             prefix = CONFIG.get("web.path", "/")[:-1]
-            path = f"{self.usage.value}/{connection.schema_name}/{name}"
+            path = f"{self.usage.value}/{CONFIG.get('postgresql.default_schema')}/{name}"
```
rissson, line 86: Can we instead of changing all of those, set
```
schema_name = CONFIG.get("postgresql.default_schema")
```
in `authentik/root/db/base.py::DatabaseWrapper`?

authentik/admin/utils.py
```diff
@@ -0,0 +1,16 @@
+"""authentik admin utils"""
+
+from authentik.admin.models import SystemSettings
+
+
+def get_system_settings(only: list[str] | None = None) -> SystemSettings:
```
rissson, line 6: As discussed, make this use a ContextVar that is lazy-set per-request from a middleware

rissson: And if we do that, then we can probably remove the `only` parameter

authentik/core/tests/test_impersonation.py
```diff
@@ -102,7 +102,7 @@ def test_impersonate_denied(self):
 
     def test_impersonate_disabled(self):
         """test impersonation that is disabled"""
-        tenant = get_current_tenant()
+        tenant = get_system_settings()
         tenant.impersonation = False
         tenant.save()
```
rissson, line 107: Maybe rename the var?

authentik/events/consumer.py
rissson: it's fine to remove it here. There are actually a bunch of places where we can remove connection.schema_name from

authentik/flows/signals.py
rissson: that is fine

authentik/lib/sync/outgoing/models.py
rissson: yeah, in the (sync) locks we unfortunately can't remove connection.schema_name

authentik/tasks/models.py
```diff
@@ -47,12 +46,6 @@ class Task(InternallyManagedMixin, SerializerModel, TaskBase):
         through_fields=("task", "dependency"),
     )
 
-    tenant = models.ForeignKey(
```
rissson, line 50: This might be a breaking change on update. We might want to set `default=None, null=True` first, and remove the column in a future update

authentik/tenants/migrations/0010_delete_domain_delete_tenant.py
```diff
@@ -0,0 +1,21 @@
...
+        ("authentik_tenants", "0009_alter_tenant_base_url"),
+        ("authentik_admin", "0001_initial"),
+        ("authentik_tasks", "0009_remove_task_tenant"),
+    ]
+
+    operations = [
+        migrations.DeleteModel(
+            name="Domain",
+        ),
+        migrations.DeleteModel(
```
rissson, line 18: Again, this is a breaking change that prevents smooth upgrades

rissson (changes requested):

PR #26177, opened 2026-09-15
web/admin: update licensing page

dewi-tik:
### What does this PR change?

- Adds a copy button for the Install ID.
- Replaces the forecast user tiles with current active user totals and counts of active users added within specified time ranges.
- Adds a user-count API endpoint that supports relative and absolute date ranges.
- Updates the Current license status card to show usage bars only when a license is installed; unlicensed installations show only the Unlicensed status.
- Removes the now unused forecast users API endpoint and references to it
- Updates the docs

#### Before:
image

#### After:

Without license:
image

With license:
image

The old tiles were misleading

### Linked issues

---

## Checklist


melizeche:
How does it look for an expired license?

dewi-tik:
> How does it look for an expired license?

@melizeche 

image

Hasn't been changed

PR #26210, opened 2026-09-16
core: add server-log endpoint for debugging

BeryJu:
a small endpoint without auth that the frontend can post to for logging.

*ONLY* ever enabled in debug mode, never in production

Specifically added for allowing the frontend to log things in environments where no developer tools are available, like Google's MinuteMaid or Apple PSSO


web/src/components/log.ts
```diff
@@ -0,0 +1,21 @@
+import { aki } from "#common/api/client";
+
+import { DebugApi } from "@goauthentik/api";
+
+let _serverLog = async (_: string): Promise<void> => {};
+
+if (process.env.NODE_ENV === "development") {
```
BeryJu:
```suggestion
if (process.env.NODE_ENV !== "production") {
```

PR #26252, opened 2026-09-18
packages/django-dramatiq-postgres/broker: don't unlock messages that weren't locked

ryanpesek:
### What does this PR change?

`_PostgresConsumer._consume_one()` now takes the message's advisory lock with its own `SELECT pg_try_advisory_lock(...)` before running the `UPDATE`. If the lock isn't acquired, it returns without calling `pg_advisory_unlock()`. If the lock is acquired but the `UPDATE` matches no row (message done, rejected or not due yet), it still unlocks, since the lock is now owned.

Every consumer on a queue gets the `NOTIFY` for every message and races to claim it. The losers' `UPDATE ... AND pg_try_advisory_lock(...)` matches 0 rows, and they then call `pg_advisory_unlock()` for a lock they never took. PostgreSQL logs this each time:

```
WARNING:  you don't own a lock of type ExclusiveLock
```

### How was this tested?

New tests in `authentik/tasks/tests/test_broker.py`:
- Mocked cursor: lock not acquired -> no `UPDATE` and no unlock; lock acquired but message not consumable -> unlock; message consumed -> no unlock.
- `TestPostgresConsumerAdvisoryLocks` uses real consumers and captures PostgreSQL notices on the locks connection:
  - an unowned unlock emits the warning (sanity check for the notice capture)
  - a losing consumer emits no warning, and the winner still holds its lock
  - a consumer attempting a non-consumable message leaves no advisory lock held

### Linked issues

None.

---

## Checklist


packages/django-dramatiq-postgres/django_dramatiq_postgres/broker.py
```diff
@@ -411,6 +411,17 @@ def _consume_one(self, message_id: str) -> Message[Any] | None:
             return None
 
         with self.locks_connection.cursor() as cursor:
+            # Take the lock on its own first. Every consumer is notified of every message, so most
+            # attempts lose the race to another consumer. Unlocking a lock that isn't held logs a
+            # "you don't own a lock" warning in PostgreSQL, so only unlock when we acquired it.
+            cursor.execute(
+                "SELECT pg_try_advisory_lock(%s)",
```
upramod: Just wondering, with the lock happening before the UPDATE now, could delayed messages cause each consumer to grab and release the lock even though the message isn't ready yet? 

Could this add unnecessary lock churn with a lot of delayed messages?

ryanpesek: Good catch, thanks. Delayed messages mostly don't get here: the enqueue NOTIFY only fires for eta IS NULL, and _fetch_pending_messages() already excludes eta >= now + timeout, which is the same window the UPDATE checks.

But you're right that there was still an acquire/release for messages that weren't consumable, e.g. finished or rejected messages when a consumer handles a notification late. I've moved the eligibility conditions into the lock query (SELECT pg_try_advisory_lock(...) FROM table WHERE <consumable conditions>). Missing, finished and not-yet-due messages now cost a single query and never take the lock. There's a new test covering missing / done / rejected / not due, which asserts one query, no unlock and no lock held.

upramod: Nice, moving the eligibility check into the lock query makes sense. Thanks for adding the regression coverage too.

rissson (approved):

Issue #25291, opened 2026-08-19
SCIM should sync on bind, unbind, and group-membership change


dominic-r:
## What

SCIM should automatically sync when a user is bound/unbound or group membership changes, not wait for the next full/manual sync.

## Want

- SCIM should automatically sync when a user is bound/unbound or group membership changes, not wait for the next full/manual sync

## Ideas this came from

- [#21754 SCIM: Group membership changes do not trigger automatic sync in 2026.2](https://github.com/goauthentik/authentik/issues/21754) — When users are removed from a group that controls SCIM sync scope (via application policy bindings), a sync is not automatically triggered. This means deprovisi
- [#23645 Trigger sync of SCIM user when it's bound to SCIM application](https://github.com/goauthentik/authentik/issues/23645) — Automatically sync a user as soon as it's bound to a SCIM application. Rather than having to wait for a full sync or having to trigger a manual sync

PR #26265, opened 2026-09-18
providers/scim: deprovision users after policy group removal

upramod:
Group removal updates SCIM group membership but leaves account deletion until a full sync.

This patch checks affected users' current access and deletes accounts that are no longer in scope. Other valid bindings preserve access. Failed deletes retain their mappings for retry.

All 78 SCIM tests pass, including both previously failing regressions. Tests use mocked HTTP. [Test run](https://github.com/upramod/authentik/actions/runs/35309894838)

Refs #25291. Covers membership removal only.


authentik/providers/scim/tasks.py
```diff
@@ -41,8 +52,56 @@ def scim_sync_delete_dispatch(*args, **kwargs):
 
 
 @actor(description=_("Sync a related object (memberships) for SCIM provider."))
-def scim_sync_m2m(*args, **kwargs):
-    return sync_tasks.sync_signal_m2m(*args, **kwargs)
+def scim_sync_m2m(group_pk: str, provider_pk: int, action: str, pk_set: list[int]):
+    # Update group membership while the removed users' remote IDs are still available.
```
BeryJu: this change should be in lib/sync/outgoing instead so that all sync providers can benefit from the fix

BeryJu (changes requested):

authentik/providers/scim/tasks.py
upramod: Done

Issue #23660, opened 2026-07-03
Sign-in hangs on the loading spinner in Android's "Add Google Workspace account"


KKHYeon:
### Describe the bug

When adding a Google Workspace account on Android (Settings → Add account → Google), Google opens the IdP sign-in in its embedded account-setup WebView (`com.google.android.gms.auth.uiflows.minutemaid.MinuteMaidActivity`; UA ends with `MinuteMaid`, engine SystemWebViewGoogle 149 — a current Chromium, so this is not an old-engine problem). authentik is the SAML IdP for the Workspace domain.

The user completes the full flow (identification → password → MFA). Server-side everything succeeds — the `login` event is created and the session is established — but the flow UI stays on the loading spinner forever. The expected follow-up request `GET /application/saml/<app>/?SAMLRequest=...` (the flow's `?next=`, which would produce the autosubmit POST back to Google's ACS) is never made. The same flow works fine in desktop and mobile browsers.

### How to reproduce

1. Google Workspace with SAML SSO via authentik (SP-initiated). Brand default authentication flow: identification → password → authenticator validation (webauthn and/or totp) → user_login.
2. On Android (tested: Android 16, Galaxy S24+, SystemWebViewGoogle 149.0.7827.164), add the Workspace account via Settings → Add account → Google.
3. Enter email and password, complete MFA (TOTP or passkey — same result).
4. Observe: server logs show the `login` event created; the WebView stays on the spinner; `/application/saml/<app>/` is never requested.

### Expected behavior

After the flow completes, the executor's redirect challenge navigates to `/application/saml/<app>/`, which autosubmits the SAMLResponse to Google's ACS and finishes the account setup. Cancelling WebAuthn should fall back to the device picker.

### Deployment Method

Kubernetes

### Version

2026.5.3

### Relevant log output

```shell

```

PR #26269, opened 2026-09-18
web/flows: never let continuous login block the final redirect

dpitts03:
The final redirect awaits continuous-login orchestration and the exit bookkeeping before navigating, so any error or hang there leaves the page on the spinner after the server has already finished the flow. In Google's Android account setup WebView, localStorage is null and the orchestrator throws (#23660, #26267).

- Catch and log orchestration errors in RedirectStage, then always navigate.
- Return a copy from discoverTabs(). waitForTabExit() rediscovers while the leader iterates the live Set, and entries re-added during for...of are visited again, so a follower that never exits was resumed forever.
- Resolve the wait when rediscovery rejects in the timer callback, and release the tab lock in a finally block.

### Linked issues

---

## Checklist


mo74m3ed:
> The final redirect awaits continuous-login orchestration and the exit bookkeeping before navigating, so any error or hang there leaves the page on the spinner after the server has already finished the flow. In Google's Android account setup WebView, localStorage is null and the orchestrator throws (#23660, #26267).
> 
> 
> 
> - Catch and log orchestration errors in RedirectStage, then always navigate.
> 
> - Return a copy from discoverTabs(). waitForTabExit() rediscovers while the leader iterates the live Set, and entries re-added during for...of are visited again, so a follower that never exits was resumed forever.
> 
> - Resolve the wait when rediscovery rejects in the timer callback, and release the tab lock in a finally block.
> 
> 
> 
> 
> 
> 
> 
> ## Details
> 
> 
> 
> ### What does this PR change?
> 
> 
> 
> ### Why is this change needed?
> 
> 
> 
> ### How was this tested?
> 
> 
> 
> ### Linked issues
> 
> 
> 
> 
> 
> 
> 
> ---
> 
> 
> 
> ## Checklist
> 
> 
> 
> -   [ ] The project has been linted, built, and tested (`make all`)
> 
> -   [ ] The documentation has been updated and formatted (`make docs`)
> 
> -   [ ] I have read the [AI usage policy](https://github.com/goauthentik/authentik/blob/main/AI_POLICY.md).
> 
>

PR #26272, opened 2026-09-18
web/flows/sfe: add Duo push support to simplified flow executor

rubyfoster5841:
## What does this PR change?

Adds Duo push support to the simplified flow executor (SFE).

`web/packages/sfe/src/index.ts`

- The SFE's `AuthenticatorValidateStage` already renders the
  `authenticator-validate` stage for the `static`, `totp` and `webauthn`
  device classes. However it had no `case` for `duo`, so the `render()` switch
  fell through to `default: break;` and rendered nothing.
- Adds a `case "duo"` to that switch, calling a new `renderDuo()`.
- Adds a `"Duo push"` label to the device-class picker, shown when a
  user has more than one compatible device.
- `renderDuo()` shows a "Sending Duo push notification..." message with
  a spinner and submits `{ duo: deviceChallenge.deviceUid }` — the same
  payload the default flow executor's `AuthenticatorValidateStageDuo`
  submits.

## Why is this change needed?

Because the SFE had no `duo` case, any user whose only configured (or
only remaining) authenticator-validate device is Duo got stuck
indefinitely on a blank SFE stage after the password stage, SFE would render
nothing and never progressed, with no error shown.

This was found via a Windows WAM/WebView2-embedded sign-in (SAML
passive auth through `Microsoft.AAD.BrokerPlugin.exe`), whose user agent
(`Edge/18.26200`, parsed as `family=Edge, major=18`) causes authentik to
serve the SFE per `compat_needs_sfe()`. Any Duo-only user hitting that
path had no way to complete MFA.

## How was this tested?

Manually, against a `default-authentication-flow` with an
`authenticator-validate` stage configured for the `duo` device class,
for a user with an enrolled Duo device. 

This was tested on both Docker Compose, as well as Kubernetes on
both 2026.8.2 and 2026.8.3 via a manual patch

- **Before:** after the password stage, the SFE rendered nothing and
  never progressed. The executor's next challenge was
  `ak-stage-authenticator-validate` with a `duo` device class, but
  `render()` had no matching case.

- **After:** the SFE shows "Sending Duo push notification...", the push
  is sent to the enrolled device, and on approval the flow completes
  normally.

**Known bugs:** on a denied push, the backend
re-returns the same `authenticator-validate` challenge, and since Duo can be
the only selected device, `renderDuo()` will auto-resubmit it

This will effectively resend the push instead of surfacing the denial. This
seems to be an existing quirk in the default flow executor (goauthentik#11368)

## AI Use
This patch was created with the help of Claude, specifically models Opus 4.8 and
Sonnet 5 using Claude Code.

## Checklist

- [x] The project has been linted, built, and tested (`make all`)
- [ ] The documentation has been updated and formatted (`make docs`) 
   (Not required as far as i can see(
- [x] I have read the AI usage policy.


BeryJu (approved):

web/packages/sfe/src/index.ts
```diff
@@ -598,6 +604,25 @@ class AuthenticatorValidateStage extends Stage<AuthenticatorValidationChallenge>
...
+                <p class="text-center">Sending Duo push notification...</p>
+                <div class="d-flex justify-content-center">
+                    <div class="spinner-border" role="status">
+                        <span class="sr-only">Loading...</span>
+                    </div>
+                </div>
+            </form>
+        `);
+
+        this.executor.submit({
```
upramod, line 622: Could we add a small SFE regression test around the Duo path? 

renderDuo() submits immediately from render(), so it would be good to lock down that a single Duo challenge sends exactly one { duo: deviceUid } request, and that selecting Duo from the picker uses the selected device UID. That would also catch accidental duplicate push submissions if this stage is re-rendered later.

PR #26275, opened 2026-09-20
core: split users list superuser annotation in API into two EXISTS subqueries

ryanpesek:
### What does this PR change?

Splits the `_annotated_is_superuser` annotation in `UserViewSet.get_queryset()` from one `EXISTS` with an `OR` across two join paths into two `EXISTS` subqueries combined with `OR`: one for direct membership of a superuser group, one for membership of a descendant group.

The result is unchanged: a user is a superuser if they're a member of a superuser group or of any of its descendants, the same as `User.is_superuser`.

PostgreSQL can only run a correlated `EXISTS` once per query as a hashed subplan when the correlation is simple equality. With an `OR` mixing two correlated conditions in one subquery, it runs once per user row instead, including rows skipped by `OFFSET` and rows inside the `DISTINCT` / `COUNT(*)` wrappers some filters add.

The plan for that per-row subquery also depends on the statistics of the `authentik_core_groupancestry` materialized view, which changes too little to be autoanalyzed. After the first group hierarchy is created, stale stats can flip it to a merge join that scans all groups for every user.

`EXPLAIN ANALYZE` of the users list with ~10k users:

| | Before | After |
|---|---|---|
| Annotation subplan | `SubPlan`, once per row | `hashed SubPlan` x2, once per query |
| Last page | ~190 ms | ~70 ms |
| Last page, stale ancestry view stats | ~20 s | ~75 ms |
| Superuser group with large child groups, page 20 | ~32 s | ~12 ms |

In a real deployment, the mean users list query time dropped by ~78%.

### How was this tested?

- New tests in `authentik/core/tests/test_users_api.py`:
  - `test_list_is_superuser`: list and detail `is_superuser` for direct, child and grandchild superuser group members, a member of a child of a non-superuser group, and a user without groups
  - `test_list_is_superuser_query_count`: more users don't add group queries (fails if the annotation is removed)
- Existing users and groups API tests pass; `black`, `ruff` and `mypy --strict` are clean
- On a real dataset: the old and new annotation and `User.is_superuser` agree for all users

The same code is on `version-2026.5` and `version-2026.8`, so a backport would be welcome.

AI usage: I used an AI coding agent to help investigate the query plans and draft the change, tests and description. I reviewed and edited all of it and ran the checks above.

### Linked issues

None.

---

## Checklist


ryanpesek:
@gergosimonyi

rissson (approved):

PR #26281, opened 2026-09-20
outposts/proxy: don't negotiate HTTP/2 to upstreams

### LINKED ISSUE #25388: Proxy Provider causes HTTP 400 error with IIS on HTTP/2
### Describe the bug

On Authentik 2026.8 the Proxy Provider while in proxy mode seems to break when proxying to a webpage that has HTTP/2 enabled. The error that was received on IIS is 
Bad Request

HTTP Error 400. The request is badly formed.

Unsure if this is happening on every HTTP/2 connection or just IIS ones. On my Authentik instance we have NGINX acting as a proxy for Authentik, however prior to update 2026.8 we had no issues with this. 

### How to reproduce

1. Configure a new application with Proxy as the provider
2. In the provider section set it to proxy, not the Forward Auth options
3. Configure external host however you want and for the internal host make sure that it is a http/2 webpage
4. Bind the provider to the necessary outpost 
5.  Attempt to reach the webapp and you should receive the HTTP 400 error

### Expected behavior

Expected for the proxy to work as intended and continue on to the website instead of receiving a 400 error. 

### Screenshots

_No response_

### Additional context

_No response_

### Deployment Method

Docker

### Version

2026.8

### Relevant log output

```shell

```

### LINKED ISSUE #25409: Proxy outpost sends a literal Host header on HTTP/2 upstreams, breaking https backends (2026.8.0 regression)
## AI assistance disclosure

Per [AI_POLICY.md](https://github.com/goauthentik/authentik/blob/main/AI_POLICY.md):

**Tool:** Claude (Anthropic), driven agentically via Claude Code.

**Extent:** AI-assisted throughout. The agent did the investigation — reading the 2026.8.0 outpost source, fingerprinting the 400 body against the upstream's own error format, bisecting the failure against logs, and running the `h2` reproduction — and drafted this report. Every result quoted below is real command output rather than model-generated text: the status codes and the `ProtocolError` came from executing those requests against the affected controller.

**Human-in-the-loop:** I have read and verified this report and take responsibility for its contents.

---

### Describe the bug

After upgrading to 2026.8.0, every request to a proxy provider whose `internal_host` is **https** returns `400`, passed through from the upstream. In my case the upstream is a UniFi (UDM) controller at `https://192.168.1.1:443`.

Every application path 400s, while the outpost's own endpoints still work:

| path | result |
| --- | --- |
| `/` | 400 |
| `/manage` | 400 |
| `/outpost.goauthentik.io/start` | 302 (correct redirect, correct `client_id`) |
| `/outpost.goauthentik.io/ping` | 204 |

Providers with plain-`http://` internal hosts on the same outpost are unaffected — they still redirect to auth normally.

The 400 body is the *upstream's* error format, not authentik's — that same controller answers `/api/self` with `{"error":{"code":401,"message":"Unauthorized"}}`. So the request is reaching the backend and being rejected there.

### How to reproduce

1. Create a proxy provider whose `internal_host` is `https://…` and whose backend negotiates **HTTP/2** via ALPN (any nginx-fronted appliance; mine is a UniFi UDM on `:443`).
2. Assign it to an outpost in `proxy` mode and route an external hostname to it.
3. Request any application path → `400`, emitted by the upstream.

The rejection is reproducible against the backend directly, with no authentik in the path. Using Python's `h2` with `validate_outbound_headers=False` so the malformed pair is actually transmitted:

```
h2, no host header (:authority only)      -> 200
h2, host: 192.168.1.1     (agrees)        -> 400
h2, host: unifi.example.com (mismatch)    -> 400
```

The same server returns 200 for all of these over HTTP/1.1.

Note this upstream rejects **any** literal `host` header on h2, even one that agrees with `:authority` — so merely keeping the two consistent is not sufficient here; the header should not be sent on h2 at all.

`h2` refuses to send the mismatched pair without that flag, citing the relevant rule:

```
h2.exceptions.ProtocolError: Request header block has mismatched
:authority and Host headers: b'192.168.1.1' / b'unifi.example.com'
```

### Expected behavior

Proxying to an https backend should keep working as it did through 2026.5.x.

On HTTP/2 upstreams the outpost should either drop the inbound `Host` and let `:authority` carry the value, or derive `:authority` from the same value it places in `Host` — not emit both inconsistently.

### Additional context

`build_client()` in `src/outpost/proxy/upstream.rs` enables HTTP/2 on the upstream connector *and* disables hyper's `Host` handling:

```rust
.https_or_http()
.enable_http1()
.enable_http2()
.build();
// Forward the request's own `Host` upstream instead of deriving it from the
// (internal) upstream URI authority. The proxy sets `Host` explicitly.
Ok(Client::builder(TokioExecutor::new())
    .set_host(false)
    .build(connector))
```

`create_proxied_request()` in `reverse_proxy.rs` strips only hop-by-hop headers, so the client's inbound `Host` is forwarded verbatim. For an https upstream that negotiates h2, the request then carries both `:authority` (from the rewritten URI) and a literal `host` header holding the *external* hostname.

Per RFC 9113 §8.3.1 that combination is malformed when the two disagree, and servers reject it. `set_host(false)` looks correct for HTTP/1.1 and wrong for HTTP/2.

**Workarounds tried, none sufficient:**

- Provider **Host header** override — still emits a literal `host` header, still 400s on this upstream.
- **`http://` internal_host** — upstream 301s to https.
- **A different upstream port** — not serving the application.

There is no provider field or environment variable exposing upstream HTTP version or `set_host`, so this is not configurable around. I ended up routing past the outpost entirely at the ingress layer.

### Deployment Method

Kubernetes

### Version

2026.8.0 (regression — 2026.5.6 was unaffected)

### Relevant log output

Worth flagging that authentik logs nothing anomalous — the outpost records the 400 as an ordinary request, with no error or warning emitted, which makes this hard to spot from the authentik side:

```shell
{"target":"authentik_axum::tracing","event":"/","runtime":"1","status":400,
 "host":"unifi.example.com","method":"GET","path":"/","scheme":"https"}
```

Timeline, for the regression window: last successful request at `2026-08-18 22:xx`, pods restarted onto 2026.8.0 at `23:20`, first 400 at `2026-08-19 01:06`, and no successful request since.

### LINKED ISSUE #25643: Proxy outpost (Rust, 2026.8.x) sends duplicate Host header on Proxy Provider requests to Internal Host — nginx rejects with 400 (distinct from #25226 / not fixed by #25234)
### Describe the bug

After upgrading to 2026.8.0 (Rust proxy outpost), every authenticated request to an app behind a Proxy Provider fails with a 400 Bad Request from the backend's nginx. This reproduces for every user, 100% of the time — login/auth with the upstream IdP succeeds, but the proxied request to the application itself fails immediately after.

This looks related to #25226 (fixed by #25234) but is a different code path: that issue was about the token-introspection/backchannel request duplicating the same Host value; this is the outpost's request to the Proxy Provider's configured Internal Host, and the duplicated values are two different hosts.

nginx access log shows the same fingerprint as #25226 — empty request line, bare 400:

`172.17.2.3 - - [01/Sep/2026:16:34:07 +0100] "-" 400 157 "-" "-" "-"`

With debug logging enabled on the backend's nginx, the actual cause:
`
client sent duplicate host header: "host: 192.168.0.4:9443", previous value: "host: 192.168.0.4:8971" while reading client request headers, client: 172.17.2.3, server: , host: "192.168.0.4:8971"`

192.168.0.4:9443 is the outpost's own bind port — not a value sent by anything upstream of the outpost (confirmed via the reverse proxy config in front of authentik, which preserves the original client Host header and does not override it). 192.168.0.4:8971 is the correct Internal Host configured on the Proxy Provider. The outpost appears to insert its own listener host:port as a Host header before separately setting the correct one for internal_host, without clearing the first — the same defect class as #25226, in a different request-construction path.

Confirmed via packet capture that TLS negotiation between the outpost and the internal host completes cleanly (full TLS 1.3 handshake, valid encrypted application data exchanged both directions) — ruling out any TLS/network cause; this is purely a malformed outgoing HTTP request from the outpost.

Still present on 2026.8.1 (retested after upgrading specifically to check whether #25234 resolved it — it did not).

### How to reproduce

1. Proxy Provider (not forward-auth) protecting an app, Internal Host set to https://<app-ip>:<port>, Internal Host SSL Validation disabled.
2. A reverse proxy (Caddy, in our case) in front of authentik's outpost, preserving the client Host header.
3. Log in as any user through the upstream IdP; authentik authenticates successfully.
4. On redirect to the protected app, the backend returns 400.

### Expected behavior

The outpost sends exactly one Host header — matching Internal Host — on requests it proxies to the configured backend application.

Deployment Method: Docker (embedded outpost)

Version: 2026.8.1 (also present on 2026.8.0)

Relevant log output

```
client sent duplicate host header: "host: 192.168.0.4:9443", previous value: "host: 192.168.0.4:8971" while reading client request headers, client: 172.17.2.3, server: , host: "192.168.0.4:8971"
172.17.2.3 - - [01/Sep/2026:16:34:07 +0100] "-" 400 157 "-" "-" "-" request_time="0.000" upstream_response_time="-"
```

### Screenshots

_No response_

### Additional context

Related to #25226 / #25234, but that fix targeted the token-backchannel request specifically and does not cover this path. Happy to provide full pcap, nginx debug logs, and reverse-proxy config on request.

### Deployment Method

Docker

### Version

2026.8.1

### Relevant log output

```shell

```

robisonjoel:
## AI assistance disclosure

Per [AI_POLICY.md](https://github.com/goauthentik/authentik/blob/main/AI_POLICY.md):

**Tool:** Claude (Anthropic), driven agentically via Claude Code.

**Extent:** AI-assisted throughout — reading the 2026.8.x and 2026.5.7 outpost
sources, building and deploying patched images against a live deployment, and
running the reproductions below. Every status code, log line and table entry
quoted here is real command output, not model-generated text.

**Human-in-the-loop:** I have read and verified this change and take
responsibility for its contents.

---

### What does this PR change?

The proxy outpost no longer negotiates HTTP/2 to upstream applications:

```rust
-    .enable_all_versions()
+    .enable_http1()
```

`set_host(false)` is deliberately left alone, so upstreams keep receiving the
client's `Host` header exactly as before.

Two small pre-existing bugs in the same path are fixed alongside:

- `*request.version_mut() = Version::HTTP_11` in `create_proxied_request` — an
  inbound HTTP/2 request would otherwise be rejected by the HTTP/1.1 pool with
  `UserUnsupportedVersion`.
- the upgrade-type comparison is now case-insensitive. Protocol names are
  case-insensitive (RFC 9110 §7.8), so a backend answering `Upgrade: WebSocket`
  currently gets a spurious 502.

This restores the Go outpost's behaviour. At `version/2026.5.7`,
`internal/outpost/proxyv2/application/mode_proxy.go`:

```go
func (a *Application) getUpstreamTransport() http.RoundTripper {
    return &http.Transport{
        TLSClientConfig: &tls.Config{InsecureSkipVerify: ...},
    }
}
```

No `ForceAttemptHTTP2` — which per `net/http.Transport` "conservatively disables
HTTP/2". The Go outpost never spoke HTTP/2 to upstreams. `.enable_all_versions()`
in the Rust port is the entire delta, and it breaks HTTP/2-capable backends in
two distinct ways.

**1. Ordinary requests are malformed.** The outpost forwards the client's
inbound `Host` while hyper derives `:authority` from the upstream URI. When they
differ that is malformed under RFC 9113 §8.3.1 and compliant upstreams reject
it. nginx with `error_log debug` names it exactly:

```
client sent ":authority" and "Host" headers with different values
while reading client request headers, request: "GET / HTTP/2.0"
```

The 400 the browser sees is the *upstream's own error page*, proxied back, which
makes this look like an application fault rather than an outpost one.

**2. WebSocket upgrades are silently dropped.** `Upgrade` and `Connection` are
connection-specific and forbidden in HTTP/2 (§8.2.2), so hyper strips them before
sending; the backend receives a plain `GET`, answers normally, and `call()` never
sees a `101`. HTTP/2 has no `101` at all (§8.6).

Demonstrated with one WebSocket echo backend behind three listeners on the same
nginx — same certificate, same proxy config — differing only in negotiated
protocol, logging what arrives:

```
h2      (TLS, ALPN h2)         proto=HTTP/2.0 up="-"         conn="-"       status=200   WS fails
h1      (TLS, ALPN http/1.1)   proto=HTTP/1.1 up="websocket" conn="upgrade" status=101   WS works
http:// (plain, control)       proto=HTTP/1.1 up="websocket" conn="upgrade" status=101   WS works
```

`up="-"` shows the headers are absent *on arrival*, so the backend is not
responsible.

Keeping HTTP/2 and fixing the authority mismatch is not a small change: it means
decoupling `:authority` from the dial target, which `hyper_util::client::legacy`
cannot do — its pool key is `(scheme, authority)` and the connector dials that
same tuple. That is a design change, not a regression fix.

### How was this tested?

On a live deployment with eight proxy providers, before and after:

| Internal host | Upstream | Before | After |
|---|---|---|---|
| `https://…:8443` | OPNsense / lighttpd (h2) | 400 on every request | serves, logged-in dashboard |
| `https://…/` | OPNsense / lighttpd (h2) | 400 on every request | serves |
| `https://…:11443/` | UniFi OS (h2) | 400 on every request | WebSocket opens, live dashboard |
| `http://…:3000` | Grafana (h1) | worked | unaffected |
| `http://…:30116/` | OpenSpeedTest (h1) | worked | unaffected |

The UniFi case exercises the WebSocket path specifically: its SPA will not
populate without `/api/ws/system`, and it additionally requires the `Origin`
host to equal `Host` on the upgrade — which holds precisely because this PR keeps
forwarding the client's `Host`.

Also verified against the echo rig above that all three listeners, including the
h2-capable one, now return `101`.

### Linked issues

closes #25388
closes #25409
closes #25643


src/outpost/proxy/upstream.rs
```diff
@@ -22,7 +22,23 @@ pub(super) fn build_client(insecure: bool) -> Result<UpstreamClient> {
...
+    // proxy forwards (see `set_host(false)` below) sits alongside the
+    // `:authority` hyper derives from the upstream URI; when they differ that
+    // is malformed (RFC 9113 8.3.1) and compliant upstreams answer 400. And
+    // `Upgrade`/`Connection` are forbidden in HTTP/2 (8.2.2), so hyper strips
+    // them and every WebSocket silently degrades into an ordinary request --
+    // HTTP/2 has no 101 either (8.6).
+    //
+    // Speaking h2 upstream *correctly* would mean decoupling `:authority` from
+    // the dial target, which `hyper_util::client::legacy` cannot do: its pool
+    // key is `(scheme, authority)` and the connector dials that same tuple.
```
rissson:
```suggestion
    // No HTTP2 and above as that breaks forwarding a Host header,
    // as it would conflict with `:authority`.
```

rissson (changes requested):
apart from a nit, the tests are not passing

robisonjoel:
> apart from a nit, the tests are not passing

Looking into the tests

robisonjoel:
Anything else youd want from my end on this?

rissson (approved):
all good for me!

PR #26284, opened 2026-09-20
website/docs: clarify SAML transient user lifecycle

dominic-r:
The Shibboleth guide recommends transient NameIDs without explaining that authentik creates temporary accounts.

Document account creation, deletion on logout, and expiration cleanup, including deletion while a session remains active. Explain the “Delete temporary users after” setting and recommend persistent NameIDs when accounts need to survive across logins.


website/docs/users-sources/sources/protocols/saml/index.mdx
```diff
@@ -33,6 +33,16 @@ If you have the provider metadata, you can extract the values that you need from
...
+
+The source's **Delete temporary users after** setting determines the expiration time, measured from account creation. It defaults to `days=1`. Use a duration such as `hours=8` or `hours=1;minutes=30`. This setting applies to temporary users created for transient NameIDs; it does not delay deletion on logout. Expiration is a fixed account lifetime, not an inactivity timeout. Deletion happens when the cleanup task runs after expiration, rather than at the exact expiration time.
+
+Use **Persistent** NameIDs when users need to keep the same authentik account across logins, including account-specific settings and group memberships. Configure the IdP to send a stable persistent NameID and set the source's **NameID Policy** to **Persistent**. authentik uses the NameID to look up the source connection on subsequent logins. Mapping a transient user's username or email to a stable value does not make the account persistent or remove its deletion behavior.
```
dewi-tik: Makes it a bit shorter while retaining important point.

```suggestion
When the IdP sends a `transient` NameID, authentik creates a temporary user linked to the SAML source. This is determined by the NameID format in the assertion, even if the source's **NameID Policy** requests another format.

Temporary users are deleted:

- when they log out of authentik, or
- by a cleanup task after they expire.

The **Delete temporary users after** setting determines the expiration time from account creation. The default is `days=1`; you can set values such as `hours=8` or `hours=1;minutes=30`. This setting does not delay deletion on logout. Cleanup tasks run periodically, rather than at the exact expiration time, and do not check for active sessions, so an expired temporary user can be deleted while still logged in.

Mapping a transient user's username or email to a stable value does not make the account persistent or prevent deletion.

Use **Persistent** NameIDs when users need to keep the same authentik account across logins. Configure the IdP to send a stable persistent NameID and set the source's **NameID Policy** to **Persistent**. authentik uses the NameID to find the existing source connection on subsequent logins.
```

dewi-tik (approved):
One small optional change

PeshekDotDev:
Ugh, there is a whole thing with this and features requested for this. IDR all the details offhand but don't merge this yet please. Slight can of worms

PeshekDotDev:
Nothing stated in the documentation is wrong, its smart to mention this, but there were oddities with how transient gets used, especially in academic circles. Just want to double check some things before talking more about it

dominic-r:
I've marked you as a requested reviewer @PeshekDotDev Feel free to push to this PR directly/merge it when ready/block it for now. You know way more about this than me, so this is probably best

PR #26320, opened 2026-09-21
lifecycle: rewrite the container entrypoint in Rust and add a debug image

emilburzo:
### Changes

- The container entrypoint moves into the authentik binary
- Everything that does still need a shell moves to a new `server-debug` image


src/boot.rs
```diff
@@ -0,0 +1,387 @@
...
+        .find(|shell| shell.is_file())
+}
+
+fn no_shell_error(command: &str) -> Report {
+    eyre!(
+        "'{command}' is not available in the production image, which ships no shell. Use the \
+         server-debug image: it is the same build with apt, a shell and the dev dependency group."
+    )
+}
+
```
rissson, line 133:
```suggestion
```

src/boot.rs
```diff
@@ -0,0 +1,387 @@
...
+            }
+            // The bash entrypoint opened up /root first, because the suite
+            // writes there, then ran `manage test authentik`.
+            if getuid().is_root() {
+                add_mode(Path::new("/root"), 0o777)?;
+            }
+            wait_for_db(&prometheus_dir)?;
+            let target = ["manage", "test", "authentik"].map(str::to_owned).to_vec();
+            run_as_authentik(&target, &prometheus_dir)
+        }
```
rissson, line 291:
```suggestion
        "bash" | "sh" => {
            let mut target = vec![command.to_owned()];
            target.extend_from_slice(&args[1..]);
            exec(&target, &env)
        }
        "test-all" => {
            // The bash entrypoint opened up /root first, because the suite
            // writes there, then ran `manage test authentik`.
            if getuid().is_root() {
                add_mode(Path::new("/root"), 0o777)?;
            }
            wait_for_db(&prometheus_dir)?;
            let target = ["manage", "test", "authentik"].map(str::to_owned).to_vec();
            run_as_authentik(&target, &prometheus_dir)
        }
```

rissson (changes requested):
some suggestions about the shell stuff we just discussed. I'll have a complete review ready tomorrow

.github/workflows/ci-main.yml
```diff
@@ -53,6 +53,9 @@ jobs:
         include:
           - name: server
             dockerfile: Dockerfile
+          - name: server-debug
+            dockerfile: Dockerfile
+            target: debug-image
```
rissson, line 58:
```suggestion
            target: debug
```

nit

.github/workflows/release-publish.yml
```diff
@@ -43,6 +43,7 @@ jobs:
       matrix:
         name:
           - server
+          - server-debug
```
rissson, line 46: Unsure if we actually want to publish a debug image on release. @BeryJu?

authentik/blueprints/management/commands/blueprint_shell.py
```diff
@@ -25,6 +30,8 @@ class Command(BaseCommand):
 
     def __init__(self, *args, **kwargs) -> None:
         super().__init__(*args, **kwargs)
+        if readline is None:
```
rissson, line 33: we probably want to print something here saying what is going wrong

lifecycle/container/Dockerfile
```diff
@@ -226,19 +212,77 @@ COPY ./pyproject.toml /
...
 
+# Setup 'ak'
+# /lifecycle comes first on PATH, and the binary runs its boot subcommand when invoked as `ak`
+RUN ln -s /ak-root/packages /packages && \
+    ln -sf /bin/authentik /lifecycle/ak
+
+# Stage: Debug
+# For `ak test-all`, AUTHENTIK_DEBUGGER and `docker exec ... bash`
+# Build it with --target debug-image as it is not the default
+FROM runtime-assemble AS debug-image
```
rissson, line 234:
```suggestion
FROM runtime-assemble AS debug
```

nit

lifecycle/container/Dockerfile
```diff
@@ -226,19 +212,77 @@ COPY ./pyproject.toml /
...
+    GOFIPS=1 \
+    RUST_BACKTRACE=full
+
+HEALTHCHECK --interval=30s --timeout=30s --start-period=120s --retries=3 CMD [ "ak", "healthcheck" ]
+
+ENTRYPOINT [ "dumb-init", "--", "ak" ]
+
+# Stage: Run
+# The default target, so it has to stay last
+FROM runtime-assemble AS final-image
```
rissson, line 269:
```suggestion
FROM runtime-assemble AS final
```

nit

.github/workflows/release-publish.yml
rissson: okay, I've read the rest of the PR, yeah it's definitely needed. I do wonder about the name tho. Maybe `server-full` instead? No strong feelings there from me

lifecycle/ak.py
```diff
@@ -1,17 +1,45 @@
...
+        os.execv(binary, [binary, "boot", *args])
+
+    if args and args[0] in NATIVE_COMMANDS:
+        os.execvp("cargo", ["cargo", "run", "--", "boot", *args])  # nosec
+
+    if args and args[0] == "dump_config":
+        os.execv(sys.executable, [sys.executable, "-m", "authentik.lib.config", *args[1:]])
+
+    subprocess.run([sys.executable, "-m", "lifecycle.wait_for_db"], check=True)  # nosec
+    os.execv(sys.executable, [sys.executable, "-m", "manage", *args])
```
rissson, line 42: This is a generic comment about how commands are handled here. We can either include these changes in this PR, or there will need to be a follow-up PR to address them.

I am generally unhappy about how commands and arguments are handled here.

All subcommands that are not `./manage.py` commands should be declared as such in rust directly (so server, worker, test-all, sh, bash, dump_config, debug, etc.).

The Python wrapper should have a list of those, and then if the first argument is any of those, it calls rust (not boot, just `authentik <cmd> <args...>`). If it does not recognise any of them, it should call rust with a `manage` command included (`authentik manage <cmd> <args...>`).

And then, in Rust, we do what we need to do to 1. boot, and 2. run what needs to be run.

Potentially we can add a `authentik boot <cmd>` command that gets called by the python wrapper to initialise the required things prior to actually starting the potentially long-running executable.

I also do not want the current argument parsing in boot.rs. We depend on a command parsing library for a reason, it's not to re-invent the wheel ourselves.

rissson (changes requested):

rissson (approved):
apart from the command parsing from before and the question about image names, LGTM

.github/workflows/release-publish.yml
emilburzo: I guess it depends how much we want to signal that they should be using the non-distroless one or not

`server-debug` will steer people away, `server-full` probably less so

I also thought about something like `server-with-tooling` but it's a bit much

lifecycle/ak.py
emilburzo: it does open some more design questions, but I'm attempting it in this PR, coming soon

emilburzo: addressed in https://github.com/goauthentik/authentik/pull/26320/changes/7f8a2e4a06c3627338561633085cbc31d3a2dfe2, let me know if that matches what you had in mind

some design decisions where I'm explicitly curios for feedback:
- boot re-execs the process to apply it's environment changes after dropping privs because changing it inplace needs unsafe (set_var)
- the django-only makefile/CI workflows call `python -m manage` directly so they don't need Rust / compile

(and of course any Rust feedback at all, if I'm doing something unidiomatic)

there's also some side-effects like the healthcheck no longer running `chown` for each call, but I think those aren't worth mentioning

.github/workflows/release-publish.yml
rissson: I'm not against `server-with-tooling` or `server-extra-tooling`. Even if long, it's clear what is going on.

lifecycle/ak.py
rissson: Thank you! I'm much happier with this.

> boot re-execs the process to apply it's environment changes after dropping privs because changing it inplace needs unsafe (set_var)

as it should, that's perfectly fine

> the django-only makefile/CI workflows call python -m manage directly so they don't need Rust / compile

that's okay for now. We will probably change it with freethreading and completely remove `manage.py` as we will depend on rust code to run python

> there's also some side-effects like the healthcheck no longer running chown for each call, but I think those aren't worth mentioning

those are actually improvements I think, healthchecks shouldn't do this x)

rissson (approved):

PR #26324, opened 2026-09-21
website/docs: Unify Docusaurus brand plugins.

GirlBossRush:
## Details

This PR moves a common step within our docs, copying static brand assets, to a Docusaurus-compatible plugin.

Once this published, a similar step in our website can be deprecated: https://github.com/goauthentik/website/blob/main/plugins/brand-assets.mts


dominic-r (approved):

packages/docusaurus-config/lib/plugins/brand.js
```diff
@@ -0,0 +1,98 @@
...
+/**
+ * Files within {@linkcode SourcePackage}, paired with their destination
+ * in the root static directory.
+ *
+ * Lockup names are intentionally swapped so `icon_left_brand.svg` is white by default.
+ *
+ * @type {ReadonlyArray<AssetEntry>}
+ */
+const BrandAssets = [
+    ["icon.png", "img/icon.png"],
```
kensternberg-authentik, line 29: Hmm.  This is mixing two things: the *behavior* of the plug-in, and then the *configuration* of the plug-in.  I feel like the actual BrandAssets definition shouldn't be part of the package; that puts it far from the thing it's branding, no?

If you change any of these, instead of just touching one file in the docs folder, you've gotta touch both this and any consumers (because the version number bumps), and that changes everyone's lock files, and ...  I mean, maybe you want that so every site that uses this info gets the same set of assets, but it still feels like a compromise.

website/static/img/.gitignore
```diff
@@ -1,7 +1,7 @@
 icon.png
 icon.svg
 icon_left_brand.svg
-icon_left_brand_colour.svg
+icon_left_brand_color.svg
 icon_top_brand.svg
-icon_top_brand_colour.svg
+icon_top_brand_color.svg
 social.png
```
kensternberg-authentik, line 7: This patch doesn't delete these old files; they're still there.

```
$ fd -I icon_left_brand_colour.svg
website/build/img/icon_left_brand_colour.svg
website/docs/build/img/icon_left_brand_colour.svg
website/static/img/icon_left_brand_colour.svg
```

kensternberg-authentik (approved):
Not broken, but there were two things that kinda bugged me.

PR #26327, opened 2026-09-21
web/elements: add Paginator component and revise Table and DualSelect to use them.

kensternberg-authentik:
web/elements/paginator: Add “paginator” to list of standarized elements

# What

This commit adds a new element, `ak-paginator`, and replaces the two uses of pagination in our product: the progressive reveal in DualSelect, and the page selector in Table.

The Paginator is a revision of the React version component. It has been cut down to our use case, and the term “bottom” has been replaced with “compact”; a “compact” paginator has only the summary (hidden below 640px) and the next/prev indicators; a full size one has the summary, the “first/prev/next/last” indicators, and an input box so the user can navigate to any page. A pair of helper functions convert from the Django paginator format to the one used by Patternfly’s paginator. The paginator emits an event that clients are expected to respond to, and then update the paginator component when the page is valid and has been changed to.

The converters have 10 tests; there are 30 unit tests covering edge cases, keyboard interaction, and ARIA validation.

Both Table and DualSelect have had their ad-hoc implementations replaced with Paginator. The DualSelect storybook has been updated to work correctly with the PageChangeEvent.

# Why

Well, for one thing, it’s not a great idea to multiple incompatible versions of the same component lying around.

For another, I’m about to add a third component that uses a paginator, and I didn’t want to add yet another paginator implementation to the mix. May as well clean it all up now.

---

## Checklist

-   [📖 The project has been linted, built, and tested (`make all`)
-   [📖] The documentation has been updated and formatted (`make docs`)
-   [📖] I have read the [AI usage policy](https://github.com/goauthentik/authentik/blob/main/AI_POLICY.md).


web/src/elements/Paginator_impl/Paginator.ts
```diff
@@ -0,0 +1,249 @@
...
+        const { page } = this.bounds;
+
+        // Client requests override internal tracking. Reset to track future events.
+        if (changed.has("page")) {
+            this.lastPage = null;
+        }
+
+        // Tell the client we're changing pages.  This shouldn't result in a loop
+        // since we ignore pointless "change to the same page" events.
+        if (page !== this.page && page !== this.lastPage) {
```
kensternberg-authentik: There is a test to make sure there's no infinite loop.

web/src/elements/Paginator_impl/Paginator.browser.test.ts
```diff
@@ -0,0 +1,427 @@
...
+        const element = await mount({ itemCount: 427, page: 7 });
+        const seen: number[] = [];
+
+        element.addEventListener(PageChangeEvent.eventName, (event) => {
+            seen.push(event.page);
+        });
+
+        element.itemCount = 50;
+        await element.updateComplete;
+        await element.updateComplete;
```
kensternberg-authentik, line 412: Should fall through immediately instead of being trapped.  Checks to make sure that a "change to the same page" event doesn't cause an infinite loop.

GirlBossRush (approved):

